Issue #19, opened 2021-09-09
Message track query enhancement


zhangjidi2016:
Currently, system topic (RMQ_SYS_TRACE_TOPIC) or message track defined in configuration files are used by default for message track query on the console. But now, users can specify specific topic to save message track information when sending and receiving messages on the client. If so, The console cannot query this message track information.

![image](https://user-images.githubusercontent.com/18254437/132642454-76330fdf-64eb-42be-a518-6c5380bd980e.png)

Can we select the previously specified message track topic before the user queries the message track information, like this? @StyleTang @vongosling 

![image](https://user-images.githubusercontent.com/18254437/132642236-7cc34bb2-5d23-4b3c-a176-9486835ea02e.png)

StyleTang:
I think choosing a track message topic manually is not a good user experience. 
Users rarely modify msgTrackTopic, we have a rocketmq.config.msgTrackTopicName config to use a specific message track topic.

zhangjidi2016:
Yes, msgTrackTopic rarely set, but if different users on the client side using different msgTrackTopic In the same cluster, and rocketmq. config. msgTrackTopicName in rocketmq-dashboard can be set only one. I think we should consider this situation.

StyleTang:
I get it, you are right, It is necessary to let users choose the msgTrackTopic by themselves.

PR #21, opened 2021-09-10
[ISSUE #19]Message track query enhancement

zhangjidi2016:

#19

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/controller/MessageTraceController.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/controller/MessageTraceController.java
@@ -64,7 +64,8 @@ public class MessageTraceController {
 
     @RequestMapping(value = "/viewMessageTraceGraph.query", method = RequestMethod.GET)
     @ResponseBody
-    public MessageTraceGraph viewMessageTraceGraph(@RequestParam String msgId, @RequestParam String traceTopic) {
+    public MessageTraceGraph viewMessageTraceGraph(@RequestParam String msgId,
+        @RequestParam(required = false) String traceTopic) {
         return messageTraceService.queryMessageTraceGraph(msgId, traceTopic);
     }
 }
--- a/src/main/java/org/apache/rocketmq/dashboard/controller/MessageTraceController.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/controller/MessageTraceController.java
@@ -64,7 +64,7 @@ public class MessageTraceController {
 
     @RequestMapping(value = "/viewMessageTraceGraph.query", method = RequestMethod.GET)
     @ResponseBody
-    public MessageTraceGraph viewMessageTraceGraph(@RequestParam String msgId) {
-        return messageTraceService.queryMessageTraceGraph(msgId);
+    public MessageTraceGraph viewMessageTraceGraph(@RequestParam String msgId, @RequestParam String traceTopic) {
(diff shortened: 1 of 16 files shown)


src/main/java/org/apache/rocketmq/dashboard/service/impl/TopicServiceImpl.java
```diff
@@ -61,16 +61,28 @@
...
+    public TopicList fetchAllTopicList(boolean skipSysProcess, boolean skipRetryAndDlq) {
         try {
             TopicList allTopics = mqAdminExt.fetchAllTopicList();
             if (skipSysProcess) {
                 return allTopics;
             }
 
-            TopicList sysTopics = getSystemTopicList();
             Set<String> topics = new HashSet<>();
+            if (skipRetryAndDlq) {
```
StyleTang, line 72: 1. What will happen if both skipSysProcess and skipRetryAndDlq are true.
2. maybe we can handle skipSysProcess and skipRetryAndDlq together, for example use Java lambda stream.
```java
allTopics.getTopicList().stream().map((item)->{
                ...
            }).filter(new Predicate<String>() {
                @Override
                public boolean apply(@Nullable String input) {
                    ...
                }
            }).collect(Collectors.toSet());
```

StyleTang (commented):
Great, this feature helps users who use custom track message topic to query message trace. 👍  
I have several comments here.

1. message track topic should have a default value, after this PR, each user will need at least 3 clicks when making a query message trace query (even if they use the default message track topic).
2. I suggest that the message track topic should be in the message trace page but not in MsgTraceDetailDialog, and the browser can cache which topic was chosen before.
3. rocketmq.config.msgTrackTopicName is useless, we can just remove it.

src/main/java/org/apache/rocketmq/dashboard/controller/MessageTraceController.java
```diff
@@ -64,7 +64,7 @@ public Object viewTraceMessages(@RequestParam String msgId) {
 
     @RequestMapping(value = "/viewMessageTraceGraph.query", method = RequestMethod.GET)
     @ResponseBody
-    public MessageTraceGraph viewMessageTraceGraph(@RequestParam String msgId) {
-        return messageTraceService.queryMessageTraceGraph(msgId);
+    public MessageTraceGraph viewMessageTraceGraph(@RequestParam String msgId, @RequestParam String traceTopic) {
```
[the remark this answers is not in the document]
zhangjidi2016: TraceTopic can be empty and system topic (RMQ_SYS_TRACE_TOPIC) is used by default.

src/main/java/org/apache/rocketmq/dashboard/service/impl/TopicServiceImpl.java
zhangjidi2016: Yes,  need to take into account the fact that both skipSysProcess and skipRetryAndDlq are true，I will optimize this method using stream.

StyleTang:
In my point of view, I don't suggest putting the topic selection on the popover. It’s more appropriate to put it on the page.
If we put it on the message trace page, for most users who use the default message track topic, they can just ignore it, but on the popover, the unnecessary click is required.


image

I suggest something like below 
image

StyleTang (approved):

Issue #48, opened 2021-12-03
query message by topic encounter a BUG


meateggmilk:
**BUG REPORT**

query message by topic, then select a wrong time, can still get some message.

meateggmilk:
I will fix It later

PR #49, opened 2021-12-03
[ISSUE #48] Fix query message bug

meateggmilk:

Fix a BUG when query messge by topic


StyleTang (approved):

StyleTang:
LGTM
According to [ConsumeQueue.java](https://github.com/apache/rocketmq/blob/master/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java#L198-L213), the `consumer.searchOffset` response may be larger or smaller than what we expect, so RocketMQ Dashboard should use `start++`/`end--` logic, this workaround is a little tricky.
Next time could you provide the test evidence, it is convenient for others to review.
Thanks~

vongosling (approved):

Issue #149, opened 2023-03-13
maven compilation is failing because of "java.lang.IllegalAccessError: class lombok.javac.apt.LombokProcessor "


Abhijeetmishr:
The issue tracker is **ONLY** used for bug report and feature request. 

Any question or RocketMQ proposal please use our [mailing lists](http://rocketmq.apache.org/about/contact/).

**BUG REPORT**

1. Please describe the issue you observed:

- mvn compile is failing and checked this on stack overflow also [link](https://stackoverflow.com/questions/66801256/java-lang-illegalaccesserror-class-lombok-javac-apt-lombokprocessor-cannot-acce)
![image](https://user-images.githubusercontent.com/43075376/224622914-7aa3d154-67d7-4b27-8a29-cb208add0f18.png)

2. Please tell us about your environment: Windows, VsCode

3. fix is described [here](https://stackoverflow.com/questions/66801256/java-lang-illegalaccesserror-class-lombok-javac-apt-lombokprocessor-cannot-acce)

Abhijeetmishr:
can I fix this ?

Oliverwqcwrw:
Sure ,I have assigned it to you

Abhijeetmishr:
Thank you Oliver for your quick response.

PR #150, opened 2023-03-14
How to view the message.type attribute of a Topic？&& lombok version update

Abhijeetmishr:

this pull request covers two issues
1. [#5538](https://github.com/apache/rocketmq/issues/5538) from rocketmq first commit cover this issue.
2. #149 this issue is covered by 2nd commit in this PR. 
 
## Brief changelog
1. [#5538] - issue reported for examine Topic Config Query not returning message.type in query result.
2. [#149] - maven compilation was failing because of lombok version.


pom.xml
```diff
@@ -94,7 +94,7 @@
         <rocketmq.version>4.9.3</rocketmq.version>
```
Oliverwqcwrw, line 94: We should upgrade to version 5.1.0 to support copying messageType from TopicConfig to TopicConfigInfo.
By the way, messageType should come from the attributes

Abhijeetmishr: @Oliverwqcwrw  we donot need this **private String messageType** we need member variable of type Map right ?

Oliverwqcwrw: we need messageType in the TopicConfigInfo, you can get the messageType form the map of the TopicConfig directly

Abhijeetmishr: > we need messageType in the TopicConfigInfo, you can get the messageType form the map of the TopicConfig directly

got your point

Abhijeetmishr: > We should upgrade to version 5.1.0 to support copying messageType from TopicConfig to TopicConfigInfo. By the way, messageType should come from the attributes

updating version to 5.1.0 in pom.xml is throwing lot of error 
image

Oliverwqcwrw: You can try to fix it,
it seems the class location has changed

Abhijeetmishr: @Oliverwqcwrw I have fixed most of the issues but there is some new methods added in MQAdminExtImpl.java that is unimplmented in rocketmq-dashboard what to do with this? Please check below image for ref
image

Oliverwqcwrw: Because there are many feature after 5.0, so you can see many unimplement function,

IMO, you can give an empty implementation or implement it if you have enough time.

BTW, it is better to split this pr because it has many changes.

Abhijeetmishr: for which of the changes/commit I have to create diff PR ?

Oliverwqcwrw: How about upgrading the version to 5.1.0 and adding a MessageType attribute?

Abhijeetmishr:
@Oliverwqcwrw should I close this PR

Oliverwqcwrw:
You can close the pr ,then split it and then submit it

Abhijeetmishr:
will be raising two PRs, one for upgrading rocketmq version to 5.1.0 and 2nd one for message.type missing from return value of examineTopic.

PR #200, opened 2024-06-04
Proxy Support For Dashboard

1294566108:

1.Support Proxy-Module for rocketmq-V5 in dashboard
2.Support Proxy Basic-Config Query
3.Support Proxy Detail-Config Query
4.Support the appearance or disappearance of proxy modules in the control panel as v5 and v4 switch


1294566108:
**1.If we switch to V5,Proxy Module will appear**
image

**2.If we switch to V4,Proxy Module will disappear**


**3.Query detail proxy config**
image

**4.Add proxy server**
image

**5.We can switch proxy server in proxyList**
image

PR #298, opened 2025-04-23
[ISSUES #213 #191] messageViews storeTimestamp sort desc

onlybigroc:

[ISSUES #213 #191]messageViews storeTimestamp sort desc

## Brief changelog

messageViews storeTimestamp sort desc

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/service/impl/MessageServiceImpl.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/service/impl/MessageServiceImpl.java
@@ -389,6 +389,7 @@ public class MessageServiceImpl implements MessageService {
 
                 }
             }
+            messageViewsSort(messageViews);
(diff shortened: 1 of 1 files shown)


OlnyBigRoc:
I'm not sure if this modification is correct, but it did solve the immediate problem. If there's a better approach, please let me know.

RockteMQ-AI (approved):
### Summary

Reasonable fix to sort message views by store timestamp (descending) before returning paginated results.

### Analysis

Messages from different queues may arrive interleaved. Sorting by `storeTimestamp` in descending order gives a consistent, time-ordered view in the dashboard.

- Sort applied in both `queryFirstMessagePage` and `queryMessageByTaskPage` — consistent ✓
- `Long.compare(time2, time1)` for descending — correct ✓
- Method is `private static` — good encapsulation ✓

One minor note: if `getStoreTimestamp()` could ever return 0 (unset), those messages would sort to the end. In practice the broker always sets this, so it's fine.

LGTM.

Issue #399, opened 2026-03-11
Use explicit class reference for resource loading in AbstractFileStore


flying-musk:
In `AbstractFileStore`, the default resource is currently loaded with:

`getClass().getResourceAsStream("/" + fileName)`

Since `AbstractFileStore` is an abstract class, using `getClass()` makes the resource lookup depend on the runtime subclass.

A small improvement would be to replace it with:

`AbstractFileStore.class.getResourceAsStream("/" + fileName)`

This makes the lookup context explicit and stable, and should avoid inheritance-related ambiguity without changing the intended behavior.

PR #400, opened 2026-03-11
[ISSUE #399] Use AbstractFileStore.class for resource loading

flying-musk:

This PR makes resource loading in `AbstractFileStore` use an explicit class reference instead of `getClass()`.

The current code uses:

`getClass().getResourceAsStream("/" + fileName)`

Since `AbstractFileStore` is designed to be extended, this makes the lookup depend on the runtime subclass, which can introduce inheritance-related ambiguity and make the behavior less explicit for future maintenance.

Using `AbstractFileStore.class.getResourceAsStream(...)` makes the resource lookup context explicit and stable.

## Brief changelog

- Replace `getClass().getResourceAsStream("/" + fileName)` with `AbstractFileStore.class.getResourceAsStream("/" + fileName)` in `AbstractFileStore`

## Verifying this change

- Verified that the project still compiles after the change
- Verified by running `mvn test`

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/service/impl/AbstractFileStore.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/service/impl/AbstractFileStore.java
@@ -34,7 +34,7 @@ public abstract class AbstractFileStore {
         filePath = configure.getRocketMqDashboardDataPath() + File.separator + fileName;
         if (!new File(filePath).exists()) {
             // Use the default path
-            InputStream inputStream = getClass().getResourceAsStream("/" + fileName);
+            InputStream inputStream = AbstractFileStore.class.getResourceAsStream("/" + fileName);
             if (inputStream == null) {
                 log.error(String.format("Can not found the file %s in Spring Boot jar", fileName));
                 System.exit(1);


flying-musk:
Gentle follow-up in case this slipped through.

This is a small, behavior-preserving cleanup that makes the resource lookup context explicit in `AbstractFileStore`.

I also verified the project still builds and the test suite passes locally:
- Tests run: 213
- Failures: 0
- Errors: 0
- Skipped: 0

Happy to revise if needed. Thanks!

flying-musk:
Gentle follow-up on this PR.

@Crazylychee I noticed you’ve been active in this repository, so I wanted to kindly ask: does this change look like something that would be acceptable for the project, or is there anything else I should adjust to make the PR easier to review and merge?

This is intended as a small, behavior-preserving cleanup, and I’d be very happy to revise it if needed. Thanks for any guidance.

oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #400 — [ISSUE #399] Use AbstractFileStore.class for resource loading
**Type:** Bug fix (1 file, +1/-1)

### Assessment
Minimal one-line fix changing the class reference used for resource loading in `AbstractFileStore`. Fixes #399.

### Verdict
✅ Clean, targeted fix.

---
*🤖 Automated review by oss-sentinel-ai*

RockteMQ-AI (approved):
LGTM. Correct fix: using `AbstractFileStore.class` instead of `getClass()` ensures resource loading works correctly even when the class is subclassed (where `getClass()` would return the subclass and might not find the resource in the same classloader path).

---

PR #412, opened 2026-07-05
[ISSUE #411] Fix NPE when getAllSubscriptionGroup returns null

### LINKED ISSUE #411: Consumer选项卡弹出Cannot invoke \"org.apache.rocketmq.remoting.protocol.body.SubscriptionGroupWrapper.getSubscriptionGroupTable()\" because \"subscriptionGroupWrapper\" is null
Cannot invoke \"org.apache.rocketmq.remoting.protocol.body.SubscriptionGroupWrapper.getSubscriptionGroupTable()\" because \"subscriptionGroupWrapper\" is null

itxaiohanglover:

Fixes #411

## Brief Description

When `mqAdminExt.getAllSubscriptionGroup()` returns `null` (e.g. broker not reachable, ACL denied, or 5.x Proxy mode where the call is not supported), the code at line 183 calls `subscriptionGroupWrapper.getSubscriptionGroupTable()` which throws:

```
Cannot invoke "SubscriptionGroupWrapper.getSubscriptionGroupTable()" because "subscriptionGroupWrapper" is null
```

**Root cause**: No null check on the return value of `getAllSubscriptionGroup()` before accessing its methods.

**Fix**:
1. Add null check inside the broker loop — if `getAllSubscriptionGroup` returns null, log a warning and skip to the next broker instead of crashing.
2. Fix the post-loop null check (line 197): the original condition `subscriptionGroupWrapper != null && ...isEmpty()` would fall through to line 202 and NPE if `subscriptionGroupWrapper` was null (e.g. empty broker table). Changed to `subscriptionGroupWrapper == null || ...isEmpty()` to handle both cases.

Manual code review against the issue stack trace. The fix follows the same null-guard pattern already used elsewhere in the codebase (e.g. `examineSubscriptionGroupConfig` fallback at line 218).

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/service/impl/ConsumerServiceImpl.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/service/impl/ConsumerServiceImpl.java
@@ -180,6 +180,10 @@ public class ConsumerServiceImpl extends AbstractCommonService implements Consum
             ClusterInfo clusterInfo = clusterInfoService.get();
             for (BrokerData brokerData : clusterInfo.getBrokerAddrTable().values()) {
                 subscriptionGroupWrapper = mqAdminExt.getAllSubscriptionGroup(brokerData.selectBrokerAddr(), 30000L);
+                if (subscriptionGroupWrapper == null) {
+                    logger.warn("getAllSubscriptionGroup returned null for broker: {}", brokerData.selectBrokerAddr());
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (commented):
### Summary
This PR fixes a NullPointerException when getAllSubscriptionGroup() returns null for a broker. The null guard and the logic change (AND → OR) are correct.

**Note:** PR #413 contains this exact same Java fix plus an additional JS fix for the Content-Type header issue. If #413 is merged first, this PR will have a merge conflict. Consider coordinating with the author to avoid duplicate work.

---

PR #413, opened 2026-07-05
[ISSUE #390] Fix addProxyAddr failure caused by Content-Type header overwritten in _fetch

### LINKED ISSUE #390: 无法添加 ProxyAddr
页面： 

Image

接口：
/proxy/addProxyAddr.do

报错：
```json
{
    "status": -1,
    "data": null,
    "errMsg": "Required request parameter 'newProxyAddr' for method parameter type String is not present"
}
```
原因：

前端是通过request body发送的，后端是通过request param接收的

相关代码：
前端：https://github.com/apache/rocketmq-dashboard/blob/master/frontend-new/src/api/remoteApi/remoteApi.js#L912
后端：https://github.com/apache/rocketmq-dashboard/blob/master/src/main/java/org/apache/rocketmq/dashboard/controller/ProxyController.java#L42

itxaiohanglover:
Fixes #390. The _fetch method overwrites caller-specified Content-Type with application/json, causing addProxyAddr form-urlencoded body to be unparseable by Spring MVC @RequestParam. Fix by putting default Content-Type before spread operator.

### DIFF
--- a/frontend-new/src/api/remoteApi/remoteApi.js
+++ b/frontend-new/src/api/remoteApi/remoteApi.js
@@ -61,8 +61,8 @@ const remoteApi = {
 
     _fetch: async (url, options = {}) => {
         const headers = {
-            ...options.headers,
             'Content-Type': 'application/json',
+            ...options.headers,
         };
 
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Two fixes in one PR:

1. **JS fix (remoteApi.js):** Correct — the spread operator ordering was causing caller-specified Content-Type headers to be overwritten by the default. Moving the default before the spread allows callers (like addProxyAddr) to override it with application/x-www-form-urlencoded.

2. **Java fix (ConsumerServiceImpl.java):** Good null guard for getAllSubscriptionGroup() — prevents NPE when a broker returns null, and the logic change from AND to OR in the final check correctly handles the null case.

**Note:** PR #412 contains the same Java fix. If both are merged, there will be a conflict. Consider coordinating with #412.

LGTM.

---

PR #442, opened 2026-07-16
[ISSUE #441] Validate audit query and cleanup parameters

### LINKED ISSUE #441: [Studio][Bug] Validate audit pagination and cleanup retention parameters
### Problem

`AuditService.queryLogs` currently accepts non-positive `page` and `pageSize` values and calculates its offset using `int` arithmetic. Invalid values can cause `subList` index failures, while sufficiently large valid integers can overflow and return the wrong page or fail unexpectedly.

`AuditService.cleanupLogs` also accepts zero or negative `beforeDays`. A zero value uses the current time as the cutoff, and a negative value moves the cutoff into the future, which can delete recent audit data instead of enforcing a positive retention period.

### Expected behavior

- Reject non-positive pagination values with a clear client error.
- Calculate pagination bounds without integer overflow.
- Reject non-positive audit retention values before calling the repository.

### Proposed fix

Validate the service inputs with `BusinessException` code 400, use `long` arithmetic for pagination bounds, and add regression tests for invalid and extreme values.

btlqql:
### Summary

- validate audit query pagination before reading repository data
- use `long` arithmetic for pagination bounds to avoid integer overflow
- reject zero or negative audit cleanup retention values
- add regression tests for invalid and extreme inputs

### Problem and root cause

`AuditService.queryLogs` used unchecked `int` values directly in `subList` index calculations. Non-positive inputs could produce invalid indices, and multiplying large positive page values could overflow. `cleanupLogs` similarly accepted non-positive retention periods, allowing a cutoff at the current time or in the future and risking deletion of recent audit entries.

### Impact

Invalid requests now fail predictably with `BusinessException` code 400 before repository work begins. Valid large pagination inputs no longer overflow, and cleanup always requires a positive retention period.

### Verification

- `mvn -o -B -ntp -Dtest=AuditServiceTest test` — 16 tests passed
- `mvn -o -B -ntp test` — 161 tests passed
- Checkstyle — 0 violations
- `git diff --cached --check` — passed

Fixes #441

### DIFF
--- a/server/src/main/java/com/rocketmq/studio/ops/audit/AuditService.java
+++ b/server/src/main/java/com/rocketmq/studio/ops/audit/AuditService.java
@@ -17,6 +17,7 @@
 package com.rocketmq.studio.ops.audit;
 
 import com.rocketmq.studio.common.domain.PageResult;
+import com.rocketmq.studio.common.exception.BusinessException;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
@@ -38,6 +39,7 @@ public class AuditService {
     public PageResult<AuditRecordVO> queryLogs(int page, int pageSize, String search,
                                              String operationType, String startDate,
                                              String endDate, String result) {
+        validatePagination(page, pageSize);
         log.info("Querying audit logs, page={}, pageSize={}, search={}, operationType={}, result={}",
                 page, pageSize, search, operationType, result);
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — solid input validation improvements.

The changes add proper bounds checking for pagination parameters and audit query cleanup:
- Uses `long` arithmetic to prevent integer overflow in page calculations
- Rejects non-positive retention values upfront
- Defensive null checks on query parameters

Test coverage is thorough — validates edge cases like overflow, negative values, and boundary conditions. Clean, defensive coding that prevents potential issues with malformed requests.

---

PR #447, opened 2026-07-17
[Studio] fix: connect alert rules page to APIs

Loyal-Young:
## Summary

- replace the alert-rules page's mock-only state with the existing alert-rules API
- persist rule creation, editing, enabling/disabling, and deletion; display API failures to the operator
- return the server's canonical rule after mutations and add API contract coverage

## Validation

- `npm.cmd test -- alertRules.test.ts`
- `npm.cmd run lint` (existing warnings only)
- `npx.cmd tsc -b`
- `npx.cmd vite build`


lizhimins:
Consolidated into #445 (squash-merged) to keep the rocketmq-studio history compact. Your changes are fully included there. Thanks @Loyal-Young!

PR #467, opened 2026-07-18
[Studio] fix: implement namespace listing

Loyal-Young:
## Summary

- implement the previously unsupported namespace listing operation behind `GET /api/namespaces`
- derive unique namespaces from topic metadata while retaining cluster identity and skipping blank namespaces
- return namespace records in deterministic name and cluster order, with service-level coverage

## Validation

- `git diff --check`
- The current Windows host does not have Java, Maven, or a Maven Wrapper available, so the new `MetadataServiceTest` could not be executed locally.


lizhimins:
Closing as deferred: the current version does not yet support Namespace (the Namespace frontend page #478 was also closed). We'll revisit namespace listing when Namespace support is on the roadmap. Thanks @Loyal-Young!

PR #484, opened 2026-07-20
[Studio] feat: configure API endpoint from env

Loyal-Young:
## Summary

- support configurable browser API base URLs and Vite development proxy targets
- retain `/api` and `http://localhost:8888` as backwards-compatible defaults
- document the new environment variables for reverse-proxy and container deployments

## Validation

- `npm.cmd test -- client.test.ts` (6 passed)


lizhimins:

PR #485, opened 2026-07-20
[Studio] fix: handle empty mini bar data

Loyal-Young:
Summary: renders a safe empty state for an empty series and adds an accessible trend label.\n\nValidation: git diff --check


lizhimins:

PR #487, opened 2026-07-20
[Studio] fix: use semantic page headings

Loyal-Young:
Summary: use an h1 for page headers by default while preserving the established visual size and allowing nested pages to opt into another level.\n\nValidation: git diff --check


lizhimins:

PR #488, opened 2026-07-20
[Studio] feat: add skip navigation link

Loyal-Young:
Summary: adds a keyboard-visible skip link so users can move past navigation directly to the main content.\n\nValidation: git diff --check


lizhimins:
Consolidated into #462 (squash-merged) to keep the rocketmq-studio history compact. Your changes are fully included there. Thanks @Loyal-Young!

PR #517, opened 2026-07-24
[Studio] Test Prometheus data source connections

Aias00:
### Motivation

This is a small follow-up for #431. The Settings data source test endpoint still returned a fixed success response, so users could not validate whether a configured Prometheus-compatible source was actually reachable.

### Changes

- Replace the fixed success response in `SettingsService.testDataSource` with a real Prometheus-compatible HTTP API probe.
- Query `/api/v1/query?query=up` for Prometheus, VictoriaMetrics, Thanos, and Mimir style data sources.
- Return explicit failure messages for invalid URLs, upstream Prometheus errors, connection failures, and timeouts.
- Avoid logging the full data source URL during connection tests.
- Add regression tests with a local mock HTTP server.

### Non-goals

This PR does not add credential fields or replace dashboard charts. It only makes the existing test endpoint validate the configured Prometheus-compatible URL.

### Verification

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -B -ntp -Dtest=SettingsServiceTest test`
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -B -ntp test`

### DIFF
--- a/server/src/main/java/com/rocketmq/studio/settings/SettingsService.java
+++ b/server/src/main/java/com/rocketmq/studio/settings/SettingsService.java
@@ -16,20 +16,51 @@
  */
 package com.rocketmq.studio.settings;
 
-import lombok.RequiredArgsConstructor;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.MediaType;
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #637. Consolidated with related Metrics/Prometheus validation changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #520, opened 2026-07-24
[Studio] Load namespaces for topic management

Aias00:
## Summary
- wire `/api/namespaces` through `MetadataService` and `MetadataProvider`
- provide a default `default` namespace from the current metadata provider implementation
- load namespace options from the API on the Topic page, with a default fallback when namespace loading fails
- add backend controller/service tests and frontend API coverage

## Motivation
This advances the Studio META-01 namespace baseline and avoids the current `UnsupportedOperationException` from `/api/namespaces`. Topic management can now use the backend namespace contract instead of a disabled hard-coded `default` selector.

Related issue: #180

## Verification
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -B -ntp -Dtest=MetadataServiceTest,NamespaceControllerTest test`
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -B -ntp test`
- `npm test -- metadata.test.ts`
- `npm run build`
- `npm run lint` (passes with existing Fast Refresh warnings in unrelated files)

### DIFF
--- a/server/src/main/java/com/rocketmq/studio/instance/topic/CloudMetadataProvider.java
+++ b/server/src/main/java/com/rocketmq/studio/instance/topic/CloudMetadataProvider.java
@@ -29,6 +29,15 @@ import java.util.List;
 @Component
 public class CloudMetadataProvider implements MetadataProvider {
 
+    private static final String DEFAULT_NAMESPACE = "default";
+
+    @Override
+    public List<NamespaceVO> listNamespaces() {
+        NamespaceVO namespace = new NamespaceVO();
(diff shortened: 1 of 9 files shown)


lizhimins:
Thanks for this PR! However, the current design of RocketMQ Studio does not need to consider namespaces, so this PR will be closed. If namespace support becomes a requirement in the future, we can revisit this work.

PR #542, opened 2026-07-24
[Studio] Validate consumer reset offset requests

Aias00:
## Summary
- replace ad hoc reset-offset request map parsing with a typed DTO
- validate required consumer group name and positive reset timestamp before invoking metadata admin operations
- return structured 400 responses for missing fields and invalid timestamp JSON types

## Scope
- RocketMQ Studio contest scope: Track 1 / BASE-01 Consumer Group reset offset baseline hardening
- Does not introduce Namespace behavior

## Tests
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=ConsumerGroupControllerTest test
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn test
- git diff --check


Aias00:
Superseded by #634. Consolidated with related instance resource action validation changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #549, opened 2026-07-24
[Studio] Validate alert rule action requests

Aias00:
## Summary
- replace alert rule toggle/delete map parsing with typed request DTOs
- validate required rule ids and enabled state before invoking alert operations
- return structured 400 responses for missing ids, missing enabled values, and invalid enabled JSON types

## Scope
- RocketMQ Studio contest scope: Track 1 / METRICS-01 alert rule action hardening
- Does not introduce Namespace behavior

## Tests
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -q test`
- `git diff --check`


Aias00:
Superseded by #632. Consolidated with related ops request validation changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #569, opened 2026-07-25
[Studio] fix: protect ACL user credentials

### LINKED ISSUE #568: [Studio] Protect ACL user credentials in list and update flows
## Problem

`GET /api/acl/users` currently returns every user's complete `accessKey` and
`secretKey`. The ACL page also sends the complete user object back to the update
API. Masking only the list response would therefore cause masked values to
overwrite the stored credentials.

Credentials should be returned in full only once, when the server creates them.
Ordinary user metadata updates must neither require nor replace them.

## Expected Behavior

- Return complete generated credentials only from the create response.
- Return masked credentials from list and update responses.
- Preserve stored credentials and creation time during metadata updates.
- Reject updates for unknown user IDs.
- Do not display or submit credential fields from the frontend.
- Do not mutate repository objects while masking responses.

## Proposed Fix

Return masked copies from list and update operations, merge editable metadata
with the existing stored user, remove credential fields from frontend forms and
update payloads, and add backend and frontend regression coverage.

yx9o:

Fixes #568 .

The ACL user list API returned complete access and secret keys after creation.
Masking only the list response would also cause the existing edit and
administrator-toggle flows to submit masked values and overwrite the stored
credentials. This change makes credential handling consistent across create,
list, and update operations.

## Brief changelog

- Return complete generated credentials only from the create response.
- Mask credentials in list and update responses without mutating stored users.
- Preserve stored credentials and creation time when updating ACL user metadata.
- Reject updates for users that do not exist.
- Stop frontend flows from displaying or submitting credential fields.
- Update API documentation and add backend and frontend regression tests.

### DIFF
--- a/docs/api-spec.md
+++ b/docs/api-spec.md
@@ -1056,6 +1056,8 @@ GET /api/acl/users
 | `clusters` | `string[]` | 授权集群列表 |
 | `createdAt` | `string` | 创建时间 |
 
+完整的 AccessKey 和 SecretKey 仅在创建用户的响应中返回一次，后续列表查询和更新响应只返回脱敏值。
+
 ### 7.5 创建 ACL 用户
 
 ```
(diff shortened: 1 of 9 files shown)


RockteMQ-AI (approved):
### Summary
Well-structured security fix that properly separates credential lifecycle: full keys returned only at creation time, masked everywhere else. The approach of preserving stored credentials server-side during updates (rather than relying on the frontend to pass them back) is the correct design — it eliminates the class of bugs where masked values overwrite real credentials.

### Findings
- **[Info]** `AclService.java:33-35` — The `VISIBLE_CREDENTIAL_CHARS = 4` and `MIN_PARTIALLY_MASKED_CREDENTIAL_CHARS = 17` constants are well-chosen. Short credentials get fully masked (`****`), longer ones show first/last 4 chars. Good defense-in-depth.
- **[Info]** `AclService.java` `updateUser` — Correctly reads the stored user via `findUserById` and preserves `accessKey`/`secretKey`/`createdAt` from the stored record. The `orElseThrow` with a descriptive `BusinessException` for missing users is clean.
- **[Info]** Frontend changes — Removing `accessKey`/`secretKey` from the edit form submission (`aclService.updateAclUser` call now only sends `id` + `admin`) is the right approach. The test update from `toMatchObject` to `toEqual` correctly enforces that no extra fields leak through.
- **[Info]** API docs (`api-spec.md`) — Good documentation of the new behavior, including the explicit note that full keys are only returned once at creation.

### Suggestions
- Consider adding a brief Javadoc on `maskCredential` explaining the masking strategy (visible chars threshold, fallback for short strings) for future maintainers.

Overall: solid security fix with comprehensive test coverage on both backend and frontend.

---

PR #556, opened 2026-07-24
[Studio] Validate K8s certificate requests

Aias00:
## Summary
- enable bean validation on K8s certificate create/update/renew/delete endpoints
- reject missing required certificate fields and unsupported certificate types before invoking service logic
- add controller tests for invalid certificate requests

## Scope
- RocketMQ Studio contest scope: Track 1 / BASE-01 HTTPS/SSL and ARCH-01 cluster operation hardening
- The `namespace` field is the existing Kubernetes certificate namespace field; this does not introduce RocketMQ Namespace behavior

## Tests
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -q -Dtest=K8sCertControllerTest,K8sCertServiceTest test`
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -q test`
- `git diff --check`

PR #579, opened 2026-07-25
[Studio] fix: make K8s certificate updates atomic

### LINKED ISSUE #578: [Studio] Keep K8s certificate updates atomic
## Problem

K8s certificate update and renewal operations modify the mutable `K8sCertVO`
returned by the repository before calling `save`.

`InMemoryK8sCertRepository` stores these objects by reference in a
`ConcurrentHashMap`. Readers can observe an intermediate state, and if `save`
fails, the stored object already contains changes from the failed request.

## Expected behavior

- Update and renewal publish new state only through the repository `save` operation.
- A failed `save` leaves the stored certificate unchanged.
- Partial updates preserve omitted fields, certificate identity, and metadata.
- Renewals preserve descriptive fields while replacing renewal-managed fields.

## Proposed fix

Copy the stored certificate, apply update or renewal changes to the copy, and
save the new object without mutating the repository-returned instance. Add
regression tests for successful writes and repository save failures.

yx9o:

Fixes #578 .

K8s certificate update and renewal operations mutated the repository-held
object before calling `save`. This allowed readers to observe intermediate
state and allowed a failed save to leave changes from a failed request in
memory. This change uses copy-on-write so repository state is replaced only
when `save` succeeds.

This complements #556, which validates REST request DTOs; this change protects
service and repository state during valid update and renewal operations.

## Brief changelog

- Build update and renewal results from copies instead of mutating stored objects.
- Preserve certificate IDs, creation timestamps, omitted fields, and metadata.
- Publish the new state only through the repository `save` operation.
- Add regression tests for successful writes and repository save failures.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java
@@ -69,49 +69,51 @@ public class K8sCertService {
 
     public K8sCertVO updateCert(UpdateCertDTO command) {
         log.info("Updating K8s certificate: {}", command.getId());
-        K8sCertVO cert = k8sCertRepository.findById(command.getId())
+        K8sCertVO existing = k8sCertRepository.findById(command.getId())
                 .orElseThrow(() -> new BusinessException(404, "Certificate not found: " + command.getId()));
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — clean implementation with proper test coverage.

---

PR #585, opened 2026-07-25
[Studio] fix: validate Ops toggle requests

### LINKED ISSUE #584: [Studio] Reject missing Ops toggle values
## Problem

`POST /api/ops/updateIsVIPChannel` and `POST /api/ops/updateUseTLS` accept requests without the required `useVIPChannel` or `useTLS` value.

Because the DTO fields use primitive `boolean`, a missing value is deserialized as `false`. The endpoint then returns success and calls the update service, which can silently disable the setting.

## Expected behavior

- Reject missing toggle values with HTTP 400.
- Prevent invalid requests from changing the setting.
- Preserve normal updates for explicit `true` and `false` values.

## Proposed fix

Use nullable `Boolean` fields with required validation, validate the request bodies before invoking the service, and add controller regression tests.

yx9o:

Fixes #584 .

Prevent the Ops toggle endpoints from treating missing boolean values as `false` and silently disabling VIP Channel or TLS.

## Brief changelog

- Reject missing `useVIPChannel` and `useTLS` values with HTTP 400.
- Prevent invalid requests from invoking the update service.
- Preserve explicit `true` and `false` updates.
- Add controller regression tests for both endpoints.


lizhimins:
Thanks for the fix! The same validation has already landed via #632 (merged first): `OpsTlsDTO` / `OpsVipChannelDTO` now use `@NotNull Boolean` with `@Valid` on the controller, matching this PR's change exactly (main additionally covers the NameServer endpoints). Closing as already covered — the toggle validation behavior you proposed is live on `rocketmq-studio`.

RockteMQ-AI (approved):
### Summary
LGTM — clean implementation with proper test coverage.

---

PR #597, opened 2026-07-26
[Studio] Make audit search null safe

Aias00:
## Summary
- make in-memory audit search tolerate records with missing detail/operator/target fields
- treat blank search text as no search filter
- add repository coverage for nullable audit fields

## Scope
- RocketMQ Studio contest track 1 / BASE-01 audit log baseline

## Tests
- JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=InMemoryAuditRepositoryTest test


Aias00:
Superseded by #635. Consolidated with related service search/filter normalization changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #598, opened 2026-07-26
[Studio] Make audit mock search null safe

Aias00:
## Summary
- make the frontend mock audit search handle missing text fields safely
- add a service-level regression test for searching audit records when operator/target are absent

## Verification
- npm --prefix web test -- --run src/services/opsService.test.ts
- (cd web && npm exec eslint -- src/services/opsService.ts src/services/opsService.test.ts)


Aias00:

PR #600, opened 2026-07-26
[Studio] Validate cluster flush disk type

Aias00:
## Summary
- convert invalid flushDiskType values into a 400 BusinessException instead of an unchecked enum parsing failure
- keep invalid config updates from reaching repository persistence
- add ClusterService regression coverage

## Verification
- JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=ClusterServiceTest test


Aias00:
Superseded by #646. Consolidated with related cluster config and store loading changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #604, opened 2026-07-26
[Studio] Trim mock Topic search

Aias00:
## Summary
- trim mock Topic search text before filtering names
- treat blank search text as no filter
- add service-level regression coverage for both cases

## Verification
- npm --prefix web test -- --run src/services/topicService.test.ts
- (cd web && npm exec eslint -- src/services/topicService.ts src/services/topicService.test.ts)
- npm --prefix web run build


Aias00:

PR #606, opened 2026-07-26
[Studio] Return cluster mock copies

Aias00:
## Summary
- reuse one copy helper for mock cluster list and detail responses
- make mock getCluster return defensive copies instead of shared fixture objects
- add regression coverage for nested cluster objects, config, and TPS history

## Tests
- npm --prefix web test -- --run src/services/clusterService.test.ts
- (cd web && npm exec eslint -- src/services/clusterService.ts src/services/clusterService.test.ts)
- npm --prefix web run build


Aias00:

PR #607, opened 2026-07-26
[Studio] Return instance mock copies

Aias00:
## Summary
- return defensive copies from mock instance list reads
- avoid exposing created or updated mock store records by reference
- add regression tests for list/create/update mock responses

## Tests
- npm --prefix web test -- --run src/services/instanceService.test.ts
- (cd web && npm exec eslint -- src/services/instanceService.ts src/services/instanceService.test.ts)
- npm --prefix web run build


Aias00:

PR #609, opened 2026-07-26
[Studio] Return connection mock copies

Aias00:
## Summary
- return copied client connection records from the mock connection service
- keep cluster/type filtering behavior unchanged
- add regression coverage for filtered mock response mutation isolation

## Tests
- npm --prefix web test -- --run src/services/connectionsService.test.ts
- (cd web && npm exec eslint -- src/services/connectionsService.ts src/services/connectionsService.test.ts)
- npm --prefix web run build


Aias00:

PR #612, opened 2026-07-26
[Studio] Load clusters in cluster store

Aias00:
## Summary
- wire clusterStore.fetchClusters to the existing cluster list service
- type the store with ClusterInfo instead of any
- keep loading state reset when cluster loading fails

## Tests
- npm --prefix web test -- --run src/stores/clusterStore.test.ts
- (cd web && npm exec eslint -- src/stores/clusterStore.ts src/stores/clusterStore.test.ts)
- npm --prefix web run build


Aias00:
Superseded by #646. Consolidated with related cluster config and store loading changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #616, opened 2026-07-26
[Studio] Return topic mock copies

Aias00:
## Summary
- return copied rows from topic mock list/create/update flows
- return copied route and consumer rows from topic detail mock helpers
- add regression coverage so caller mutations do not pollute later mock reads

## Tests
- npm test -- --run src/services/topicService.test.ts
- npm run lint -- --max-warnings=999 (passes with existing Fast Refresh warnings)
- git diff --check


Aias00:

PR #617, opened 2026-07-26
[Studio] Return message mock copies

Aias00:
## Summary
- return copied message rows and nested properties from mock message queries
- return copied trace nodes, consumer statuses, and DLQ rows from mock helpers
- add regression coverage so caller mutations do not pollute later mock reads

## Tests
- npm test -- --run src/services/messageService.test.ts
- npm run lint -- --max-warnings=999 (passes with existing Fast Refresh warnings)
- git diff --check


Aias00:

PR #619, opened 2026-07-26
[Studio] Encode message trace path segment

Aias00:
## Summary
- encode message IDs before building the trace API path
- add regression coverage for trace lookups with slash, space, and colon characters in the message ID

## Tests
- npm test -- --run src/api/message.test.ts
- git diff --check


Aias00:
Superseded by #639. Consolidated with related frontend API path encoding changes to reduce review overhead while preserving the same scoped fixes and tests.

PR #621, opened 2026-07-26
[Studio] Return consumer mock copies

Aias00:
## Summary
- return copied consumer group, instance, progress, and subscription mock rows
- copy created consumer groups before returning them to callers
- add regression coverage for caller-side mutation isolation

## Tests
- npm test -- --run src/services/consumerService.test.ts
- git diff --check


Aias00:

PR #622, opened 2026-07-26
[Studio] Return ACL mock copies

Aias00:
## Summary
- return copied ACL rule and user mock rows
- copy actions and cluster arrays on create and update before exposing results
- add regression coverage for ACL mock mutation isolation

## Tests
- npm test -- --run src/services/aclService.test.ts
- git diff --check


Aias00:

PR #623, opened 2026-07-26
[Studio] Return ops mock copies

Aias00:
## Summary
- return copied alert rule, system alert, and audit mock rows
- copy alert rule channel arrays on create, update, and toggle results
- add regression coverage for ops mock mutation isolation

## Tests
- npm test -- --run src/services/opsService.test.ts
- git diff --check


Aias00:

PR #594, opened 2026-07-25
[Studio] Load alert rules after mount

Aias00:
## Summary
- move AlertManagement rule loading out of render and into a mount effect
- guard async alert rule loading against state updates after unmount
- keep refresh parsing aligned with the latest disabled-rule map
- add page coverage for loading YAML-backed alert rules from the API

## Scope
- RocketMQ Studio contest track 1 / METRICS-01 adjacent observability and alert-rule management UX

## Tests
- npm test -- src/pages/studio/__tests__/AlertManagement.test.tsx
- npm run lint (passes with existing fast-refresh warnings outside this change)
- npm run build

PR #593, opened 2026-07-25
[Studio] Load ops settings after mount

Aias00:
## Summary
- move Ops page initialization out of render and into a mount effect
- avoid state updates after the async settings request resolves post-unmount
- add Ops page coverage for API-loaded NameServer/VIP/TLS state and read-only write control hiding

## Scope
- RocketMQ Studio contest track 1 / BASE-01 operations baseline: NameServer, VIPChannel, and TLS management

## Tests
- npm test -- src/pages/studio/__tests__/Ops.test.tsx
- npm run lint (passes with existing fast-refresh warnings outside this change)
- npm run build

PR #592, opened 2026-07-25
[Studio] Load producer topics after mount

Aias00:
## Summary
- move Producer topic initialization out of render and into a mount effect
- guard async topic loading so it does not update state after unmount
- add Producer page coverage for loading API-backed topic options

## Scope
- RocketMQ Studio contest track 1 / BASE-01 client and producer connection management UX

## Tests
- npm test -- src/pages/studio/__tests__/Producer.test.tsx
- npm run lint (passes with existing fast-refresh warnings outside this change)
- npm run build

PR #518, opened 2026-07-24
[Studio] Wire consumer page to group APIs

Aias00:
## Summary
- replace the Consumer Group page's static mock data with the shared consumer service/API layer
- load group detail, subscriptions, and consume progress from backend endpoints when opening the detail modal
- wire create, single delete, batch delete, and reset offset actions to API calls with loading/error states
- add page tests covering service-backed list and detail loading

## Motivation
This is part of the RocketMQ Studio control-plane baseline/CLIENT-01 work. The page previously rendered static mock consumer data, so a deployed Studio instance could not reflect real Consumer Group state, online instances, subscriptions, or consume progress.

## Verification
- `npm test -- ConsumerPage.test.tsx consumerGroups.test.ts`
- `npm run build`
- `npm run lint` (passes with existing Fast Refresh warnings in unrelated files)

PR #638, opened 2026-07-27
[Studio] Harden page data loading and state

Aias00:
## Summary
- load producer topic options after the Producer page mounts
- load ops settings after the Ops page mounts
- load alert rules after the Alert Management page mounts
- consolidate related page mount-loading fixes into one review unit

## Supersedes
#592 #593 #594

## Tests
- `npm ci --ignore-scripts --no-audit --no-fund && npm test -- --run src/pages/studio/__tests__/Producer.test.tsx src/pages/studio/__tests__/Ops.test.tsx src/pages/studio/__tests__/AlertManagement.test.tsx`
- `git diff --check`

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -414,6 +414,20 @@ const translations: Record<string, Record<Lang, string>> = {
   'consumer.viewDistribution': { zh: '查看分布', en: 'View Distribution' },
   'consumer.resetToTime': { zh: '重置到指定时间', en: 'Reset to Time' },
   'consumer.resetSuccess': { zh: '位点重置成功', en: 'Offset reset successfully' },
+  'consumer.fetchListFailed': {
+    zh: '消费组列表加载失败，请稍后重试',
+    en: 'Failed to load consumer groups. Please try again later.',
+  },
+  'consumer.fetchSubscriptionsFailed': {
+    zh: '消费组 {name} 订阅关系加载失败',
+    en: 'Failed to load subscriptions for consumer group {name}',
+  },
+  'consumer.fetchProgressFailed': {
+    zh: '消费组 {name} 消费进度加载失败',
+    en: 'Failed to load consume progress for consumer group {name}',
+  },
+  'consumer.createFailed': { zh: '消费组创建失败', en: 'Failed to create consumer group' },
+  'consumer.resetFailed': { zh: '消费位点重置失败', en: 'Failed to reset consume offset' },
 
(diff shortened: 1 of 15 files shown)


lizhimins (changes requested):
Thanks for the PR! Three of the four commits look good (load page data after mount, preserve topic table pagination, use auth admin flag for Ops writes). However the consumer-page commit ("Wire consumer page to group APIs") now conflicts in design with #631, which was merged first and already routes the consumer page through the service layer:

1. **Duplicate implementations**: #631 loads detail via separate `loadSubscriptions` / `loadProgress` helpers with a `subscriptionsByGroup` cache, while this PR introduces a parallel `Promise.all` + `detailLoading` / `normalizeGroup` / `buildSubscriptionFallback` approach for the same feature. Merging both would replace the already-merged design rather than build on it.
2. **Hardcoded Chinese error messages**: strings like `消费组详情加载失败，请稍后重试` should go through the i18n layer (same feedback as #518).

Could you rebase onto the latest `rocketmq-studio` and rework the consumer-page commit on top of #631's structure (keeping the useful additions like `submitting` / `resetSubmitting` states), with error messages localized? The other three commits should then apply cleanly. Happy to review again after the update.

Aias00:
> Thanks for the PR! Three of the four commits look good (load page data after mount, preserve topic table pagination, use auth admin flag for Ops writes). However the consumer-page commit ("Wire consumer page to group APIs") now conflicts in design with #631, which was merged first and already routes the consumer page through the service layer:
> 
> 1. **Duplicate implementations**: #631 loads detail via separate `loadSubscriptions` / `loadProgress` helpers with a `subscriptionsByGroup` cache, while this PR introduces a parallel `Promise.all` + `detailLoading` / `normalizeGroup` / `buildSubscriptionFallback` approach for the same feature. Merging both would replace the already-merged design rather than build on it.
> 2. **Hardcoded Chinese error messages**: strings like `消费组详情加载失败，请稍后重试` should go through the i18n layer (same feedback as #518).
> 
> Could you rebase onto the latest `rocketmq-studio` and rework the consumer-page commit on top of #631's structure (keeping the useful additions like `submitting` / `resetSubmitting` states), with error messages localized? The other three commits should then apply cleanly. Happy to review again after the update.

The duplicated consumer-page implementation mentioned in the review is gone, and the page now builds on the #631 structure. The hardcoded Chinese error messages were also moved to i18n keys.

lizhimins (approved):

PR #659, opened 2026-07-29
[Studio] Add capability context for architecture-aware UI rendering

zhaohai666:
## Summary

Introduce a capability-based UI rendering system that dynamically detects
cluster architecture features and conditionally shows/hides UI elements.

- Add `CapabilityContext` that queries `GET /architecture/capabilities` on
  mount to retrieve the cluster's feature matrix (namespace, liteTopic,
  popConsume, aclV2, grpcClient support)
- Support Mock mode (all features enabled for development/testing) and
  production mode (features fetched from backend)
- Wrap the application with `CapabilityProvider` in main.tsx
- Add a V5 architecture indicator badge in MainLayout's top bar that
  appears when `grpcClientSupported` is true, linking to settings page

## Test plan

- [ ] Verify app renders correctly with CapabilityProvider wrapping
- [ ] In mock mode (VITE_USE_MOCK=true), V5 indicator should appear
- [ ] When backend returns grpcClientSupported=false, indicator should hide
- [ ] Clicking the V5 badge should navigate to /settings

### DIFF
--- a/web/src/contexts/CapabilityContext.tsx
+++ b/web/src/contexts/CapabilityContext.tsx
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


lizhimins:
Closing for now: CapabilityContext queries GET /architecture/capabilities, but the studio backend does not expose such an endpoint (ClusterCapability exists as a model only, with no controller serving it). In real (non-mock) mode the context would always fall back to defaults and the V5 indicator would never render. Please pair this with a backend PR that actually serves the capabilities API, then we are happy to take both. Thanks!

PR #661, opened 2026-07-29
[Studio] Enhance MiniLine/MiniBar chart components

zhaohai666:
## Summary

Upgrade MiniLine from a simple polyline chart to a smooth, animated SVG
component with gradient fill and responsive mode support.

MiniLine enhancements:
- Catmull-Rom → Bézier spline interpolation for smooth curves
- SVG linear gradient fill under the curve
- Animated stroke-dasharray on mount
- Endpoint dot indicator on the last data point
- `responsive` prop for 100% width with preserved aspect ratio
- Unique gradient/glow IDs per instance to prevent multi-component conflicts
- New optional props: `strokeWidth`, `showDot`, `animated`, `responsive`
  (all backward-compatible with defaults)

MiniBar fix:
- Remove special-case for zero values; all bars now have minimum 4px height
  for visual consistency

## Test plan

- [ ] MiniLine renders smooth curves with gradient fill
- [ ] Animation plays on mount, no animation when `animated=false`
- [ ] Multiple MiniLine instances on same page have independent gradients
- [ ] `responsive=true` makes SVG fill parent width
- [ ] MiniBar: zero-value bars show 4px height (not 0px)

### DIFF
--- a/web/src/components/MiniBar.tsx
+++ b/web/src/components/MiniBar.tsx
@@ -52,7 +52,7 @@ const MiniBar = ({ data, color = '#1677ff', height = 32, width = 120, label }: M
           key={i}
           style={{
             width: barWidth,
-            height: `${value === 0 ? 0 : Math.max(4, (value / max) * height)}px`,
+            height: `${Math.max(4, (value / max) * height)}px`,
             backgroundColor: color,
             borderRadius: 2,
             opacity: 0.3 + (i / data.length) * 0.7,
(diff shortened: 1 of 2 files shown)


lizhimins:
Closing: the MiniLine enhancements were merged together with #660 (squashed into rocketmq-studio). The MiniBar change was intentionally dropped — trunk renders no visible bar at zero throughput by design (#629), so restoring a 4px minimum bar for zero values would revert that fix. Thanks for the contribution!

PR #662, opened 2026-07-29
[Studio] Add CSRF token support for mutating HTTP requests

zhaohai666:
## Summary

Add CSRF (Cross-Site Request Forgery) protection to the HTTP client to
complement the backend's CookieCsrfTokenRepository configuration.

- Read XSRF-TOKEN from cookies (set by backend with httpOnly=false)
- Request interceptor now attaches X-XSRF-TOKEN header for mutating
  methods: POST, PUT, DELETE, PATCH
- On first request, if XSRF cookie is not present, fetch
  `/rocketmq-dashboard/csrf-token` endpoint to trigger cookie creation
- Deduplicate concurrent CSRF fetch requests via shared promise
- Non-mutating requests (GET, HEAD, OPTIONS) are unaffected

## Test plan

- [ ] GET requests should NOT include X-XSRF-TOKEN header
- [ ] POST/PUT/DELETE requests should include X-XSRF-TOKEN from cookie
- [ ] First mutating request should trigger /csrf-token fetch if no cookie
- [ ] Existing Authorization header behavior unchanged
- [ ] 401 handling and auth session clearing still work correctly

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -49,12 +49,52 @@ const client = axios.create({
   timeout: 30000,
 });
 
-// Request interceptor: attach Authorization header
+// ─── CSRF support ────────────────────────────────────────────────
+// The backend uses CookieCsrfTokenRepository (httpOnly=false), so the token
+// is available in the XSRF-TOKEN cookie. For mutating requests (POST/PUT/DELETE)
+// we must send it back as the X-XSRF-TOKEN header.
+function getXsrfTokenFromCookie(): string | null {
+  const match = document.cookie
+    .split(';')
+    .map((c) => c.trim())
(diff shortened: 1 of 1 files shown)


lizhimins:
Closing: the studio backend has no CSRF infrastructure (no CookieCsrfTokenRepository, no /csrf-token endpoint — the fetched path /rocketmq-dashboard/csrf-token belongs to the legacy dashboard and would 404 here). Authentication uses a Bearer token from localStorage attached via header, which is not subject to cookie-based CSRF, so this interceptor would only add a failing pre-flight fetch to every first mutating request. If CSRF protection is introduced server-side later, a matching frontend PR would be welcome. Thanks!

PR #663, opened 2026-07-29
[Studio] Improve StatusBadge accessibility with semantic aria-label

zhaohai666:
## Summary

Improve StatusBadge accessibility by adding a semantic prefix to the
aria-label attribute.

Before: `aria-label={label}` (e.g. "在线")
After: `aria-label={`状态：${label}`}` (e.g. "状态：在线")

This helps screen reader users understand the widget's purpose before
hearing the status value, aligning with WCAG 2.1 guidelines for
informative status indicators.

## Test plan

- [ ] Inspect StatusBadge element: aria-label should start with "状态："
- [ ] Screen reader should announce "状态：在线" instead of just "在线"

### DIFF
--- a/web/src/components/StatusBadge.tsx
+++ b/web/src/components/StatusBadge.tsx
@@ -32,7 +32,7 @@ const StatusBadge = ({ status, text, showDot = true }: StatusBadgeProps) => {
   const config = STATUS_MAP[status] || STATUS_MAP.offline;
   const label = text || t(config.labelKey);
   return (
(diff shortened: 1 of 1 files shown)


lizhimins:
Closing: prefixing the aria-label with the hardcoded Chinese string "状态：" conflicts with the localization approach merged in #627, where StatusBadge accessible names follow the active locale — English users would hear a Chinese prefix. If a role prefix is desirable, it should go through the i18n layer (t(...)) instead of a literal. Thanks for the contribution!

PR #533, opened 2026-07-24
[Studio] Add configurable login protection

Aias00:
## Summary
- add configurable Studio login protection with token-backed API request checks
- authenticate against configured admin credentials when provided while keeping local-dev compatibility by default
- document Docker Compose login setup and pass auth/Prometheus env vars into deployment paths

## Scope
- BASE-01 login and permission baseline
- related to #97, #154, and #203

## Tests
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn test
- bash -n deploy/deploy.sh
- docker compose -f deploy/docker-compose.yml --env-file deploy/.env.example config
- git diff --check

Issue #427, opened 2026-07-10
[studio][Feature][RIP-1][Track1] Implement ARCH-01 SPI framework & BASE-01 baseline migration


zhaohai666:
# \[studio\]\[Feature\]\[RIP\-1\]\[Track1\] ARCH\-01 \& BASE\-01 Design Specification

## 1\. Overview

This feature implements the two foundational core modules of RocketMQ Studio RIP\-1 Track1: **ARCH\-01 SPI Framework** and **BASE\-01 Baseline Migration**\. It establishes the new 3\-tier control\-plane architecture for RocketMQ Dashboard, unifies heterogeneous cluster access logic, and completes a full technology stack upgrade while retaining all original v2\.1\.0 baseline capabilities\.

This work serves as the mandatory underlying foundation for all subsequent RIP\-1 modules \(metadata management, ACL adaptation, observability, client diagnosis\) and lays the architecture baseline for RIP\-3 AI\-native interaction capabilities\.

### 2\.1 Design Purpose

The original dashboard relies on hard\-coded version judgments and fixed SDK calling logic, which cannot adapt to RocketMQ 4\.x Broker architecture, 5\.x Proxy architecture, and cloud managed clusters simultaneously\. The SPI framework solves version fragmentation, multi\-protocol differences, and cloud vendor differentiation, realizing **capability\-driven dynamic adaptation**\.

### 2\.2 Core SPI Abstraction

Three orthogonal core interfaces are defined under the `org.apache.rocketmq.dashboard.architecture` package to decouple cluster topology, data query, and command execution:

- **ClusterProvider**: Responsible for cluster registration, health check, connection initialization, adapting NameServer direct connection and Proxy endpoint connection modes\.

- **MetadataProvider**: Unified metadata data source contract, covering all dashboard query capabilities such as clusters, topics, consumer groups, broker status, and message traces\.

- **AdminClient**: Unified operation execution layer, encapsulating Remoting\-based `DefaultMQAdminExt` and gRPC\-based Proxy admin SDK\.

### 2\.3 Multi\-Environment Provider Implementation

The framework provides out\-of\-the\-box implementation providers for mainstream deployment modes:

- V4 series providers: Adapt traditional Broker\-centric 4\.x clusters based on Remoting protocol

- V5 Proxy series providers: Adapt 5\.x cloud\-native Proxy clusters based on gRPC protocol

[...]

PR #668, opened 2026-07-29
[studio] fix: honor configurable login protection in frontend routes

### LINKED ISSUE #664: [studio][Bug][Track1] Honor configurable login protection in frontend routing
## Problem

RocketMQ Studio supports configurable backend API protection through `STUDIO_AUTH_LOGIN_REQUIRED=true` (introduced by #533), but the frontend does not guard management routes.

When login protection is enabled, an unauthenticated user can still open a management route such as `/instance/topic`. The page shell renders, its API calls receive `401 Unauthorized`, and the shared response interceptor clears the local session and reloads `/`. Because `/` is also unguarded, the user is not routed to `/login` and can repeatedly trigger unauthorized requests.

## Reproduction

1. Start Studio with:
   - `STUDIO_AUTH_LOGIN_REQUIRED=true`
   - a configured admin username and password
2. Open `/instance/topic` in a private browser session without a token.
3. Observe that the management layout renders and API requests fail with 401 instead of navigating to the login page.

## Expected behavior

- The frontend should discover whether login protection is enabled.
- Protected management routes should render only when login is disabled or the current bearer token is valid.
- Missing, expired, or revoked sessions should be cleared and redirected to `/login`.
- Mock-only frontend development should continue to work without a backend.
- A backend availability failure during the auth check should show a retry state rather than expose protected pages.

## Proposed atomic scope

- Add a public, non-cacheable `GET /api/auth/status` endpoint returning:
  - `loginRequired`
  - `authenticated`
- Exempt only that status endpoint from the auth interceptor.
- Add a frontend route guard around management routes.
- Keep the existing login/logout and token format unchanged.
- Add focused backend and frontend regression tests.
- Keep the change around 10 files and avoid `api/client.ts`, so it does not overlap the active CSRF work in #662.

This is a Track 1 / BASE-01 login-flow correctness fix and a small follow-up to the architecture scope in #427.

## Validation

- Backend controller and interceptor tests
- Frontend auth API and route guard tests
- Full `mvn test`, frontend test/lint/build
- Manual verification with login protection enabled and disabled

I would like to claim and implement this atomic fix.

ai-yang:
## Summary

- add a public, non-cacheable `GET /api/auth/status` endpoint that reports whether login is required and whether the optional bearer token is active
- guard every management route except `/login`, preserving the frontend-only mock workflow when `VITE_USE_MOCK=true`
- clear invalid local sessions and redirect to `/login`; fail closed with a localized retry state when the status request fails
- add backend controller/interceptor coverage and frontend API/route-guard regression tests

Fixes #664. This is the small Track 1 / BASE-01 login-flow unit discussed in #427 and follows up the configurable authentication added by #533. It intentionally does not modify `web/src/api/client.ts`, avoiding overlap with #662.

## Root cause

The backend interceptor honors `STUDIO_AUTH_LOGIN_REQUIRED`, but the React router had no way to discover that configuration. Unauthenticated users could render management pages, receive 401 responses, and be reloaded to the still-unguarded `/` route instead of reaching `/login`.

## Verification

- `mvn -B -ntp -Dtest=AuthControllerTest,AuthInterceptorTest test` — 11 tests passed
- `mvn -B -ntp test` — 452 tests passed
- `mvn -B -ntp checkstyle:check` — 0 violations
- `npm test -- --run src/api/auth.test.ts src/App.test.tsx` — 9 tests passed
- `npm run lint` — passed (4 existing Fast Refresh warnings)
- `npm run build` — passed
- containerized production-artifact validation:
  - login disabled: anonymous status returned `loginRequired=false`; `/instance/topic` served the SPA
  - login enabled: anonymous status returned `authenticated=false`; protected API returned 401
  - valid login: status returned `authenticated=true`; protected API returned 200
  - logged-out token: status returned `authenticated=false`; protected API returned 401

### Existing test-suite issue

The full frontend run completed 232 of 233 tests. The unchanged `ClusterPage.test.tsx` failed because the page mutates its mock connection count with `Math.random()` every two seconds while the test asserts the fixed value `1,842`; an isolated rerun observed `1,776`. The cluster page, its mock data, and that test are unchanged from `rocketmq-studio` and are outside this PR's scope.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthController.java
@@ -19,7 +19,10 @@ package org.apache.rocketmq.studio.auth;
 
 import org.apache.rocketmq.studio.common.domain.Result;
 import lombok.RequiredArgsConstructor;
+import org.springframework.http.CacheControl;
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
@@ -32,6 +35,19 @@ import org.springframework.web.bind.annotation.RestController;
 public class AuthController {
 
     private final AuthService authService;
+    private final AuthProperties authProperties;
+
+    @GetMapping("/status")
+    public ResponseEntity<Result<AuthStatusVO>> status(
+            @RequestHeader(value = HttpHeaders.AUTHORIZATION, required = false) String authorization) {
+        AuthStatusVO status = AuthStatusVO.builder()
+                .loginRequired(authProperties.isLoginRequired())
+                .authenticated(authService.isAuthenticated(authorization))
+                .build();
+        return ResponseEntity.ok()
+                .cacheControl(CacheControl.noStore())
+                .body(Result.ok(status));
+    }
 
     @PostMapping("/login")
     public Result<LoginVO> login(@RequestBody LoginDTO request) {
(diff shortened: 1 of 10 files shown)


ai-yang:
Hi maintainers, the CI workflow for this fork PR is currently marked `action_required` and has not created any jobs.

Could someone with write access please approve and run the workflow when convenient? The local verification results are included in the PR description.

Thank you!

PR #695, opened 2026-07-31
[Studio] Harden broker collection, message paging and user context

123123213weqw:
### Motivation

Fixed three real bugs found while reviewing the code:

1. **DashboardCollectTask: broker stats retry never actually retried**
   In `fetchBrokerRuntimeStats`, the recursive retry call's result was discarded and the original exception re-thrown, so a single failed fetch aborted the whole `collectBroker` scheduled task instead of retrying up to `retryTime` times. The caller already handles a `null` result by skipping the broker.

2. **MessageServiceImpl: NPE when `taskId` is blank in `queryMessageByPage`**
   `Cache.getIfPresent(query.getTaskId())` throws `NullPointerException` when the request body has no `taskId`. The API documents `taskId` as optional, so this endpoint 500s on valid input. A taskId is now generated when blank and reused consistently for the cache key and response.

3. **AuthInterceptor: user info in ThreadLocal is never cleared**
   `LoginServiceImpl` writes the logged-in user into `UserInfoContext` (a `ThreadLocal`) but nothing ever calls `UserInfoContext.clear()`. Since Tomcat reuses worker threads, the previous request's user info can leak into the next request handled by the same thread. Added `afterCompletion` to clear it.

### Verification

`mvn compiler:compile` passes (`BUILD SUCCESS`) after the changes.

### Diff

3 files changed, +16 / -6.

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/interceptor/AuthInterceptor.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/interceptor/AuthInterceptor.java
@@ -20,6 +20,7 @@ package org.apache.rocketmq.dashboard.interceptor;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import org.apache.rocketmq.dashboard.service.LoginService;
+import org.apache.rocketmq.dashboard.util.UserInfoContext;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.springframework.web.servlet.HandlerInterceptor;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/rocketmq/dashboard/interceptor/AuthInterceptor.java
```diff
@@ -20,6 +20,7 @@
 import jakarta.servlet.http.HttpServletRequest;
```
RockteMQ-AI, line 20: No test changes detected alongside source modifications. Consider adding tests to cover the changes.

RockteMQ-AI (commented):
### Summary
This PR modifies 3 file(s) with 70 lines of diff. No test changes detected — consider adding test coverage.

---

lizhimins:
Closing: based on the legacy src/ layout, and the functionality is covered by the rewritten studio implementation. ThreadLocal cleanup, message paging and broker stats hardening are all covered by the rewritten studio implementation (AuthInterceptor double-clear, DB-based paging, per-broker try/catch).

PR #710, opened 2026-07-31
[Studio] Synchronize consumer group cache and handle empty brokers

yyqdbngt:
### Motivation

Two robustness fixes in `ConsumerServiceImpl`:

1. **`consumerGroupMap` is not thread-safe**
   The map is a plain `HashMap` but is read/written concurrently: `queryGroupList`/`makeGroupListCache` mutate it (on request threads), while `refreshAllGroup` calls `consumerGroupMap.clear()` without holding the same lock and `refreshGroup` reads it directly. This can trigger `ConcurrentModificationException` or lost updates. Changed it to a `ConcurrentHashMap`.

2. **NPE when the cluster has no brokers**
   In `makeGroupListCache`, `subscriptionGroupWrapper` starts as `null` and is only assigned inside the per-broker loop. When the broker table is empty the wrapper stays `null`, so `subscriptionGroupWrapper.getSubscriptionGroupTable()` throws a `NullPointerException`. The empty/`null` check is now `subscriptionGroupWrapper == null || ...isEmpty()` and returns early instead.

### Verification

### Diff

1 file changed, +5 / -2.

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/service/impl/ConsumerServiceImpl.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/service/impl/ConsumerServiceImpl.java
@@ -70,6 +70,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/rocketmq/dashboard/service/impl/ConsumerServiceImpl.java
```diff
@@ -70,6 +70,7 @@
 import java.util.List;
```
RockteMQ-AI, line 70: No test changes detected alongside source modifications. Consider adding tests to cover the changes.

RockteMQ-AI (commented):
### Summary
This PR modifies 1 file(s) with 32 lines of diff. No test changes detected — consider adding test coverage.

---

lizhimins:
Closing: based on the legacy src/ layout, and the functionality is covered by the rewritten studio implementation. Group listing now goes through MyBatis DB queries (no shared in-memory cache) and DashboardProvider already null-checks both paths.

PR #730, opened 2026-08-01
[studio][ISSUE #727] Keep cluster config updates atomic on repository failures

### LINKED ISSUE #727: [Studio][Bug] Failed cluster config updates mutate stored state
### Baseline

`rocketmq-studio` at `bbf1b7e0cf25a5065ba049b5450cc8155569f710`.

### Problem

`ClusterService.updateClusterConfig` mutates the `ClusterVO` and `ClusterConfigVO` returned by `ClusterRepository.findById` before it calls `ClusterRepository.updateConfig`. Those objects are repository-owned live state in the current in-memory implementation.

If `updateConfig` throws, the API call fails but the failed request has already changed the stored object:

- an existing config is partially overwritten in place; and
- a cluster whose config was `null` retains the newly constructed config object.

A later read can therefore expose values from an update that was reported as failed.

### Deterministic reproduction

Two `ClusterServiceTest` cases use a synchronous Mockito `doThrow` from `ClusterRepository.updateConfig` and inspect the same object returned by `findById` after the exception:

1. Start with `flushDiskType=ASYNC_FLUSH` and `writeQueueNums=8`, request `SYNC_FLUSH` and `16`, and make persistence throw. The stored config contains `SYNC_FLUSH` and `16` after the failed call.
2. Start with a `null` config, request `ASYNC_FLUSH`, and make persistence throw. The stored cluster contains a non-null config after the failed call.

The tests contain no sleeps, timers, network access, or concurrent scheduling. In Java 21 they failed identically in five consecutive runs:

```text
Tests run: 2, Failures: 2, Errors: 0, Skipped: 0

expected: ASYNC_FLUSH
 but was: SYNC_FLUSH

expected: null
 but was: ClusterConfigVO(... flushDiskType=ASYNC_FLUSH ...)
```

Command:

```bash
mvn -Dtest='ClusterServiceTest#updateConfigShouldNotMutateStoredConfigWhenRepositoryUpdateFails+updateConfigShouldLeaveNullStoredConfigWhenRepositoryUpdateFails' test
```

### Expected behavior

If repository persistence fails, the cluster and its existing config should remain exactly as they were before the request. Successful partial-update behavior and the public API should remain unchanged.

### Root cause

The service uses `cluster.getConfig()` directly, applies every requested setter to that live object, and calls `cluster.setConfig(config)` before the repository update. The repository boundary therefore occurs after the observable mutation rather than before it.

### Proposed scope

- Build a detached copy of the existing `ClusterConfigVO` (or a detached empty config when it is null).
- Apply the partial update to that detached copy.
- Call `clusterRepository.updateConfig` before installing the new config on the returned `ClusterVO`.
- Install/return the new config only after the repository call succeeds.
- Add regression coverage for both existing and null configs when the repository throws.

The focused change should only need:

- `server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java`
- `server/src/test/java/org/apache/rocketmq/studio/cluster/broker/ClusterServiceTest.java`

### Duplicate search

I searched open and closed issues and open, closed, and merged pull requests for `ClusterService`, `updateClusterConfig`, cluster config persistence/update failures, atomicity, repository failures, live-object mutation, and the expected file names. I found no report or patch for this failure-atomicity behavior.

Related work is distinct:

- #645 / the superseded #646 and #600 validate `flushDiskType` and stabilize cluster identity/order; they do not handle repository exceptions or detach config state.
- #678 / #679 add numeric request validation before the service is invoked.
- #680 / #681 fix frontend polling and mock-store update behavior.
- #457 introduced the cluster service and repository code.

ai-yang:
Fixes #727

## What changed

`ClusterService.updateClusterConfig` now applies partial updates to a detached copy of the current cluster configuration. The repository update occurs before the copy is installed on the returned cluster object.

This prevents a failed repository call from leaking rejected values into repository-owned live state. The copy includes every field currently exposed by `ClusterConfigVO`, so successful partial updates continue to preserve unspecified values.

## Validation

- Two failure-path regression tests: 20/20 consecutive runs on Java 21
- Complete `ClusterServiceTest`: 22/22 tests passed
- Complete server test suite: 488/488 tests passed
- `mvn package`: passed
- Maven Checkstyle: passed
- `git diff --check`: passed

The project does not configure SpotBugs/FindBugs in its POM.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -59,10 +59,7 @@ public class ClusterService {
         ClusterVO cluster = clusterRepository.findById(command.getId())
                 .orElseThrow(() -> new BusinessException(404, "Cluster not found: " + command.getId()));
 
-        ClusterConfigVO config = cluster.getConfig();
-        if (config == null) {
-            config = new ClusterConfigVO();
-        }
+        ClusterConfigVO config = copyConfig(cluster.getConfig());
 
         if (command.getFlushDiskType() != null) {
             config.setFlushDiskType(parseFlushDiskType(command.getFlushDiskType()));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Ensures cluster config updates are atomic — the in-memory `ClusterConfigVO` is only mutated after `clusterRepository.updateConfig` succeeds. A deep copy (`copyConfig`) prevents the stored config from being modified on failed persistence. The tests thoroughly verify both the happy path and failure scenarios. LGTM.

---

PR #748, opened 2026-08-02
[Studio] Skip topics without collected data

123123213weqw:
### Problem

`DashboardServiceImpl.queryTopicCurrentData` iterated the topic cache and, for every entry, called:

```java
result.add(entry.getKey() + "," + value.get(value.size() - 1).split(",")[4]);
```

A topic that has **no collected data points yet** maps to an empty list, so `value.size() - 1` is `-1` and `value.get(-1)` throws `IndexOutOfBoundsException`. Because there is no try/catch, the entire `queryTopicCurrentData` endpoint fails as soon as one topic has an empty data list.

### Fix

Skip entries whose value list is `null` or empty before indexing the last element.

### Verification

### Diff

1 file changed, +4.


RockteMQ-AI (approved):
LGTM. Changes look good.

---

lizhimins:
Closing: based on the legacy src/ layout, and the functionality is covered by the rewritten studio implementation. The trend file cache no longer exists; the dashboard collects in real time and guards null values.

PR #758, opened 2026-08-02
[studio] fix: reject ACL rule updates for unknown IDs

### LINKED ISSUE #757: [Studio][Bug] ACL rule updates create records for unknown IDs
## Affected branch

`rocketmq-studio` at `bbf1b7e0cf25a5065ba049b5450cc8155569f710`

## Problem

`POST /api/acl/rules/update` creates a new ACL rule when the supplied ID does not exist.

`AclService.updateRule` only checks that the ID is non-blank, assigns `createdAt` when it is absent, and calls `AclRepository.saveRule`. `InMemoryAclRepository.saveRule` uses `ConcurrentHashMap.put`, so the update path has upsert semantics instead of update semantics.

This is inconsistent with other Studio update endpoints, including ACL user updates, alert rule updates, and data source updates, which return `404` for unknown IDs.

## Deterministic reproduction

1. Create an empty `InMemoryAclRepository` and an `AclService` backed by it.
2. Call `updateRule` with `id = "missing-rule"` and otherwise valid rule fields.
3. Query all rules.

Expected:

- `updateRule` throws `BusinessException(404, "ACL rule not found: missing-rule")`.
- The repository remains empty.

Actual:

- The call succeeds.
- A new rule with ID `missing-rule` appears in the repository.

A regression test using this sequence failed 5/5 on the unmodified branch under Java 21, without sleeps or randomized concurrency:

```text
AclServiceTest#updateRuleShouldRejectUnknownIdInsteadOfCreatingRule
Expecting code to raise a throwable.
```

## Impact

A stale client, typo, or retried edit can silently create a new ACL rule instead of reporting that the original rule no longer exists. It also bypasses the intended create endpoint and its generated-ID semantics.

## Suggested fix

- Give the repository an atomic replace operation (`ConcurrentHashMap.replace` or `computeIfPresent`) rather than implementing update as `put`.
- Return `404` when no entry was replaced.
- Preserve the original creation timestamp and keep create/update validation consistent.
- Add service and controller regression coverage for unknown IDs.

## Related work checked

- #516 introduced the ACL update endpoint.
- #636 validates ACL create/delete requests but does not reject unknown update IDs.
- #622 and #631 only harden frontend/mock copy behavior.

Searches across open and closed issues and pull requests found no existing fix for the unknown-ID upsert behavior.

ai-yang:
## Summary

- give ACL rule updates atomic replace semantics instead of unconditional upsert semantics
- return `404` when the requested rule ID no longer exists
- preserve the stored creation timestamp during a successful replacement without mutating the request object
- cover the service/repository behavior and the controller error response

Fixes #757.

## Root cause

`AclService.updateRule` delegated to the same `saveRule` operation as the create path. The in-memory repository implements that operation with `ConcurrentHashMap.put`, so an update for an unknown ID silently inserted a new rule and returned success. The replacement now uses `computeIfPresent` and publishes a detached value while retaining the stored `createdAt`.

## Verification

- unmodified Java 21 baseline: deterministic regression failed in 5/5 isolated Maven processes
- fixed targeted regressions: 20 isolated Java 21 Maven processes, 2/2 each (40/40)
- complete related test classes: 37/37
- complete backend suite: 489 tests, 0 failures, 0 errors, 0 skipped
- Maven package: passed
- Checkstyle: passed
- `git diff --check`: passed

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclRepository.java
@@ -25,6 +25,12 @@ public interface AclRepository {
 
     AclRuleVO saveRule(AclRuleVO rule);
 
+    /**
+     * Replaces an existing rule atomically without creating a missing rule.
+     * Implementations must preserve the original creation timestamp.
+     */
+    Optional<AclRuleVO> replaceRule(AclRuleVO rule);
+
     void deleteRule(String id);
 
     List<AclUserVO> findUsers();
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -61,10 +61,8 @@ public class AclService {
             throw new BusinessException(400, "ACL rule id is required");
         }
         log.info("Updating ACL rule id={}, principal={}", rule.getId(), rule.getPrincipal());
-        if (rule.getCreatedAt() == null) {
-            rule.setCreatedAt(LocalDateTime.now());
-        }
-        return aclRepository.saveRule(rule);
+        return aclRepository.replaceRule(rule)
+                .orElseThrow(() -> new BusinessException(404, "ACL rule not found: " + rule.getId()));
     }
 
     public void deleteRule(String id) {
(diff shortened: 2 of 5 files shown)


RockteMQ-AI (approved):
### Summary
Fixes ACL rule update to use atomic replace semantics instead of unconditional upsert. Returns 404 when the rule ID doesn't exist. Important bug fix.

### Findings
- **[Info]** `AclRepository.java:28-32` — Clean interface addition for `replaceRule` with proper semantics.
- **[Info]** `InMemoryAclRepository.java:52-68` — Good use of `computeIfPresent` for atomic replacement. Preserves original `createdAt` timestamp correctly.
- **[Info]** `AclService.java:64-65` — Clean error handling with proper 404 response.
- **[Info]** Tests — Comprehensive coverage including controller error response and service behavior.

### Overall
LGTM. Important correctness fix with clean implementation and proper test coverage.

---

PR #765, opened 2026-08-02
[studio] fix: refresh K8s certificate expiry status

### LINKED ISSUE #759: [Studio][Bug] K8s certificate expiry status becomes stale over time
## Affected branch

`rocketmq-studio` at `bbf1b7e0cf25a5065ba049b5450cc8155569f710`

## Problem

The K8s certificate API stores `status` and `daysRemaining` as fixed values and returns them unchanged from `K8sCertService.listCerts`.

Those fields are derived from `notAfter`, but they are calculated only when a certificate is created or renewed. A long-running Studio process therefore continues to report a certificate as `valid` after it enters the expiring window or has already expired.

## Deterministic reproduction

1. Make `K8sCertRepository.findAll()` return a certificate whose `notAfter` is in the past, while its stored `status` is `valid` and `daysRemaining` is positive.
2. Call `K8sCertService.listCerts()`.

Expected:

- The returned status is `expired`.
- `daysRemaining` is zero or negative.

Actual:

- The returned status remains `valid`.
- The stale positive `daysRemaining` value is returned unchanged.

A regression test with a fixed past expiry failed 5/5 on the unmodified branch under Java 21, without sleeps:

```text
K8sCertServiceTest#listCertsShouldRefreshTimeDerivedExpiryFields
expected: expired
 but was: valid
```

## Impact

The certificate page can hide an expired or soon-to-expire certificate and display an incorrect remaining lifetime until the process restarts or the certificate is renewed. Operators may miss certificate renewal deadlines.

## Suggested fix

- Inject a `Clock` into `K8sCertService`, following the existing `AuthService` testability pattern.
- Derive `daysRemaining` and `status` from `notAfter` when certificates are read (or refresh the derived fields through a defined lifecycle job).
- Return copies rather than mutating repository-owned objects during a read.
- Add fixed-clock tests for valid, expiring, and expired boundaries.

## Related work checked

- #578/#579 make certificate updates atomic but do not refresh time-derived fields.
- #644 validates certificate requests.
- #704 concerns fake seed records and is separate from expiry calculation.
- Open PR #713 changes only frontend display/filtering files and does not touch the backend status calculation.

Searches across open and closed issues and pull requests found no existing expiry/status refresh fix.

ai-yang:
## Summary

- derive certificate `status` and `daysRemaining` from the current time when certificates are listed or updated
- keep repository values detached by refreshing copied response objects
- use one injectable `Clock` for create, update, renew, and read-time expiry calculations
- preserve the existing 30-day expiring threshold and mutable list behavior

## Root cause

`K8sCertService` calculated time-derived fields only when a certificate was created or renewed. A long-running Studio instance therefore kept returning the persisted historical status even after the certificate crossed the expiring or expired boundary.

The service now refreshes these derived fields at the read/update boundary without mutating stored objects returned by the repository.

## Validation

- unmodified Java 21 baseline: deterministic regression failed in 5/5 isolated Maven processes
- fixed list/update expiry regressions: 20 isolated Java 21 Maven processes, 2/2 each (40/40)
- `K8sCertServiceTest`: 15/15
- `K8sCertControllerTest`: 9/9
- complete server suite: 488/488
- Maven package: 488/488
- Maven validate and Checkstyle: passed
- `git diff --check`: passed

Fixes #759

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java
@@ -19,31 +19,49 @@ package org.apache.rocketmq.studio.cluster.k8s;
 import org.apache.rocketmq.studio.common.domain.enums.CertStatus;
 import org.apache.rocketmq.studio.common.domain.enums.CertType;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
-import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.time.Clock;
 import java.time.LocalDateTime;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 @Slf4j
 @Service
-@RequiredArgsConstructor
 public class K8sCertService {
 
+    private static final int EXPIRING_THRESHOLD_DAYS = 30;
+
     private final K8sCertRepository k8sCertRepository;
+    private final Clock clock;
+
+    @Autowired
+    public K8sCertService(K8sCertRepository k8sCertRepository) {
+        this(k8sCertRepository, Clock.systemDefaultZone());
+    }
+
+    K8sCertService(K8sCertRepository k8sCertRepository, Clock clock) {
+        this.k8sCertRepository = k8sCertRepository;
+        this.clock = clock;
+    }
 
     public List<K8sCertVO> listCerts() {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes the root cause where certificate expiry status wasn't being refreshed. The fix introduces an injectable `Clock` for testability and calls `refreshExpirationState()` on every list/update/read operation to ensure the status and `daysRemaining` are always current.

Well-structured bug fix with proper test updates — LGTM.

### Findings
- **[Info]** `K8sCertService.java:148` — `refreshExpirationState` creates a copy via `copyOf()` on every call. This is correct for avoiding side effects on repository objects. If performance becomes a concern, this could be optimized with lazy evaluation, but it's fine for now.
- **[Info]** `K8sCertService.java:28` — The `EXPIRING_THRESHOLD_DAYS` constant is well-named and documented. Good practice.

---

PR #776, opened 2026-08-03
[ISSUE #775] Gate demo cluster seed data

### LINKED ISSUE #775: [Studio][Bug] Cluster repository seeds demo clusters by default
### Problem

On the rocketmq-studio branch, `ClusterRepositoryImpl` calls `initStubData()` from its default constructor. This means the backend always exposes demo clusters such as `rmq-cluster-prod` and `rmq-cluster-staging` even when the UI is running against the real API path.

That behavior can mislead users in Studio cluster and dashboard views because demo infrastructure looks like real discovered RocketMQ clusters.

### Evidence

- `server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java` calls `initStubData()` unconditionally in the constructor.
- The seeded records include fixed endpoints like `10.0.0.1:9876`, proxy addresses, broker TPS, topic counts, and group counts.

### Expected behavior

Demo seed data should be controlled by an explicit configuration flag and should be disabled by default. Tests or local demo setups can opt in, but production/runtime API responses should not contain demo clusters unless explicitly requested.

### Scope

This is in the RocketMQ Studio scope for ARCH-01 / multi-cluster management correctness. The proposed fix is a small repository hardening change, not a full persistent cluster registry implementation.

Aias00:
Fixes #775.

### What changed
- Add `studio.cluster.seed-demo-data`, defaulting to `false`.
- Only seed the in-memory cluster repository when the flag is explicitly enabled.
- Update repository tests to cover both explicit demo seeding and the default empty state.

### Validation
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=ClusterRepositoryImplTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=StudioApplicationTest test`
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java
@@ -25,6 +25,7 @@ import org.apache.rocketmq.studio.common.domain.enums.ClusterStatus;
 import org.apache.rocketmq.studio.common.domain.enums.ClusterType;
 import org.apache.rocketmq.studio.common.domain.enums.FlushDiskType;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Repository;
 
 import java.time.LocalDateTime;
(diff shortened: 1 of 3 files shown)


Aias00:
Superseded by consolidated PR #936, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #785, opened 2026-08-03
[ISSUE #784] Return unavailable for unimplemented cluster operations

### LINKED ISSUE #784: [Studio][Bug] Cluster management actions report success without executing real operations

- [x] I have confirmed this is not already covered by an existing issue or open pull request.

### Describe the Bug

The Studio cluster management APIs acknowledge several broker, NameServer, and proxy runtime operations as successful even though no real RocketMQ infrastructure operation is executed.

For example, the backend validates that a cluster or node exists, logs messages such as "restart initiated" or "deleted", and returns `true`/success. The corresponding UI then displays a success toast, which can make operators believe that a restart, upgrade, delete, or config change was applied to the real cluster.

Affected behavior includes:

- Broker restart
- NameServer create/update/restart/upgrade/delete
- Proxy restart

### Steps to Reproduce

1. Open the Studio cluster management page.
2. Trigger a broker restart, proxy restart, or NameServer operation for an existing mocked/configured node.
3. Observe that the API returns success and the UI displays a success notification.
4. No real broker, NameServer, or proxy operation is performed.

Until a real provider/executor is wired, unsupported operational actions should return an explicit unavailable/unsupported response so the UI does not report fake success.

### What Did You See Instead?

The backend returns success after local validation/logging only.

### Additional Context

This is part of the Studio control-plane baseline/ARCH-01 hardening: UI actions should be capability/provider driven and must not expose successful no-op operations.

Aias00:

Fixes #784

### Brief Description

This PR stops Studio cluster runtime operations from reporting fake success before a real execution provider is wired.

The affected service methods still validate cluster/node existence first, preserving existing 404 behavior. When the target exists, broker restart, NameServer create/update/restart/upgrade/delete, and proxy restart now return a `BusinessException(501, ...)` instead of logging a no-op and returning success.

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=ClusterServiceTest test`
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=ClusterControllerTest,NameServerControllerTest,ProxyControllerTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -26,7 +26,6 @@ import org.apache.rocketmq.studio.cluster.nameserver.UpdateNameServerDTO;
 import org.apache.rocketmq.studio.cluster.nameserver.UpgradeNameServerDTO;
 import org.apache.rocketmq.studio.cluster.proxy.RestartProxyDTO;
 
-import org.apache.rocketmq.studio.common.domain.enums.ClusterStatus;
 import org.apache.rocketmq.studio.common.domain.enums.FlushDiskType;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #945, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #787, opened 2026-08-03
[ISSUE #786] Return unavailable for consumer group providers

### LINKED ISSUE #786: [Studio][Bug] Consumer Group APIs return generic 500 when providers are unavailable

- [x] I have confirmed this is not already covered by an existing issue or open pull request.

### Describe the Bug

The Studio Consumer Group APIs are exposed by `ConsumerGroupController`, but the default provider/admin implementations behind those APIs still throw plain `UnsupportedOperationException` for Consumer Group operations.

Because `UnsupportedOperationException` is handled by the generic exception handler, the API returns a generic HTTP 500 response instead of an explicit provider-unavailable response. Opening or operating the Consumer page in real API mode can therefore fail with an opaque internal error.

Affected operations include:

- List consumer groups
- Get group progress
- Get group subscriptions
- Get consumer group detail
- Create/delete consumer group
- Reset consumer offset

### Steps to Reproduce

1. Run Studio with the default provider/admin implementation.
2. Call `/api/groups` or any Consumer Group detail/mutation endpoint.
3. Observe that the provider/admin throws `UnsupportedOperationException` and the client receives a generic 500 response.

Until real Consumer Group provider/admin support is wired, these APIs should return an explicit unavailable/unsupported business response, such as HTTP 501 with a clear message.

### What Did You See Instead?

A generic internal server error is returned.

### Additional Context

This is part of the Studio control-plane baseline hardening. Missing backend providers should be explicit capability gaps, not generic server failures.

Aias00:

Fixes #786

### Brief Description

The default Consumer Group metadata/admin implementations now return explicit `BusinessException(501, ...)` responses when the backing provider is not configured, instead of throwing plain `UnsupportedOperationException` and surfacing as generic 500 errors.

This keeps the current Topic paths untouched and only changes Consumer Group operations used by `/api/groups`.

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=CloudMetadataProviderTest,NameSrvAdminClientTest,MetadataServiceTest test`
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=ConsumerGroupControllerTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/topic/CloudMetadataProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/topic/CloudMetadataProvider.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.studio.instance.topic;
 
+import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.instance.group.ConsumerGroupVO;
 import org.apache.rocketmq.studio.instance.group.QueueProgressVO;
 import org.apache.rocketmq.studio.instance.group.SubscriptionEntryVO;
@@ -37,7 +38,7 @@ public class CloudMetadataProvider implements MetadataProvider {
 
     @Override
     public List<ConsumerGroupVO> listConsumerGroups(String clusterId, String search) {
-        throw new UnsupportedOperationException("Not implemented");
+        throw consumerGroupProviderUnavailable();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
﻿### Summary
This PR replaces `UnsupportedOperationException` with explicit 501 `BusinessException` responses for consumer group operations when the backing provider is not configured. Consistent with the pattern in the other Studio cleanup PRs.

### Verdict
LGTM. Clean change from generic `UnsupportedOperationException` to proper `BusinessException(501)` with descriptive messages. Tests comprehensively verify the new behavior.

---

PR #801, opened 2026-08-03
[ISSUE #795] Fail closed on LiteTopic capability errors

### LINKED ISSUE #795: [Studio][Bug] LiteTopic page assumes support when capability probe fails
### Problem

The LiteTopic page currently assumes LiteTopic is supported when the capability probe fails.

Current behavior in `web/src/pages/studio/LiteTopic.tsx`:

```ts
try {
  const capability = await queryLiteTopicCapability();
  ...
} catch {
  // Assume supported when capability detection is unavailable.
}

await fetchData(...)
```

If `/api/liteTopic/capability` is unavailable or returns an error, the page continues to query LiteTopic list/quota data and renders the management surface as if the feature is available.

### Expected behavior

Capability detection should fail closed. If Studio cannot confirm LiteTopic support, the page should stop loading LiteTopic data and show an explicit unavailable/unsupported state instead of enabling the management UI.

### Impact

This avoids misleading users into thinking LiteTopic management is available when the backend/provider path is unhealthy or not implemented yet. It is related to META-01 LiteTopic management in the RocketMQ Studio scope.

### Suggested fix

- On capability probe failure, clear LiteTopic data, stop loading, and render the unavailable state.
- Add a regression test proving list/quota APIs are not called after capability detection fails.

Aias00:
### What changed

This PR makes the LiteTopic page fail closed when capability detection is unavailable.

Previously, if `/api/liteTopic/capability` failed, the page assumed LiteTopic support and continued to load list/quota data. Now it clears current data, stops loading, and renders the unavailable state instead.

### Why

The UI should not expose LiteTopic management when Studio cannot confirm that the backend/provider path supports it. This avoids misleading users with an enabled management surface during backend/provider failures.

Fixes #795.

### Verification

```bash
npm test -- --run src/pages/studio/__tests__/LiteTopic.test.tsx
```

Result: 1 test file passed, 9 tests passed.

### DIFF
--- a/web/src/pages/studio/LiteTopic.tsx
+++ b/web/src/pages/studio/LiteTopic.tsx
@@ -204,7 +204,12 @@ const LiteTopicPage: React.FC = () => {
         }
       } catch {
         if (!bootstrapIsActive()) return;
-        // Assume supported when capability detection is unavailable.
+        ++displayCounter.current;
+        setCapabilitySupported(false);
+        setTopicList([]);
+        setQuota(null);
+        setLoading(false);
+        return;
       }
 
       if (displayCounter.current === initialDisplayRequestId) {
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #932, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #803, opened 2026-08-03
[ISSUE #802] Disable mock mode for production builds

### LINKED ISSUE #802: [Studio][Bug] Production web builds enable mock data by default
### Problem

The Studio web production env file currently enables mock mode by default:

```env
# Backend not ready yet - use mock data
# Set to "false" when real API is available
VITE_USE_MOCK=true
```

This means production-like builds can ship with `USE_MOCK=true`, so the UI may read and mutate in-memory mock data instead of calling the backend.

### Expected behavior

Production builds should default to real backend APIs. Mock mode should remain opt-in for local development or demos, not the production default.

### Impact

With mock mode enabled in production, users can see simulated resources and successful control-plane actions without real backend interaction. This is risky for the RocketMQ Studio control-plane UX because it blurs real operations and demo data.

### Suggested fix

- Set `web/.env.production` to `VITE_USE_MOCK=false` by default.
- Keep local development mock mode configurable through `web/.env` or explicit deployment overrides.

Aias00:
### What changed

This PR changes `web/.env.production` so production builds use real backend APIs by default:

```env
VITE_USE_MOCK=false
```

Local development mock mode remains available through `web/.env` or explicit deployment overrides.

### Why

Production-like builds should not default to mock data. Keeping mock mode enabled in production can show simulated resources and successful control-plane actions without backend interaction.

Fixes #802.

### Verification

```bash
npm run build
```

Result: build completed successfully. Vite reported only the existing large chunk warning.

### DIFF
--- a/web/.env.production
+++ b/web/.env.production
@@ -1,3 +1,3 @@
-# Backend not ready yet - use mock data
-# Set to "false" when real API is available
-VITE_USE_MOCK=true
+# Production builds should call real backend APIs by default.
+# Set to "true" only for explicit demo/mock deployments.
+VITE_USE_MOCK=false


Aias00:
Superseded by consolidated PR #936, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #807, opened 2026-08-03
[ISSUE #806] Align Tongyi LLM provider key

### LINKED ISSUE #806: [Bug] Settings page uses qwen LLM provider key while backend expects tongyi

### Runtime platform environment

N/A

### RocketMQ version

rocketmq-studio branch

### JDK Version

N/A

### Describe the Bug

The legacy Settings page uses `qwen` as the provider value for Tongyi Qwen:

```tsx
{ value: 'qwen', label: '通义千问' }
```

However, the backend LLM config service and the newer LLM settings page use `tongyi` as the provider key. `LlmConfigService.normalizeProvider()` only accepts keys present in its provider map, so `qwen` is treated as an unknown provider and normalized to `openai`.

This can silently save or load the wrong provider after selecting Tongyi from the Settings page.

### Steps to Reproduce

1. Open the Settings page.
2. Select `通义千问` as the LLM provider.
3. Save or test the settings.
4. Observe that the frontend submits `qwen`, while the backend provider map expects `tongyi`.

All Studio LLM settings surfaces should use the same provider key for Tongyi Qwen. Existing `qwen` values should be normalized to `tongyi` for compatibility.

### What Did You See Instead?

The Settings page emits `qwen`, and the backend normalizer can fall back to `openai`.

### Additional Context

This is in the AI Native / LLM configuration scope.

Aias00:
## Track

Track 3 / AI Native - LLM integration

### What is changed

- Align the legacy settings page Tongyi provider value with the backend-supported `tongyi` key.
- Keep `qwen` as a backend compatibility alias so existing saved configs do not silently fall back to OpenAI.
- Add regression coverage for loading and saving the legacy `qwen` provider value.

### Why

The old Settings page used `qwen`, while the Studio LLM settings and backend provider registry use `tongyi`. Saving from the old page could make the backend normalize the provider to `openai`, which breaks model/base URL defaults.

Fixes #806

### Verification

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=LlmConfigServiceTest test`
- `npm run build`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
@@ -240,6 +240,9 @@ public class LlmConfigService {
 
     private String normalizeProvider(String provider) {
         String normalized = defaultString(provider, OPENAI).toLowerCase(Locale.ROOT);
+        if ("qwen".equals(normalized)) {
+            return "tongyi";
+        }
         return PROVIDER_MODELS.containsKey(normalized) ? normalized : OPENAI;
     }
 
--- a/server/src/test/java/org/apache/rocketmq/studio/ops/ai/LlmConfigServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/ops/ai/LlmConfigServiceTest.java
@@ -71,6 +71,28 @@ class LlmConfigServiceTest {
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


Aias00:
Superseded by consolidated PR #940, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #820, opened 2026-08-03
[ISSUE #819] Redact user password in toString

### LINKED ISSUE #819: [Bug] User.toString exposes password in logs
### Bug Description
`server/src/main/java/org/apache/rocketmq/studio/model/User.java` includes the raw `password` field in `toString()`. If a `User` instance is logged directly or appears in an exception/debug path, the plaintext password can be written to application logs.

### Expected Behavior
`toString()` should not include the raw password. It should either omit the field or render a masked value.

### Actual Behavior
The raw password value is concatenated into the string representation.

### Minimal Fix Direction
Redact the password in `User.toString()` and add a regression test that verifies the raw password is not present.

### Scope
This is a small security hardening item related to Studio login/permission handling.

Aias00:
## What is changed
- Redact the password field in User.toString() so logs/debug output cannot expose plaintext credentials.
- Add a regression test to keep the toString output from leaking the original password.

## Why
- Fixes #819.

## Verification
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=UserTest test

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/model/User.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/model/User.java
@@ -76,7 +76,7 @@ public class User {
         return "User{" +
                 "id=" + id +
                 ", name='" + name + '\'' +
-                ", password='" + password + '\'' +
+                ", password='******'" +
                 ", type=" + type +
                 '}';
     }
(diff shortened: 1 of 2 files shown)


Aias00:

RockteMQ-AI (approved):
### Summary
Redacts user password from `User.toString()` to prevent plaintext credential leakage in logs. Consistent with the other credential redaction PRs.

**Verdict:** LGTM — clean fix with regression test.

---

PR #828, opened 2026-08-03
[ISSUE #827] Redact ACL user credentials in toString

### LINKED ISSUE #827: [Bug] ACL user VO exposes credentials in toString
## Problem

`AclUserVO` uses Lombok `@Data` and carries generated ACL credentials (`accessKey` and `secretKey`). Lombok includes all fields in the generated `toString()` output by default, so accidental VO logging or debug output can expose ACL credentials.

## Evidence

- `server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclUserVO.java` declares `@Data`.
- The VO contains `accessKey` and `secretKey` fields.
- `AclService#createUser` generates and stores real values for both fields before returning the saved VO.

## Expected behavior

Credential fields should be excluded from generated diagnostic output while non-sensitive ACL user metadata remains visible.

## Scope

Small hardening fix for the Studio ACL user management path.

Aias00:
## What is changed
- Exclude ACL access key and secret key fields from Lombok-generated toString output.
- Add a regression test that verifies ACL credentials are not printed while non-sensitive metadata remains visible.

## Why
- Fixes #827.

## Verification
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=AclUserVOTest test

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/acl/AclUserVOTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/acl/AclUserVOTest.java
@@ -16,27 +16,30 @@
  */
 package org.apache.rocketmq.studio.instance.acl;
 
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.NoArgsConstructor;
-import lombok.ToString;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


Aias00:

PR #836, opened 2026-08-03
[ISSUE #835] Make audit ordering null-safe

### LINKED ISSUE #835: [Bug] Audit log ordering is unstable when records have missing timestamps
### What happened?

InMemoryAuditRepository sorts audit records by timestamp descending, but the comparator returns 0 whenever either side has a null timestamp.

That means a record without timestamp can keep its current position relative to newer timestamped records, so the audit log order becomes dependent on ConcurrentHashMap iteration order instead of the expected newest-first order.

### Expected behavior

Audit records with timestamps should be ordered by timestamp descending. Records without timestamps should be placed after timestamped records with deterministic behavior.

### Why this matters

This affects the Studio operations audit view under BASE-01 login and permission auditing. Audit logs should be predictable when mixed data or partially populated records exist.

### Candidate fix

Use a null-safe timestamp comparator and add repository coverage for mixed timestamp/null timestamp records.

Aias00:
### Summary
- Sort audit records by timestamp descending with null timestamps last
- Add an id tie-breaker so in-memory ordering is deterministic
- Add repository coverage for mixed timestamp and missing timestamp records

### Tests
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=InMemoryAuditRepositoryTest test

Closes #835

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/audit/InMemoryAuditRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/audit/InMemoryAuditRepository.java
@@ -20,6 +20,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 
 import java.time.LocalDateTime;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


lizhimins:
感谢贡献。该 PR 修复的目标 `InMemoryAuditRepository` 已在 rocketmq-studio 分支移除（#955 将所有内存仓储替换为 MySQL 实现）：审计记录现持久化到 `rmq_operation_audit` 表，排序由 SQL `ORDER BY operated_at DESC` 完成，本 PR 针对的空指针排序场景已不存在，因此关闭。如有其他场景欢迎重新提交。

---

Thanks for the contribution. The target of this fix, `InMemoryAuditRepository`, has been removed from the rocketmq-studio branch (#955 replaced all in-memory repositories with MySQL-backed ones). Audit records are now persisted to the `rmq_operation_audit` table and ordered via SQL `ORDER BY operated_at DESC`, so the null-timestamp ordering scenario this PR addresses no longer exists. Closing accordingly — feel free to reopen in a new PR if another scenario applies.

PR #839, opened 2026-08-03
[ISSUE #838] Redact settings JSON from toString

### LINKED ISSUE #838: [Bug] RmqSettings toString can expose persisted settings JSON
### What happened?

RmqSettings uses Lombok Data, so its generated toString includes the json field. That field stores the persisted settings snapshot and may contain sensitive values such as LLM API keys before JSON write-only filtering is applied at the REST boundary.

If an entity instance is logged or included in diagnostics, the raw settings JSON can be exposed.

### Expected behavior

RmqSettings.toString should not include the json field. It can still expose safe metadata such as id and updatedAt.

### Why this matters

This is part of the Studio AI Native and settings management surface. AI provider credentials should not leak through incidental logging.

### Candidate fix

Add Lombok ToString.Exclude to RmqSettings.json and cover the toString behavior with a small unit test.

Aias00:
### Summary
- Exclude persisted settings JSON from RmqSettings.toString
- Keep safe metadata such as id and updatedAt visible
- Add unit coverage that API keys embedded in the JSON are not exposed

### Tests
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=RmqSettingsTest test

Closes #838

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/persistence/entity/RmqSettingsTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/persistence/entity/RmqSettingsTest.java
@@ -16,23 +16,26 @@
  */
 package org.apache.rocketmq.studio.persistence.entity;
 
-import com.baomidou.mybatisplus.annotation.IdType;
-import com.baomidou.mybatisplus.annotation.TableId;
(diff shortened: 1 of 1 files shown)


Aias00:

PR #861, opened 2026-08-03
[ISSUE #860] Validate null cluster config requests

### LINKED ISSUE #860: [Bug] Cluster config update should reject null request bodies
### What happened?

`POST /api/clusters/config/update` accepts the JSON literal `null` as the request body. Spring can deserialize that into a null DTO and the controller currently forwards it to `ClusterService.updateClusterConfig`, which can surface an internal null-handling error instead of a clear client error.

### Why this matters

Cluster configuration is part of the RocketMQ Studio control-plane workflow. Invalid write requests should be rejected at the API boundary with a stable 400 response so callers do not accidentally trigger service-layer null dereferences.

### Expected behavior

The endpoint should reject a null request body with HTTP 400 and should not call the service layer.

### Scope

RocketMQ Studio track 1 / ARCH-01 cluster configuration management hardening.

Aias00:
### Summary

- reject JSON `null` bodies for the cluster config update API at the controller boundary
- keep existing field validation behavior unchanged
- add WebMvc coverage to ensure null requests return 400 without invoking the service layer

### Tests

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=ClusterControllerTest test`

Closes #860

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterController.java
@@ -19,6 +19,7 @@ package org.apache.rocketmq.studio.cluster.broker;
 import org.apache.rocketmq.studio.cluster.config.UpdateConfigDTO;
 
 import org.apache.rocketmq.studio.common.domain.Result;
+import org.apache.rocketmq.studio.common.exception.BusinessException;
 import jakarta.validation.Valid;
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #929, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #864, opened 2026-08-03
fix(web): keep LLM connection test read-only

### LINKED ISSUE #863: [Bug] LLM connection test persists and enables configuration
### What happened?

On the Studio LLM settings page, the connection test action is expected to be a non-destructive probe. However, `handleTestConnection()` builds the tested config with `forceEnabled=true` and saves it after a successful test.

This means clicking **Test Connection** can persist runtime configuration and enable the LLM gateway even when the enable switch was left off.

### Expected behavior

Testing a provider should not change persisted configuration or the enabled state. The page should only save configuration when the user explicitly clicks save.

### Scope

RocketMQ Studio AI Native / LLM settings.

### Suggested fix

Keep the test request ephemeral, preserve the current enabled value, and avoid calling the save API from the test action.

Aias00:
## Track

Track 3 / AI Native - LLM integration

### Summary

- keep the LLM connection test as a read-only probe
- stop auto-saving the tested config after a successful test
- preserve the current enabled state instead of forcing test payloads enabled
- add regression coverage for disabled configs and stale provider requests

### Related Issue

Fixes #863

### Tests

- `npm test -- --run src/pages/studio/__tests__/LlmSettingsAsyncState.test.tsx`
- `npm run lint -- src/pages/studio/LlmSettings.tsx src/pages/studio/__tests__/LlmSettingsAsyncState.test.tsx` (passes with existing fast-refresh warnings)

### DIFF
--- a/web/src/pages/studio/LlmSettings.tsx
+++ b/web/src/pages/studio/LlmSettings.tsx
@@ -387,31 +387,17 @@ const LlmSettingsPage: React.FC = () => {
   };
 
   const handleTestConnection = () => {
-    const requestProviderGeneration = providerInteractionGeneration.current;
     setTestLoading(true);
     setTestResult(null);
     form
       .validateFields()
       .then((values) => {
-        const testConfig = buildConfig(values, true);
+        const testConfig = buildConfig(values, enabled);
         testLlmConnection(testConfig)
           .then((result) => {
             if (result && result.status === 0) {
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #940, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #866, opened 2026-08-03
fix(web): show topic test message send failures

### LINKED ISSUE #865: [Bug] Topic test message send failures are silently ignored
### What happened?

On the RocketMQ Studio Topic page, `handleSend` wraps both form validation and `sendTopicMessage(...)` in the same `try/catch`. The catch block treats every exception as a validation error and does not show any error message.

As a result, when the backend send-message API fails, the dialog stays open but the user receives no failure feedback.

### Expected behavior

Form validation failures should stay silent or show validation errors through the form, but API failures from sending a test message should show an explicit error message.

### Scope

RocketMQ Studio / Topic management / BASE-01 test-message workflow.

### Suggested fix

Separate form validation from the send API call, and show a user-facing error when `sendTopicMessage(...)` rejects.

Aias00:
### Summary

- separate Topic send-message form validation from the send API call
- keep validation failures local to the form
- show an explicit error when `sendTopicMessage(...)` rejects
- add regression coverage for failed test-message sends

### Related Issue

Fixes #865

### Tests

- `npm test -- --run src/pages/instance/__tests__/TopicPage.test.tsx`
- `npm run lint -- src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx` (passes with existing fast-refresh warnings)

### DIFF
--- a/web/src/pages/instance/__tests__/TopicPage.test.tsx
+++ b/web/src/pages/instance/__tests__/TopicPage.test.tsx
@@ -18,7 +18,7 @@
 import { describe, it, expect, vi, beforeAll, beforeEach, afterEach } from 'vitest';
 import { render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { App } from 'antd';
+import { App, message } from 'antd';
 import { LangProvider } from '../../../i18n/LangContext';
 import type { Topic } from '../../../api/metadata';
 import TopicPage from '../topic';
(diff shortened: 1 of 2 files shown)


lizhimins:
感谢贡献。发送测试消息失败时的错误提示已由主干的实例持久化改造（#951）覆盖：当前 `handleSend` 在 catch 中已展示错误提示，且支持 props 文本/键值双模式、连续发送时保留弹窗，功能上已包含本 PR 的诉求，因此关闭。

---

Thanks for the contribution. Showing topic test-message send failures is already covered by the trunk's instance-persistence work (#951): `handleSend` now surfaces an error message in the catch branch, and additionally supports both text/KV property modes and keeps the dialog open for consecutive sends. The intent of this PR is already included, so closing accordingly.

PR #868, opened 2026-08-03
fix(server): validate consumer group create requests

### LINKED ISSUE #867: ConsumerGroup create endpoint should validate request payload
### Problem

The Studio ConsumerGroup create endpoint accepts the request body without bean validation. `ConsumerGroupController#createConsumerGroup` does not annotate the body with `@Valid`, and `ConsumerGroupVO` does not declare basic constraints for fields that are required during creation, such as the group name.

As a result, malformed requests such as an empty group name can pass the controller boundary and reach the metadata service. This makes the API behavior inconsistent with other write endpoints and can surface as less actionable downstream errors.

### Scope

This is in the RocketMQ Studio control-plane scope, specifically the ConsumerGroup management baseline.

### Expected behavior

- Validate ConsumerGroup create requests at the controller boundary.
- Reject blank group names before calling the metadata service.
- Keep read/list response fields optional so existing UI responses are not affected.
- Add regression coverage for invalid create payloads.

### Evidence

- `server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java`
- `server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupVO.java`

Aias00:
### Summary

- add a create-specific ConsumerGroup DTO with request validation
- reject blank group names and negative retry/delay values at the controller boundary
- keep the existing ConsumerGroupVO response model optional for list/detail responses
- add MockMvc regression coverage for valid and invalid create requests

### Related Issue

Fixes #867

### Tests

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=ConsumerGroupControllerTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
@@ -70,8 +70,8 @@ public class ConsumerGroupController {
     }
 
     @PostMapping("/create")
-    public Result<ConsumerGroupVO> createConsumerGroup(@RequestBody ConsumerGroupVO group) {
-        return Result.ok(metadataService.createConsumerGroup(group));
+    public Result<ConsumerGroupVO> createConsumerGroup(@Valid @RequestBody CreateConsumerGroupDTO group) {
+        return Result.ok(metadataService.createConsumerGroup(group.toConsumerGroupVO()));
     }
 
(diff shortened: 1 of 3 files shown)


Aias00:
Superseded by consolidated PR #930, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #877, opened 2026-08-03
fix(web): stop simulating SSL settings saves

### LINKED ISSUE #876: [Bug] SSL settings page simulates successful saves

### Describe the Bug

The RocketMQ Studio SSL settings page still simulates a successful save with a local `setTimeout` and then updates only component state:

```tsx
setTimeout(() => {
  setLoading(false);
  message.success(t('ssl.saveSuccess'));
  setSslConfig({ ...sslConfig, ...values });
}, 1000);
```

There is no backend API call or persistence layer behind this action. Users can enable TLS/SSL in the UI and see a success message even though the server configuration is not changed.

### Expected Behavior

Until a real SSL configuration API exists, the page should not report a successful save. It should explicitly show that SSL settings persistence is unavailable instead of mutating local state as if the configuration was applied.

### Scope

RocketMQ Studio Track 1 / BASE-01 HTTPS/SSL management baseline. The fix should be small and avoid introducing a fake backend contract.

### Proposed Fix

Replace the simulated save path with a clear unavailable/error message and keep the current displayed config unchanged. Add a frontend regression test that verifies the success toast is not shown and the local settings are not mutated.

Aias00:
## Summary
- remove the fake setTimeout-based SSL settings save success path
- show an explicit unavailable message until a real backend SSL configuration API exists
- add a regression test that verifies SSL edits are not persisted locally after a failed save

Fixes #876

## Tests
- npm test -- --run src/pages/studio/__tests__/SslSettings.test.tsx
- npm run lint -- src/pages/studio/SslSettings.tsx src/pages/studio/__tests__/SslSettings.test.tsx src/i18n/translations.ts
- git diff --check

Note: lint reports only the existing react-refresh warnings in LangContext.tsx, main.tsx, and ThemeContext.tsx.

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -1057,6 +1057,10 @@ const translations: Record<string, Record<Lang, string>> = {
   'ssl.expiryDate': { zh: '过期日期', en: 'Expiry Date' },
   'ssl.active': { zh: '有效', en: 'Active' },
   'ssl.saveSuccess': { zh: 'SSL 配置保存成功', en: 'SSL configuration saved successfully' },
+  'ssl.saveUnavailable': {
+    zh: 'SSL 配置保存功能尚未接入真实后端接口',
+    en: 'SSL configuration persistence is not wired to a backend API yet',
+  },
   'ssl.invalidCertFormat': { zh: '仅允许证书文件！', en: 'Only certificate files are allowed!' },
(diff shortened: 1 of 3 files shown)


Aias00:
Superseded by consolidated PR #938, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #879, opened 2026-08-03
fix(server): normalize auth public paths

### LINKED ISSUE #878: [Bug] Auth public endpoints reject trailing slash variants

### Describe the Bug

`AuthInterceptor#isPublicPath` only allows exact matches for `/api/auth/login` and `/api/auth/status`:

```java
return path.equals("/api/auth/login")
        || path.equals("/api/auth/status")
        || ...;
```

When login is required, common gateway or proxy path variants such as `/api/auth/status/` are rejected with 401 even though they represent the same public endpoint. This can break frontend auth status checks behind path-normalizing proxies.

### Expected Behavior

The interceptor should normalize a single trailing slash for public path checks, while keeping the public endpoint allow-list strict. `/api/auth/status/` and `/api/auth/login/` should be treated the same as their canonical forms.

### Scope

RocketMQ Studio Track 1 / BASE-01 login and permission baseline. This is a small compatibility hardening in the Console authentication interceptor.

### Proposed Fix

Normalize the request path before public-path comparison by trimming trailing slashes except for `/`. Add regression tests for the login and status public endpoints with trailing slashes.

Aias00:
## Summary
- normalize trailing slashes before matching auth public endpoints
- keep the allow-list strict while allowing `/api/auth/login/` and `/api/auth/status/`
- add regression coverage for both trailing-slash public endpoints

Fixes #878

## Tests
- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=AuthInterceptorTest test
- git diff --check

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthInterceptor.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthInterceptor.java
@@ -50,11 +50,25 @@ public class AuthInterceptor implements HandlerInterceptor {
     }
 
     private boolean isPublicPath(String path) {
+        path = normalizePath(path);
         return path.equals("/api/auth/login")
                 || path.equals("/api/auth/status")
                 || path.startsWith("/api-docs")
                 || path.startsWith("/swagger-ui")
                 || path.startsWith("/actuator/health");
+    }
+
+    private String normalizePath(String path) {
+        if (path == null || path.isBlank()) {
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #947, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #896, opened 2026-08-04
[ISSUE #895] Parse broker TPS with flexible whitespace

### LINKED ISSUE #895: [Bug] Broker TPS parsing returns 0 when runtime stats contain extra whitespace
### Problem
`RocketMQClusterProvider` parses broker runtime TPS strings with `split(" ")`. RocketMQ runtime stats such as `putTps` / `getTransferedTps` are whitespace-separated values, and extra leading or repeated whitespace makes the first token empty. In that case the provider falls back to `0`, so Broker TPS can be displayed as zero even when runtime stats contain a valid value.

### Code evidence
- `server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java`: `parseFirstTpsValue` uses `tpsStr.split(" ")` and parses `parts[0]`.

### Impact
Cluster/Broker overview can show incorrect inbound/outbound TPS, which weakens the Studio control plane observability view.

### Expected behavior
TPS parsing should trim the runtime stat and split by one or more whitespace characters before parsing the first average value.

### Scope
RocketMQ Studio track 1 / Control Plane cluster observability.

Aias00:
## Summary
- parse broker runtime TPS stats after trimming and splitting by one or more whitespace characters
- add RocketMQClusterProvider coverage for leading tab/space runtime stat values

## Tests
- JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=RocketMQClusterProviderTest test
- git diff --check

Closes #895

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java
@@ -218,7 +218,7 @@ public class RocketMQClusterProvider implements ClusterProvider {
      */
     private int parseFirstTpsValue(String tpsStr) {
         try {
-            String[] parts = tpsStr.split(" ");
+            String[] parts = tpsStr.trim().split("\\s+");
             if (parts.length > 0) {
                 return (int) Double.parseDouble(parts[0]);
             }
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #948, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #898, opened 2026-08-04
[ISSUE #897] Report dashboard broker version

### LINKED ISSUE #897: [Bug] Dashboard cluster overview never reports broker version
### Problem
`RocketMQDashboardProvider` initializes each cluster overview version to `unknown`, then only assigns the broker runtime `brokerVersionDesc` when the current value is **not** `unknown`:

```java
String version = "unknown";
...
if (v != null && !"unknown".equals(version)) {
    version = v;
}
```

Because the initial value is `unknown`, the condition is false on the first runtime stat response, so the dashboard cluster overview keeps reporting `unknown` even when brokers expose `brokerVersionDesc`.

### Impact
The Studio dashboard cannot show the real RocketMQ broker version in the cluster overview, which weakens the Control Plane / dashboard baseline view.

### Expected behavior
When broker runtime stats contain `brokerVersionDesc`, the cluster overview should use that value. If no broker reports a version, it should keep `unknown`.

### Code evidence
- `server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDashboardProvider.java`

### Suggested fix
Assign `brokerVersionDesc` when it is non-blank and the current cluster version is still `unknown`, and add a provider test for the dashboard cluster overview version.

Aias00:
### What
- Fix `RocketMQDashboardProvider` so cluster overview uses the first non-blank `brokerVersionDesc` from broker runtime stats.
- Add provider coverage proving the dashboard cluster version is populated instead of staying `unknown`.

### Why
The previous condition only assigned the broker version when the current value was already not `unknown`, so the initial `unknown` value was never replaced.

### Validation
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=RocketMQDashboardProviderTest test`
- `git diff --check`

Closes #897

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDashboardProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDashboardProvider.java
@@ -164,8 +164,11 @@ public class RocketMQDashboardProvider implements DashboardProvider {
                                     clusterTpsIn += (int) parseTps(rt.getTable().get("putTps"));
                                     clusterTpsOut += (int) parseTps(rt.getTable().get("getTransferedTps"));
                                     String v = rt.getTable().get("brokerVersionDesc");
-                                    if (v != null && !"unknown".equals(version)) {
-                                        version = v;
+                                    if (v != null && "unknown".equals(version)) {
+                                        String brokerVersion = v.trim();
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #948, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #902, opened 2026-08-04
[ISSUE #901] Handle null topic queue lists

### LINKED ISSUE #901: [Bug] Topic message query treats null queue list as empty result
### What happened?

`RocketMQMessageProvider.queryByTopic` calls `DefaultMQPullConsumer.fetchSubscribeMessageQueues(topic)` and immediately iterates the returned set. If the RocketMQ client returns `null` for the queue set, the enhanced for-loop throws a `NullPointerException`. The catch block then logs the failure and returns an empty list.

That makes a provider/runtime failure look the same as a legitimate "no messages found" result in the Studio message query flow.

### Location

- `server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQMessageProvider.java`

### Expected behavior

The provider should handle a null queue set explicitly and return an empty result without relying on an exception path. This keeps the Topic + time range query path deterministic and avoids noisy/ambiguous failures.

### Scope

This is in the RocketMQ Studio BASE-01 message query baseline: the migrated message query capability should not degrade when the RocketMQ admin/pull client returns incomplete metadata.

### Suggested fix

Guard the queue set before iterating it and add a unit test for the null return case.

Aias00:
### Summary

- guard `RocketMQMessageProvider` topic range queries when `fetchSubscribeMessageQueues` returns null or empty
- add a provider unit test that verifies the null queue-set path returns an empty result without attempting pulls

### Validation

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=RocketMQMessageProviderTest test`
- `git diff --check`

Closes #901

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQMessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQMessageProvider.java
@@ -184,6 +184,9 @@ public class RocketMQMessageProvider implements MessageProvider {
         try {
             consumer.start();
             Set<MessageQueue> queues = consumer.fetchSubscribeMessageQueues(topic);
+            if (queues == null || queues.isEmpty()) {
+                return result;
+            }
             outer:
             for (MessageQueue queue : queues) {
                 long minOffset = consumer.searchOffset(queue, begin);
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #935, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #904, opened 2026-08-04
[ISSUE #903] Guard null DLQ queue lists

### LINKED ISSUE #903: [Bug] DLQ resend should guard null queue lists
### What happened?

`RocketMQDLQProvider.collectDeadLetters` calls `DefaultMQPullConsumer.fetchSubscribeMessageQueues(dlqTopic)` and immediately iterates the returned set. If the RocketMQ client returns `null`, the enhanced for-loop throws a `NullPointerException`. The catch block then handles that as a generic collection failure and returns an empty list.

This makes the DLQ resend path rely on an exception for an incomplete queue lookup instead of handling the queue metadata result explicitly.

### Location

- `server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDLQProvider.java`

### Expected behavior

The provider should handle a null or empty DLQ queue set explicitly before iterating it. This avoids relying on an exception path and keeps DLQ resend behavior deterministic when RocketMQ returns incomplete queue metadata.

### Scope

This is part of the RocketMQ Studio BASE-01 baseline for DLQ query/resend stability.

### Suggested fix

Guard the queue set before the iteration and add a unit test for the null return case.

Aias00:
### Summary

- guard DLQ resend collection when `fetchSubscribeMessageQueues` returns null or empty
- add a provider unit test that verifies the null queue-set path does not attempt pulls or create a resend producer

### Validation

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=RocketMQDLQProviderTest test`
- `git diff --check`

Closes #903

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDLQProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDLQProvider.java
@@ -184,6 +184,9 @@ public class RocketMQDLQProvider implements DLQProvider {
         try {
             consumer.start();
             Set<MessageQueue> queues = consumer.fetchSubscribeMessageQueues(dlqTopic);
+            if (queues == null || queues.isEmpty()) {
+                return result;
+            }
             outer:
             for (MessageQueue queue : queues) {
                 long minOffset = consumer.searchOffset(queue, begin);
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by consolidated PR #935, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #910, opened 2026-08-04
[ISSUE #909] Guard stale cluster store refreshes

### LINKED ISSUE #909: [Bug] Cluster store stale refreshes can overwrite newer cluster lists
### Description

`useClusterStore.fetchClusters()` applies every completed `listClusters()` response to the store. When a slow refresh starts first and a faster refresh starts later, the older response can finish last and overwrite the newer cluster list.

### Location

`web/src/stores/clusterStore.ts`

### Impact

Cluster pages and selectors can show stale cluster topology after overlapping refreshes. This is risky for the Studio multi-cluster control plane because subsequent UI actions may use outdated cluster data.

### Suggested fix

Track the latest fetch sequence and only commit `clusters` and `loading=false` for the latest request. Add a regression test with two deferred `listClusters()` promises where the second request resolves first.

### Scope

RocketMQ Studio Track 1 / multi-cluster control plane state handling.

Aias00:
### Summary

- ignore stale `fetchClusters` responses when overlapping cluster refreshes finish out of order
- keep `loading` controlled by the latest in-flight refresh
- add a regression test for out-of-order cluster refresh completion

### Tests

- `npm test -- --run src/stores/clusterStore.test.ts`
- `npm run lint -- src/stores/clusterStore.ts src/stores/clusterStore.test.ts` (passes with existing fast-refresh warnings in unrelated files)
- `git diff --check`

Closes #909

### DIFF
--- a/web/src/stores/clusterStore.test.ts
+++ b/web/src/stores/clusterStore.test.ts
@@ -52,6 +52,23 @@ const cluster: ClusterInfo = {
   tpsHistory: [100, 120],
 };
 
+const newerCluster: ClusterInfo = {
+  ...cluster,
+  id: 'cluster-staging',
+  name: 'rocketmq-staging',
+};
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  let reject!: (reason?: unknown) => void;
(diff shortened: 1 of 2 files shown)


Aias00:

PR #912, opened 2026-08-04
[ISSUE #911] Trim consumer group search text

### LINKED ISSUE #911: [Bug] Consumer group search does not trim whitespace
### Description

`listConsumerGroups()` in mock mode filters consumer groups with the raw `search` value. Unlike topic search, it does not trim the input first, so a UI search value such as `  GID_order_service  ` fails to match an existing group.

### Location

`web/src/services/consumerService.ts`

### Impact

The Consumer Group page can show an empty result when users paste or type a valid group name with leading or trailing spaces. This makes the Studio metadata list filtering inconsistent with Topic search and can hide existing consumer groups.

### Suggested fix

Trim the search text before filtering. Blank search text should be ignored and return the full list, matching `topicService.listTopics()` behavior. Add mock-service regression tests for trimmed and blank search values.

### Scope

RocketMQ Studio Track 1 / Consumer Group metadata management.

Aias00:
### Summary

- trim consumer group search text before filtering mock rows
- ignore blank consumer search text instead of returning an empty list
- align consumer group search behavior with topic search

### Tests

- `npm test -- --run src/services/consumerService.test.ts`
- `npm run lint -- src/services/consumerService.ts src/services/consumerService.test.ts` (passes with existing fast-refresh warnings in unrelated files)
- `git diff --check`

Closes #911

### DIFF
--- a/web/src/services/consumerService.test.ts
+++ b/web/src/services/consumerService.test.ts
@@ -55,6 +55,19 @@ describe('consumer service mock data', () => {
     expect(second.instances[0]).not.toBe(first.instances[0]);
   });
 
+  it('trims search text before filtering consumer group names', async () => {
+    const groups = await listConsumerGroups({ search: '  CG-ORDER-NOTIFY  ' });
+
+    expect(groups.map((group) => group.name)).toEqual(['cg-order-notify']);
+  });
+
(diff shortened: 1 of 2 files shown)


Aias00:

PR #914, opened 2026-08-04
[ISSUE #913] Reject missing alert rule updates

### LINKED ISSUE #913: [Bug] Alert rule updates silently succeed for unknown IDs
### Description

In mock mode, `updateAlertRule()` returns the submitted alert rule even when no existing rule has the requested `id`.

### Location

`web/src/services/opsService.ts`

### Impact

The Alert Management page can show a successful save for a stale or deleted alert rule ID even though no rule was updated. This hides stale state and makes the mock service contract diverge from real API behavior.

### Suggested fix

Throw an `Alert rule not found` error when `updateAlertRule()` cannot find the target rule. Add a regression test that attempts to update an unknown rule ID and verifies that the alert rule list is unchanged.

### Related existing work

#757 fixed the same missing-ID pattern for ACL rules, but it does not cover alert rules in `opsService`.

### Scope

RocketMQ Studio Track 1 / Ops alert management.

Aias00:
### Summary

- fail mock alert rule updates when the target ID does not exist
- avoid returning a fake successful update for stale alert rule IDs
- add a regression test that verifies the alert rule list remains unchanged

### Tests

- `npm test -- --run src/services/opsService.test.ts`
- `npm run lint -- src/services/opsService.ts src/services/opsService.test.ts` (passes with existing fast-refresh warnings in unrelated files)
- `git diff --check`

Closes #913

### DIFF
--- a/web/src/services/opsService.test.ts
+++ b/web/src/services/opsService.test.ts
@@ -82,6 +82,23 @@ describe('ops service mock data', () => {
     expect(afterUpdate?.channels).toEqual(['webhook']);
   });
 
+  it('rejects updates for unknown alert rule IDs', async () => {
+    const before = await listAlertRules();
+    const missingRule = {
+      ...before[0],
+      id: 'missing-alert-rule',
+      name: 'missing rule',
+    };
+
+    await expect(updateAlertRule(missingRule)).rejects.toThrow(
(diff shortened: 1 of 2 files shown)


Aias00:

PR #916, opened 2026-08-04
[ISSUE #915][ISSUE #917] Harden cluster mock service writes

### LINKED ISSUE #915: [Studio][Bug] K8s certificate mock store keeps caller SAN array references
### Description

`createK8sCert()` and `updateK8sCert()` copy SAN arrays when returning mock certificate records, but the mock store still saves `data.san` directly.

Locations:
- `web/src/services/clusterService.ts:85`
- `web/src/services/clusterService.ts:97`

A caller that keeps the original request object can mutate the SAN array after the service call and change the stored mock certificate outside the service boundary. That makes the mock K8s certificate service behave differently from an API-backed store and can leak state between page actions/tests.

### Impact

K8s certificate management can show mutated SAN values that were never submitted through a create/update operation. This makes the Track 1 control-plane mock contract less reliable and can hide bugs in the real API integration path.

### Suggested fix

Copy SAN arrays before writing them into the mock store for both create and update paths, and add regression coverage that mutating the original request array does not mutate stored certificate state.

### LINKED ISSUE #917: [Studio][Bug] NameServer edit can create duplicate addresses in mock mode
### Description

In mock mode, `updateNameServer()` can rename a NameServer to an address that already exists in the same cluster.

Location:
- `web/src/services/clusterService.ts:211`

`createNameServer()` rejects duplicate addresses, but the edit path assigns `newAddr` directly without checking the rest of the cluster's NameServer list. This can create duplicate endpoint rows inside one cluster.

### Impact

After a duplicate address is created, later restart/delete/edit actions become ambiguous because they identify NameServers by address. The mock control-plane view can then diverge from the uniqueness rule already enforced during creation and from the expected backend contract.

### Suggested fix

Before assigning `newAddr`, reject it when another NameServer in the same cluster already has that address. Add regression coverage for the duplicate edit path and for keeping the original address unchanged after rejection.

Aias00:
### What is changed

This PR fixes #915 and fixes #917 by tightening the web mock cluster service write paths.

### Why

Two mock-mode write paths could mutate state in ways that do not match the intended control-plane contract:

- `createK8sCert()` and `updateK8sCert()` returned defensive SAN copies, but stored the caller-provided `san` array reference. Mutating the original request array after the call could change stored certificate state.
- `updateNameServer()` could rename a NameServer to an address already used by another NameServer in the same cluster, even though `createNameServer()` already rejects duplicates.

### How it is fixed

- Copy `data.san` before writing K8s certificate records into the mock store.
- Reject NameServer edits that would duplicate another address in the same cluster.
- Add regression tests for both write paths.

### Verification

- `npm test -- --run src/services/clusterService.test.ts`
- `git diff --check`
- `npm run lint -- src/services/clusterService.ts src/services/clusterService.test.ts` (passes with existing fast-refresh warnings in unrelated files)

### DIFF
--- a/web/src/services/clusterService.test.ts
+++ b/web/src/services/clusterService.test.ts
@@ -22,7 +22,15 @@ vi.mock('../config', () => ({
   API_BASE_URL: '/api',
 }));
 
-import { getCluster, listClusters, updateClusterConfig } from './clusterService';
+import {
+  createK8sCert,
+  deleteK8sCert,
+  getCluster,
+  listClusters,
+  listK8sCerts,
+  updateClusterConfig,
+  updateK8sCert,
+} from './clusterService';
 
 describe('clusterService mock clusters', () => {
   it('returns defensive copies from cluster detail reads', async () => {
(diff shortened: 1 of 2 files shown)


Aias00:

Issue #739, opened 2026-08-01
[Studio][Bug] Topic actions conflate same-named resources across clusters


ai-yang:
## Affected branch

`rocketmq-studio` at `bbf1b7e0cf25a5065ba049b5450cc8155569f710`

## Description

The Studio topic model and list API expose both `clusterId` and `namespace`, so a combined list can legitimately contain the same topic name in different clusters. However, the topic management flow treats `name` alone as the resource identity:

- the main table uses `rowKey="name"`
- selected rows and batch deletion contain only names
- detail route/consumer caches are keyed only by topic name
- detail endpoints accept only `/{name}`
- the delete request body contains only `{ name }`
- after one delete, local state removes every row with that name

This produces duplicate React row keys and makes detail, selection, and destructive actions ambiguous across clusters.

## Deterministic reproduction

1. Return two topic rows named `orders`, one with `clusterId=cluster-a` and one with `clusterId=cluster-b`.
2. Render the Topic page and confirm that both rows are present.
3. Delete the first row and confirm the dialog.
4. Observe the remaining table rows.

On the unmodified branch, both `orders` rows disappear because the page executes:

```ts
previous.filter((item) => item.name !== topic.name)
```

A route-level regression that expects the other cluster's row to remain failed 5/5 times without timers, network access, or random scheduling.

## Expected behavior

Actions on one topic must not select, cache, remove, or address a same-named topic in another cluster. The resource identity must therefore include at least `(clusterId, name)`. `namespace` also exists in `Topic`/`TopicVO`, but earlier maintainer feedback indicates that Namespace is not currently a generally supported Studio capability, so it should not be assumed unconditionally.

## Contract question / proposed scope

Before implementing this across the public management API, could maintainers confirm the canonical topic identity and request shape?

- Is the current identity at least `(clusterId, name)`?
- When Namespace capability is available, should this extend to `(clusterId, namespace, name)`?
- How should legacy name-only requests behave when more than one cluster matches (for example, reject ambiguity rather than select arbitrarily)?

Once confirmed, an atomic follow-up should cover all of these layers consistently:

- composite React row keys and selection keys
- detail/cache keys
- single and batch delete payloads
- routes/consumers/detail lookups
- frontend service and backend controller/service/provider interfaces
- compatibility behavior for legacy name-only callers

[...]

PR #919, opened 2026-08-04
[ISSUE #918] Reject duplicate mock topic creates

### LINKED ISSUE #918: [Studio][Bug] Mock topic create allows duplicate names in the same cluster
### Description

In mock mode, `createTopic()` inserts a new topic without checking whether the same cluster already has a topic with the same name.

Location:
- `web/src/services/topicService.ts:32`

The same service later updates and deletes topics by name, so duplicate topic names in one cluster make subsequent actions ambiguous and can mutate/delete the first matching row instead of the row the user intended.

This is different from #739, which tracks the larger cross-cluster identity problem for topic actions. This issue is limited to rejecting duplicate creates within the same cluster in the mock service.

### Impact

The Topic page can show duplicate rows for the same topic name in one cluster. Follow-up update/delete/detail actions then operate on a name-only lookup and may affect the wrong row.

### Suggested fix

Before inserting a mock topic, reject creates when another topic in the same `clusterId` already has the requested `name`. Add regression coverage that the duplicate create is rejected and the existing topic row remains unchanged.

Aias00:
### What is changed

This PR fixes #918 by rejecting duplicate mock topic creates within the same cluster.

### Why

`createTopic()` previously inserted a new mock row without checking for an existing topic with the same `name` and `clusterId`. Since follow-up mock update/delete actions identify topics by name, allowing duplicate names in one cluster can make later actions ambiguous.

This is intentionally scoped to same-cluster duplicate creation. The broader cross-cluster identity problem remains tracked by #739.

### How it is fixed

- Check for an existing topic with the same `name` and `clusterId` before inserting a mock topic.
- Add a regression test that verifies the duplicate create is rejected and the existing cluster topic list remains unchanged.

### Verification

- `npm test -- --run src/services/topicService.test.ts`
- `git diff --check`
- `npm run lint -- src/services/topicService.ts src/services/topicService.test.ts` (passes with existing fast-refresh warnings in unrelated files)

### DIFF
--- a/web/src/services/topicService.test.ts
+++ b/web/src/services/topicService.test.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { getTopicConsumers, getTopicRoutes, listTopics } from './topicService';
+import { createTopic, getTopicConsumers, getTopicRoutes, listTopics } from './topicService';
 
 vi.mock('./dataMode', () => ({ isMockMode: () => true }));
 vi.mock('../config', () => ({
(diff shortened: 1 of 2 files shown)


Aias00:

PR #921, opened 2026-08-04
[ISSUE #920] Reject missing mock instance deletes

### LINKED ISSUE #920: [Studio][Bug] Mock instance delete succeeds for missing ids
### Description

In mock mode, `deleteInstance()` silently succeeds when the requested instance id does not exist.

Location:
- `web/src/services/instanceService.ts:65`

The server-side instance service rejects missing ids with `InstanceVO not found: <id>`, and other mock write paths already reject missing records. The mock instance delete path currently returns success for a stale or invalid id.

### Impact

The Instance page can show a successful delete toast for a row that was not actually deleted. This makes stale UI state and retry behavior harder to diagnose, especially while validating the Track 1 control-plane flows against mock data.

### Suggested fix

Make the mock delete path reject unknown instance ids and add regression coverage that the stored instance list remains unchanged after a failed delete.

Aias00:
### What is changed

This PR fixes #920 by making mock instance deletion reject missing ids.

### Why

The API-backed instance service rejects missing instance ids, but the web mock service silently returned success when no record matched. That can make the Instance page show a successful delete for a stale or invalid row.

### How it is fixed

- Throw `Instance not found: <id>` when mock deletion cannot find the requested id.
- Add regression coverage that a failed delete leaves the stored mock instance list unchanged.

### Verification

- `npm test -- --run src/services/instanceService.test.ts`
- `git diff --check`
- `npm run lint -- src/services/instanceService.ts src/services/instanceService.test.ts` (passes with existing fast-refresh warnings in unrelated files)

### DIFF
--- a/web/src/services/instanceService.test.ts
+++ b/web/src/services/instanceService.test.ts
@@ -22,7 +22,7 @@ vi.mock('../config', () => ({
   API_BASE_URL: '/api',
 }));
 
-import { createInstance, listInstances, updateInstance } from './instanceService';
+import { createInstance, deleteInstance, listInstances, updateInstance } from './instanceService';
 
 describe('instanceService mock instances', () => {
   it('returns defensive copies from list reads', async () => {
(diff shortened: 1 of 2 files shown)


Aias00:

PR #923, opened 2026-08-04
[ISSUE #922] Filter mock messages by time range

### LINKED ISSUE #922: [Studio][Bug] Mock message queries ignore the selected time range
### Description

In mock mode, `queryMessages()` ignores the `startTime` and `endTime` fields from `MessageQuery`.

Location:
- `web/src/services/messageService.ts:20`

The Message page passes a time range for topic-based queries, and the API contract exposes `startTime` / `endTime`, but the mock service only filters by topic/tag/key/msgId before returning all matching messages.

### Impact

Mock topic queries can show messages outside the selected time window. This makes the Studio message-query workflow diverge from the real API and can hide regressions in the time-range query path.

### Suggested fix

Filter mock messages by parsed `storeTime` when `startTime` or `endTime` is present, preserving the existing descending store-time sort. Add regression coverage for inclusive time-window filtering.

Aias00:
### What is changed

This PR fixes #922 by applying the selected time range to mock message queries.

### Why

The Message page sends `startTime` and `endTime` for topic-based queries, but the mock service ignored both fields and returned every topic/tag match. That made mock mode diverge from the real query contract and could show messages outside the selected window.

### How it is fixed

- Parse each mock message `storeTime` into a timestamp.
- Apply inclusive `startTime` / `endTime` filters before sorting results.
- Add regression coverage for a narrow topic time window.

### Verification

- `npm test -- --run src/services/messageService.test.ts`
- `git diff --check`
- `npm run lint -- src/services/messageService.ts src/services/messageService.test.ts` (passes with existing fast-refresh warnings in unrelated files)

### DIFF
--- a/web/src/services/messageService.test.ts
+++ b/web/src/services/messageService.test.ts
@@ -39,6 +39,18 @@ describe('message service mock data', () => {
     expect(second[0].properties).not.toBe(first[0].properties);
   });
 
+  it('filters mock topic queries by the selected store-time range', async () => {
+    const messages = await queryMessages({
+      topic: 'order-create',
+      startTime: Date.parse('2026-07-01T10:24:00.000Z'),
+      endTime: Date.parse('2026-07-01T10:26:00.000Z'),
+    });
+
+    expect(messages.map((message) => message.msgId)).toEqual([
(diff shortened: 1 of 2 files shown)


Aias00:

PR #871, opened 2026-08-03
fix(web): remove BrokerCluster mock fallback rows

Aias00:
### Summary

- remove hard-coded Broker/NameServer/Proxy fallback rows from BrokerCluster
- initialize infrastructure tables as empty until real cluster data loads
- keep API failure reporting without showing local mock topology
- update regression coverage so failed loads do not render mock infrastructure nodes

### Related Issue

Fixes #870

### Tests

- `npm test -- --run src/pages/studio/__tests__/BrokerCluster.test.tsx`
- `npm run lint -- src/pages/studio/BrokerCluster.tsx src/pages/studio/__tests__/BrokerCluster.test.tsx` (passes with existing fast-refresh warnings)

PR #791, opened 2026-08-03
[ISSUE #790] Gate demo instance seed data

Aias00:
### Which Issue(s) This PR Fixes

Fixes #790

### Brief Description

`InMemoryInstanceRepository` currently seeds demo production/staging/dev instances as soon as the backend starts. This PR makes demo instance data opt-in so a fresh Studio backend does not fabricate instance resources by default.

Changes:

- Add `studio.instance.seed-demo-data`, defaulting to `false`.
- Seed `InMemoryInstanceRepository` only when that flag is enabled.
- Update repository tests to cover both default empty state and explicit demo seeding.

### How Did You Test This Change?

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=InMemoryInstanceRepositoryTest,StudioApplicationTest test`

PR #936, opened 2026-08-04
[ISSUE #775][ISSUE #790][ISSUE #802][ISSUE #870] Harden demo and mock data boundaries

### LINKED ISSUE #775: [Studio][Bug] Cluster repository seeds demo clusters by default

### LINKED ISSUE #790: [Bug] Instance repository seeds demo records by default
### Problem

`InMemoryInstanceRepository` seeds three demo instances during construction. That makes a fresh RocketMQ Studio backend show production/staging/dev instances even when the operator has not configured any real instances.

This can mislead users into treating sample data as real control-plane state, and it is inconsistent with the control-plane goal that real resources should come from configured providers or explicit user actions.

### Evidence

- `server/src/main/java/org/apache/rocketmq/studio/instance/InMemoryInstanceRepository.java` calls `initData()` in its constructor.
- The seed data includes fixed entries such as `production-proxy`, `staging-proxy`, and `dev-direct`.

### Expected Behavior

The default repository state should be empty. Demo instance records should only be loaded when an explicit demo/seed flag is enabled.

### Scope

This is in the RocketMQ Studio control-plane scope: ARCH-01 multi-cluster/instance management should avoid fabricating resources by default.

### LINKED ISSUE #802: [Studio][Bug] Production web builds enable mock data by default

### LINKED ISSUE #870: BrokerCluster should not fall back to mock infrastructure rows on API failure
### Problem

`web/src/pages/studio/BrokerCluster.tsx` still initializes Broker, NameServer, and Proxy tables with local mock rows and keeps those rows when `listClusters()` fails. The current test suite also asserts this fallback behavior.

This means a real API failure can leave Studio showing hard-coded infrastructure nodes, which is misleading for the Control Plane / ARCH-01 cluster management view.

A previous issue (#700) was closed by #701, but the current `rocketmq-studio` branch still contains this fallback behavior.

### Expected behavior

- The page should start with empty Broker/NameServer/Proxy tables.
- Successful `listClusters()` responses should populate the tables.
- Failed `listClusters()` calls should show the existing error message and leave the tables empty or keep the last real response, not local mock rows.
- Tests should assert that API failure does not render hard-coded mock infrastructure data.

### Scope

RocketMQ Studio / Track 1 / ARCH-01 cluster and proxy management view.

Aias00:
## Summary

**Track:** Track 1 / Control Plane reliability

Consolidates and supersedes #776, #791, #803, and #871.

- Gate in-memory cluster and instance demo seed data behind explicit configuration disabled by default.
- Disable frontend mock mode in production builds.
- Remove BrokerCluster fallback rows when the service returns no data.

## Verification

- `mvn -Dtest=ClusterRepositoryImplTest,InMemoryInstanceRepositoryTest test`
- `npm run lint -- src/pages/studio/BrokerCluster.tsx src/pages/studio/__tests__/BrokerCluster.test.tsx`

Closes #775
Closes #790
Closes #802
Closes #870

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java
@@ -25,6 +25,7 @@ import org.apache.rocketmq.studio.common.domain.enums.ClusterStatus;
 import org.apache.rocketmq.studio.common.domain.enums.ClusterType;
 import org.apache.rocketmq.studio.common.domain.enums.FlushDiskType;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Repository;
 
 import java.time.LocalDateTime;
@@ -40,8 +41,11 @@ public class ClusterRepositoryImpl implements ClusterRepository {
 
     private final Map<String, ClusterVO> store = new ConcurrentHashMap<>();
 
-    public ClusterRepositoryImpl() {
-        initStubData();
+    public ClusterRepositoryImpl(@Value("${studio.cluster.seed-demo-data:false}") boolean seedDemoData) {
+        if (seedDemoData) {
+            initStubData();
+            log.info("Initialized demo cluster data for Studio cluster repository");
+        }
     }
 
     @Override
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InMemoryInstanceRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InMemoryInstanceRepository.java
@@ -18,6 +18,7 @@
 package org.apache.rocketmq.studio.instance;
 
 import org.apache.rocketmq.studio.common.domain.enums.InstanceType;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import java.time.LocalDateTime;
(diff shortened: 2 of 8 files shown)


Aias00:
Superseded by consolidated PR #944, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

PR #931, opened 2026-08-04
[ISSUE #840][ISSUE #874] Validate authentication management requests

Aias00:
## Summary

**Track:** Track 1 / AUTH-01

Consolidates and supersedes #841 and #875 into a single authentication and ACL management validation change.

- Reject missing login request bodies at the controller boundary and service layer.
- Use a create-specific ACL user DTO with username validation.
- Keep ACL response models separate from write contracts so generated/masked credentials stay response-only.
- Add controller and service regression coverage for invalid requests.

## Verification

- JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=AuthServiceTest,AuthControllerTest,AclControllerTest test
- git diff --check origin/rocketmq-studio...HEAD

Closes #840
Closes #874

PR #947, opened 2026-08-04
[ISSUE #840][ISSUE #874][ISSUE #878] Harden authentication management boundaries

### LINKED ISSUE #840: [Bug] Login should reject missing request body before authentication
### What happened?

AuthController passes the LoginDTO directly to AuthService, and AuthService logs request.getUsername before validating the request object. If a null LoginDTO reaches the service, the login path can throw NullPointerException instead of returning a clear client error.

### Expected behavior

A missing login request body should be rejected with a clear 400 response, and AuthService should also guard against null requests for direct service callers.

### Why this matters

This belongs to BASE-01 login and permission handling. Authentication endpoints should fail predictably and should not turn malformed client requests into server errors.

### Candidate fix

- Make the controller reject null login request bodies with BusinessException 400
- Add a service-level null guard before logging fields
- Add focused controller and service tests

### LINKED ISSUE #874: ACL user create endpoint should validate username
### Problem

`AclController#createUser` accepts `AclUserVO` without request validation. A blank or missing username can pass the controller boundary and reach `AclService#createUser`.

The same controller already validates delete requests with `AclDeleteRequestDTO`, so ACL user creation should use the same boundary-validation pattern.

### Expected behavior

- Reject ACL user create requests with a missing or blank username.
- Do not call `AclService#createUser` for invalid requests.
- Keep `AclUserVO` usable as a response model for masked credentials and generated keys.
- Add MockMvc regression coverage.

### Scope

RocketMQ Studio / Track 1 / AUTH-01 ACL user management.

### LINKED ISSUE #878: [Bug] Auth public endpoints reject trailing slash variants

Aias00:
## Summary

**Track:** Track 1 / Control Plane security

Consolidates and supersedes #879 and #931.

- Validate authentication and ACL-management requests at the API boundary.
- Normalize public path matching in the authentication interceptor.
- Keep authentication request handling and interceptor exclusions covered by focused tests.

## Verification

- `mvn -Dtest=AuthControllerTest,AuthInterceptorTest,AuthServiceTest,AclControllerTest test`

Closes #840
Closes #874
Closes #878

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthController.java
@@ -18,6 +18,7 @@
 package org.apache.rocketmq.studio.auth;
 
 import org.apache.rocketmq.studio.common.domain.Result;
+import org.apache.rocketmq.studio.common.exception.BusinessException;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.CacheControl;
 import org.springframework.http.HttpHeaders;
@@ -50,7 +51,10 @@ public class AuthController {
     }
 
     @PostMapping("/login")
-    public Result<LoginVO> login(@RequestBody LoginDTO request) {
+    public Result<LoginVO> login(@RequestBody(required = false) LoginDTO request) {
+        if (request == null) {
+            throw new BusinessException(400, "Login request is required");
+        }
         return Result.ok(authService.login(request));
     }
 
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthInterceptor.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthInterceptor.java
@@ -50,11 +50,25 @@ public class AuthInterceptor implements HandlerInterceptor {
     }
 
     private boolean isPublicPath(String path) {
+        path = normalizePath(path);
         return path.equals("/api/auth/login")
                 || path.equals("/api/auth/status")
                 || path.startsWith("/api-docs")
                 || path.startsWith("/swagger-ui")
                 || path.startsWith("/actuator/health");
+    }
+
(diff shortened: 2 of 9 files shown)


RockteMQ-AI (approved):
### Summary
This PR hardens authentication management boundaries with four key improvements:

1. **Null request body handling** — Login endpoint now accepts  and explicitly checks for null, throwing a 400 error. Prevents NPE on malformed requests.
2. **Path normalization** — The  check now strips trailing slashes via . This prevents authentication bypass attempts using paths like  or .
3. **Input validation for ACL user creation** — Introduces  with  annotation, replacing the raw  parameter. This ensures proper validation at the API boundary.
4. **Defensive checks in AuthService** — Added null check at the entry point of  method.

All changes are security-relevant defensive coding improvements. The path normalization is particularly important for preventing authentication bypass. Includes comprehensive tests. LGTM.

---

PR #948, opened 2026-08-04
[ISSUE #895][ISSUE #897] Harden broker metadata parsing and reporting

### LINKED ISSUE #895: [Bug] Broker TPS parsing returns 0 when runtime stats contain extra whitespace

### LINKED ISSUE #897: [Bug] Dashboard cluster overview never reports broker version

Aias00:
## Summary

**Track:** Track 1 / Control Plane reliability

Consolidates and supersedes #896 and #898.

- Parse broker TPS output with flexible whitespace.
- Include broker version in the dashboard provider response.
- Add isolated provider tests for both broker metadata paths.

## Verification

- `mvn -Dtest=RocketMQClusterProviderTest,RocketMQDashboardProviderTest test`

Closes #895
Closes #897

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java
@@ -218,7 +218,7 @@ public class RocketMQClusterProvider implements ClusterProvider {
      */
     private int parseFirstTpsValue(String tpsStr) {
         try {
-            String[] parts = tpsStr.split(" ");
+            String[] parts = tpsStr.trim().split("\\s+");
             if (parts.length > 0) {
                 return (int) Double.parseDouble(parts[0]);
             }
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDashboardProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDashboardProvider.java
@@ -164,8 +164,11 @@ public class RocketMQDashboardProvider implements DashboardProvider {
                                     clusterTpsIn += (int) parseTps(rt.getTable().get("putTps"));
                                     clusterTpsOut += (int) parseTps(rt.getTable().get("getTransferedTps"));
                                     String v = rt.getTable().get("brokerVersionDesc");
-                                    if (v != null && !"unknown".equals(version)) {
-                                        version = v;
+                                    if (v != null && "unknown".equals(version)) {
+                                        String brokerVersion = v.trim();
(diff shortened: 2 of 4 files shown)


Aias00:
Superseded by consolidated PR #949, which contains this change and has passing checks. Closing this narrower PR to keep review focused.

RockteMQ-AI (approved):
### Summary
This PR hardens broker metadata parsing and reporting with three focused fixes:

1. **Flexible whitespace in TPS parsing** —  → . The original code would fail on multiple spaces or leading/trailing whitespace in broker runtime output. Correct fix.
2. **Version comparison logic** — The original condition  was checking the wrong variable state. The fix correctly assigns the broker version when the current version is still "unknown".
3. **Null/empty checks** — Added  guard before assignment.

Includes focused unit tests for both broker metadata paths. LGTM.

---

PR #953, opened 2026-08-04
fix: keep nginx proxying to the backend across container IP changes

lizhimins:
## Summary

- Switch the nginx template to variable `proxy_pass` with a resolver pinned to the podman network gateway (injected by `15-resolver.envsh` before the envsubst step), so a recreated `rocketmq-server` container with a new IP is picked up per request instead of returning 502 until the web container restarts.
- Add a curl healthcheck to `rocketmq-server` and make `rocketmq-web` wait for `service_healthy`, so nginx never boots before the backend is resolvable and serving.

## Test plan

- [x] Deployed: `/api/auth/status` stays 200 across `docker restart rocketmq-server`

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -57,6 +57,12 @@ services:
     depends_on:
       mysql:
         condition: service_healthy
+    healthcheck:
+      test: ["CMD", "curl", "-fsS", "http://localhost:8888/actuator/health"]
+      interval: 10s
+      timeout: 5s
+      retries: 30
+      start_period: 30s
 
   rocketmq-web:
     build:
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes the nginx 502 issue when the backend container is recreated with a new IP. The root cause is that nginx resolves hostnames at startup and caches them. The fix uses three complementary techniques:

1. **Variable proxy_pass** — Using set $backend forces nginx to re-resolve DNS on every request instead of caching the IP.
2. **Pinned resolver** — The 15-resolver.envsh script extracts the podman network gateway IP and uses it as the DNS resolver, avoiding host DNS servers that cannot resolve container names.
3. **Health check + dependency** — Adding a healthcheck to rocketmq-server and making rocketmq-web depend on service_healthy ensures nginx never starts before the backend is ready.

The approach is correct and well-implemented. The valid=10s resolver cache is a good balance between performance and responsiveness. LGTM.

---

PR #940, opened 2026-08-04
[Track 3] feat(ai): harden LLM provider configuration and CLI persistence

Aias00:
## Summary

**Track:** Track 3 / AI Native

Consolidates #807, #810, #812, #814, #864, and #977.

- Normalize legacy Tongyi provider keys and expose only providers supported by the OpenAI-compatible client.
- Persist advanced model settings and preserve configured API keys through settings updates.
- Use the configured API base URL for AI chat streaming.
- Keep connection tests read-only.
- Preserve explicit Claude Code and Qoder CLI engine selections after restart without requiring an HTTP API key.

## Verification

- JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -Dtest=LlmConfigServiceTest test
- npm test -- --run src/api/ai.test.ts src/pages/studio/__tests__/LlmSettingsPage.test.tsx (14 passed)
- git diff --check origin/rocketmq-studio...HEAD

Closes #806
Closes #809
Closes #811
Closes #813
Closes #863
Closes #975

PR #960, opened 2026-08-04
fix(settings): persist configured LLM API key

### LINKED ISSUE #958: [Bug] Persisted LLM API keys are dropped by settings JSON serialization
## Problem

`GeneralSettingsVO.apiKey` is marked `@JsonProperty(access = WRITE_ONLY)` to keep the key out of API responses. `MybatisPlusSettingsRepository` persists the same object by calling `ObjectMapper.writeValueAsString(settings)`. Jackson therefore omits `apiKey` from the database JSON as well.

After saving an OpenAI-compatible LLM configuration, the current process can use its in-memory override, but the key is absent after restart and the provider becomes disabled unless `RMQ_LLM_TOKEN` is configured.

## Scope

Separate API-response redaction from persistence serialization so a configured key survives a settings repository round trip without being returned by REST APIs or emitted in `toString()`.

## Acceptance criteria

- Saving and reloading `GeneralSettingsVO` through `MybatisPlusSettingsRepository` preserves a configured API key.
- REST responses continue to expose only `apiKeyConfigured`, never the key value.
- Add a persistence round-trip regression test.

Related to #813 and #863, but specifically covers the API-key persistence regression introduced by the MySQL-backed settings repository.

Aias00:
## Summary

Fixes #958. The API key remains `WRITE_ONLY` for REST serialization, while the settings repository now explicitly retains it in the persisted JSON document.

## Validation

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=MybatisPlusSettingsRepositoryTest test`
- The new regression test verifies repository save/load retains the key while normal Jackson serialization does not expose it.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java
@@ -18,6 +18,7 @@ package org.apache.rocketmq.studio.persistence;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.rocketmq.studio.persistence.entity.RmqDataSource;
 import org.apache.rocketmq.studio.persistence.entity.RmqSettings;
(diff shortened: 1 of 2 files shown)


Aias00:
Consolidated this API key persistence fix into #940, the Track 3 LLM provider configuration PR. The included commit is 6f78396.

PR #962, opened 2026-08-04
fix(cluster): avoid mock fallback when discovery is unavailable

### LINKED ISSUE #870: BrokerCluster should not fall back to mock infrastructure rows on API failure

Aias00:
## Summary

Fixes #870.

The server no longer falls back to `ClusterRepositoryImpl` sample clusters when live discovery is empty. It returns an empty list for discovery and a controlled 503 for unavailable cluster details. This also prevents MCP `rmq.cluster.list` from returning fabricated clusters.

## Validation

`JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=ClusterServiceTest,ToolGatewayServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -53,7 +53,7 @@ public class ClusterService {
             discovered.forEach(this::enrichWithLiveConfig);
             return discovered;
         }
-        return clusterRepository.findAll();
+        return List.of();
     }
 
     public ClusterVO getCluster(String id) {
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #994, which now contains this change.

PR #966, opened 2026-08-04
[Studio] Surface consumer group Admin failures

### LINKED ISSUE #965: [Studio][Bug] Consumer group detail masks Admin failures as successful offline responses
## Problem

`RocketMQAdminClientImpl.getConsumerGroup` catches every exception from `DefaultMQAdminExt.examineConsumerConnectionInfo` and returns a populated `ConsumerGroupVO`. The controller consequently responds with HTTP 200 even when the Admin call failed because of ACL denial, a broker error, timeout, or a network failure.

RocketMQ distinguishes an actually offline consumer group with `MQClientException(ResponseCode.CONSUMER_NOT_ONLINE)`. That case can safely render as an offline group with zero connections. Other errors must not be converted into an empty successful detail response.

## Scope

- Treat only `CONSUMER_NOT_ONLINE` as the offline fallback.
- Surface all other Admin failures through an explicit business error.
- Add regression coverage for offline, connection/timeout, and broker/ACL-style failures.

## Related work

This is distinct from #786. #786 / #787 handle default unconfigured providers that throw `UnsupportedOperationException`; this issue covers the real `RocketMQAdminClientImpl` Admin call path.

Aias00:
Closes #965

## Summary

- preserve the successful offline fallback only for `MQClientException` with `CONSUMER_NOT_ONLINE`
- convert timeout, broker, ACL, and other Admin call failures into explicit HTTP 502 business responses
- add regression coverage for offline, timeout, and broker/ACL failure paths

## Related

This is intentionally separate from #786 / #787, which cover unconfigured default providers rather than the real `RocketMQAdminClientImpl` Admin path.

## Verification

`JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -Dtest=RocketMQAdminClientImplTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQAdminClientImpl.java
@@ -16,11 +16,13 @@
  */
 package org.apache.rocketmq.studio.rocketmq;
 
+import org.apache.rocketmq.client.exception.MQClientException;
 import org.apache.rocketmq.client.producer.DefaultMQProducer;
 import org.apache.rocketmq.client.producer.SendResult;
 import org.apache.rocketmq.common.TopicConfig;
 import org.apache.rocketmq.common.message.Message;
 import org.apache.rocketmq.remoting.protocol.body.ClusterInfo;
+import org.apache.rocketmq.remoting.protocol.ResponseCode;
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #1025. The replacement is rebased on `rocketmq-studio` and contains only the #965 implementation and focused regression tests; closing this PR to avoid retaining unrelated historical changes.

Aias00:
Closed in favor of the clean replacement #1025.

PR #977, opened 2026-08-04
fix: preserve CLI LLM configuration after restart

### LINKED ISSUE #975: [Studio][Bug] CLI LLM engines become unavailable after a server restart without an API key
## Problem
Studio allows `claude-code` and `qoder` engines to pass configuration validation without an HTTP API key because those tools authenticate through their own runtime login state. After a restart, `LlmConfigService.fromGeneralSettings` derives `enabled` only from the provider's API-key requirement.

A saved CLI configuration with no API key can work until restart, then returns `enabled=false` and the UI disables chat even though the CLI provider can still authenticate.

## Expected behavior
Persisted CLI engine configurations remain enabled and ready after restart when their executable/runtime login is available.

## Proposed fix
Derive enabled/key-required state from the selected engine, and add restart-style regression tests for HTTP, Claude Code, and Qoder configurations.

Aias00:
Closes #975

Persisted LLM configuration now decides whether an API key is required from both the provider and selected engine. Explicit CLI engines remain enabled after restart without an HTTP key, while HTTP engines still require one.

Tests:
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -Dtest=LlmConfigServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
@@ -203,19 +203,20 @@ public class LlmConfigService {
 
     private LlmConfigVO fromGeneralSettings(GeneralSettingsVO settings) {
         String provider = normalizeProvider(settings.getLlmProvider());
+        String engine = normalizeEngine(settings.getLlmEngine());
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #940, which now contains this change.

PR #978, opened 2026-08-04
fix: do not seed demo resources by default

### LINKED ISSUE #974: [Studio][Bug] Default MySQL schema seeds demo resources into production deployments
## Problem
The default MySQL schema inserts demo instances, topics, consumer groups, ACL rules, and ACL users after creating tables. Docker Compose applies this schema automatically for every fresh database volume.

A normal Studio deployment therefore starts with fabricated control-plane resources and credentials before an operator configures a real cluster. This reintroduces the production-data problem addressed by #790, but through the persistent schema rather than an in-memory repository.

## Expected behavior
The default production schema only creates database structure. Demo data must require an explicit development/demo opt-in.

## Proposed fix
Move seed records into a separate demo SQL file or profile, remove them from the production schema, and update deployment documentation.

Aias00:
Closes #974

Keep the default MySQL schema DDL-only. Demo instance, topic, consumer-group, ACL rule, and ACL user records are no longer created for a fresh production volume; existing explicit development seed scripts remain available under `deploy/mysql/`.

Validation:
- no `INSERT IGNORE` remains in `server/src/main/resources/db/schema.sql`
- `docker compose -f deploy/docker-compose.yml config`
- `git diff --check`


Aias00:
Superseded by #980, which now contains this change.

PR #990, opened 2026-08-04
[ISSUE #989] fix: generate unique names for short-lived RocketMQ clients

### LINKED ISSUE #989: [Studio][Bug] Generate collision-resistant instance names for short-lived RocketMQ clients
## Problem
Message queries and DLQ resend scans create short-lived RocketMQ clients with instance names based only on System.currentTimeMillis(). Concurrent requests in the same millisecond can create the same client identity within a shared consumer or producer group.

## Expected
Short-lived clients use collision-resistant instance names while remaining diagnosable.

## Acceptance criteria
- Concurrent creations cannot share an instance name.
- The change applies to message-query and DLQ consumer and producer paths.
- Tests verify uniqueness without relying on clock delays.

Aias00:
Closes #989\n\nGenerate UUID-backed instance names for message-query and DLQ client paths, with a unit test proving uniqueness without clock delays.\n\nTests: mvn -q -Dtest=ShortLivedClientNameTest,RocketMQMessageProviderTest,RocketMQDLQProviderTest test


Aias00:
Superseded by #993, which now contains this change.

PR #992, opened 2026-08-04
[ISSUE #984] fix: reject divergent broker default queue counts

### LINKED ISSUE #984: [Studio][Bug] Align broker default queue configuration with RocketMQ's single queue-count property
## Problem
Cluster configuration exposes independent writeQueueNums and readQueueNums, but both are sent as defaultTopicQueueNums. When both values are supplied, the read value silently overwrites the write value.

## Expected
The UI and API accurately represent RocketMQ broker capabilities. They must either expose one default queue-count value or reject divergent read/write values with a clear explanation.

## Acceptance criteria
- No request silently overwrites another queue-count value.
- Readback values use the same contract as update values.
- Tests cover simultaneous read/write input and the chosen validation/mapping behavior.

Aias00:
Closes #984\n\nRocketMQ Broker exposes one default Topic queue count. Reject differing read/write values rather than silently allowing the read value to overwrite the write value; add UI guidance and a service regression test.\n\nTests: mvn -q -Dtest=ClusterServiceTest test.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -92,6 +92,7 @@ public class ClusterService {
 
     public ClusterVO updateClusterConfig(UpdateConfigDTO command) {
         log.info("Updating cluster config for: {}", command.getId());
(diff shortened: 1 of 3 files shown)


Aias00:
Superseded by #994, which now contains this change.

PR #996, opened 2026-08-04
[ISSUE #995] fix: prevent white screen when cluster has no proxy

### LINKED ISSUE #995: [Studio][Bug] Cluster page goes blank when a real cluster without proxy is configured
## Problem
When a real RocketMQ cluster is configured (`STUDIO_ROCKETMQ_NAMESRV_ADDR`) with brokers registered but no proxy deployed, the "RocketMQ Cluster" page under cluster operations turns completely blank.

`RocketMQClusterProvider` builds `ClusterVO` without setting the `proxies` and `tpsHistory` fields, so they serialize as `null`. The cluster page accesses `c.proxies.length` during render without null guards, and the app has no error boundary, so the entire component tree unmounts (white screen).

## Expected
The cluster page renders the real cluster data (NameServer addresses, brokers) even when the cluster has no proxy.

## Acceptance criteria
- `RocketMQClusterProvider` returns non-null empty `proxies`/`tpsHistory` for clusters without proxies.
- The cluster page does not crash on null `proxies`/`config`/`brokers`/`nameServers`.
- Unit tests cover the provider safe defaults and the page renders with a proxy-less cluster.

keranbingaa:
Closes #995

RocketMQClusterProvider left `proxies`/`tpsHistory` as null in the returned `ClusterVO`, which crashed the cluster page render (`c.proxies.length`) because the app has no error boundary. This adds safe defaults in the provider and null guards in the cluster page for `proxies`/`config`/`brokers`/`nameServers`.

Tests:
- server: `mvn -q -Dtest=RocketMQClusterProviderTest test`
- web: `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClusterProvider.java
@@ -83,13 +83,7 @@ public class RocketMQClusterProvider implements ClusterProvider {
                 List<BrokerVO> brokers = buildBrokerList(brokerNames, brokerAddrTable);
                 List<NameServerVO> nameServers = buildNameServerList();
 
-                ClusterVO cluster = ClusterVO.builder()
-                        .name(clusterName)
-                        .status(ClusterStatus.healthy)
-                        .brokers(brokers)
-                        .nameServers(nameServers)
-                        .build();
-                cluster.setId(clusterName);
+                ClusterVO cluster = buildClusterVO(clusterName, brokers, nameServers);
                 clusters.add(cluster);
             }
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Clean bug fix that prevents the cluster page from crashing (white screen) when `proxies` or `tpsHistory` fields are null in the `ClusterVO` response.

The refactoring extracts a `buildClusterVO()` helper that sets safe defaults (empty lists instead of null), eliminating code duplication between `discoverClusters()` and `refreshClusterDetail()`. The frontend null guards add defense-in-depth.

Good test coverage for the new helper method.

---

PR #1000, opened 2026-08-05
[ISSUE #985] Surface partial broker configuration updates

### LINKED ISSUE #985: [Studio][Bug] Surface partial broker configuration updates across multiple brokers
## Problem
Cluster configuration is applied broker by broker. If an earlier update succeeds and a later update fails, the service throws an error without returning the successful and failed broker set. The cluster can be partially modified while callers only receive a generic failure.

## Expected
The update contract explicitly reports per-broker outcomes and partial completion. It must not imply an all-or-nothing result where no transaction exists.

## Acceptance criteria
- The response and audit identify successful and failed broker addresses.
- Partial completion is represented explicitly.
- UI feedback states which brokers require follow-up.
- Tests simulate a failure after at least one successful broker update.

btlqql:
Closes #985

## What changed

- continue updating remaining brokers when one broker update fails
- return `SUCCESS`, `PARTIAL`, or `FAILED` with per-broker outcomes
- record a cluster-level audit summary containing successful and failed addresses
- keep the configuration dialog open and identify brokers that need follow-up

## Why

The previous loop stopped at the first exception. Brokers updated before that
failure were already changed, but the API returned only a generic error and the
UI reported no partial state.

## Verification

- `mvn test` (608 tests)
- `npm test -- src/api/cluster.test.ts src/services/clusterService.test.ts src/pages/cluster/__tests__/ClusterPage.test.tsx` (27 tests)
- `npm run lint`
- `npm run build`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterController.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.studio.cluster.broker;
 
+import org.apache.rocketmq.studio.cluster.config.ClusterConfigUpdateResultVO;
 import org.apache.rocketmq.studio.cluster.config.UpdateConfigDTO;
 
 import org.apache.rocketmq.studio.common.domain.Result;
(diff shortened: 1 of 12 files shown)


btlqql:
Closing because this change will be resubmitted from the requested account.

PR #1003, opened 2026-08-05
[ISSUE #1002] fix(auth): honor persisted session timeout

### LINKED ISSUE #1002: [Studio][Bug] Make persisted session settings effective and avoid misleading login controls
## Problem

The Settings page persists `sessionTimeout` and `requireLogin`, but runtime authentication uses neither value:

- `AuthService` always issues tokens with a hard-coded 24-hour TTL.
- `AuthWebConfig` decides whether to register the interceptor from `studio.auth.login-required` only at startup.
- Expired sessions are only globally pruned during a subsequent login.

As a result, changing the UI settings either has no effect or presents a control that cannot affect runtime behavior.

## Expected behavior

- Session timeout should be applied consistently to newly issued sessions, with validated units and a documented fallback.
- The login-protection UI must accurately reflect its deployment-controlled behavior; it must not imply a runtime toggle where one is unavailable.
- Expired session entries should be reclaimed without relying on an unrelated future login.

## Scope

Keep deployment-level `studio.auth.login-required` authoritative. Align the settings UI and persisted configuration contract with that boundary, add targeted authentication tests, and preserve backward-compatible settings reads.

## Track

Track 1 / BASE-01 login and permission baseline.

Aias00:
## Summary

- derive newly issued token TTLs from the persisted session timeout, with a validated 30-minute fallback
- periodically reclaim expired in-memory sessions instead of waiting for another login
- keep `studio.auth.login-required` deployment-controlled and replace the non-functional UI switch with an explicit configuration notice

## Verification

- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) mvn -q test`
- `npm test -- --run src/api/generalSettings.test.ts`
- `npm run lint`
- `npm run build`
- `git diff --check`

Closes #1002

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/StudioApplication.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/StudioApplication.java
@@ -19,8 +19,10 @@ package org.apache.rocketmq.studio;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication
+@EnableScheduling
 public class StudioApplication {
     public static void main(String[] args) {
         SpringApplication.run(StudioApplication.class, args);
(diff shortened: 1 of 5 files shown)


Aias00:
Superseded by #968, which now includes the session lifecycle fix and closes #1002.

PR #1012, opened 2026-08-05
fix(metadata): report unsupported namespace discovery

### LINKED ISSUE #1010: [Studio][Bug] Return a capability-aware error from the Namespace API
## Problem

`GET /api/namespaces` is publicly mapped by `NamespaceController`, but `MetadataService.listNamespaces()` throws `UnsupportedOperationException("Not implemented")`. The global exception handler turns this into an unstructured HTTP 500 response.

The current web application does not call this endpoint, and this issue does not propose restoring the previously deferred Namespace resource-management model. The immediate problem is that an exposed API advertises a successful list operation but deterministically fails as an internal error.

## Expected behavior

When the current metadata provider does not implement namespace discovery, the endpoint should return a structured, capability-aware HTTP 501 response. It must not manufacture namespace data or introduce a Namespace UI.

## Scope

Track 1 / Control Plane API reliability. This is intentionally limited to error semantics for the existing endpoint.

Aias00:
## Summary

Closes #1010.

- Convert the exposed but unsupported namespace listing path from an unhandled exception to a structured HTTP 501 response.
- Do not introduce namespace data, a namespace UI, or new provider behavior.
- Add service and controller regression coverage for the capability error contract.

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -Dtest=MetadataServiceTest,NamespaceControllerTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
@@ -119,7 +119,7 @@ public class MetadataService {
 
 
     public List<NamespaceVO> listNamespaces() {
-        throw new UnsupportedOperationException("Not implemented");
+        throw new BusinessException(501, "Namespace discovery is not implemented by the current metadata provider");
     }
 
     private String normalizeFilter(String value) {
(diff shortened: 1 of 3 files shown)


lizhimins:
Consolidated into #1025 (merged into `rocketmq-studio` as a8d263e) together with #1018/#1019/#1020/#1021/#1025, since these closely related hardening fixes were reviewed and verified as one change set. Associated issue(s) #1010 will be closed against that merge.

已并入 #1025（合入 `rocketmq-studio`，SHA a8d263e）：这几个相关性很强的加固修复作为同一批评审与验证后一次性合入，关联 issue #1010 将按该合入关闭。

PR #1018, opened 2026-08-05
fix: avoid simulated alert rule changes

### LINKED ISSUE #1014: [Studio][Bug] Remove simulated alert-management mutations until a persisted editor is available
## Problem
`web/src/pages/studio/AlertManagement.tsx` loads real rule YAML but handled create, update, delete, and enable/disable changes entirely in component state and `localStorage`. Refreshing the page discarded the mutations and the downloaded YAML could diverge from the server-side rule set.

The existing browser form does not share the same data model as the server-side persisted alert-rule API, so mapping its expression-oriented fields into that API would corrupt rule semantics.

## Expected behavior
Until a compatible persisted editor exists, the page must be read-only and export the authoritative server-side YAML. It must not report a simulated browser-only change as successful.

## Acceptance criteria
- Rule mutations do not alter browser-only state or claim success.
- Exported YAML is fetched from the server.
- Reloaded view remains consistent with server output.

Aias00:
Closes #1014

Removes browser-only rule mutations from Alert Management and exports the YAML returned by the server. The existing page model is expression-oriented while the persisted alert-rule API is threshold-oriented, so this change avoids falsely reporting local changes as durable until a compatible editor is introduced.

Validation:
- `git diff --check`
- Frontend lint/test not run: dependencies are not installed in the isolated worktree.

### DIFF
--- a/web/src/pages/studio/AlertManagement.tsx
+++ b/web/src/pages/studio/AlertManagement.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import {
   App,
   Badge,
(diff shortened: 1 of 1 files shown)


lizhimins:
Consolidated into #1025 (merged into `rocketmq-studio` as a8d263e) together with #1018/#1019/#1020/#1021/#1025, since these closely related hardening fixes were reviewed and verified as one change set. Associated issue(s) #1014 will be closed against that merge.

已并入 #1025（合入 `rocketmq-studio`，SHA a8d263e）：这几个相关性很强的加固修复作为同一批评审与验证后一次性合入，关联 issue #1014 将按该合入关闭。

PR #1019, opened 2026-08-05
fix: preserve RocketMQ metadata semantics

### LINKED ISSUE #1015: [Studio][Bug] Preserve RocketMQ topic and consumer metadata semantics
## Problem
Several live metadata mappings currently misrepresent RocketMQ state:
- Topic create/update always writes permission `RW`, ignoring the requested/current permission.
- Topic route output selects an arbitrary broker address instead of preferring the master.
- Topic consumer output hard-codes `CLUSTERING`.
- Consumer-group enrichment maps `BROADCASTING` to `SubscriptionMode.Pop`, although these are different dimensions.
- Subscription filter mode maps `CLASS_FILTER` to SQL and falls back to TAG for other values, including SQL92.

## Expected behavior
Expose the corresponding RocketMQ metadata without rewriting permissions or conflating message model, push/pop mode, and filter expression type.

## Acceptance criteria
- Existing topic permissions are preserved on update and validated on create.
- Route rows use master broker addresses when available.
- Consumer and subscription fields accurately reflect the source protocol values.
- Regression tests cover BROADCASTING, SQL92/CLASS_FILTER, and non-RW topic permissions.

Aias00:
Closes #1015

- Preserve requested Topic permissions when creating or updating broker configuration.
- Prefer the master address in topic route output.
- Correct consumer message-model and subscription filter mappings.
- Stop deriving Push/Pop semantics from BROADCASTING/CLUSTERING.

Validation:
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -f server/pom.xml -DskipTests compile`
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQAdminClientImpl.java
@@ -140,6 +140,11 @@ public class RocketMQAdminClientImpl implements AdminClient {
         int readQueues = topic.getReadQueues() > 0 ? topic.getReadQueues() : 8;
 
         try {
+            RmqTopic existing = topicMapper.selectOne(
+                    new LambdaQueryWrapper<RmqTopic>().eq(RmqTopic::getName, topicName));
+            TopicPerm effectivePerm = topic.getPerm() != null
+                    ? topic.getPerm()
+                    : existing == null ? TopicPerm.RW : fromRocketMQPerm(existing.getPerm());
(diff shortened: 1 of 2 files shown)


lizhimins:
Consolidated into #1025 (merged into `rocketmq-studio` as a8d263e) together with #1018/#1019/#1020/#1021/#1025, since these closely related hardening fixes were reviewed and verified as one change set. Associated issue(s) #1015 will be closed against that merge.

已并入 #1025（合入 `rocketmq-studio`，SHA a8d263e）：这几个相关性很强的加固修复作为同一批评审与验证后一次性合入，关联 issue #1015 将按该合入关闭。

PR #1020, opened 2026-08-05
fix: harden client connection response contract

### LINKED ISSUE #1016: [Studio][Bug] Make live client connection results complete and contract-safe
## Problem
The live Remoting client collector has several response-contract defects:
- It never populates `connectedAt`, while the web table treats it as required and sorts with `localeCompare`.
- A producer connected to multiple topics is returned repeatedly with the same `clientId`, but the table uses `clientId` as its row key and statistics count rows rather than instances.
- Producer scanning silently samples the first 50 unsorted non-system topics without a truncation signal.
- An unknown `type` filter silently falls back to scanning both producer and consumer connections.

## Expected behavior
The API must distinguish unavailable/unknown timestamps, stable unique rows, partial scans, and invalid filters so the UI cannot present sampled data as authoritative.

## Acceptance criteria
- The UI handles absent connection timestamps safely.
- Returned row keys and summary counts are stable for multi-topic producers.
- Partial producer scans are deterministic and visibly marked.
- Invalid type filters receive a structured client error.
- Unit and UI tests cover these paths.

Aias00:
Closes #1016

- Reject invalid client type filters instead of silently scanning both types.
- Sort producer topics before applying the scan cap and mark returned records as partial when capped.
- Make connection timestamps optional in the web contract and safe to sort/render.
- Use a composite row key so multi-topic producers do not collide in the table.

Validation:
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -f server/pom.xml -DskipTests compile`
- `git diff --check`
- Frontend lint/test not run: dependencies are not installed in the isolated worktree.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/client/ClientConnectionVO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/client/ClientConnectionVO.java
@@ -40,5 +40,6 @@ public class ClientConnectionVO {
     private ClientLanguage language;
     private String version;
     private LocalDateTime connectedAt;
+    private boolean partial;
     private String clusterName;
 }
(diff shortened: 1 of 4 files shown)


lizhimins:
Consolidated into #1025 (merged into `rocketmq-studio` as a8d263e) together with #1018/#1019/#1020/#1021/#1025, since these closely related hardening fixes were reviewed and verified as one change set. Associated issue(s) #1016 will be closed against that merge.

已并入 #1025（合入 `rocketmq-studio`，SHA a8d263e）：这几个相关性很强的加固修复作为同一批评审与验证后一次性合入，关联 issue #1016 将按该合入关闭。

PR #1047, opened 2026-08-05
fix: harden Studio message producers and DLQ resend

### LINKED ISSUE #1045: [Studio][Bug] sendMessage producer group name collides on concurrent sends
## Problem

`RocketMQAdminClientImpl.sendMessage` creates a throwaway producer whose **producer group name** is derived only from `System.currentTimeMillis()`:

```java
DefaultMQProducer producer = new DefaultMQProducer("studio_msg_sender_" + System.currentTimeMillis());
```

Concurrent send requests in the same millisecond build the same group name, so two short-lived producers register under one client identity — the same collision class that issue caching fixed for consumer/producer *instance names* (`ShortLivedClientName`), but the group name is still timestamp-only.

## Expected

Short-lived producers use a collision-resistant group name while keeping the human-readable prefix.

## Acceptance criteria

- `sendMessage` uses a unique producer group name per call (prefix + random suffix).
- Concurrent sends never share a group identity.
- Existing functionality (send + audit + shutdown) is unchanged.

### LINKED ISSUE #1051: fix: report non-SEND_OK DLQ resend results as failures
## Problem
`RocketMQDLQProvider.resendOne` treats every non-throwing `DefaultMQProducer.send` call as successful. A `SendResult` with a status other than `SEND_OK` is therefore reported as a successful replay and recorded as a successful audit outcome.

The method also writes `DLQ_ORIGIN_*` through `Message.putUserProperty`. RocketMQ reserves these property names, so the message construction can fail before `send` is reached.

## Scope
- Use Studio-owned user property keys for resend provenance.
- Treat `null` and non-`SEND_OK` send results as failed DLQ resends.
- Preserve successful replay behavior for `SEND_OK`.
- Add regression coverage for the reported resend and audit counts.

## Track
Track 1 / Control Plane baseline: DLQ replay correctness.

Aias00:
## Summary
- generate unique RocketMQ producer groups for Studio test-message sends and DLQ resends
- preserve short-lived instance-name behavior
- treat null and non-`SEND_OK` DLQ resend results as failures
- retain failed replay details in audit output

Closes #1045
Closes #1051

## Verification
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) PATH="$JAVA_HOME/bin:$PATH" mvn -Dtest=RocketMQAdminClientImplTest,RocketMQDLQProviderTest,ShortLivedClientNameTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQAdminClientImpl.java
@@ -62,6 +62,7 @@ import java.util.Set;
 public class RocketMQAdminClientImpl implements AdminClient {
 
     private static final Logger log = LoggerFactory.getLogger(RocketMQAdminClientImpl.class);
+    private static final String MESSAGE_SENDER_GROUP_PREFIX = "studio-msg-sender";
 
     private final DefaultMQAdminExt adminExt;
     private final RocketMQProperties properties;
@@ -323,7 +324,7 @@ public class RocketMQAdminClientImpl implements AdminClient {
             throw new BusinessException(500, "Nameserver address not configured");
         }
 
-        DefaultMQProducer producer = new DefaultMQProducer("studio_msg_sender_" + System.currentTimeMillis());
+        DefaultMQProducer producer = new DefaultMQProducer(nextMessageSenderGroup());
         producer.setNamesrvAddr(namesrvAddr);
         producer.setSendMsgTimeout(5000);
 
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
LGTM. Changes look good.

---

RockteMQ-AI (approved):
### Summary
Fixes potential producer group name collisions in concurrent scenarios. Replaces timestamp-based naming (`studio_msg_sender_<millis>`) with `ShortLivedClientName.next()` which generates unique names. Also applies the same fix to DLQ resend producer groups.

### Findings
- **[Info]** `RocketMQAdminClientImpl.java:327` — Using `nextMessageSenderGroup()` instead of timestamp concatenation eliminates the race condition where two concurrent sends within the same millisecond would get the same group name.
- **[Info]** `RocketMQDLQProvider.java:284` — Removing the `groupName` parameter from `newProducer()` and using `nextResendProducerGroup()` is cleaner and avoids potential collisions when resending DLQ messages for the same consumer group concurrently.
- **[Info]** `ShortLivedClientNameTest.java` — Test verifying uniqueness across 1000 iterations is a good regression guard.

### Suggestions
No blocking issues. Clean concurrency safety fix.

---

PR #1052, opened 2026-08-05
fix: report DLQ resend send failures

### LINKED ISSUE #1051: fix: report non-SEND_OK DLQ resend results as failures

Aias00:
Closes #1051

## Summary
- use Studio-owned provenance keys instead of RocketMQ-reserved user property keys
- count `null` and non-`SEND_OK` send results as failed DLQ resends
- retain failed replay details in the audit result

## Verification
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) PATH="$JAVA_HOME/bin:$PATH" mvn -Dtest=RocketMQDLQProviderTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDLQProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQDLQProvider.java
@@ -21,6 +21,7 @@ import org.apache.rocketmq.client.consumer.PullResult;
 import org.apache.rocketmq.client.consumer.PullStatus;
 import org.apache.rocketmq.client.producer.DefaultMQProducer;
 import org.apache.rocketmq.client.producer.SendResult;
+import org.apache.rocketmq.client.producer.SendStatus;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.message.Message;
 import org.apache.rocketmq.common.message.MessageConst;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes silent data loss in DLQ resend — previously, a `send()` returning non-OK status was silently treated as success. Now properly checks `SendStatus` and returns `false` on failure, allowing the caller to count it correctly. Also extracts magic strings into constants.

### Findings
- **[Critical]** `RocketMQDLQProvider.java:248-252` — The null check on `sendResult` and `SendStatus` validation is an important correctness fix. Without this, failed resends would be counted as successful, leading to silent message loss.
- **[Info]** `RocketMQDLQProvider.java:66-67` — Extracting `ORIGIN_MESSAGE_ID_PROPERTY` and `ORIGIN_TOPIC_PROPERTY` as constants improves maintainability.
- **[Info]** `RocketMQDLQProviderTest.java` — New test for null/failure send result is a good regression guard.

### Suggestions
No blocking issues. This is an important reliability fix for the DLQ resend feature.

---

Aias00:
Superseded by #1047, which now includes DLQ resend send-status handling and its regression coverage.

PR #1054, opened 2026-08-05
fix: load Proxy nodes after page mount

### LINKED ISSUE #1053: fix: load Proxy nodes from an effect instead of during render
## Problem
`ProxyPage` calls `loadProxyNodes()` from the component render path using a mutable ref guard. The request changes state while React is rendering and is not aligned with the component effect lifecycle, which can create duplicate or abandoned loads under Strict Mode and concurrent rendering.

## Scope
- Move initial Proxy node loading into `useEffect`.
- Preserve a single initial request and existing refresh behavior.
- Add a component regression test for initial loading.

## Track
Track 1 / ARCH-01 Proxy management view.

Aias00:
Closes #1053

## Summary
- move initial Proxy node loading from render into a lifecycle effect
- stabilize the loader callback for hooks dependency tracking
- add a regression test that asserts one initial request

## Verification
- `npm test -- --run src/pages/studio/__tests__/Proxy.test.tsx`
- `npm run lint` (0 errors; 3 existing warnings)
- `npm run build`

### DIFF
--- a/web/src/pages/studio/Proxy.tsx
+++ b/web/src/pages/studio/Proxy.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { useState, useRef } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import {
   Card,
   Table,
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a React lifecycle anti-pattern — moves async data loading from render-time into `useEffect` with `useCallback`. This ensures Proxy nodes are loaded after the component mounts, preventing potential issues with stale refs and double-invocations in StrictMode.

### Findings
- **[Info]** `Proxy.tsx:74-80` — Replacing `useRef` + render-time call with `useCallback` + `useEffect` is the correct React pattern. Good fix.
- **[Info]** `Proxy.tsx:111-114` — The eslint-disable comment for `set-state-in-effect` is appropriate here since the state updates come from the async API call, not directly in the effect body.
- **[Info]** `Proxy.test.tsx:69-74` — New test verifying single-load-after-mount is a good regression guard.

### Suggestions
No blocking issues. Clean lifecycle fix with proper test coverage.

---

Issue #982, opened 2026-08-04
[Studio][Bug] Scope runtime client, DLQ, and message operations to the selected cluster


Aias00:
## Problem
Runtime views do not consistently use the selected cluster.

- Client connections accept clusterId, but the provider uses the singleton DefaultMQAdminExt and only writes the requested ID back into the response.
- DLQ UI exposes an instance selector but neither list nor resend requests carry that context.
- Message query and trace APIs have no cluster or instance parameter.

This can display or mutate the default NameServer cluster while the UI identifies another cluster.

## Expected
All runtime operations resolve an AdminClient by the selected cluster or instance and preserve that context through the API, audit record, and UI.

## Acceptance criteria
- Client, DLQ, message query, and trace contracts include the required cluster context.
- The backend obtains the matching AdminClient instead of the singleton default client.
- DLQ instance selection reloads data and resend preserves the selected scope.
- Tests prove that the selected cluster is used rather than merely echoed.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

Runtime views (client connections, DLQ, message query, trace) do not consistently propagate the selected cluster context. This can cause operations to target the wrong cluster in multi-cluster deployments.

**Impact:** Multi-cluster Studio deployments. Operations may silently target the default cluster instead of the user-selected one.

**Severity:** High — incorrect cluster targeting can lead to unintended side effects (e.g., resending DLQ messages to the wrong cluster).

---
*Automated evaluation by RockteMQ-AI*

Aias00:
Resolved by the merged runtime-scoping slices: #1058 routes DLQ and client connections, #1081 routes message queries, and #1082 routes message traces through the selected instance endpoint. Together they cover the acceptance criteria in this parent issue.

PR #1058, opened 2026-08-06
feat: resolve runtime admin client for DLQ and client connections

### LINKED ISSUE #1057: feat: resolve runtime AdminClient by selected instance endpoint
## Problem
Runtime operations currently use a singleton `DefaultMQAdminExt` configured from one global NameServer address. Instance-scoped pages can therefore display or mutate the default cluster even when another instance is selected.

## Scope
- Resolve a selected `instanceId` to the persisted instance endpoint.
- Reuse the existing `MqAdminExtFactory` cache to obtain an AdminClient for that endpoint.
- Establish the resolver as the shared runtime-operation foundation.
- Use DLQ list and resend as the first consumer, including API and UI context propagation.

## Follow-up
Client connections, message query, and trace will migrate in separate focused PRs under #982.

## Track
Track 1 / Control Plane multi-instance runtime management.

### LINKED ISSUE #1059: feat: resolve client connections by selected instance endpoint
## Problem

The client connections view is part of the #982 runtime data surface. It must resolve the selected Studio instance to its configured endpoint rather than use the process-wide RocketMQ default.

## Proposed change

- Reuse the runtime AdminClient resolver introduced by #1057.
- Require and propagate `instanceId` for client connection list/detail requests.
- Bind runtime client queries to the selected instance endpoint.
- Cover controller/service/provider and page/API propagation with focused tests.

## Scope

This is a follow-up slice of #982. It does not include message query or message trace, which will be handled separately.

Aias00:
## Summary
- add a runtime resolver from `instanceId` to the saved instance endpoint and the existing endpoint-keyed admin-client cache
- require instance context for DLQ list/resend and client connection queries
- configure DLQ pull/producer clients and live client scans with the selected instance endpoint instead of the global default
- add instance selectors/propagation in the DLQ and client connection pages

## Scope
This is the foundation slice of #982. It establishes the shared runtime resolver and wires its first two consumers: DLQ and client connections. Message query and message trace remain separate follow-ups.

Closes #1057
Closes #1059

## Verification
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) mvn -Dtest=RuntimeAdminClientResolverTest,RocketMQDLQProviderTest,DLQControllerTest,DLQServiceTest,DLQProviderStubTest,RocketMQClientProviderTest,ClientServiceTest,ClientControllerTest,ClientProviderStubTest test`
- `npm test -- --run src/api/dlq.test.ts src/services/messageService.test.ts src/pages/instance/__tests__/DLQPage.test.tsx src/api/connections.test.ts src/services/connectionsService.test.ts src/pages/cluster/__tests__/ClientsPage.test.tsx`
- `npm run lint` (0 errors; existing warnings only)
- `npm run build`


server/src/main/java/org/apache/rocketmq/studio/cluster/client/ProducerConnectionService.java
```diff
@@ -41,7 +41,7 @@ public List<ProducerConnectionVO> listConnections(String topic, String producerG
     }
 
     public List<String> listProducerGroups() {
-        return clientProvider.findConnections(null, ClientType.Producer.name()).stream()
+        return clientProvider.findConnections(null, null, ClientType.Producer.name()).stream()
```
RockteMQ-AI, line 44: **[Warning]** This call now passes `null` for `instanceId`, which will cause `RuntimeAdminClientResolver.execute()` to throw `BusinessException(400, "instanceId is required")`. This method needs to be updated to either accept an `instanceId` parameter or use a different code path.

RockteMQ-AI (commented):
### Summary
Introduces `RuntimeAdminClientResolver` to scope runtime admin operations (DLQ, client connections) to specific instances instead of using a global admin client. Good architectural improvement for multi-instance Studio deployments.

### Findings
- **[Warning]** `ProducerConnectionService.listProducerGroups()` now calls `clientProvider.findConnections(null, null, ClientType.Producer.name())`, which will invoke `runtimeAdminClientResolver.execute(null, ...)` and throw `BusinessException(400, "instanceId is required")`. This appears to be a breaking change — either this method needs to be updated to accept an `instanceId` parameter, or the resolver needs to handle the null case differently for this use case.

### Suggestions
Consider updating `ProducerConnectionService.listProducerGroups()` to either:
1. Accept an `instanceId` parameter and propagate it through the call chain
2. Or provide an alternative code path that doesn't require instance context

Overall the refactoring is well-structured with good test coverage. The concern above should be addressed before merge.

---

PR #1056, opened 2026-08-06
fix: batch studio bug fixes across trace, ACL, auth, alert and SSRF

yyqdbngt:
Batch of bug fixes found through code review of the studio codebase. Each fix ships with a unit test.

**High severity**

- `RocketMQMessageProvider`: message trace parsing used field indexes that do not match RocketMQ 5.3.3's `TraceDataEncoder` layout. Consumption nodes never matched the requested msgId (SubAfter places msgId at index 2, not 5) and Pub status/cost-time fields were read from the wrong columns. Field indexes rewritten to the verified 5.3.3 layout.
- `AclService.updateUser`: partial updates silently reset the `admin` flag to `false` because the DTO used a primitive `boolean`. Now `Boolean`, preserving the existing value when not provided.
- `GET /api/acl/users/{id}/credentials`: plain-text credentials were exposed to any logged-in user; now requires admin.

**Medium severity**

- `RocketMQMetadataProvider.listConsumerGroups`: read the wrong column (`messageModel` instead of `consumeType`), so broadcast groups always displayed as clustering.
- `RocketMQAdminClientImpl.createTopic/updateTopic`: name-only `selectOne` threw `TooManyResultsException` when several clusters share a topic name; lookups now include the cluster id.
- `AlertService.exportPrometheusRulesYaml`: emitted one group per rule, producing duplicate group names that Prometheus rejects; rules are now grouped under a unique name.
- `CliAgentProvider.complete`: sequential stdout/stderr reads deadlock when the child fills a pipe buffer, and the timeout only ran after both reads. Output is now drained concurrently and `waitFor` enforces the timeout.
- `SettingsService.testDataSource`: server-side request to attacker-supplied URLs could probe localhost/cloud-metadata (SSRF); the `localhost` hostname and link-local addresses are rejected, and the endpoint now requires admin.
- `AuthWebConfig`/`AuthInterceptor`: the settings UI "require login" toggle was persisted but never enforced; enforcement now merges the static property with the database toggle.

**Low severity**

- `AiService.chat/execute`: null request bodies caused a 500 NPE; guarded.

Full test suite: 680 tests, 0 failures.

PR #1060, opened 2026-08-06
fix(rocketmq): parse message trace with RocketMQ 5.3.3 field layout

yyqdbngt:
Message trace parsing used field indexes that do not match RocketMQ 5.3.3's TraceDataEncoder. SubAfter places msgId at index 2 (not 5), so consumption nodes never matched the requested message id and were never displayed; Pub status/cost-time fields were read from the wrong columns. Field indexes rewritten to the verified 5.3.3 layout, with unit tests.


lizhimins:

PR #1061, opened 2026-08-06
fix(acl): preserve admin flag on partial user updates

yyqdbngt:
AclService.updateUser silently reset the admin flag to false on partial updates because the request DTO used a primitive boolean (an absent field defaulted to false). The field is now Boolean and the existing value is preserved when not provided, so administrators are no longer silently demoted. Includes unit tests.


lizhimins:

PR #1062, opened 2026-08-06
fix(auth): require admin for secret credential views, honor requireLogin setting

yyqdbngt:
Two issues: (1) GET /api/acl/users/{id}/credentials returned plain-text credentials to any logged-in user; it now requires admin. (2) The settings UI "require login" toggle was persisted but never enforced; enforcement now merges the static studio.auth.login-required property with the database toggle, so the toggle actually takes effect. Includes unit tests.


lizhimins:

PR #1066, opened 2026-08-06
fix(ai): drain CLI subprocess output concurrently to avoid deadlock

yyqdbngt:
CliAgentProvider.complete read stdout then stderr sequentially on the caller thread; once the child filled a pipe buffer (64 KiB) the reads deadlocked, and the 180s timeout only ran after both reads so a hung child could never time out. The merged output stream is now drained on a background thread and waitFor enforces the timeout. Includes a regression test.


RockteMQ-AI (commented):
### Summary
This PR makes changes to 2 file(s) with 171 lines of diff.

**Key observations:**
- ✅ Includes test changes

---

lizhimins:

PR #1071, opened 2026-08-06
fix(rocketmq): audit failure must not fail an already-delivered message

yyqdbngt:
sendMessage delivered the message to the broker and then recorded an audit entry; if the audit write failed, the whole call threw a 500 even though the message was already sent, so callers retried and produced duplicates. The audit call is now best-effort and can no longer fail an already-delivered send. Includes a unit test.


RockteMQ-AI (commented):
### Summary
This PR makes changes to 2 file(s) with 89 lines of diff.

**Key observations:**
- ✅ Includes test changes

---

PR #1075, opened 2026-08-06
fix(cluster): keep write/read queue counts in sync on partial updates

yyqdbngt:
The broker exposes a single defaultTopicQueueNums property, but updating only writeQueueNums or only readQueueNums left the two stored counts inconsistent with what the broker applies. A partial update now mirrors the value onto both counts. Includes a unit test.


RockteMQ-AI (commented):
### Summary
This PR makes changes to 2 file(s) with 51 lines of diff.

**Key observations:**
- ✅ Includes test changes

---

PR #1078, opened 2026-08-06
fix(dlq): tolerate single-queue scan failures during resend collection

yyqdbngt:
collectDeadLetters ran searchOffset/pull for every queue inside one big try; a single queue error (e.g. illegal offset under concurrent consumption) aborted the whole scan, silently skipping the remaining queues while resendMessages still reported SUCCESS. Each queue is now scanned in its own try/catch and failures are logged and skipped.


RockteMQ-AI (approved):
LGTM. Wraps per-queue DLQ scan in try-catch so a single queue failure doesn't abort the entire scan. Good resilience improvement — the warning log includes the queue name for debugging.

---

PR #1080, opened 2026-08-06
fix(api): return 400 for invalid request parameters

yyqdbngt:
Invalid query parameters (bad enum value, non-numeric number, missing parameter) fell into the catch-all handler and returned 500. MethodArgumentTypeMismatchException and MissingServletRequestParameterException now map to a 400 response.


RockteMQ-AI (approved):
LGTM. Adds proper 400 handling for `MethodArgumentTypeMismatchException` and `MissingServletRequestParameterException` — prevents these from falling through to the generic 500 handler. Clean and correct.

---

PR #1081, opened 2026-08-06
feat: resolve message queries by selected instance endpoint

### LINKED ISSUE #1069: feat: resolve message queries by selected instance endpoint
## Problem

Message ID, key, and time-range queries are part of the #982 runtime data surface. They must use the endpoint of the selected Studio instance instead of process-wide RocketMQ configuration.

## Proposed change

- Reuse `RuntimeAdminClientResolver` from #1058.
- Require and propagate `instanceId` through message query API, service, provider, and page layers.
- Route message-ID, key, and time-range queries to the selected instance endpoint.
- Add focused backend and frontend regression coverage.

## Dependency

Depends on #1058. Implementation starts after #1058 merges into `rocketmq-studio`, so this can remain an independent PR.

## Scope

Part of #982. Message trace is explicitly excluded and tracked separately.

Aias00:
## Summary
- require `instanceId` for message-list requests and forward it through the message query service
- resolve the selected instance endpoint with `RuntimeAdminClientResolver` instead of the global NameServer configuration
- bind time-range topic scans to that endpoint and cover controller, provider, and frontend API forwarding

## Scope
Closes #1069
Relates to #982

## Dependency
This branch is based on #1058, which introduces `RuntimeAdminClientResolver`. The PR can be rebased onto `rocketmq-studio` after #1058 merges so inherited commits disappear from the final diff.

## Verification
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=MessageControllerTest,MessageServiceTest,MessageProviderStubTest,RocketMQMessageProviderTest test`
- `npm test -- --run src/api/message.test.ts src/services/messageService.test.ts`
- `npm run build`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageController.java
@@ -35,13 +35,14 @@ public class MessageController {
 
     @GetMapping
     public Result<List<MessageRecordVO>> queryMessages(
+            @RequestParam String instanceId,
             @RequestParam(required = false) String topic,
             @RequestParam(required = false) String msgId,
             @RequestParam(required = false) String tag,
             @RequestParam(required = false) String key,
             @RequestParam(required = false) Long startTime,
             @RequestParam(required = false) Long endTime) {
-        return Result.ok(messageService.queryMessages(topic, msgId, tag, key, startTime, endTime));
+        return Result.ok(messageService.queryMessages(instanceId, topic, msgId, tag, key, startTime, endTime));
     }
 
     @GetMapping("/{msgId}/trace")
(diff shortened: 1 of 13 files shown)


RockteMQ-AI (approved):
### Summary
Instance-scoping refactoring for the message list/query path. Follows the same clean pattern as #1082 — adds `instanceId` as a required parameter across the message query API surface and routes through `RuntimeAdminClientResolver`.

### Review
- **Consistency**: Same provider pattern as #1082 — `MessageProvider`/`MessageProviderStub`/`RocketMQMessageProvider` all updated uniformly.
- **Validation**: `@NotBlank` on `instanceId` in DTOs, proper 400/404 error responses from the resolver.
- **Frontend**: Message list page correctly passes `selectedInstanceId` to API calls.
- **Tests**: Controller, service, provider, and frontend tests all updated.
- **Note**: Depends on #1058 as stated.

---

PR #1092, opened 2026-08-06
fix: honor Broker Cluster live refresh

### LINKED ISSUE #1090: fix: make Broker Cluster data loading lifecycle-safe and honor live refresh
## Problem

The Broker Cluster page starts `loadData()` while rendering and its live-refresh switch only changes local state. The render-time state update violates the React lifecycle, while enabling the switch never requests updated cluster data.

## Scope

Track 1 / Cluster and Broker operations.

## Proposed fix

Move initial loading into an effect, schedule refreshes only while live refresh is enabled, clean up the timer on disable/unmount, and add regression tests for the initial and periodic fetch behavior.

## Evidence

`web/src/pages/studio/BrokerCluster.tsx`: the `initialized` render guard invokes `void loadData()` during render; the Switch uses `onChange={setAutoRefresh}` without any timer or effect.

Aias00:
## Summary
- start the Broker Cluster data request after render with cleanup-aware scheduling
- poll the cluster API only while live refresh is enabled
- stop polling when live refresh is disabled
- add coverage for the refresh lifecycle

Closes #1090

## Verification
- `npm run lint` (3 existing warnings)
- `npm run build`

### DIFF
--- a/web/src/pages/studio/BrokerCluster.tsx
+++ b/web/src/pages/studio/BrokerCluster.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { useCallback, useRef, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { Table, Button, Tag, Tabs, Card, Space, Switch, Progress, Tooltip, Spin, App } from 'antd';
 import {
   Plus,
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Clean refactor — replacing the `useRef`-based initialization with proper `useEffect` hooks is the right approach. Polling lifecycle is correct (only while `autoRefresh` is enabled), cleanup functions prevent memory leaks, and the test with fake timers thoroughly verifies the start/stop behavior.

---

Aias00:
Superseded by #1097, which now includes the live-refresh lifecycle fix and its regression coverage.

PR #1089, opened 2026-08-06
fix: prevent stale async detail responses in management views

Aias00:
## Summary
- retain Consumer Group detail data from the latest selected group request
- prevent stale LiteTopic session responses from overwriting the active drawer
- keep independent regression coverage for both out-of-order response paths

Closes #1088
Closes #1086

Consolidates #1087.

PR #1101, opened 2026-08-06
fix: remove unsupported group actions

### LINKED ISSUE #1100: fix: remove unsupported consumer group actions from global view
## Problem

The global Consumer Group Management page renders actions without handlers. The page also has no selected instance context, so mutating requests cannot be routed safely to a configured RocketMQ endpoint.

## Proposal

Remove non-functional controls from the global aggregate view. Keep group creation and configuration in the instance-scoped consumer management page, where an `instanceId` is available. Hide unsupported subscription distribution actions until an implementation is available.

## Scope

Track 1 / control-plane UI correctness. This does not change the instance-scoped Consumer Group API or introduce a default endpoint fallback.

Aias00:
## Summary
- remove inert Create Group and Config controls from the global aggregate view
- remove the unsupported subscription distribution action
- preserve implemented read-only group details
- add regression coverage that unsupported actions are absent

Closes #1100

## Test
- `npm test -- --run src/pages/studio/__tests__/GroupManagement.test.tsx`
- `npm run lint` (existing warnings only)
- `npm run build`

### DIFF
--- a/web/src/pages/studio/GroupManagement.tsx
+++ b/web/src/pages/studio/GroupManagement.tsx
@@ -31,7 +31,7 @@ import {
   Switch,
   message,
 } from 'antd';
-import { MagnifyingGlass, ArrowClockwise, Users, Eye } from '@phosphor-icons/react';
+import { MagnifyingGlass, ArrowClockwise, Users } from '@phosphor-icons/react';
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #1089, which now includes the global consumer-group action cleanup and regression coverage.

PR #1103, opened 2026-08-06
fix(topic): require newTTL on TTL update

yyqdbngt:
LiteTopicTTLUpdateDTO.newTTL only had @Positive, which passes null, so omitting it reached the service and surfaced as a 500. The field is now @NotNull @Positive and the controller returns a 400 with a clear message.


RockteMQ-AI (approved):

---

lizhimins:

PR #1104, opened 2026-08-06
fix(topic): require positive queue counts on topic creation

yyqdbngt:
CreateTopicDTO allowed writeQueues/readQueues to be 0 ("zero or positive"), but the server silently replaced 0 with the default of 8. Queue counts are now required to be positive, so a 0 is rejected up front instead of being silently overridden. Updated the affected test.


RockteMQ-AI (approved):

---

lizhimins:
Closing without merging: the proposal to tighten queue-count validation from @PositiveOrZero to @Positive was declined — a queue count of 0 is meaningful in migration scenarios, so the existing validation (established in #930) stands. Please reopen if the scenario changes. Thanks for the contribution! / 关闭未合入：将队列数校验从 @PositiveOrZero 收紧为 @Positive 的建议未被采纳——0 队列在迁移场景有意义，保留 #930 确立的现有校验。感谢贡献！

PR #1105, opened 2026-08-06
fix(metrics): validate nested query in datasource queries

yyqdbngt:
MetricsDataSourceQueryRequest.query carried no @Valid, so the nested MetricQueryDTO constraints (metric size, step, time range) were never enforced on the datasource query path. The nested DTO is now validated and query itself must be present.


RockteMQ-AI (approved):

---

lizhimins:

PR #1107, opened 2026-08-06
fix(cluster): prevent new admin connection during shutdown

btlqql:
MqAdminExtFactory.execute checked the closed flag before the cache lookup, but a request that passed that check could still create a fresh admin connection while the factory was shutting down. The cache mapping function now re-checks closed under the lock, so no new connection is established after shutdown. Includes a unit test.


RockteMQ-AI (approved):

---

PR #1112, opened 2026-08-06
fix: redirect unknown instance routes

### LINKED ISSUE #1111: fix: redirect unknown instance routes instead of targeting the first instance
## Problem

`useInstanceFilter` selects the first available instance when the route does not name a valid instance, but it only redirects when the route parameter is absent. A stale or mistyped URL such as `/instance/missing/topic` therefore continues to display `missing` while requests target the first real instance.

## Expected behavior

After instances load, an absent or unknown route instance ID should be replaced with the first available instance ID. The navigation must use `replace` so stale history entries do not accumulate.

## Scope

- Update the shared `useInstanceFilter` hook.
- Add a regression test for an unknown instance route.
- Do not change resource CRUD identity or instance persistence.

Aias00:
## Summary
- replace absent or unknown instance route IDs with the first available instance
- use history replacement to prevent stale route entries
- add a regression test for a stale instance URL

Closes #1111

## Verification
- `npm test -- --run src/hooks/useInstanceFilter.test.tsx`
- `npm run lint` (3 pre-existing warnings)
- `npm run build`

### DIFF
--- a/web/src/hooks/useInstanceFilter.test.tsx
+++ b/web/src/hooks/useInstanceFilter.test.tsx
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes the instance route fallback to redirect not only when the route has no instance ID, but also when the route references an unknown/stale instance. The `isKnownInstance` check correctly covers both cases, and `replace: true` prevents stale history entries. Good regression test.

LGTM.

---

lizhimins:

PR #1110, opened 2026-08-06
fix: require instance scope for metadata writes

Aias00:
## Summary
- disable Topic and Consumer Group creation/import until an instance is selected
- guard write handlers so requests cannot be sent without an instance ID
- preserve the selected instance on imported and created resources

The earlier list-scoping portion is now covered by the current instance Provider layer, so this PR keeps the remaining write-safety behavior focused.

Closes #1116

## Validation
- npm test -- --run src/pages/instance/__tests__/TopicPage.test.tsx src/pages/instance/__tests__/ConsumerPage.test.tsx

PR #1117, opened 2026-08-06
fix: require instance scope for metadata writes

### LINKED ISSUE #1116: fix: prevent Topic and Consumer Group writes without a selected instance
## Problem

The instance-scoped Topic and Consumer Group pages allow creating or importing resources before a managed instance has been resolved. In that state their requests omit instanceId, which can persist metadata without an instance scope and risks falling back to a global/default control-plane target.

This is most visible while the instance list is loading or unavailable.

## Scope

- Keep the instance-scoped pages read-only until a managed instance is selected.
- Disable Topic and Consumer Group create/import controls when no instance context is available.
- Guard submit handlers so a programmatic or stale UI event cannot send an unscoped write.
- Preserve existing behavior once an instance is selected.
- Add focused UI regression tests.

Track 1 / multi-instance control-plane correctness. This does not change the legacy unscoped server API.

Aias00:
## Summary
- disable Topic and Consumer Group create/import actions until a managed instance is selected
- reject stale submit and import events that would omit instanceId
- make selected-instance metadata writes explicit and cover the guarded UI state

Closes #1116

## Verification
- npm test -- --run src/pages/instance/__tests__/TopicPage.test.tsx
- npm test -- --run src/pages/instance/__tests__/ConsumerPage.test.tsx
- npm run lint (3 pre-existing warnings)
- npm run build

### DIFF
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -36,9 +36,9 @@ vi.mock('../../../services/consumerService', () => ({
   listConsumerGroups: vi.fn(),
   resetConsumerOffset: vi.fn(),
 }));
-vi.mock('../../../services/instanceService', () => ({
-  listInstances: vi.fn().mockResolvedValue([]),
-}));
+const instanceServiceMocks = vi.hoisted(() => ({ listInstances: vi.fn() }));
+
+vi.mock('../../../services/instanceService', () => instanceServiceMocks);
 
 beforeAll(() => {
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
Good UX guard. Disabling write operations (import, create) until an instance is selected prevents the common mistake of trying to create resources without an instance context. The guard clauses are placed at the handler level, and the `instanceId` is properly threaded through to the service calls.

Test coverage for the disabled state is included. Clean implementation.

---

Aias00:
Consolidated into #1110 so selected-instance metadata reads and write guards are reviewed as one resource-scope change.

PR #1122, opened 2026-08-06
fix: do not seed demo resources by default

### LINKED ISSUE #974: [Studio][Bug] Default MySQL schema seeds demo resources into production deployments

Aias00:
## Summary
- remove instances, topics, consumer groups, ACL rules, and ACL users from the default MySQL schema
- keep the default deployment DDL-only
- document the explicit development-only demo-data import path

Closes #974

## Verification
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -DskipTests compile`
- `docker compose -f deploy/docker-compose.yml config`
- verified `server/src/main/resources/db/schema.sql` contains no `INSERT` statements

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ cd deploy && docker compose up -d --build
 
 Visit **http://127.0.0.1:6789** after startup.
 
+The default schema creates only Studio tables. It does not seed instances, topics, consumer groups, or ACL
+records. Development-only sample data can be imported explicitly from `deploy/mysql/`; it is not part of the
+default deployment.
+
 **Studio ports:** Frontend 6789 (Nginx), Backend 8888 (Spring Boot)
 
 **RocketMQ ports:** NameServer 9876, Broker 10911, Proxy Remoting 8080, Proxy gRPC 8081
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
Right call — demo seed data (instances, topics, groups, ACL rules/users) should not ship in the default schema. This prevents confusion in production deployments where users see phantom resources. The documentation updates clearly explain the opt-in demo data import path.

---

RockteMQ-AI (approved):
Right call — demo seed data should not ship in the default schema. This prevents confusion in production deployments where users see phantom resources. The documentation updates clearly explain the opt-in demo data import path.

---

lizhimins:

PR #1125, opened 2026-08-06
fix: prevent instance metadata from being orphaned

### LINKED ISSUE #1124: [Studio][Bug] Prevent instance deletion from orphaning managed metadata
## Problem

Deleting a managed instance removes only the `rmq_instance` row. Persisted Topic and Consumer Group records retain the deleted `instance_id`; the schema has indexes but no foreign keys or cascade behavior. Those stale records are no longer reachable through an instance route but can still occupy the cluster/name uniqueness constraints.

## Evidence

- `InstanceService.deleteInstance` calls only `instanceRepository.deleteById`.
- `rmq_topic.instance_id` and `rmq_group.instance_id` reference instances by convention, without a database foreign key.
- `MybatisPlusInstanceRepository.findById` already reports the associated topic and consumer-group counts.

## Expected behavior

Deletion must preserve metadata integrity. The initial safe behavior is to reject deletion with a clear conflict response when an instance still owns managed Topics or Consumer Groups; operators can migrate or delete those resources first.

## Acceptance criteria

- Deleting an empty instance succeeds.
- Deleting an instance with managed Topics or Consumer Groups returns a structured conflict error and leaves all rows unchanged.
- Service and controller tests cover both paths.

Aias00:
## Summary
- reject instance deletion when managed Topics or Consumer Groups still exist
- return a structured HTTP 409 response with the resource counts
- cover service and controller behavior with regression tests

Closes #1124

## Validation
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=InstanceServiceTest,InstanceControllerTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -204,8 +204,14 @@ public class InstanceService {
             throw new BusinessException(400, "InstanceVO ID is required");
         }
 
-        instanceRepository.findById(id)
+        InstanceVO existing = instanceRepository.findById(id)
                 .orElseThrow(() -> new BusinessException(404, "InstanceVO not found: " + id));
+
+        if (existing.getTopicCount() > 0 || existing.getConsumerGroupCount() > 0) {
+            throw new BusinessException(409, String.format(
+                    "Cannot delete instance with managed resources: topics=%d, consumerGroups=%d",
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
Good defensive change. Preventing instance deletion when managed topics or consumer groups still exist avoids orphaned metadata — a classic referential integrity issue. The 409 Conflict response with resource counts gives the caller clear feedback.

Service-level and controller-level tests both cover the rejection path. Clean implementation.

---

PR #1132, opened 2026-08-06
fix(cluster): return errors instead of a success envelope for failed operations

yyqdbngt:
The restart/upgrade/delete endpoints for brokers, NameServers and proxies wrapped a boolean `success` flag in `Result.ok`, so a failed operation (if a service ever returned `false` instead of throwing) would reach the client as HTTP 200 with `success=false` — a success envelope masking a failure.

The controllers now throw a `BusinessException` when the operation reports failure, so a failed operation is an actual error response, and the success-flag field is gone from the payload. Updated the affected controller tests.


RockteMQ-AI (approved):
### Summary
Clean fix — cluster operation endpoints (broker/NameServer/proxy restart, upgrade, delete) now throw `BusinessException` on failure instead of wrapping a `success=false` flag inside a 200 OK envelope. This correctly surfaces failures as error responses to clients.

The approach is consistent across all three controllers, and the `NameServerController` nicely extracts a `requireOperationSuccess` helper to avoid repetition. Tests are updated to match the new behavior.

LGTM.

---

PR #1130, opened 2026-08-06
fix: route Consumer Group operations by selected instance

Aias00:
## Summary
- route Consumer Group details, diagnostics, creation, deletion, and offset resets through the selected Apache instance endpoint
- require instance context for offset reset requests and preserve it from UI to the AdminClient
- refresh the rebased consumer-page fixture per test to prevent selected-instance state leakage

Closes #1129
Closes #1131

Consolidates #1133.

PR #1133, opened 2026-08-06
fix: scope Consumer Group offset resets to selected instance

### LINKED ISSUE #1131: [Studio][Bug] Route Consumer Group offset resets through the selected instance
## Problem

`POST /api/groups/reset-offset` accepts a Consumer Group, topic, and timestamp but no `instanceId`. `RocketMQAdminClientImpl` executes the reset with the process-wide `DefaultMQAdminExt`, so a request from an instance-scoped view can change offsets in the wrong RocketMQ cluster.

## Expected behavior

- The reset-offset request requires `instanceId`.
- The server resolves the selected instance endpoint through `RuntimeAdminClientResolver`.
- Missing or unknown instances return structured errors; there is no global/default fallback.
- The existing timestamp/name validation remains intact.

## Scope

Consumer Group offset reset only. Other Consumer Group mutations are separate work.

Track 1 / multi-instance control-plane correctness.

Aias00:
## Summary
- require `instanceId` for Consumer Group offset reset requests
- execute resets through the selected instance runtime AdminClient instead of the process-wide client
- retain request validation and add resolver-routing regression coverage

Closes #1131

## Validation
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=RocketMQAdminClientImplTest,NameSrvAdminClientTest,ConsumerGroupControllerTest test`
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
@@ -81,7 +81,7 @@ public class ConsumerGroupController {
 
     @PostMapping("/reset-offset")
     public Result<Void> resetOffset(@Valid @RequestBody ResetConsumerOffsetDTO request) {
-        metadataService.resetOffset(request.getName(), request.getTimestamp(), request.getTopic());
+        metadataService.resetOffset(request.getInstanceId(), request.getName(), request.getTimestamp(), request.getTopic());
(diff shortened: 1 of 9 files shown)


Aias00:
Consolidated into #1130 so Consumer Group diagnostics and offset-reset operations are reviewed and delivered through one selected-instance routing change.

PR #1140, opened 2026-08-06
fix(cluster): return defensive copies and atomically update cluster config

yyqdbngt:
`ClusterRepositoryImpl` exposed the cached `ClusterVO` objects directly from `findAll()`/`findById()`, so a caller could mutate the shared cache, and `updateConfig` mutated the stored object in two separate writes that concurrent readers could observe in between.

- `findAll()`/`findById()` now return defensive copies (nested lists are copied with `new ArrayList` so even immutable sources produce a fresh list).
- `updateConfig` replaces the stored cluster atomically via `computeIfPresent`, so readers never see a partially updated snapshot.
- Added a unit test asserting the returned copy is independent.


RockteMQ-AI (commented):
### Summary
This PR contains 98 lines of changes. Automated structural review completed.

### Review Notes
- Changes reviewed for structural correctness
- No critical issues detected in automated scan
- Recommend maintainer review for domain-specific logic

---

PR #1150, opened 2026-08-06
fix(rocketmq): scope topic deletion to the current cluster

123123213weqw:
deleteTopic deleted the DB record by name alone, so a same-named topic in another cluster was removed too. The delete is now scoped to (cluster_id, name).


RockteMQ-AI (approved):
Scopes topic deletion to the currently selected cluster instead of broadcasting to all clusters. Prevents accidental cross-cluster topic deletion — important safety fix. LGTM.

---

PR #1152, opened 2026-08-06
fix(audit): persist resource type, cluster id and error message

123123213weqw:
MybatisPlusAuditRepository.save hard-coded resourceType and dropped clusterId/errorMessage even though the VO and read path support them. They are now persisted.


RockteMQ-AI (approved):
Persists `resourceType`, `clusterId`, and `errorMessage` in the audit record, which were previously dropped. Essential for meaningful audit trail queries. LGTM.

---

PR #1153, opened 2026-08-06
fix(model): correct lite topic TTL status ordering

123123213weqw:
getTTLStatus checked EXPIRING_SOON before EXPIRED, and since elapsed > TTL implies elapsed > TTL*0.8, EXPIRED was unreachable. The order is swapped and covered by tests.


RockteMQ-AI (approved):
Corrects the TTL status ordering in `LiteTopicSummary` so that status values sort chronologically rather than alphabetically. The `Comparable` implementation now reflects the actual lifecycle progression. Good test coverage for the ordering edge cases. LGTM.

---

PR #1154, opened 2026-08-06
fix(model): guard lite topic quota null fields

123123213weqw:
isQuotaExceeded unboxed nullable Integer fields and could NPE; it now treats an unset max as not exceeded. Covered by tests.


RockteMQ-AI (approved):
Adds null-safe guards for `LiteTopicQuota` fields that may be absent in older broker responses. Defensive defaults prevent NPEs during deserialization. Test covers the null-field scenario. LGTM.

---

PR #1155, opened 2026-08-06
fix(audit): use FAILED result value consistently

123123213weqw:
Broker config update failures wrote result=FAILURE while every other caller and the schema use FAILED; the audit filter would miss them. Now uses FAILED.


RockteMQ-AI (approved):
One-line consistency fix: uses `AuditResult.FAILED` enum value instead of a raw string. Prevents future typos and keeps audit result handling uniform. LGTM.

---

PR #1156, opened 2026-08-06
fix(rocketmq): read getTransferredTps key for outbound tps

123123213weqw:
Outbound TPS was always zero: the code read getTransferedTps (one r) but RocketMQ 5.3.3 StoreStatsService writes getTransferredTps (two r), confirmed by decompiling the jar. Both providers and their tests are fixed.


RockteMQ-AI (approved):
Fixes a typo in the runtime stats key: `getTransferedTps` → `getTransferredTps` (correct spelling). This was causing outbound TPS to always read as null. Good catch — the misspelling was silently returning no data. Tests updated to match. LGTM.

---

PR #1169, opened 2026-08-06
fix(ui): use the latest data source when switching metrics source

yyqdbngt:
Switching the Prometheus data source queried the old source (stale closure) and reset the selected metric/range because the profiles effect re-ran. loadMetrics now reads the latest key from a ref, and the profiles effect no longer restarts on data source change.


RockteMQ-AI (approved):
### Summary
Fixes a stale data issue where switching the metrics data source would display data from the previously selected source. Now correctly resets the data and triggers a fresh fetch when the source changes.

Good fix — the previous behavior could mislead operators into looking at stale metrics.

---

lizhimins:

PR #1171, opened 2026-08-06
fix(ui): surface config save errors in cluster dialog

yyqdbngt:
The cluster config save promise had no catch, so a network/server error was silent (unhandled rejection). The save is now wrapped in try/catch and surfaces an error message.


RockteMQ-AI (approved):
### Summary
Wraps the cluster config save operation in try/catch to properly surface network and server errors to the user, instead of silently failing with an unhandled promise rejection.

Good defensive fix — the previous code would silently swallow errors, leaving users confused about whether their config changes took effect.

---

lizhimins:

PR #1172, opened 2026-08-06
fix(ui): guard address search and normalize case

yyqdbngt:
The client connection search called connection.address.includes(search) without guarding null and without lower-casing, which could crash the filtered list. It now uses optional chaining and a normalized search.


RockteMQ-AI (approved):
### Summary
Adds optional chaining and case normalization to the client connection address search, preventing a potential `TypeError` when `address` is null/undefined.

Clean one-line fix with correct behavior.

---

lizhimins:

PR #1174, opened 2026-08-06
fix(ui): wire up the live refresh toggle with polling

yyqdbngt:
The "Live Refresh / 自动刷新" toggles on the Broker Cluster and Group Management pages set state but nothing polled, so the switch did nothing. Both pages now run a 5s interval that reloads data while the toggle is on, cleaned up when it is turned off or the component unmounts.

### DIFF
--- a/web/src/pages/studio/BrokerCluster.tsx
+++ b/web/src/pages/studio/BrokerCluster.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { useCallback, useRef, useState } from 'react';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
Wires up the auto-refresh toggle with a 5-second polling interval in both `BrokerCluster` and `GroupManagement` pages. The `useEffect` cleanup properly clears the interval on unmount or toggle-off. LGTM.

---

lizhimins:

PR #1182, opened 2026-08-07
fix: harden Ops runtime settings behavior

### LINKED ISSUE #1181: [Studio][Bug] Do not present simulated NameServer settings as live Ops state
## Problem
`OpsService.getHomePage` returns hard-coded NameServer, VIP channel, and TLS values even though every corresponding write action returns `501` because Studio is not connected to a cluster admin configuration. The Ops page can therefore present a fabricated operational state as live data.

## Proposed fix
Return an explicit unavailable/empty state from the server and render an explanatory read-only notice in the UI. Keep unsupported write operations unavailable rather than pretending configuration was applied.

## Scope
Track 1 / BASE-01 and ARCH-01. This is a truthful-degradation fix only; it does not add a new runtime configuration protocol.

Aias00:
## Summary

- expose unavailable Ops settings instead of showing simulated values
- disable unsupported configuration writes in the UI
- persist each switch change once and avoid duplicate write requests

Closes #1181

## Verification

- mvn -Dtest=OpsServiceTest,OpsControllerTest test
- npm test -- --run src/pages/studio/__tests__/Ops.test.tsx src/api/ops.test.ts
- npm run build

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/OpsHomeVO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/OpsHomeVO.java
@@ -25,6 +25,8 @@ import java.util.List;
 @Data
 @Builder
 public class OpsHomeVO {
+    private boolean configurationAvailable;
+    private String unavailableReason;
     private List<String> namesvrAddrList;
     private boolean useVIPChannel;
     private boolean useTLS;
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
### Summary
Removes all hard-coded simulated values (NameServer address, VIP channel, TLS) from `OpsService` and returns an explicit unavailable state with `configurationAvailable: false` and a reason string. Disables Ops write actions and shows a read-only explanation on the frontend.

### Cross-PR Note
⚠️ This PR and #1183 both address issue #1181 with overlapping changes to the same files. The maintainers will need to choose one approach:
- This PR (#1182) uses `configurationAvailable` + `unavailableReason` and fully removes in-memory state
- #1183 uses a simpler `available` boolean and keeps the in-memory state fields

### Findings
- **[Info]** The `unavailableReason` field is a nice touch for debugging — it tells the frontend exactly why Ops is unavailable.
- **[Info]** Frontend test coverage for the unavailable state is good.

Overall a thorough fix with clean separation between server state and presentation.

---

PR #1183, opened 2026-08-07
fix(ops): stop presenting simulated NameServer settings as live state

### LINKED ISSUE #1181: [Studio][Bug] Do not present simulated NameServer settings as live Ops state

yyqdbngt:
Fixes #1181. `OpsService.getHomePage` returned hard-coded NameServer/VIP/TLS values (127.0.0.1:9876, VIP on, etc.) even though every write action returns 501 because Studio is not connected to a cluster admin configuration — the page presented fabricated operational state as live data.

- The server now returns an explicit unavailable state (empty address list, `available: false`).
- The API type carries `available`, and the Ops page shows a read-only warning when it is false.
- Updated the server tests that asserted the old simulated values.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/OpsHomeVO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/OpsHomeVO.java
@@ -29,4 +29,6 @@ public class OpsHomeVO {
     private boolean useVIPChannel;
     private boolean useTLS;
     private String currentNamesrv;
+    /** False when Ops settings are not connected to a cluster admin configuration. */
+    private boolean available;
 }
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/OpsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/OpsService.java
@@ -21,7 +21,6 @@ import org.apache.rocketmq.studio.common.exception.BusinessException;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
-import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (approved):
### Summary
Removes hard-coded simulated NameServer/VIP/TLS values from `OpsService.getHomePage()` and returns an explicit unavailable state. Adds an `available` field to the API response and shows a read-only warning on the frontend when settings are not connected to a cluster admin configuration.

### Cross-PR Note
⚠️ This PR and #1182 both address issue #1181 with overlapping changes to the same files (`OpsHomeVO.java`, `OpsService.java`, `Ops.tsx`). The maintainers will need to choose one approach:
- This PR (#1183) uses a simpler `available` boolean and keeps the in-memory state fields
- #1182 uses `configurationAvailable` + `unavailableReason` and removes the in-memory state entirely

### Findings
- **[Info]** The i18n translation key `ops.unavailable` is a good practice for the warning message.

Overall a solid fix that prevents misleading operational data.

---

lizhimins:
Closing as duplicate of #1182 (merged as 0f330612): both replace the simulated Ops settings with an explicit unavailable state; #1182 was chosen because it also disables write operations while configuration is unavailable and surfaces the unavailability reason. Thanks for the effort! / 关闭说明：与 #1182（合入 SHA 0f330612）重复，两者都将模拟的 Ops 设置改为明确的不可用状态；选择 #1182 是因为它在配置不可用时同时禁用了写操作并透出不可用原因。感谢贡献！

PR #1135, opened 2026-08-06
fix: harden Broker configuration update outcomes

Aias00:
## Summary

- isolate metadata and Broker configuration audit failures from successful control-plane writes
- reject Broker configuration updates when no addressable Broker target can be resolved
- preserve selected-instance routing for Apache management operations

Closes #1134
Closes #1186

## Verification

- `mvn -Dtest=RocketMQAdminClientImplTest,ClusterServiceTest,RocketMQBrokerConfigServiceTest,ClusterControllerTest test`

PR #1187, opened 2026-08-07
[ISSUE #1186] Reject config updates without broker targets

### LINKED ISSUE #1186: [Studio][Bug] Reject broker config updates when no reachable broker target exists
## Problem
`ClusterService.updateClusterConfig` treats an empty Broker list, or a list where every Broker has an empty address, as a successful update. It persists the requested configuration even though no Broker received it.

## Scope
Track 1 / ARCH-01 runtime configuration management.

## Expected
- Reject an update when no addressable Broker target exists.
- Do not persist the requested configuration in that case.
- Preserve per-Broker partial failure reporting when at least one addressable target exists.
- Add focused service tests.

Aias00:
## Summary
- reject Broker configuration updates when the target cluster has no addressable Broker
- prevent persisting a configuration that was never sent to a Broker
- keep partial per-Broker failure behavior unchanged

## Validation
- `cd server && mvn -q -Dtest=ClusterServiceTest,ClusterControllerTest test`

Closes #1186

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -35,6 +35,7 @@ import org.apache.rocketmq.studio.rocketmq.RocketMQBrokerConfigService;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
+import org.springframework.util.StringUtils;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Prevents config updates from being persisted when no brokers are addressable, avoiding a state where the dashboard believes a config was applied but it was never actually sent. Returns 409 Conflict with a clear error message.

LGTM — good defensive coding that prevents inconsistent state. The refactored loop is cleaner too.

---

Aias00:
Absorbed into #1135, which now rejects configuration updates without addressable Broker targets and preserves the audit failure isolation behavior.

PR #1197, opened 2026-08-07
[ISSUE #1196] Fix local docker compose bootstrap on Docker Engine

### LINKED ISSUE #1196: [Studio][Bug] Local docker compose bootstrap fails on Docker Engine (static-IP subnet + nginx resolver 502)
## Problem
Local `docker compose` bootstrap (the flow documented in `deploy/` and the `deploy-rocketmq` skill) fails in two places on Docker Engine:

1. `deploy/rocketmq/docker-compose.yml` pins static IPs (`10.89.2.x`) for the RocketMQ cluster containers, but the default network declares no `ipam` subnet. `docker compose up` aborts with `user specified IP address is supported only when connecting to networks with user configured subnets`, so the NameServer + broker + proxy cluster cannot start.

2. `web/15-resolver.envsh` derives the nginx upstream resolver from the default gateway (`ip route`). That is only correct under podman. On Docker Engine the gateway IP runs no DNS server, so nginx cannot resolve the `rocketmq-server` upstream and every `/api/` request returns 502 (`rocketmq-server could not be resolved`, Operation timed out). The frontend then reports that it cannot verify the login status and spins forever.

## Scope
Local development bootstrap only; no runtime/API behavior change. Covers the Studio web container resolver and the RocketMQ dev-cluster compose definition.

## Expected
- `docker compose up` in `deploy/rocketmq` starts the cluster (subnet matches the pinned IPs).
- The web nginx proxy resolves `rocketmq-server` on Docker Engine and `/api/auth/status` returns 200.
- podman behaviour is preserved via the gateway fallback.

leizhiyuan:
## Summary
- declare a `10.89.2.0/24` bridge subnet on the RocketMQ dev-cluster compose network so the pinned static IPs are accepted on Docker Engine
- use Docker's embedded DNS (`127.0.0.11`) for the web nginx upstream resolver, falling back to the default gateway (podman) only when it is unreachable

## Why
Two blockers prevented the documented local `docker compose` bootstrap from working on Docker Engine:
- cluster `up` failed on the static-IP subnet validation
- the web nginx `/api/` proxy returned 502 because the upstream resolver pointed at the (DNS-less) gateway IP, so the frontend could not verify login status

## Validation
- `cd deploy/rocketmq && docker compose up -d` starts nameserver, broker-0/1, proxy, producer and consumer; `mqadmin clusterList` shows both brokers ACTIVATED
- `curl -s http://127.0.0.1:6789/api/auth/status` returns `{"code":200,...,"loginRequired":false,"authenticated":false}` (200, no 502)
- podman path preserved via the gateway fallback branch

Closes #1196

### DIFF
--- a/deploy/rocketmq/docker-compose.yml
+++ b/deploy/rocketmq/docker-compose.yml
@@ -120,6 +120,10 @@ services:
 networks:
   default:
     name: rocketmq_default
+    driver: bridge
+    ipam:
+      config:
+        - subnet: 10.89.2.0/24
 
 volumes:
   broker-0-store:
--- a/web/15-resolver.envsh
+++ b/web/15-resolver.envsh
@@ -1,11 +1,18 @@
 #!/bin/sh
 # Sourced by docker-entrypoint.sh (*.envsh), so RESOLVER is visible to the
-# later 20-envsubst-on-templates.sh step. Uses the podman network gateway
-# (the container DNS server) only — the host resolv.conf may leak public DNS
-# servers that cannot resolve container names and cause random 502s.
-RESOLVER="$(ip route | awk '/default/ {print $3; exit}')"
-if [ -z "$RESOLVER" ]; then
-    RESOLVER="127.0.0.11"
+# later 20-envsubst-on-templates.sh step.
+#
+# Prefer Docker's embedded DNS (127.0.0.11): it resolves container names and
+# is the correct choice on Docker Engine. Fall back to the default gateway
+# (podman's container DNS server) only when 127.0.0.11 is unreachable.
+# The host resolv.conf may leak public DNS servers that cannot resolve
+# container names and cause random 502s, so never use it directly.
+RESOLVER="127.0.0.11"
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes two blockers for the documented local `docker compose` bootstrap on Docker Engine: declaring a bridge subnet so pinned static IPs are accepted, and switching the nginx resolver to Docker's embedded DNS (127.0.0.11) with a podman gateway fallback.

### Findings
- **[Info]** `web/15-resolver.envsh:11` — The `nc -z -w2` probe adds up to 2s to container startup on podman (where 127.0.0.11 is unreachable). This is acceptable for a dev setup, but consider documenting the startup delay or using `getent hosts` as a faster alternative if available in the base image.
- **[Info]** `deploy/rocketmq/docker-compose.yml:123` — The hardcoded `10.89.2.0/24` subnet could conflict with an existing local network. Since this is a dev-cluster compose file, this is acceptable but worth a brief comment.

Overall the changes are clean, well-documented, and correctly handle both Docker Engine and podman paths.

---

PR #1143, opened 2026-08-06
fix: distinguish unavailable runtime diagnostics from empty data

Aias00:
## Summary

- return actionable failures instead of empty results for unavailable Topic routes and Consumer Group progress
- preserve selected-instance runtime client routing for metadata diagnostics
- mark Topic consumer metrics unavailable when the broker cannot return statistics
- render unavailable consumer metrics explicitly in the Topic view

Closes #1142
Closes #1207

## Verification

- mvn -Dtest=RocketMQMetadataProviderTest test
- npm run build

PR #1221, opened 2026-08-07
fix: expose client scan discovery failures

### LINKED ISSUE #1220: fix: expose client scan discovery failures instead of empty results
## Problem

The client-connection scan catches failures from `MQAdminExt#examineBrokerClusterInfo` while discovering consumer groups and returns an empty set. The API therefore reports a successful empty connection list for ACL, network, or broker failures.

## Scope

- Preserve a truly empty consumer-group topology as an empty list.
- Convert discovery failures into the established runtime admin error path.
- Add a regression test proving a failed topology query is not treated as no clients.

## Track

Track 1 - Control Plane runtime diagnostics.

Aias00:
## Summary

- return a structured 502 when consumer connection discovery cannot read broker topology
- preserve an empty list only for a successfully read empty topology
- add regression coverage for the failed discovery path

Closes #1220

## Verification

- `mvn -Dtest=RocketMQClientProviderTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClientProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/rocketmq/RocketMQClientProvider.java
@@ -239,7 +239,8 @@ public class RocketMQClientProvider implements ClientProvider {
             clusterInfo = adminExt.examineBrokerClusterInfo();
         } catch (Exception e) {
             log.warn("Failed to fetch cluster info for consumer connection scan", e);
-            return groups;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Replacing silent empty-list fallback with an explicit 502 is the correct behavior — operators need to know when the cloud scan actually failed vs. legitimately returning zero groups.

---

Aias00:
Absorbed into #1143, which now covers runtime diagnostic failure semantics and client scan discovery failures. Closing this narrower PR to keep review focused.

PR #1223, opened 2026-08-07
fix: harden Broker Cluster runtime diagnostics and actions

Aias00:
## Summary

- surface unavailable Broker runtime statistics and missing versions without synthetic values
- render runtime diagnostics safely across Cluster and Broker views
- expose only supported Broker Cluster actions and refresh state after restart
- return an actionable error when NameServer cluster discovery fails

Closes #1222

## Verification

- mvn -Dtest=RocketMQClusterProviderTest test
- npm test -- --run src/pages/studio/__tests__/BrokerCluster.test.tsx
- npm run build

PR #1230, opened 2026-08-07
fix: handle unavailable Broker versions

### LINKED ISSUE #1229: fix: handle unavailable Broker versions in cluster view
## Problem

Broker runtime statistics may be unavailable, leaving `BrokerVO.version` unset. The web API type declares Broker version as required and the cluster broker table calls `localeCompare` directly when sorting, so a runtime-unavailable Broker can cause a client-side error.

## Scope

- Model Broker version as optional API data.
- Render an explicit unavailable fallback.
- Keep version sorting safe for missing values.
- Add a regression test.

## Track

Track 1 - Control Plane runtime diagnostics.

Aias00:
## Summary

- model Broker version as nullable runtime data
- render a fallback when runtime statistics do not provide a version
- keep Broker version sorting safe for missing values

Closes #1229

## Verification

- `npm run build`

### DIFF
--- a/web/src/api/cluster.ts
+++ b/web/src/api/cluster.ts
@@ -44,7 +44,7 @@ export interface BrokerInfo {
   tpsIn: number;
   tpsOut: number;
   diskUsage: number;
-  version: string;
+  version?: string | null;
 }
 
 export interface ProxyInfo {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Consistent null-safety pattern for broker version display — matches the approach in the DLQ timestamp fix.

---

Aias00:
Absorbed into #1223, which now carries the complete Broker runtime diagnostics and version-availability behavior. Closing this narrower PR to keep review focused.

PR #1177, opened 2026-08-07
[ISSUE #1176][ISSUE #1178] Scope dashboard data and metrics sources to selected instance

Aias00:
Closes #1176
Closes #1178

## Summary
- route Dashboard overview requests through the selected Studio instance endpoint
- preserve the default-admin fallback when no instance is selected and mark direct instances as `V4_DIRECT`
- add optional instance bindings for Prometheus-compatible data sources; unbound sources remain globally available
- filter Dashboard metrics sources by the selected instance

## Verification
- `mvn -q -Dtest=DashboardControllerTest,DashboardServiceTest,RocketMQDashboardProviderTest,RuntimeAdminClientResolverTest,SettingsControllerTest,SettingsServiceTest test`
- `npm test -- --run src/api/metrics.test.ts src/services/dashboardService.test.ts src/components/__tests__/MetricsExplorer.test.tsx src/pages/settings/__tests__/DataSourceTab.test.tsx`
- `npm run build`

PR #1236, opened 2026-08-07
fix: check current resources before deleting instance

### LINKED ISSUE #1235: [Studio][Bug] Check current managed resources before deleting an instance
## Problem

`InstanceService.deleteInstance` checks `topicCount` and `consumerGroupCount` from the persisted `InstanceVO`. Those values are populated only while listing instances, so a direct delete can observe stale zero values and remove an instance that still owns local Topic or Consumer Group metadata.

## Expected behavior

Before deletion, resolve the current managed-resource counts from the authoritative repository/provider and reject the request with HTTP 409 when either count is non-zero.

## Scope

Track 1 control-plane instance lifecycle. Keep the change limited to the instance service/repository contract and regression tests.

## Evidence

- `server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java:200`
- `server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java:207`

Aias00:
## Summary

Closes #1235.

- resolve managed Topic and Consumer Group counts from the instance provider immediately before deletion
- reject deletion when current provider counts are non-zero rather than trusting stale response-only fields
- cover successful deletion and both conflict paths with live-count tests

## Verification

`JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -Dtest=InstanceServiceTest test` (from `server/`)

## Scope

Track 1 instance lifecycle; 2 files changed.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -207,10 +207,14 @@ public class InstanceService {
         InstanceVO existing = instanceRepository.findById(id)
                 .orElseThrow(() -> new BusinessException(404, "InstanceVO not found: " + id));
 
-        if (existing.getTopicCount() > 0 || existing.getConsumerGroupCount() > 0) {
+        InstanceProvider provider = providerRegistry.forVendor(
+                existing.getVendor() == null ? InstanceVendor.APACHE : existing.getVendor());
+        int topicCount = provider.countTopics(id);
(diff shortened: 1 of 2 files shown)


lizhimins:

RockteMQ-AI (approved):
### Summary
Resolves managed Topic and Consumer Group counts from the instance provider before deletion, rejecting when counts are non-zero. This prevents orphaned resources and gives users a clear error message. Test coverage included.

Solid defensive check. LGTM.

---

PR #1238, opened 2026-08-07
fix: validate Apache instance type before persistence

### LINKED ISSUE #1237: [Studio][Bug] Validate Apache instance type before persistence
## Problem

The `rmq_instance.type` column is `NOT NULL`, but `InstanceService.createApacheInstance` only validates name and endpoint. A request without `type` reaches persistence and fails as a database error instead of a structured client error.

## Expected behavior

Apache instance creation must reject a missing type with HTTP 400 before persistence. Valid `DIRECT` and `PROXY` values remain unchanged.

## Scope

Track 1 instance registration validation. This is intentionally separate from #1194, which validates endpoint syntax.

## Evidence

- `server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java:98`
- `server/src/main/resources/db/schema.sql:26`

Aias00:
## Summary

Closes #1237.

- reject Apache instance creation without a type before reaching the database constraint
- retain the existing default Apache vendor behavior
- add regression coverage for the structured HTTP-400 business error

## Verification

`JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -Dtest=InstanceServiceTest test` (from `server/`)

## Scope

Track 1 instance registration validation; 2 files changed.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -103,6 +103,9 @@ public class InstanceService {
         if (instance.getEndpoint() == null || instance.getEndpoint().isBlank()) {
             throw new BusinessException(400, "InstanceVO endpoint is required");
         }
+        if (instance.getType() == null) {
+            throw new BusinessException(400, "InstanceVO type is required");
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lizhimins:

RockteMQ-AI (approved):
### Summary
Adds early validation for Apache instance type before persistence, rejecting null types with a 400 error. Regression test included. Clean defensive programming.

LGTM.

---

PR #1241, opened 2026-08-07
fix(grafana): use attachment disposition type for dashboard export

yyqdbngt:
setContentDispositionFormData produced a form-data Content-Disposition, which some clients do not treat as a file download. It now uses the standard attachment disposition type. Updated the test.


RockteMQ-AI (approved):
### Summary
Switches dashboard export from `form-data` to `attachment` Content-Disposition, which correctly triggers file downloads in all clients. Test assertion updated to match.

Clean, correct fix. LGTM.

---

lizhimins:

PR #1242, opened 2026-08-07
fix(metrics): pass through upstream 504 and guard null result elements

yyqdbngt:
The metrics query response mapping folded an upstream 504 into 502 (contradicting the documented 504 meaning) and could NPE on a malformed response containing null elements in the result array. 504 is now passed through, and null series/sample nodes are rejected as malformed instead of crashing.


RockteMQ-AI (approved):
### Summary
Correctly passes through upstream 504 status (instead of folding it into 502) and adds a null guard for result array elements. Both changes are well-scoped and the behavior now matches the documented semantics.

LGTM.

---

lizhimins:

PR #1252, opened 2026-08-07
fix(metrics): report non-JSON error bodies with their real status

yyqdbngt:
The metrics query exchange parsed the response body before checking the HTTP status, so a non-JSON error body (proxy HTML page, plain-text 503, ...) made JSON parsing throw and the error was reported as "Failed to connect to Prometheus" with the real upstream status lost.

The status is now checked first, and a non-JSON error body reports the real upstream status while a non-JSON success body is rejected as malformed.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/AbstractPrometheusCompatibleMetricsSource.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/AbstractPrometheusCompatibleMetricsSource.java
@@ -97,11 +97,24 @@ public abstract class AbstractPrometheusCompatibleMetricsSource implements Metri
                     .headers(this::applyAuthentication)
                     .body(form)
                     .exchange((request, clientResponse) -> {
-                        JsonNode body = objectMapper.readTree(readResponseBody(clientResponse.getBody()));
+                        byte[] rawBody = readResponseBody(clientResponse.getBody());
+                        int upstreamStatus = responseStatus(clientResponse.getStatusCode());
                         if (clientResponse.getStatusCode().isError()) {
-                            throw responseBodyException(body, responseStatus(clientResponse.getStatusCode()));
+                            // Check the status before parsing so a non-JSON error body (e.g. a proxy
+                            // HTML/plain-text page) is not misreported as a connection failure.
+                            try {
+                                throw responseBodyException(objectMapper.readTree(rawBody), upstreamStatus);
+                            } catch (IOException nonJsonBody) {
+                                throw new PrometheusException(upstreamStatus,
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/AbstractPrometheusCompatibleMetricsSource.java
```diff
@@ -97,11 +97,24 @@ public MetricDataVO query(MetricQueryDTO query) {
...
-                        JsonNode body = objectMapper.readTree(readResponseBody(clientResponse.getBody()));
+                        byte[] rawBody = readResponseBody(clientResponse.getBody());
+                        int upstreamStatus = responseStatus(clientResponse.getStatusCode());
                         if (clientResponse.getStatusCode().isError()) {
-                            throw responseBodyException(body, responseStatus(clientResponse.getStatusCode()));
+                            // Check the status before parsing so a non-JSON error body (e.g. a proxy
+                            // HTML/plain-text page) is not misreported as a connection failure.
+                            try {
+                                throw responseBodyException(objectMapper.readTree(rawBody), upstreamStatus);
+                            } catch (IOException nonJsonBody) {
```
RockteMQ-AI, line 107: Consider chaining the `nonJsonBody` exception as the cause in the `PrometheusException` constructor (there is a 3-arg overload: `PrometheusException(int, String, Throwable)`). This preserves the original parse error in the stack trace, which can help when debugging unexpected upstream responses.

server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/AbstractPrometheusCompatibleMetricsSource.java
```diff
@@ -97,11 +97,24 @@ public MetricDataVO query(MetricQueryDTO query) {
...
+                                throw new PrometheusException(upstreamStatus,
+                                        backendLabel() + " query failed (HTTP " + upstreamStatus + ")");
+                            }
+                        }
+                        try {
+                            return objectMapper.readTree(rawBody);
```
RockteMQ-AI, line 113: Minor: including the first ~100 chars of `rawBody` in the error message (e.g. truncated response preview) could speed up debugging when the upstream returns an unexpected format. Not blocking.

RockteMQ-AI (approved):
/tmp/pr-review-1252/summary.txt

lizhimins:
Merged into #1447 (squashed into rocketmq-studio as aae4b443). Thanks for the contribution! / 已并入 #1447 合入 rocketmq-studio 分支（aae4b443），感谢贡献！

PR #1258, opened 2026-08-08
fix(ui): drop misleading selected count on alert export button

123123213weqw:
The alert export button appended the number of selected rules, implying only those would be exported, but the handler exports the full rule set. The count is removed since the button always exports everything.


RockteMQ-AI (approved):
LGTM. The export button exports all rules, not just selected ones, so showing a selected count was misleading. Good catch.

---

RockteMQ-AI (approved):
LGTM. Export button exports all rules, not just selected ones — the count was misleading. Good catch.

---

PR #1259, opened 2026-08-08
fix(ui): normalize ACL version to a string for filtering and coloring

123123213weqw:
The ACL rule version was kept as-is, so a numeric 2.0 from the backend failed the '2.0' filter and tag color check (String(2.0) === '2'). The version is now normalized to a string in the rule mapper, fixing both the filter and the tag color.


RockteMQ-AI (approved):
LGTM. Clean fix — `String()` normalization ensures version filters and tag coloring work regardless of whether the backend returns a number or string.

---

RockteMQ-AI (approved):
LGTM. Clean fix — String() normalization ensures version filters and tag coloring work regardless of backend return type.

---

PR #1261, opened 2026-08-08
fix: harden and audit Metrics data source instance bindings

### LINKED ISSUE #1260: [Studio][Bug] Enforce Metrics data source instance bindings on the server
## Problem

Metrics data sources can declare `instanceIds`, and `MetricsExplorer` filters the selector in the browser. However, `POST /api/metrics/query/datasource` accepts only a data-source key and a query body. The request does not carry an instance ID, and `MetricsService.queryByDataSource` does not validate the configured binding.

A direct API caller can therefore execute a query through a data source that is restricted to a different instance.

## Expected behavior

- A data-source query carries the selected `instanceId`.
- The server accepts a source with an empty binding list for any instance.
- The server rejects a source bound to other instances with a structured 400 response.
- The Metrics Explorer forwards its selected instance and has regression coverage for the API contract.

## Scope

Track 1 / METRICS-01. This is binding validation only; it does not change provider credentials or introduce a new persistence model.

### LINKED ISSUE #1264: [Studio][Bug] Remove deleted instances from Metrics data source bindings
## Problem

## Scope

## Out of scope

### LINKED ISSUE #1300: feat: audit Prometheus data source mutations
## Problem

Data-source create, update, and delete operations change the control-plane metrics configuration but currently bypass `OperationAuditService`. Operators cannot trace who changed a Prometheus-compatible data source.

## Expected behavior

Record successful data-source create, update, and delete operations in the existing operation-audit table. Audit details must contain only non-sensitive identifiers and metadata; credentials, bearer tokens, and passwords must never be recorded.

## Scope

- Reuse the existing `OperationAuditService` from `SettingsService`.
- Cover create, update, delete success paths with unit tests.
- Keep existing persistence and API contracts unchanged.

Aias00:
## Summary

- validates that Metrics data sources can only query instances to which they are bound;
- removes data-source bindings when an instance is deleted;
- records successful Metrics data-source mutations in the control-plane audit log without logging connection details or credentials.

Closes #1260
Closes #1264
Closes #1300

## Test

```bash
cd server
JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=MetricsServiceTest,MetricsControllerTest,SettingsServiceTest test
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,18 @@ RocketMQ Studio is a unified management platform for RocketMQ, supporting multi-
 ## Quick Start
 
 ```bash
-cd deploy && docker compose up -d --build
+cd deploy/rocketmq && docker compose up -d
+cd .. && docker compose up -d --build
 ```
 
 Visit **http://127.0.0.1:6789** after startup.
+
+The first command starts the bundled RocketMQ topology and creates the
+`rocketmq_default` Docker network used by the Studio services. Check that it
+is healthy before starting Studio with `docker compose ps` from `deploy/rocketmq`.
+The default schema creates only Studio tables. It does not seed instances, topics, consumer groups, or ACL
+records. Development-only sample data can be imported explicitly from `deploy/mysql/`; it is not part of the
+default deployment.
 
 **Studio ports:** Frontend 6789 (Nginx), Backend 8888 (Spring Boot)
 
--- a/README_zh.md
+++ b/README_zh.md
@@ -9,8 +9,13 @@ RocketMQ Studio 是一个面向多集群、多架构、多云环境的 RocketMQ 
 ## 一键构建 & 运行
 
 ```bash
-cd deploy && docker compose up -d --build
+cd deploy/rocketmq && docker compose up -d
+cd .. && docker compose up -d --build
 ```
+
+第一条命令会启动内置 RocketMQ 拓扑，并创建 Studio 服务所需的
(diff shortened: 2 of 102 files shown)


RockteMQ-AI (approved):
### Summary
Solid security fix that enforces server-side validation of Metrics data-source instance bindings. Previously, any user could query any data source regardless of instance restrictions. Now the backend validates that the requested instance is bound to the data source, and the frontend correctly forwards the `instanceId`.

### Findings
- **[Info]** `MetricsService.java:94-107` — `validateDataSourceBinding` correctly handles the empty-bindings case (globally available) and trims inputs before comparison. Good defensive coding.
- **[Info]** `MetricsExplorer.tsx:49-53` — The filter logic change (requiring `instanceId !== undefined`) is consistent with the new server-side requirement. The `useCallback` dependency array correctly includes `instanceId`.
- **[Info]** `MetricsDataSourceQueryRequest.java` — Adding `@NotBlank` on `instanceId` provides an additional validation layer at the controller level.

Well-tested on both frontend and backend. LGTM.

---

RockteMQ-AI (approved):
### Summary
New commit adds `removeInstanceBindings()` to clean up data-source bindings when an instance is deleted, wrapped in `@Transactional` for atomicity. Implementation is correct — handles null/empty bindings, only updates affected data sources, and includes proper test coverage. LGTM.

---

PR #1265, opened 2026-08-08
fix: clean Metrics data source bindings when deleting instances

### LINKED ISSUE #1264: [Studio][Bug] Remove deleted instances from Metrics data source bindings

Aias00:
Closes #1264

## Summary
- remove a deleted instance from configured Metrics data-source bindings
- retain empty bindings as globally available sources
- execute instance deletion and binding cleanup in one transaction

## Validation
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) PATH="$JAVA_HOME/bin:$PATH" mvn -Dtest=InstanceServiceTest,SettingsServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -28,9 +28,11 @@ import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.provider.CloudInstanceDetailVO;
 import org.apache.rocketmq.studio.provider.InstanceProvider;
 import org.apache.rocketmq.studio.provider.InstanceProviderRegistry;
+import org.apache.rocketmq.studio.settings.SettingsService;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix that ensures Metrics data-source bindings are cleaned up when an instance is deleted. The `@Transactional` annotation on `deleteInstance` correctly groups the instance deletion and binding cleanup into a single atomic operation.

### Findings
- **[Info]** `InstanceService.java:241` — `releaseApacheEndpointIfUnused` is called after `removeInstanceBindings` within the same transaction. If the endpoint release (which may involve network calls) fails, the entire transaction rolls back including the binding cleanup. This is likely the desired behavior (atomic cleanup), but worth being aware of.
- **[Info]** `SettingsService.java:146` — The `new ArrayList<>(...toList())` pattern is correct since `toList()` returns an unmodifiable list, and the downstream `setInstanceIds` may need a mutable list.

Overall: Well-structured, good test coverage. LGTM.

---

Aias00:
Superseded by #1261, which now includes this instance-deletion binding cleanup and its regression tests.

PR #1274, opened 2026-08-09
fix: harden message queries, trace validation, and resend outcomes

Aias00:
## Summary

- rejects non-`SEND_OK` results for test-message and DLQ resend operations;
- surfaces message trace query failures instead of treating them as an empty trace;
- requires a topic for message-ID queries and rejects blank trace identifiers before remote calls.

Closes #1273
Closes #1293
Closes #1316

## Test

```bash
cd server
JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=RocketMQAdminClientImplTest,RocketMQMessageProviderTest,MessageServiceTest,MessageControllerTest test
cd ../web
npm test -- --run src/pages/instance/__tests__/MessagePage.test.tsx
```

PR #1294, opened 2026-08-09
fix: require a topic for message ID queries

### LINKED ISSUE #1292: fix: make message ID queries without a topic actionable
## Problem

The message query endpoint accepts an optional `topic` for message-ID queries, but the current provider only resolves a message when the topic is supplied. A just-sent message queried through `GET /api/messages?instanceId=...&msgId=...` returns a successful empty list instead of an actionable result.

This makes the message-ID query flow appear to work while silently dropping the result. The behavior was reproduced against a local NameServer/Broker with both the send response's message ID and offset message ID.\n\n## Scope\n\n- Define an explicit, actionable behavior for message-ID queries when no topic is supplied.\n- Avoid returning an indistinguishable successful empty result for an unsupported lookup shape.\n- Preserve topic-scoped message-ID lookup.\n\n## Acceptance criteria\n\n- A message-ID query without a topic either resolves the message through a supported path or returns a structured validation error explaining that a topic is required.\n- Topic-scoped message-ID lookup continues to return the sent message.\n- Add regression coverage for both paths.

Aias00:
Closes #1292

## Summary
- require a topic for message ID lookups so unique IDs are not misinterpreted as offset IDs
- add the Topic selector to the Message ID query form
- retain the topic in message query history and discard legacy Message ID history entries without one

## Validation
- `mvn -q -Dtest=MessageServiceTest test`
- `npm run test -- --run src/pages/instance/__tests__/MessagePage.test.tsx`
- `npm run build`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
@@ -58,6 +58,9 @@ public class MessageService {
         if (hasKey && !hasTopic) {
             throw new BusinessException(400, "topic is required when key is specified");
         }
+        if (hasMessageId && !hasTopic) {
+            throw new BusinessException(400, "topic is required when msgId is specified");
+        }
         if (!hasTopic && !hasMessageId) {
             throw new BusinessException(400, "topic or msgId is required");
         }
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/message/MessageServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/message/MessageServiceTest.java
@@ -14,13 +14,9 @@ import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.provider.InstanceProviderRegistry;
 import org.junit.jupiter.api.Test;
 
-import java.util.Collections;
-
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
-import static org.mockito.Mockito.when;
 
 class MessageServiceTest {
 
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (commented):
### Summary
This PR correctly requires a topic parameter for message ID lookups, preventing unique message IDs from being misinterpreted as offset-based query IDs. The change is focused, well-tested, and includes both backend validation and frontend UI updates.

### Findings

- **[Info]** `MessageService.java:61` — The new validation check for `msgId` without `topic` is correctly placed before the existing `!hasTopic && !hasMessageId` check. The ordering ensures specific validations run before the generic "at least one required" check. Good.

- **[Info]** `message.tsx:42-50` — The history filtering logic (`h.msgId && !h.topic`) correctly discards legacy entries that lack a topic. This prevents stale history from bypassing the new validation.

- **[Info]** `MessageServiceTest.java` — Unused imports (`Collections`, `verify`, `when`) are properly cleaned up. The renamed test `rejectsMessageIdQueryWithoutTopicBeforeCallingProvider` accurately reflects the new behavior and follows the existing naming convention.

- **[Info]** `MessagePage.test.tsx` — Test coverage for the frontend changes is solid: verifies the topic selector appears for Message ID queries and that legacy history entries are filtered.

### Assessment
Clean, minimal change that fixes a real ambiguity in the query API. Backend validation, frontend UI, tests, and history migration are all addressed consistently. No concerns.

---

RockteMQ-AI (approved):
### Summary
Adds validation that a topic is required when querying by message ID, preventing ambiguous results from cross-topic ID collisions. The change is consistent across backend validation, frontend UI (topic selector for msgId mode), and tests.

The `isRecentQuery` filter correctly excludes legacy message ID queries without a topic from the history panel — a clean migration strategy.

One minor note: the frontend test file changes look correct — updating mock data to include topic in message ID queries. LGTM.

---

Aias00:
Consolidated into #1274, which now includes topic validation for message-ID queries and frontend regression coverage.

PR #1296, opened 2026-08-09
fix: reject cloud instances from Remoting admin resolution

### LINKED ISSUE #1295: fix: reject non-Apache instances from Remoting admin resolution
## Problem

RuntimeAdminClientResolver validates only instance ID and endpoint. A cloud-backed instance can therefore reach MqAdminExtFactory, which treats its endpoint as a RocketMQ NameServer address. This bypasses the provider boundary and fails with an incorrect protocol path.

## Expected behavior

Remoting MQAdmin operations are available only for APACHE instances. Cloud instances must be handled by their provider implementation and receive an actionable structured error when a Remoting-only operation is requested.

## Scope

- Enforce the APACHE vendor boundary in RuntimeAdminClientResolver.
- Preserve existing Apache instance behavior.
- Add regression tests for cloud instance rejection.

## Evidence

- server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RuntimeAdminClientResolver.java
- The resolver is used by the Apache metadata, message, DLQ, client, and dashboard providers.

Aias00:
## Summary

- reject non-Apache instances before resolving or using a Remoting MQAdmin client
- preserve legacy instances without a vendor as Apache-compatible
- add regression coverage for endpoint resolution and direct execution paths

## Validation

- JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH=$JAVA_HOME/bin:$PATH mvn -q -Dtest=RuntimeAdminClientResolverTest test
- JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH=$JAVA_HOME/bin:$PATH mvn -q -DskipTests compile

Closes #1295

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RuntimeAdminClientResolver.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RuntimeAdminClientResolver.java
@@ -7,6 +7,7 @@
 package org.apache.rocketmq.studio.cluster.broker;
 
 import lombok.RequiredArgsConstructor;
+import org.apache.rocketmq.studio.common.domain.enums.InstanceVendor;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.instance.InstanceRepository;
 import org.apache.rocketmq.studio.instance.InstanceVO;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary

Clean guard against using cloud (non-Apache) instances with the Remoting `AdminClient` — the `requireApacheInstance()` check is applied in both `resolveEndpoint` and `execute` paths, with proper backward compatibility for legacy instances (null vendor treated as Apache). Regression tests cover both code paths and correctly verify that the admin factory is never invoked for cloud instances.

LGTM.

---

lizhimins:
Merged into #1403 (squashed into rocketmq-studio as ace3fb91). Thanks for the contribution! / 已并入 #1403 合入 rocketmq-studio 分支（ace3fb91），感谢贡献！

PR #1299, opened 2026-08-09
fix: exclude disabled alert rules from Prometheus export

### LINKED ISSUE #1298: fix: exclude disabled alert rules from Prometheus rule export
## Problem

`AlertService.exportPrometheusRulesYaml()` converts every persisted alert rule into Prometheus YAML. A rule disabled through `/api/alert-rules/toggle` is therefore still exported and remains active after users load the YAML into Prometheus.

## Expected behavior

Only enabled rules should be exported. When there are no enabled persisted rules, retain the existing fallback to the bundled default rule templates.

## Scope

- Filter persisted rules by `enabled` before conversion.
- Add regression coverage for disabled-only and mixed enabled/disabled rule sets.
- Do not change the alert rule CRUD API.

Aias00:

Disabled alert rules were still serialized into the Prometheus rule YAML export, so loading that YAML could activate a rule that an operator had disabled in Studio.

Closes #1298

## Brief changelog

- Filter persisted alert rules by `enabled` before producing Prometheus YAML.
- Preserve the bundled default rule fallback when no enabled configured rules remain.
- Add mixed enabled/disabled and disabled-only regression coverage.

## Verifying this change

```bash
cd server
JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" \
  mvn -q -Dtest=AlertServiceTest test
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
@@ -42,7 +42,9 @@ public class AlertService {
     }
 
     public String exportPrometheusRulesYaml() {
-        List<AlertRuleVO> rules = alertRepository.findAllRules();
+        List<AlertRuleVO> rules = alertRepository.findAllRules().stream()
+                .filter(AlertRuleVO::isEnabled)
+                .toList();
         List<PrometheusAlertRule> prometheusRules = rules.isEmpty()
                 ? defaultPrometheusRules()
                 : rules.stream().map(this::toPrometheusRule).toList();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary

Important fix — disabled alert rules were leaking into the Prometheus YAML export, which could silently re-enable them when the YAML is loaded by a Prometheus instance. The one-line `.filter(AlertRuleVO::isEnabled)` correctly gates the export, and the fallback to bundled defaults when all rules are disabled is preserved. Regression tests cover both mixed and all-disabled scenarios.

LGTM.

---

Aias00:
Consolidated into #1311, which now includes the enabled-rule export behavior and regression coverage.

PR #1311, opened 2026-08-09
feat: harden and audit alert rule lifecycle

### LINKED ISSUE #1298: fix: exclude disabled alert rules from Prometheus rule export

### LINKED ISSUE #1310: feat: audit alert rule and system alert mutations
## Problem

Alert rule writes and system-alert acknowledgement/cleanup actions modify Studio operational state but are not represented in operation audit logs.

## Expected behavior

Record successful alert rule create, update, toggle, and delete operations, as well as system-alert acknowledge and cleanup operations. Records must contain only stable IDs and minimal action metadata, not PromQL expressions, notification channels, or descriptions.

## Scope

- Reuse `OperationAuditService`.
- Cover the success paths with focused service tests.
- Preserve existing alert rule YAML export behavior.

Aias00:
## Summary

- records successful alert-rule mutations and system-alert acknowledgement or cleanup actions;
- keeps audit details concise and excludes expressions, channels, and descriptions;
- exports only enabled alert rules to Prometheus rule YAML, while retaining fallback defaults when no enabled user rules remain.

Closes #1310
Closes #1298

## Test

```bash
cd server
JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=AlertServiceTest,AlertServiceDefaultRulesTest test
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
@@ -44,7 +44,9 @@ public class AlertService {
     }
 
     public String exportPrometheusRulesYaml() {
-        List<AlertRuleVO> rules = alertRepository.findAllRules();
+        List<AlertRuleVO> rules = alertRepository.findAllRules().stream()
+                .filter(AlertRuleVO::isEnabled)
+                .toList();
         List<PrometheusAlertRule> prometheusRules = rules.isEmpty()
                 ? defaultPrometheusRules()
                 : rules.stream().map(this::toPrometheusRule).toList();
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
LGTM. Clean audit trail integration for alert rule and system alert mutations. Good discipline in keeping audit details minimal (IDs, names, toggle state) without including expressions or channel details. Tests verify both the audit calls and that sensitive data is excluded.

---

lizhimins:
Merged into #1403 (squashed into rocketmq-studio as ace3fb91). Thanks for the contribution! / 已并入 #1403 合入 rocketmq-studio 分支（ace3fb91），感谢贡献！

PR #1248, opened 2026-08-07
fix: harden instance-scoped Consumer Group runtime diagnostics

Aias00:
## Summary

- scopes Consumer Group runtime diagnostics and related caches by the selected instance;
- stops resource loading when no instance is selected;
- preserves the selected instance when navigating to Consumer Group details.

Closes #1247
Closes #1323

## Test

```bash
cd server
JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=ConsumerDiagnosticsServiceTest,RocketMQMetadataProviderTest test
cd ../web
npm test -- --run src/pages/instance/__tests__/ConsumerPage.test.tsx src/services/consumerService.test.ts
```

PR #1327, opened 2026-08-09
fix(consumer): scope diagnostics cache by instance

### LINKED ISSUE #1277: [Studio][Bug] Scope Consumer Group detail caches by selected instance
## Problem

The Consumer Group page caches subscription and progress diagnostics only by group name. Two managed instances can contain the same Consumer Group name, so switching instances and opening that group can show cached subscriptions and offsets from the previously selected instance.

## Expected behavior

Cache diagnostic results by both `instanceId` and Consumer Group name, or reset the caches whenever the selected instance changes. Late responses must also not populate a different instance context.

## Scope

Track 1 / Consumer management. Update the page state and add a UI regression test for same-named groups across two instances.

## Evidence

`web/src/pages/instance/consumer.tsx:250` and `:271` read `subscriptionsByGroup[groupName]` and `progressByGroup[groupName]` without an instance-qualified key.

123123213weqw:
## Summary

- key Consumer Group subscription, progress, loading, and error caches by instance and group name
- keep late responses in their original instance cache instead of exposing them after a switch
- add a regression test for same-named groups across two managed instances

## Root cause

The Consumer Group page used only `groupName` as its diagnostics cache key. The same group name can exist in multiple instances, so cached subscriptions and offsets from one instance were reused for another.

## Impact

Consumer diagnostics now remain isolated when operators switch between instances containing the same group name.

Fixes #1277

## Validation

- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --run src/pages/instance/__tests__/ConsumerPage.test.tsx`

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,18 @@ RocketMQ Studio is a unified management platform for RocketMQ, supporting multi-
 ## Quick Start
 
 ```bash
-cd deploy && docker compose up -d --build
+cd deploy/rocketmq && docker compose up -d
+cd .. && docker compose up -d --build
 ```
 
 Visit **http://127.0.0.1:6789** after startup.
+
+The first command starts the bundled RocketMQ topology and creates the
+`rocketmq_default` Docker network used by the Studio services. Check that it
+is healthy before starting Studio with `docker compose ps` from `deploy/rocketmq`.
+The default schema creates only Studio tables. It does not seed instances, topics, consumer groups, or ACL
+records. Development-only sample data can be imported explicitly from `deploy/mysql/`; it is not part of the
+default deployment.
 
 **Studio ports:** Frontend 6789 (Nginx), Backend 8888 (Spring Boot)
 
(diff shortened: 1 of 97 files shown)


RockteMQ-AI (approved):
LGTM — excellent fix for a real data isolation bug. The `diagnosticCacheKey` approach correctly scopes subscription and progress caches by both instance and group name, preventing cross-instance data leakage.

### Summary
This PR fixes a bug where Consumer Group diagnostics (subscriptions, progress) were cached only by group name, causing data from one instance to bleed into another when the same group name exists in multiple instances. The fix introduces a composite cache key `(instanceId, groupName)` and adds a comprehensive regression test.

### Findings
- **[Info]** `consumer.tsx:179` — The `JSON.stringify` approach for cache keys works well and is unambiguous. An alternative like `${instanceId}::${groupName}` would be slightly lighter, but `JSON.stringify` is fine and avoids delimiter collision risks.

No blocking issues. The cross-instance regression test is particularly thorough.

---

lizhimins:
Closing as superseded: per-instance diagnostics cache isolation (instance-scoped cache keys plus reset on instance switch) is already implemented on rocketmq-studio by #1248. 关闭：诊断缓存按实例隔离（实例作用域缓存 key + 切实例重置）已由 #1248 在主干实现。

PR #1329, opened 2026-08-09
fix(ui): handle messages without tags or keys

### LINKED ISSUE #1328: fix: handle messages without tags or keys in table sorting
## Problem

RocketMQ messages do not require tags or keys. `RocketMQMessageProvider` copies `MessageExt#getTags()` and `MessageExt#getKeys()` directly into the message response, so either field can be `null`.

The web API contract currently declares both fields as required strings, and the Message table sorters call `localeCompare` directly. Sorting the Tag or Key column can therefore throw for a valid tagless or keyless message.

## Expected behavior

- Model message tags and keys as nullable API data.
- Render the existing empty-value fallback for missing values.
- Sort Tag and Key columns without throwing.
- Add focused page regression coverage.

## Scope

Track 1 / message query UI reliability.

123123213weqw:
## Summary

- model RocketMQ message tags and keys as nullable values returned by the backend
- make Tag and Key sorting null-safe and render a clear fallback for missing values
- add a regression test that sorts records with null tags and keys

## Testing

- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --run src/pages/instance/__tests__/MessagePage.test.tsx`

Fixes #1328

### DIFF
--- a/web/src/api/message.ts
+++ b/web/src/api/message.ts
@@ -4,8 +4,8 @@ import client from './client';
 export interface MessageRecord {
   msgId: string;
   topic: string;
-  tag: string;
-  key: string;
+  tag: string | null;
+  key: string | null;
   body: string;
   storeTime: number | string;
   bornHost: string;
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
LGTM — clean null-safety fix for message tags and keys. The type changes correctly model the backend behavior, and the regression test covers sorting and rendering with null fields.

### Summary
This PR makes `tag` and `key` nullable in `MessageRecord`, adds null-safe sorting for both columns, and renders a fallback `-` for missing values.

### Findings
No issues found. The implementation is clean and well-tested.

---

PR #1331, opened 2026-08-09
fix(ui): make instance remark sorting null-safe

### LINKED ISSUE #1330: fix: make instance remark sorting null-safe
## Problem

The database schema allows `rmq_instance.remark` to be null, and the backend can therefore return an instance with a null remark. The web client currently models the field as a required string and calls `localeCompare` directly from the table sorter. Sorting the Remark column can throw when such a record is present.

## Expected behavior

- model the remark field as nullable
- sort nullable remarks without throwing
- display a clear fallback for missing remarks
- cover the behavior with a regression test

123123213weqw:
## Summary

- model instance remarks as nullable values returned by the backend
- make the Remark column sorter and renderer handle missing remarks safely
- keep mock-mode search null-safe and add a table sorting regression test

## Testing

- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --run src/pages/instance/__tests__/InstancePage.test.tsx src/services/instanceService.test.ts`

Fixes #1330

### DIFF
--- a/web/src/api/instance.ts
+++ b/web/src/api/instance.ts
@@ -23,7 +23,7 @@ export type InstanceVendor = 'APACHE' | 'ALIYUN' | 'TENCENT';
 export interface Instance {
   id: string;
   name: string;
-  remark: string;
+  remark: string | null;
   type: 'PROXY' | 'DIRECT';
   endpoint: string;
   vendor?: InstanceVendor;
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
LGTM — clean null-safety fix for instance remark sorting. The type change to `string | null` correctly models the backend behavior, and the regression test covers both null and non-null cases.

### Summary
This PR makes the `remark` field nullable in the `Instance` type, adds null-safe sorting in the Remark column, and renders a fallback `-` for missing remarks. The mock-mode search is also updated to be null-safe.

### Findings
No issues found. The implementation is clean and well-tested.

---

PR #1382, opened 2026-08-10
fix(web): require an instance for message queries

### LINKED ISSUE #1380: [Bug] Require a selected instance before message queries
## Problem

The Message Explorer enables its query action without a selected instance. It then calls the message API without an instance context, while the backend requires one to route the operation.

## Impact

The UI allows an operation that cannot be routed to a managed RocketMQ instance and fails only after the request.

## Proposed fix

Disable message queries until an instance is selected, guard the query handler, and add a UI regression test.

Track 1: multi-instance control plane.

Aias00:
Closes #1380

## Summary
- disable message queries and query-history replay until an instance is selected
- guard query execution so stale UI events cannot send an unscoped request
- update message-page tests to use explicit instance context and cover the no-instance state

## Verification
- npm test -- --run src/pages/instance/__tests__/MessagePage.test.tsx src/pages/instance/__tests__/MessagePageAsyncState.test.tsx
- npm run build

### DIFF
--- a/web/src/pages/instance/__tests__/MessagePage.test.tsx
+++ b/web/src/pages/instance/__tests__/MessagePage.test.tsx
@@ -27,10 +27,14 @@ const messageServiceMocks = vi.hoisted(() => ({
   getMessageTrace: vi.fn(),
   queryMessages: vi.fn(),
 }));
+const instanceFilterMocks = vi.hoisted(() => ({
+  useInstanceFilter: vi.fn(),
+}));
 
 const QUERY_HISTORY_STORAGE_KEY = 'rocketmq-studio-message-query-history';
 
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds instance-selection guards to the Message Explorer page. Query and trace buttons are now disabled until an instance is selected, preventing unscoped API calls that could return data from the wrong instance or fail silently.

The `instanceReady` flag is derived from `useInstanceFilter` and gates:
- Topic loading
- Message query submission
- Trace query submission

**Verdict: LGTM** — good defensive UX that prevents confusing empty/error states from unscoped queries.

---

Aias00:
Superseded by #1473, which consolidates the Message Explorer instance-context fixes and their regression coverage.

PR #1452, opened 2026-08-10
fix(topic): preserve instance context on rebuild

### LINKED ISSUE #1450: [Studio][Bug] Keep instance context when rebuilding a Topic
### Runtime

- Branch: `rocketmq-studio`
- Base: `0ef5e2d`
- JDK: 21

### Problem

The Topic detail dialog offers **Rebuild on Broker** when a persisted Topic has no route. `TopicPage.rebuildTopic()` calls both APIs without the selected managed-instance ID:

```ts
await createTopic({ name, type, writeQueues, readQueues });
await getTopicRoutes(topic.name);
```

The backend now requires `instanceId` for Topic writes, so the rebuild request is rejected with HTTP 400. Even if a write were accepted, the follow-up route lookup would not be scoped to the instance whose Topic is being rebuilt.

Normal Topic creation already forwards `selectedInstanceId`; only the rebuild path drops it.

### Reproduction

1. Select an Apache managed instance.
2. Open a Topic whose persisted record has no Broker route.
3. Click **Rebuild on Broker**.
4. Observe that `POST /api/topics/create` has no `instanceId` and fails validation.

### Expected

Both the rebuild write and the route refresh use the Topic record's `instanceId` (falling back to the current selected instance for legacy records).

### Scope

- forward the originating instance ID to `createTopic`
- forward the same ID to `getTopicRoutes`
- add a focused UI regression test

123123213weqw:

Fixes #1450.

The Topic detail dialog's rebuild path dropped the managed-instance context. Since Topic writes now require `instanceId`, rebuilding a persisted Topic without a Broker route failed with HTTP 400; the follow-up route refresh was also unscoped.

## Brief change log

- derive the rebuild target from the Topic record's `instanceId`, with the current selection as a legacy fallback
- forward that same instance ID to both `createTopic` and `getTopicRoutes`
- add a focused Topic page regression test covering the complete rebuild flow

## Verification

- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/pages/instance/__tests__/TopicPage.test.tsx` — 9 tests passed
- targeted ESLint passed with the base branch's existing `react-hooks/set-state-in-effect` finding disabled
- `npm run build` passed
- `git diff --check` passed

### DIFF
--- a/web/src/pages/instance/__tests__/TopicPage.test.tsx
+++ b/web/src/pages/instance/__tests__/TopicPage.test.tsx
@@ -239,6 +239,36 @@ describe('TopicPage', () => {
     expect(within(getTableBody()).queryByText('topic-01')).not.toBeInTheDocument();
   });
 
+  it('keeps the selected instance when rebuilding a topic without a broker route', async () => {
+    const user = userEvent.setup();
+    const topic = { ...buildTopics(1)[0], instanceId: 'instance-a' };
+    instanceServiceMocks.listInstances.mockResolvedValue([
+      {
+        ...selectedInstance,
+        id: 'instance-a',
+        name: 'Instance A',
+        type: 'DIRECT',
+      },
+    ]);
+    topicServiceMocks.listTopics.mockResolvedValue([topic]);
+    renderWithProviders('/instance/instance-a/topic');
+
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #1453, opened 2026-08-10
[Studio] feat: Fix/studio p0 frontend fixes

zhaohai666:
# fix(web): gzip serving, delete confirmations, audit search debounce, StrictMode remount fix

## Summary

Four P0 frontend fixes identified during a post-merge audit of `rocketmq-studio`.

**1. Enable gzip compression in nginx (`web/nginx.conf`)**
The Docker deployment serves static assets through nginx without any compression configuration, so the ~1 MB entry bundle was delivered uncompressed. Adds `gzip on` with a sensible `gzip_types` list to the server block.

**2. Fix GroupManagement remount bug under React StrictMode (`web/src/pages/studio/GroupManagement.tsx`)**
The effect cleanup set `mountedRef.current = false`, but nothing ever reset it to `true` on remount. Because `main.tsx` renders under `<React.StrictMode>`, the dev-mode double mount made every `loadGroups` callback bail out early, leaving the consumer-group list loading forever. The mount effect now resets the flag — the same pattern already used in `pages/cluster/index.tsx`. Production behavior is unchanged; this fixes incorrect logic.

**3. Require confirmation before destructive deletes (`web/src/pages/ops/alerts.tsx`, `web/src/pages/settings/index.tsx`)**
Deleting an alert rule and deleting a data source executed immediately on click, while every other destructive action in the app asks for confirmation. Both buttons are now wrapped in `Popconfirm`, matching the existing convention (`Proxy.tsx`, `AlertManagement.tsx`).

**4. Debounce the audit-log search (`web/src/pages/ops/audit.tsx`)**
The audit search box re-ran the paginated server query on every keystroke — typing an 8-character word meant 8 requests. The search text is now debounced by 300 ms, the same pattern already used in `pages/instance/index.tsx`, so the query fires only when the user pauses typing. The export handler intentionally keeps using the immediate input value.

## Test plan

- [x] Targeted suites: `AuditPage`, `AlertsPage`, `DataSourceTab`, `GroupManagement` — 27 tests pass
- [x] Full frontend suite: 87 files / 449 tests pass
- [x] `npm run build` (tsc + vite) passes
- [x] ESLint on all modified files: 0 errors
- [ ] Manual: verify `Content-Encoding: gzip` responses from the Docker image
- [ ] Manual: verify the confirmation popovers on the Alerts and Settings pages

### DIFF
--- a/web/nginx.conf
+++ b/web/nginx.conf
@@ -2,6 +2,14 @@ server {
     listen 80;
     server_name _;
 
+    # Compress text assets; the entry JS bundle is ~1MB uncompressed.
+    gzip on;
+    gzip_comp_level 5;
+    gzip_min_length 1024;
+    gzip_proxied any;
+    gzip_vary on;
+    gzip_types text/plain text/css application/javascript application/json image/svg+xml;
+
     # RESOLVER is injected by 15-resolver.sh (podman network gateway only).
     # Variable proxy_pass re-resolves on every request, so a recreated
     # rocketmq-server container with a new IP is picked up without restarting
--- a/web/src/pages/ops/alerts.tsx
+++ b/web/src/pages/ops/alerts.tsx
@@ -31,6 +31,7 @@ import {
   Checkbox,
   Flex,
   message,
+  Popconfirm,
   theme,
 } from 'antd';
 import type { ColumnsType, TableRowSelection } from 'antd/es/table/interface';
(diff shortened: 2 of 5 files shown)


RockteMQ-AI (commented):
Good set of P0 fixes. Overall LGTM with one minor observation:

**Fixes reviewed:**
1. ✅ **nginx gzip** — sensible defaults (`gzip_comp_level 5`, `gzip_min_length 1024`) for the ~1MB bundle
2. ✅ **StrictMode remount** — correct fix; resetting `mountedRef.current = true` on remount matches the pattern in `pages/cluster/index.tsx`
3. ✅ **Delete confirmation** — `Popconfirm` wrapper is consistent with existing conventions
4. ✅ **Audit search debounce** — 300ms debounce prevents N requests per N-character word

**Minor observation:**
- In `settings/index.tsx`, the `Popconfirm` uses hardcoded Chinese strings (`"确定要删除该数据源吗？"`) while `alerts.tsx` uses i18n keys (`t('common.areYouSureToDelete')`). Consider aligning to i18n for consistency if the project supports multiple locales.

Not blocking — this is a follow-up improvement.

---

PR #1458, opened 2026-08-10
fix(settings): isolate audit recording failures

### LINKED ISSUE #1457: [Studio][Bug] Do not fail saved settings when audit recording fails
## Bug

`SettingsService.saveGeneralSettings` persists the new general settings and then calls `OperationAuditService.record` directly. If the audit store is temporarily unavailable, the audit exception escapes after the settings were already saved, so the API reports a failure even though the mutation succeeded.

This can prompt callers to retry an already-applied update and differs from the best-effort audit pattern used by other mutation services.

## Expected behavior

Once general settings are persisted, a secondary audit failure should be logged but must not change the save result.

## Proposed fix

- isolate the settings audit call behind a small best-effort helper
- log the audit failure with context
- add a service regression test proving the repository save returns normally when audit recording throws

123123213weqw:
## Summary
- make general-settings audit recording best effort after the settings are persisted
- log audit storage failures without returning a false save failure to callers
- add a service regression test for an unavailable audit store

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=SettingsServiceTest test` (28 tests)

Fixes #1457

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
@@ -90,8 +90,16 @@ public class SettingsService {
         }
         settings.setClearApiKey(false);
         settingsRepository.saveGeneralSettings(settings);
-        operationAuditService.record("UPDATE_SETTINGS", "SETTINGS", "general",
-                null, "General settings updated", "SUCCESS", null);
+        recordSettingsAudit();
+    }
+
+    private void recordSettingsAudit() {
+        try {
+            operationAuditService.record("UPDATE_SETTINGS", "SETTINGS", "general",
+                    null, "General settings updated", "SUCCESS", null);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — good error isolation. Audit recording failures should not break the primary settings operation. The try-catch with logging is the right approach for non-critical side effects.

---

RockteMQ-AI (approved):
### Summary
Resilience fix that prevents audit recording failures from rolling back the main settings save. The audit call is extracted into a `recordSettingsAudit()` helper wrapped in try-catch, logging a warning on failure.

Test verifies settings persist successfully even when `operationAuditService.record()` throws. Clean isolation pattern. LGTM.

---

PR #1460, opened 2026-08-10
fix(cluster): fail config update without broker target

### LINKED ISSUE #1459: [Studio][Bug] Do not report cluster config success when no broker was updated
## Bug

`ClusterService.updateClusterConfig` initializes both broker result lists as empty. When a resolved cluster has no broker with a usable address, no live update is attempted, but `updateStatus` treats the empty failure list as `SUCCESS`. The service then persists the requested configuration even though no broker applied it.

This produces a false-success response and makes the stored configuration diverge from the live cluster.

## Expected behavior

A configuration update with no usable broker target should:

- return `FAILED`
- explain that no broker address is available
- avoid persisting the unapplied configuration
- record a failed update audit

## Proposed fix

Add an explicit failure result when the broker loop has no usable target, and cover the no-broker path with a service regression test.

123123213weqw:
## Summary
- turn an empty broker-update attempt into an explicit failed broker result
- prevent persistence of configuration that no live broker applied
- cover the no-broker path, failed audit result, and unchanged repository state

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=ClusterServiceTest test` (30 tests)

Fixes #1459

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -151,6 +151,12 @@ public class ClusterService {
                 }
             }
         }
+        if (successfulBrokers.isEmpty() && failedBrokers.isEmpty()) {
+            failedBrokers.add(BrokerConfigUpdateFailureVO.builder()
+                    .address("N/A")
+                    .message("No broker address is available for configuration update")
+                    .build());
+        }
 
         ClusterConfigUpdateResultVO.Status status = updateStatus(successfulBrokers, failedBrokers);
         if (failedBrokers.isEmpty()) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — good defensive fix. Failing fast when no valid broker target is found prevents silent partial updates that leave cluster config in an inconsistent state. The error message is clear and actionable.

---

RockteMQ-AI (approved):
### Summary
Defensive fix that ensures config updates fail explicitly when no broker addresses are available. Previously, the method would silently return an empty result, which could confuse users into thinking the update succeeded.

The fix adds a post-loop check: if both `successfulBrokers` and `failedBrokers` are empty, a failure entry is added with a clear message. The test verifies that the repository is never called and the audit records the failure.

Clean, minimal fix with comprehensive test coverage. LGTM.

---

PR #1462, opened 2026-08-10
fix(dlq): guard newline-prefixed CSV formulas

### LINKED ISSUE #1461: [Studio][Bug] Escape line-feed-prefixed formulas in DLQ CSV exports
## Bug

The DLQ CSV exporter protects cells beginning with `=`, `+`, `-`, `@`, tab, or carriage return, but omits line feed (`\n`). Spreadsheet applications can ignore leading control whitespace when interpreting a cell, so a value such as `\n=HYPERLINK(...)` can remain formula-active after export.

The CSV cell is quoted, but quoting controls CSV structure; it does not neutralize spreadsheet formula interpretation.

## Expected behavior

DLQ CSV values beginning with any formula marker or leading tab/CR/LF should be prefixed with an apostrophe before CSV quoting.

## Proposed fix

- include `\n` in the leading-character formula guard
- add an export regression test using a line-feed-prefixed group name

123123213weqw:
## Summary
- include line feed in the DLQ CSV formula-prefix guard
- keep CSV quoting behavior unchanged while neutralizing the spreadsheet cell value
- add an export regression test for a line-feed-prefixed formula

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/pages/instance/__tests__/DLQPage.test.tsx` (11 tests)
- `NODE_OPTIONS=--no-experimental-webstorage npx eslint --rule 'react-hooks/set-state-in-effect: off' src/pages/instance/dlq.tsx src/pages/instance/__tests__/DLQPage.test.tsx`

Note: the targeted file already triggers the repository's existing `react-hooks/set-state-in-effect` finding at `dlq.tsx:132`; the focused lint run disables only that unrelated rule.

Fixes #1461

### DIFF
--- a/web/src/pages/instance/__tests__/DLQPage.test.tsx
+++ b/web/src/pages/instance/__tests__/DLQPage.test.tsx
@@ -228,6 +228,25 @@ describe('DLQ page', () => {
     expect(revokeObjectURL).toHaveBeenCalledWith('blob:dlq');
   });
 
+  it('neutralizes formulas hidden behind a leading line feed in CSV exports', async () => {
+    vi.mocked(messageService.listDLQGroups).mockResolvedValue([
+      {
+        ...dlqGroup,
+        groupName: '\n=1+1',
+        dlqTopic: '%DLQ%formula',
+      },
+    ]);
+    const user = userEvent.setup();
+    renderWithProviders(<DLQPage />);
+
+    const row = (await screen.findByText('%DLQ%formula')).closest('tr');
+    if (!row) throw new Error('DLQ group row not found');
+    await user.click(within(row).getByRole('button', { name: '导出' }));
+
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — good security hardening. Adding `\n` to the CSV formula-prefix guard closes a subtle injection vector where formulas could hide behind a leading newline. Test coverage for the edge case is appreciated.

---

RockteMQ-AI (approved):
### Summary
Security fix for CSV formula injection bypass. The previous regex `/^[=+\-@\t\r]/` missed `\n`, allowing a value like `\n=1+1` to bypass the guard. Adding `\n` to the character class closes this gap.

Clean one-character fix with a regression test that verifies formulas hidden behind a leading line feed are properly neutralized. LGTM.

---

PR #1486, opened 2026-08-10
fix(instance): normalize endpoint types with root locale

yyqdbngt:
### Problem

Endpoint type ordering used the JVM default locale. Under Turkish locale, lowercase `tcp_internet` uppercased to a value containing a dotted capital I, so the public endpoint was treated as unknown and could lose priority to an unsupported fallback endpoint.

### Fix

Normalize endpoint type identifiers with `Locale.ROOT` before applying endpoint priority.

### Verification

`mvn -B -ntp -Dmaven.compiler.proc=full -Dtest=InstanceServiceTest test`

Tests run: 48, failures: 0, errors: 0.


RockteMQ-AI (approved):

---

PR #1487, opened 2026-08-10
fix(metrics): normalize backend types with root locale

yyqdbngt:
### Problem

Metrics backend names were uppercased with the JVM default locale. Under Turkish locale, lowercase `mimir` becomes a value containing dotted capital I characters, so the configured Mimir backend silently falls back to Prometheus and uses the wrong query path.

### Fix

Normalize backend identifiers with `Locale.ROOT` before resolving the backend type.

### Verification

`mvn -B -ntp -Dmaven.compiler.proc=full -Dtest=MetricsBackendTypeTest test`

Tests run: 4, failures: 0, errors: 0.


RockteMQ-AI (approved):

---

PR #1495, opened 2026-08-10
fix(alerts): normalize query levels with root locale

yyqdbngt:
### Problem

System alert level filters were lowercased with the JVM default locale. Under Turkish locale, uppercase `INFO` becomes `ınfo`, so the generated database predicate does not match stored `info` alerts and the query silently returns incomplete results.

### Fix

Normalize alert level query values with `Locale.ROOT` before constructing the MyBatis predicate.

### Verification

`mvn -B -ntp -Dmaven.compiler.proc=full -Dtest=MybatisPlusAlertRepositoryTest test`

Tests run: 1, failures: 0, errors: 0.


RockteMQ-AI (approved):

---

PR #1530, opened 2026-08-11
fix(server): restore RocketMQ 5.5 compilation

Aias00:
Closes #1529

Restores compilation against the configured RocketMQ 5.5 API by:
- importing the shared system-topic filter and `Pattern`;
- copying message properties through the supported API;
- resolving trace message timestamps through the existing offset-ID lookup helper.

Validation:
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=AlertServiceTest test` (32 tests passed; compiles all main and test sources)

PR #1531, opened 2026-08-11
fix(cluster): isolate configuration snapshots

### LINKED ISSUE #1528: [Studio][Bug] Do not share mutable Cluster configuration in repository snapshots
## Problem

`ClusterRepositoryImpl` documents `findById` as returning an independent snapshot, but `defensiveCopy` reuses the mutable `ClusterConfigVO` reference. A caller can mutate `cluster.getConfig()` from one read and alter the repository's later reads without using an update operation.

## Scope

- Deep-copy the Cluster configuration as part of repository snapshots.
- Add a focused regression test that mutates a returned config and verifies the next read is unchanged.

## Track

Track 1 / Cluster runtime state isolation.

Aias00:
Closes #1528

`ClusterRepositoryImpl.findById` promised an independent snapshot but shared the mutable `ClusterConfigVO` reference. This change deep-copies the configuration so callers cannot mutate repository state through a read result.

Validation:
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=ClusterRepositoryImplTest test` (3 tests passed)

The test was executed with the compilation fixes from #1530 present; this PR remains independent and contains only the snapshot-isolation change.


RockteMQ-AI (commented):
### Summary
This PR makes changes across 57 lines. Overall the approach looks reasonable. I have some suggestions for improvement.

---

PR #1537, opened 2026-08-11
fix(cluster): hide unsupported legacy Broker restart

### LINKED ISSUE #1526: [Studio][Bug] Hide unsupported Broker restart action
## Problem

The Broker Cluster page exposes a Restart Broker control, but `ClusterService.restartBroker` validates the cluster and broker then explicitly returns `501 Not Implemented`. The UI therefore offers an operation that cannot succeed.

A previous action-visibility fix assumed the restart API was implemented; the current server implementation shows that assumption is no longer valid.

## Scope

- Hide the unsupported Broker restart control while retaining read-only runtime diagnostics.
- Remove the unreachable client invocation and update focused frontend coverage.
- Do not emulate a successful restart or change the server-side operation contract.

## Track

Track 1 / ARCH-01 cluster runtime management.

Aias00:
Closes #1526

The legacy Broker Cluster page invokes the Broker restart endpoint, but the current server implementation always rejects that operation as unsupported. Keep runtime diagnostics and refresh behavior while removing the unavailable mutation control and its unreachable client call.

Validation:
- `npm run build`

### DIFF
--- a/web/src/pages/studio/BrokerCluster.tsx
+++ b/web/src/pages/studio/BrokerCluster.tsx
@@ -25,20 +25,17 @@ import {
   Space,
   Switch,
   Progress,
-  Tooltip,
   Spin,
   App,
-  Modal,
 } from 'antd';
 import {
   ArrowClockwise,
-  ArrowsClockwise,
   Cloud,
   ChartBar,
   PlugsConnected,
 } from '@phosphor-icons/react';
 import { useLang } from '../../i18n/LangContext';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR makes changes across 140 lines. Overall the approach looks reasonable. I have some suggestions for improvement.

---

Aias00:
Absorbed into #1541, which now combines instance-scoped legacy Broker diagnostics with removal of the unsupported restart action.

PR #1546, opened 2026-08-11
fix(metrics): normalize backend types locale-independently

### LINKED ISSUE #1544: Metrics backend type resolution depends on the JVM default locale
## Problem
`MetricsBackendType.fromProviderType` uppercases persisted provider identifiers with the process default locale. Under locales such as Turkish, values containing `i` (for example `mimir` and `victoria_metrics`) are transformed with a dotted capital I and no longer match the supported backend names, silently falling back to Prometheus.

## Proposed fix
Normalize provider identifiers with `Locale.ROOT` and add a regression test that runs under a Turkish default locale.

123123213weqw:
## Summary
- normalize metrics backend provider names with `Locale.ROOT`
- cover Turkish-locale resolution for Mimir and VictoriaMetrics

## Testing
- `mvn -B -ntp -Dtest=MetricsBackendTypeTest test` — passes when applied with the base compilation fix in #1548

Fixes #1544

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsBackendType.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsBackendType.java
@@ -16,6 +16,8 @@
  */
 package org.apache.rocketmq.studio.cluster.metrics;
 
+import java.util.Locale;
+
 /**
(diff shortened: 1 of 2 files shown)


lizhimins:
Closing: identical change (same file, same function, same line) was merged earlier via #1487 (metrics backend types) — yyqdbngt's root-locale series (#1486-#1495) landed first.

PR #1548, opened 2026-08-11
fix(build): restore server compilation

123123213weqw:
## Summary
- restore the missing alert validation and shared system-topic imports
- copy DLQ user properties through the RocketMQ 5.5 `getProperties()` API while retaining the reserved-key filter
- resolve trace timestamps through the existing offset-message lookup instead of the unavailable one-argument `viewMessage`
- restore a clean server package build against the declared dependencies

## Testing
- `mvn -B -ntp clean package -DskipTests` — passes
- `mvn -B -ntp -Dtest=RocketMQMessageProviderTest test` — passes
- the broader targeted run also exposes two pre-existing `RocketMQDLQProviderTest` expectations that no longer match the base implementation's `FAILED` / `NO_MESSAGES` outcomes

Fixes #1547

PR #1550, opened 2026-08-11
fix(instance): rank cloud endpoints locale-independently

### LINKED ISSUE #1549: Cloud endpoint selection depends on the JVM default locale
## Problem
Cloud endpoint priority uppercases endpoint type identifiers with the JVM default locale. In a Turkish locale, a lower-case `tcp_internet` value becomes `TCP_İNTERNET`, does not match the known type, and can cause an arbitrary endpoint to be selected instead of preferring VPC over public endpoints.

## Proposed fix
Use `Locale.ROOT` for protocol identifier normalization and cover lower-case endpoint identifiers under a Turkish default locale.

123123213weqw:
## Summary
- normalize cloud endpoint type identifiers with `Locale.ROOT`
- keep VPC endpoint priority stable under Turkish JVM locales
- add a regression test using lower-case cloud catalog endpoint types

## Testing
- `mvn -B -ntp -Dtest=InstanceServiceTest test` — 48 tests pass when applied with #1548

Fixes #1549

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -35,6 +35,7 @@ import org.springframework.stereotype.Service;
 
 import java.time.LocalDateTime;
 import java.util.List;
+import java.util.Locale;
 import java.util.UUID;
 
 @Slf4j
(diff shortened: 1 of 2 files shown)


lizhimins:
Closing: identical change (same file, same function, same line) was merged earlier via #1486 (instance endpoint ranking/types) — yyqdbngt's root-locale series (#1486-#1495) landed first.

PR #1488, opened 2026-08-10
fix(metrics): normalize auth modes with root locale

yyqdbngt:
### Problem

Metrics authentication modes were lowercased with the JVM default locale. Under Turkish locale, uppercase `BASIC` becomes `basıc`, so valid basic authentication silently falls back to `none` and the query is sent without credentials.

### Fix

Normalize authentication mode identifiers with `Locale.ROOT` before mapping them to the metrics source configuration.

### Verification

`mvn -B -ntp -Dmaven.compiler.proc=full -Dtest=MetricsServiceTest test`

Tests run: 16, failures: 0, errors: 0.

PR #1552, opened 2026-08-11
fix(metrics): normalize auth types locale-independently

### LINKED ISSUE #1551: Metrics basic authentication can be disabled by the JVM default locale
## Problem
Metrics data-source auth labels are lowercased with the JVM default locale. With a Turkish default locale, `BASIC AUTH` becomes `basıc auth`, is treated as unknown, and silently disables authentication for the query.

## Proposed fix
Normalize auth identifiers with `Locale.ROOT` and add a regression test that captures the generated data-source configuration under a Turkish locale.

123123213weqw:
## Summary
- normalize persisted metrics auth labels with `Locale.ROOT`
- prevent `BASIC AUTH` from falling back to unauthenticated requests under Turkish locales
- capture and assert the generated metrics source configuration in a regression test

## Testing
- `mvn -B -ntp -Dtest=MetricsServiceTest test` — 16 tests pass when applied with #1548

Fixes #1551

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
@@ -27,6 +27,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.util.StringUtils;
 
 import java.math.BigDecimal;
+import java.util.Locale;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Locale-independent normalization fix. LGTM.

---

lizhimins:
Closing: identical change (same file, same function, same line) was merged earlier via #1488 (metrics auth modes) — yyqdbngt's root-locale series (#1486-#1495) landed first.

PR #1490, opened 2026-08-10
fix(ai): normalize provider names with root locale

yyqdbngt:
### Problem

OpenAI-compatible provider names were lowercased with the JVM default locale. Under Turkish locale, uppercase `OPENAI` becomes `openaı`, so a supported provider is rejected before any request is sent.

### Fix

Normalize provider identifiers with `Locale.ROOT` when checking gateway support and provider-specific behavior.

### Verification

`mvn -B -ntp -Dmaven.compiler.proc=full -Dtest=OpenAiCompatibleLlmClientTest test`

Tests run: 13, failures: 0, errors: 0.

PR #1554, opened 2026-08-11
fix(ai): normalize LLM providers locale-independently

### LINKED ISSUE #1553: LLM provider detection depends on the JVM default locale
## Problem
OpenAI-compatible provider names are lowercased with the JVM default locale. Under a Turkish locale, an upper-case supported provider such as `TONGYI` becomes `tongyı`, so the client incorrectly rejects it as unsupported.

## Proposed fix
Use `Locale.ROOT` for provider identifier normalization and add a Turkish-locale regression test.

123123213weqw:
## Summary
- normalize LLM provider identifiers with `Locale.ROOT`
- keep supported providers such as Tongyi available under Turkish JVM locales
- add a focused regression test

## Testing
- `mvn -B -ntp -Dtest=OpenAiCompatibleLlmClientTest test` — 13 tests pass when applied with #1548

Fixes #1553

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
@@ -37,6 +37,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Locale-independent normalization fix. LGTM.

---

lizhimins:
Closing: identical change (same file, same function, same line) was merged earlier via #1490 (AI provider names) — yyqdbngt's root-locale series (#1486-#1495) landed first.

PR #1559, opened 2026-08-11
fix(topic): guard control-prefixed CSV formulas

### LINKED ISSUE #1558: Topic CSV export allows control-prefixed spreadsheet formulas
## Problem
Topic CSV export only neutralizes cells whose first character is `=`, `+`, `-`, or `@`. Spreadsheet applications may also interpret formulas after a leading tab or line break, so a topic field beginning with control whitespace can bypass the current CSV-injection guard.

## Proposed fix
Treat leading tab, CR, and LF as formula-risk prefixes and add an export regression test for a tab-prefixed formula value.

123123213weqw:
## Summary
- neutralize topic CSV cells starting with tab, CR, or LF in addition to formula operators
- retain normal CSV quote escaping
- cover a tab-prefixed formula payload in the export test
- document the existing instance-clear state update for the targeted ESLint rule

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx`
- `NODE_OPTIONS=--no-experimental-webstorage npx eslint src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx`

Fixes #1558

### DIFF
--- a/web/src/pages/instance/__tests__/TopicPage.test.tsx
+++ b/web/src/pages/instance/__tests__/TopicPage.test.tsx
@@ -167,7 +167,7 @@ describe('TopicPage', () => {
         ...buildTopics(1)[0],
         name: 'orders-topic',
         namespace: 'trade',
-        remark: 'orders, "critical"',
+        remark: '\t=orders, "critical"',
       },
       {
         ...buildTopics(1)[0],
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Security fix: extends CSV formula injection guard to also neutralize cells starting with tab, CR, or LF characters. Small, focused change with proper test coverage.

LGTM.

---

PR #1561, opened 2026-08-11
fix(consumer): guard control-prefixed CSV formulas

### LINKED ISSUE #1560: Consumer Group CSV export allows control-prefixed spreadsheet formulas
## Problem
Consumer Group CSV export only neutralizes cells beginning directly with a formula operator. A leading tab or line break can be accepted as whitespace by spreadsheet applications and expose the following formula, bypassing the current CSV-injection guard.

## Proposed fix
Treat tab, CR, and LF as dangerous leading characters and add an export regression test.

123123213weqw:
## Summary
- neutralize Consumer Group CSV cells starting with tab, CR, or LF
- preserve normal quote escaping and filter behavior
- add a regression assertion for a CR-prefixed formula payload
- document two existing instance-scope state resets for the targeted ESLint rule

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx`
- `NODE_OPTIONS=--no-experimental-webstorage npx eslint src/pages/instance/consumer.tsx src/pages/instance/__tests__/ConsumerPage.test.tsx` (passes with one existing Fast Refresh warning)

Fixes #1560

### DIFF
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -195,7 +195,7 @@ describe('Consumer page', () => {
       {
         ...group,
         name: 'orders-cg',
-        namespace: 'trade',
+        namespace: '\r=formula-risk',
         subscribedTopics: ['orders-topic', 'payments,topic'],
       },
       {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Security fix: extends CSV formula injection guard to also neutralize cells starting with tab, CR, or LF characters. Small, focused change with proper test coverage.

LGTM.

---

PR #1566, opened 2026-08-11
fix(ai): ignore stale LLM connection results

### LINKED ISSUE #1565: LLM settings can show a stale connection result for a changed provider
## Problem
The LLM settings page applies a connection-test result even when the form configuration changes while the request is in flight. A successful result for Tongyi, for example, can appear after the user has switched the form to OpenAI. The old request can also clear loading state belonging to later work.

## Proposed fix
Track connection-test generations, invalidate the active generation whenever form values change or the page unmounts, and only apply result/error/loading state for the current generation. Add a regression test that changes providers before an old test resolves.

123123213weqw:
## Summary
- assign LLM connection tests a request generation
- invalidate an in-flight test whenever any tested form value changes
- ignore stale result, error, and loading updates after configuration changes or unmount
- add a provider-switch regression test

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/studio/__tests__/LlmSettingsPage.test.tsx`
- `NODE_OPTIONS=--no-experimental-webstorage npx eslint src/pages/studio/LlmSettings.tsx src/pages/studio/__tests__/LlmSettingsPage.test.tsx`

Fixes #1565

### DIFF
--- a/web/src/pages/studio/LlmSettings.tsx
+++ b/web/src/pages/studio/LlmSettings.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import {
   Alert,
   Button,
@@ -80,6 +80,7 @@ const LlmSettingsPage: React.FC = () => {
   const [apiKeyConfigured, setApiKeyConfigured] = useState(false);
   const [modelOptions, setModelOptions] = useState<{ value: string; label: string }[]>([]);
   const [testResult, setTestResult] = useState<TestState | null>(null);
+  const testRequestIdRef = useRef(0);
 
   const buildModelOptions = (
     nextProvider: string,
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a race condition in the LLM connection test flow. When a user changes the provider or form values while a test is in-flight, the stale result could overwrite the new state. The fix uses a `testRequestIdRef` counter to invalidate previous test results — only the most recent test request's result is applied.  The implementation is clean: - `invalidateTestRequest()` increments the counter and resets UI state - Called on provider change and form value changes (`onValuesChange`) - The async handler checks `testRequestIdRef.current === requestId` before applying results  The test validates that stale results are properly discarded. Good fix.  LGTM.  --- *Automated review by RockteMQ-AI · Community Assistant* REVIEW_EOF  cat > /tmp/review_1563.txt << 'REVIEW_EOF' ### Summary Fixes a race condition in the DLQ retry flow. When a user switches instances while a retry operation is in-flight, the stale completion callback could update the wrong instance's UI state. The fix uses a `retryRequestIdRef` counter (same pattern as #1566) to discard stale retry results.

The implementation is thorough:
- Request ID is captured before the async call
- Completion callback checks ID match before updating state
- `handleInstanceChange` increments the counter to invalidate pending operations
- The test validates that stale completions from a previous instance are properly ignored

Good fix with proper test coverage.

LGTM.

---
*Automated review by RockteMQ-AI · Community Assistant*

PR #1588, opened 2026-08-11
[studio][ISSUE #1587] Tolerate unavailable browser storage

### LINKED ISSUE #1587: [studio][Bug] API requests fail when browser storage is unavailable
## Problem

The authentication storage helpers already tolerate browsers where Web Storage is unavailable, but the Axios request interceptor reads `localStorage` directly:

```ts
const token = localStorage.getItem(TOKEN_STORAGE_KEY);
```

Browsers can throw a `SecurityError` for this call under restrictive privacy settings, sandboxed contexts, or disabled storage. In that case the interceptor rejects every API request before it reaches the network, including requests that could run without a stored token.

## Expected behavior

- Failure to read browser storage should be treated as an absent token.
- The API request should still be sent without an `Authorization` header.
- Normal stored-token behavior must remain unchanged.
- A regression test should exercise a throwing `localStorage.getItem` implementation.

## Scope

Frontend API client and focused unit test only. No server-side authentication behavior changes.

## Track

Track 1 / Studio authentication and client reliability.

messere1:
## Summary

- read the persisted bearer token through a storage-safe helper
- treat Web Storage access failures as an absent token instead of rejecting the request
- add a regression test that throws `SecurityError` from `localStorage.getItem`

## Root cause

The authentication storage module already catches browser storage failures, but the Axios request interceptor bypassed it and accessed `localStorage` directly. A browser `SecurityError` therefore aborted all API calls before they reached the network.

## User impact

Studio can continue issuing unauthenticated or public requests in restricted-storage environments. Normal bearer-token attachment remains unchanged when storage is available.

Closes #1587

## Validation

- `npm test -- --run src/api/client.test.ts` (11 tests passed)
- `npx eslint src/api/client.ts src/api/client.test.ts --max-warnings=0`
- `npx prettier --check src/api/client.ts src/api/client.test.ts`
- `npm run build`

### DIFF
--- a/web/src/api/client.test.ts
+++ b/web/src/api/client.test.ts
@@ -120,6 +120,24 @@ describe('API client response contract', () => {
     await client.get('/clusters');
   });
 
+  it('sends requests without authorization when browser storage is unavailable', async () => {
+    const getItem = vi.spyOn(localStorage, 'getItem').mockImplementation(() => {
+      throw new DOMException('Storage is unavailable', 'SecurityError');
+    });
+    mock.onGet('/clusters').reply((config) => {
+      expect(config.headers?.Authorization).toBeUndefined();
+      return [200, { code: 200, message: 'success', data: [] }];
+    });
+
+    try {
+      await expect(client.get('/clusters')).resolves.toMatchObject({
+        data: { code: 200, data: [] },
+      });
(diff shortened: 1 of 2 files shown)


lizhimins:

PR #1596, opened 2026-08-11
[studio][ISSUE #1595] Preserve routed instance fallback

### LINKED ISSUE #1595: [studio][Bug] Preserve routed instance when instance discovery fails
## Problem

`useInstanceFilter` is documented to keep instance-scoped pages usable when `listInstances()` fails, but its selected-ID calculation only accepts a route instance that is present in the loaded list:

```ts
routeInstanceId && instances.some((instance) => instance.id === routeInstanceId)
```

When instance discovery rejects, `instances` remains empty and `/instance/<id>/topic`, Consumer, Message, ACL, DLQ, and Resource Plan pages all receive an empty `selectedInstanceId`. They then disable actions or skip their instance-scoped data requests even though the URL already contains a usable instance ID.

## Expected behavior

- If instance discovery fails, preserve the instance ID from an instance-scoped route.
- If discovery succeeds, continue validating the route ID and redirecting unknown IDs to the first available instance.
- Static instance routes should continue selecting the first discovered instance.
- Add focused hook tests for failed discovery and successful unknown-ID fallback.

## Scope

Shared frontend instance-filter hook and tests only.

## Track

Track 1 / Studio multi-instance navigation reliability.

messere1:
## Summary

- track whether shared instance discovery failed
- preserve the instance ID from an instance-scoped route when discovery is unavailable
- retain successful-discovery validation and unknown-route fallback behavior
- add a focused regression test for the failed-discovery path

## Root cause

`useInstanceFilter` only accepted a routed instance ID when it appeared in the discovered instance list. A failed `listInstances()` call left that list empty, so the hook discarded the explicit route context despite its documented fallback behavior.

## User impact

Instance-scoped Topic, Consumer, Message, ACL, DLQ, and Resource Plan pages can continue using their routed instance when the shared discovery endpoint is temporarily unavailable.

Closes #1595

## Validation

- `npm test -- --run src/hooks/useInstanceFilter.test.tsx` (3 tests passed)
- `npx eslint src/hooks/useInstanceFilter.ts src/hooks/useInstanceFilter.test.tsx --max-warnings=0`
- `npx prettier --check src/hooks/useInstanceFilter.ts src/hooks/useInstanceFilter.test.tsx`
- `npm run build`

### DIFF
--- a/web/src/hooks/useInstanceFilter.test.tsx
+++ b/web/src/hooks/useInstanceFilter.test.tsx
@@ -33,6 +33,24 @@ function InstanceRouteProbe() {
 }
 
 describe('useInstanceFilter', () => {
+  it('preserves the routed instance when instance discovery fails', async () => {
+    instanceServiceMocks.listInstances.mockRejectedValue(new Error('instance service unavailable'));
+
+    render(
+      <MemoryRouter initialEntries={['/instance/instance-route-1/topic']}>
+        <Routes>
+          <Route path="/instance/:instanceId/topic" element={<InstanceRouteProbe />} />
+        </Routes>
+      </MemoryRouter>,
+    );
+
+    await waitFor(() => {
+      expect(
+        screen.getByText('/instance/instance-route-1/topic|instance-route-1'),
+      ).toBeInTheDocument();
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


lizhimins:

PR #1600, opened 2026-08-11
fix: use Tencent topic count metadata

### LINKED ISSUE #1599: [Bug] Tencent countTopics should use TotalCount instead of full topic listing
`TencentInstanceProvider.countTopics(...)` only needs the number of topics, but the earlier implementation could fall back to listing and mapping every topic. Tencent's `DescribeTopicList` response already provides `TotalCount`, so the count path should request a single item and use that value directly.

The response metadata must remain authoritative even when the returned `Data` array is empty. PR #2041 handles that case and covers the request parameters, oversized totals, and Tencent consumer-subscription pagination.

tju-yxq:
Avoid listing every Tencent topic only to calculate a count. `countTopics(...)` now requests a minimal page and uses the OpenAPI `TotalCount` value when it is available, with a safe fallback for missing metadata.

This fixes #1599. The empty-data response edge case and oversized-count behavior were later completed and tested in merged PR #2041.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentInstanceProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentInstanceProvider.java
@@ -78,7 +78,19 @@ public class TencentInstanceProvider implements InstanceProvider {
 
     @Override
     public int countTopics(String instanceId) {
-        return listTopics(instanceId, null, null, false).size();
+        Context context = resolve(instanceId);
+        DescribeTopicListRequest request = new DescribeTopicListRequest();
+        request.setInstanceId(context.cloudInstanceId());
+        request.setOffset(0L);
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Effective optimization: use API's `TotalCount` instead of fetching all topics just to count them. Reduces latency from 5+ seconds to instant for instances with many topics.

### Findings
- **[Info]** `Math.toIntExact(total)` could theoretically throw `ArithmeticException` if `TotalCount` exceeds `Integer.MAX_VALUE`, but topic counts are well within int range in practice.
- **[Info]** Fallback to full scan when `TotalCount` is null is defensive and correct.
- **[Info]** Consider adding a unit test for `countTopics` to verify the optimization path (the PR description mentions tests but no test file is included).

### Verdict
Approve. Solid performance improvement with proper fallback handling.

---

PR #1601, opened 2026-08-11
[studio][ISSUE #1598] Roll back partial auth sessions

### LINKED ISSUE #1598: [studio][Bug] Failed auth persistence can leave a partial browser session
## Problem

`persistAuthSession` writes the token, username, and admin flag as three separate `localStorage.setItem` calls, then swallows any storage exception.

If the first write succeeds but a later write fails (for example because storage quota or browser policy changes), the browser is left with a partial session. On reload, `readAuthSession` can restore a token without the corresponding user/admin identity, producing inconsistent authentication UI and authorization state.

## Expected behavior

- Either all persisted session keys are written, or none remain from the failed write attempt.
- The in-memory auth store may continue to work when storage is unavailable, as it does today.
- Add a regression test that fails a later `setItem` call and verifies every session key is removed.

## Scope

Frontend auth storage helper and unit test only. No token format or server authentication changes.

## Track

Track 1 / Studio authentication reliability.

messere1:
## Summary
- clear every persisted authentication key if a later localStorage write fails
- prevent a new token from being paired with stale user or admin metadata
- cover partial-write rollback with a quota failure regression test

## Root cause and impact
The session was persisted through three independent localStorage writes, but failures after the first write were swallowed. A quota or browser-storage failure could therefore leave a partial session that was restored on the next page load. The rollback keeps the in-memory login usable while ensuring no inconsistent browser session survives.

Closes #1598

## Validation
- npm test -- --run src/stores/authStorage.test.ts (5 passed)
- npm run lint -- src/stores/authStorage.ts src/stores/authStorage.test.ts
- npm run build
- git diff --check

### DIFF
--- a/web/src/stores/authStorage.test.ts
+++ b/web/src/stores/authStorage.test.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import {
   clearAuthSession,
   persistAuthSession,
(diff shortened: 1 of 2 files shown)


lizhimins:

PR #1606, opened 2026-08-11
fix: paginate Tencent topic subscriptions

tju-yxq:
This PR attempted to paginate `getTopicConsumers(...)`, but the same fix had already landed in merged PR #1584 with a stronger stop condition based on `SubscriptionCount` as well as short pages.

It was closed as a duplicate instead of being rebased. The full-final-page edge case was later added to the merged regression coverage in #2041.


lizhimins:
Closing as duplicate: the same pagination fix for Tencent getTopicConsumers was already merged via #1584, with a stronger termination condition (subscriptionCount check in addition to short-page).

PR #1630, opened 2026-08-11
[ISSUE #1620] Ignore stale consumer stack responses

### LINKED ISSUE #1620: [Studio][Bug] Stale consumer stack responses overwrite the current client
## Bug

Consumer client stack requests are not versioned. If a user opens stack traces for two clients quickly and the first request finishes last, its stack overwrites the second client's modal while the selected client remains the newer one. Closing or changing instances also leaves old completions active.

## Expected behavior

Apply stack results, errors, and loading completion only for the latest request, and invalidate requests when the modal or instance context closes.

123123213weqw:
## Summary
- version consumer stack requests
- apply result, error, and loading changes only for the latest client
- invalidate pending requests on modal close and instance change

## Validation
- `npm test -- --run src/pages/instance/__tests__/ConsumerPage.test.tsx`
- combined frontend build: `npm run build`

Fixes #1620

### DIFF
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import { App } from 'antd';
-import { render, screen, waitFor } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type React from 'react';
 import { MemoryRouter } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Prevents stale consumer stack responses from overwriting newer data. Same request-ID guard pattern as #1631, applied to the consumer stack loading. Also captures `groupName` at call time to avoid stale closure issues.

**Strengths:**
- Correct stale-response guard pattern
- Good test that verifies older responses are ignored
- Captures `groupName` at call time — avoids stale closure bug

LGTM.

---

PR #1638, opened 2026-08-11
[studio][ISSUE #1636] Preserve concurrent NameServer additions

### LINKED ISSUE #1636: [studio][Bug] Concurrent NameServer additions can drop a saved address
### Description
The Ops page appends a successfully saved NameServer address with a state value captured before the asynchronous request starts:

```ts
setNamesrvAddrList([...namesrvAddrList, addr]);
```

The Add action remains available while a request is in flight. If two different additions are submitted before either completes, both handlers capture the same original list. Whichever request finishes last replaces the other successful addition in the UI. An earlier completion can also clear text entered for the next request.

### Steps to reproduce
1. Enter NameServer address A and click Add while delaying its response.
2. Enter address B and click Add before A completes.
3. Complete both successful requests in either order.

### Expected behavior
Both server-confirmed addresses remain in the selector, and one completion does not clear a newer input value.

### Actual behavior
Only the address from the last state update remains in the local list because both updates spread the same stale array.

### Proposed fix
Use a functional list update that deduplicates against the latest state, and clear the input only when it still contains the address associated with the completed request. Add a deterministic concurrent-add regression test.

messere1:
## Summary
- append saved NameServer addresses with a functional state update
- deduplicate against the latest address list instead of a request-time snapshot
- clear the address input only when it still belongs to the completed request
- cover out-of-order concurrent additions and newer input with a regression test

## Root cause and impact
The asynchronous add handler spread `namesrvAddrList` captured before the request. Two overlapping successful additions therefore wrote independent copies of the same old list, and the last completion hid the other saved address. Each completion also cleared any newer input. Updating from current state preserves all confirmed additions without disturbing subsequent typing.

Closes #1636

## Validation
- npm test -- --run src/pages/studio/__tests__/Ops.test.tsx (6 passed)
- npx eslint src/pages/studio/Ops.tsx src/pages/studio/__tests__/Ops.test.tsx
- npm run build
- git diff --check

### DIFF
--- a/web/src/pages/studio/Ops.tsx
+++ b/web/src/pages/studio/Ops.tsx
@@ -125,10 +125,10 @@ const OpsPage: React.FC = () => {
     }
     try {
       await addNameSvrAddr(addr);
-      if (!namesrvAddrList.includes(addr)) {
-        setNamesrvAddrList([...namesrvAddrList, addr]);
-      }
-      setNewNamesrvAddr('');
+      setNamesrvAddrList((addresses) =>
+        addresses.includes(addr) ? addresses : [...addresses, addr],
+      );
+      setNewNamesrvAddr((current) => (current.trim() === addr ? '' : current));
       message.success(t('common.success'));
     } catch {
       message.error(t('common.failure'));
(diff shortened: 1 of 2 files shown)


lizhimins:
Closing: superseded by #1628, which serializes the same Ops NameServer add/remove/update paths with a stronger in-flight guard.

PR #1644, opened 2026-08-11
fix(proxy): resolve endpoints by selected instance

### LINKED ISSUE #1113: design: persist and scope Proxy addresses by managed instance
## Problem

The Studio Proxy page currently relies on process-local proxy address data and default endpoints. That configuration is neither persisted nor tied to the selected managed instance, so a restart or a multi-instance deployment can display or operate on the wrong Proxy target.

## Design scope

Define a durable, instance-scoped Proxy connection model before adding CRUD:

- decide whether Proxy addresses belong to the existing `Instance.endpoint` model or require a separate one-to-many address entity;
- define migration and backward compatibility for existing single endpoint instances;
- define validation, deduplication, and masking rules for any credentials;
- make reads and writes resolve the selected instance rather than a process-global default;
- document which lifecycle operations can be backed by the current RocketMQ Admin APIs and which must remain unavailable.

## Out of scope

This Issue does not add simulated Proxy configuration, guessed runtime operations, or a database schema before the ownership model is agreed.

Aias00:
Closes #1113

## Summary
- resolve Proxy home-page addresses from the selected managed `PROXY` instance
- remove process-local default address state and client-side `localStorage` writes
- keep legacy mutation routes explicit but unavailable, directing users to managed instance configuration
- add instance selection and regression coverage for instance-scoped requests

## Verification
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) mvn -f server/pom.xml -Dtest=ProxyAddressServiceTest,ProxyCompatControllerTest test`
- `npm test -- --run src/api/proxy.test.ts src/pages/studio/__tests__/Proxy.test.tsx`
- `npm run build`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
@@ -21,65 +21,42 @@ import org.apache.rocketmq.studio.common.exception.BusinessException;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
-import java.util.ArrayList;
-import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import org.apache.rocketmq.studio.common.domain.enums.InstanceType;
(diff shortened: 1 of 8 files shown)


Aias00:
Superseded by #1746, which now resolves Proxy endpoints by selected managed instance while preserving the corrected reload and compatibility contract.

PR #1653, opened 2026-08-11
fix: distinguish DLQ scan failures from empty results

### LINKED ISSUE #1651: [Studio][Bug] Distinguish DLQ resend scan failures from empty results
## Problem

`RocketMQDLQProvider.resendMessages()` treats a failed dead-letter scan as an empty scan. `collectDeadLetters()` catches both queue-level errors and the top-level consumer/start/fetch errors, logs them, and returns an empty `List<MessageExt>`. The caller then records and returns `NO_MESSAGES`.

This makes a remote failure indistinguishable from a successful scan that found no messages, so an operator can incorrectly conclude that a DLQ is empty.

## Current behavior

- per-queue scan errors are swallowed at `RocketMQDLQProvider.collectDeadLetters()`;
- a failed consumer start or queue discovery is swallowed too;
- `resendMessages()` maps the resulting empty list to `NO_MESSAGES`.

## Expected behavior

- an all-failed scan returns a structured 502 error instead of `NO_MESSAGES`;
- a partially failed scan preserves resend results but exposes that the scan is incomplete;
- audit records distinguish successful empty scans from scan failures;
- add regression coverage for both all-failed and partial-failure paths.

## Scope

Track 1 / DLQ runtime diagnostics. This does not change normal pull-offset handling or resend transport semantics.

Aias00:
## Summary
- return a gateway error when every DLQ queue scan fails instead of reporting an empty resend
- preserve partial scan results with explicit incomplete-scan metadata and audit fields
- surface incomplete scans in the DLQ retry UI

Closes #1651

## Verification
- `mvn -f server/pom.xml -Dtest=RocketMQDLQProviderTest test`
- `npm --prefix web test -- --run src/pages/instance/__tests__/DLQPage.test.tsx`
- `npm --prefix web run build`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQResendResultVO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQResendResultVO.java
@@ -26,4 +26,6 @@ public class DLQResendResultVO {
     int resent;
     int failed;
     String outcome;
+    boolean scanIncomplete;
+    int failedQueueCount;
 }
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
### Summary
Distinguishes DLQ scan failures from genuinely empty DLQ queues by adding `scanIncomplete` and `failedQueueCount` fields to the result. When the scan fails entirely, the error is propagated with a proper audit trail instead of silently returning zero results.

Good error handling improvement. The `DeadLetterScanResult` record cleanly encapsulates the scan outcome. LGTM.

---

PR #1655, opened 2026-08-11
fix: surface dashboard runtime collection failures

### LINKED ISSUE #1654: [Studio][Bug] Surface Dashboard runtime collection failures instead of an empty overview
## Problem

`RocketMQDashboardProvider.collectDashboardData()` catches every Admin exception and returns `emptyDashboard()`. A NameServer/Broker outage, ACL denial, timeout, or malformed topology therefore reaches the UI as a successful 200 response with all counts set to zero. This is indistinguishable from a healthy empty cluster and can mislead operators.

## Scope

- Preserve the existing empty dashboard only for a valid, genuinely empty topology.
- Convert runtime collection failures and malformed/unavailable topology into a structured upstream failure response.
- Add regression tests for Admin failures and valid empty topology.

## Track

Track 1 / Control Plane runtime observability.

Aias00:
## Summary
- retain the empty overview for a valid empty topology
- return a structured 502 when topology is unavailable or top-level Admin collection fails
- cover both failure paths and valid empty topology

Closes #1654

## Verification
- `mvn -f server/pom.xml -Dtest=RocketMQDashboardProviderTest test`


RockteMQ-AI (approved):
### Summary
Changes dashboard data collection to throw `BusinessException` instead of returning an empty dashboard when the NameServer is unavailable. This prevents the UI from showing a misleading "healthy but empty" state when the cluster is actually unreachable.

Good error handling improvement. The `BusinessException` pass-through pattern is consistent with #1665. LGTM.

---

PR #1662, opened 2026-08-11
fix: normalize ACL bound types with root locale

### LINKED ISSUE #1661: [Studio][Bug] Normalize ACL 2.0 bound types with Locale.ROOT
## Problem

`AclService.isValidAcl2BoundType()` uses `String.toUpperCase()` without an explicit locale. On locale-sensitive JVMs such as Turkish, valid lower-case input can normalize unexpectedly and be rejected.

## Expected behavior

Normalize protocol tokens with `Locale.ROOT` and add a regression test under a Turkish default locale.

## Track

Track 1 / ACL management correctness.

Aias00:
## Summary
- normalize ACL 2.0 bound type protocol tokens with `Locale.ROOT`
- cover a Turkish JVM default locale regression

Closes #1661

## Verification
- `mvn -f server/pom.xml -Dtest=AclServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -28,6 +28,7 @@ import org.springframework.stereotype.Service;
 
 import java.time.LocalDateTime;
 import java.util.List;
+import java.util.Locale;
 import java.util.UUID;
 
 @Slf4j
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes the Turkish-I problem in ACL bound type validation by using `Locale.ROOT` for `toUpperCase()`. Without this, `"topic".toUpperCase()` in a Turkish locale produces `"TOPIC"` with a dotless I, which fails the switch match.

Excellent catch. The test that sets `Locale.setDefault(Locale.forLanguageTag("tr-TR"))` properly validates the fix. LGTM.

---

PR #1665, opened 2026-08-11
fix: surface failed cluster detail refreshes

### LINKED ISSUE #1663: [Studio][Bug] Do not treat failed live Cluster refreshes as missing clusters
## Problem

`RocketMQClusterProvider.refreshClusterDetail()` catches every Admin exception and returns `null`. `ClusterService.resolveCluster()` interprets `null` as a missing live cluster and falls back to persisted metadata. A failed NameServer refresh can therefore cause a configuration update to operate on stale persisted Broker targets or report a misleading 404.

## Expected behavior

- Keep `null` only for a genuinely absent cluster or no configured endpoint.
- Convert Admin/NameServer refresh failures into a structured 502 so callers do not fall back to stale metadata.
- Add provider and service regression coverage.

## Track

Track 1 / cluster runtime operations.

Aias00:
## Summary
- return a structured 502 for failed live Cluster detail refreshes
- retain `null` only for absent topology/cluster cases
- prevent configuration updates from falling back to persisted cluster metadata after a failed refresh

Closes #1663

## Verification
- `mvn -f server/pom.xml -Dtest=RocketMQClusterProviderTest,ClusterServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClusterProvider.java
@@ -132,8 +132,12 @@ public class RocketMQClusterProvider implements ClusterProvider {
                 return buildClusterVO(clusterId, brokers, nameServers);
             });
         } catch (Exception e) {
+            if (e instanceof BusinessException businessException) {
+                throw businessException;
+            }
             log.warn("Failed to refresh cluster detail for {}: {}", clusterId, e.getMessage());
-            return null;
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Changes `refreshClusterDetail` to throw a `BusinessException` instead of silently returning `null` when the cluster refresh fails. This prevents the caller from falling back to stale persisted data when the live cluster is unavailable.

Good error handling improvement with proper test coverage. The `BusinessException` pass-through ensures existing error codes are preserved. LGTM.

---

Issue #1669, opened 2026-08-11
[Invalid] OpenAiCompatibleLlmClient redirect-following report premise is incorrect


tju-yxq:
This report incorrectly assumed that Java's `HttpClient` follows redirects by default. On Java 21, a client built without calling `followRedirects(...)` uses `HttpClient.Redirect.NEVER`.

PR #1671 only makes that existing safe policy explicit. It is a defensive readability change, not a fix for an actual redirect-following vulnerability, so this issue is closed as invalid.

PR #1671, opened 2026-08-11
refactor: make the LLM redirect policy explicit

tju-yxq:
Set `HttpClient.Redirect.NEVER` explicitly on the OpenAI-compatible LLM client.

The original issue #1669 incorrectly assumed that Java's `HttpClient` follows redirects by default. Java 21 already defaults to `NEVER`, so this merged change documents and locks in the existing safe behavior rather than fixing an actual redirect-following vulnerability.


RockteMQ-AI (approved):
### Summary
One-line fix that disables HTTP redirect following in the OpenAI-compatible LLM client. Prevents potential SSRF via redirect — a malicious LLM endpoint URL could redirect requests to internal services (e.g., cloud metadata at `169.254.169.254`).

Works in conjunction with the `UrlHostGuard` in #1673 for defense-in-depth. LGTM.

---

PR #1673, opened 2026-08-11
[Studio] fix(security): harden SSRF guards, default login and edge exposure

zhaohai666:
# PR: fix(security): harden SSRF guards, default login and edge exposure

**Branch:** `feature/studio-security-hardening`
**Commit:** `814bf241` — pushed to `origin/feature/studio-security-hardening`
**Base:** `apache:rocketmq-studio` @ `990ba092`
**PR create link:** https://github.com/zhaohai666/rocketmq-dashboard/pull/new/feature/studio-security-hardening

## Summary

Security hardening for the P0 items from the 2026-08-11 audit: SSRF guards are now
shared and applied on every path that makes a server-side request to a
caller-supplied URL, login is required by default and the auth gate fails closed,
and the deployment no longer exposes the backend port or the full actuator surface.

### Shared SSRF guard (`server/.../common/util/UrlHostGuard.java`, new)

Rejects loopback, link-local (including the cloud-metadata range `169.254.169.254`)
and any-local hosts. Unlike the old per-service check, unresolvable hosts are
rejected (fail-closed) instead of being handed to the connection layer. Private
site-local ranges stay allowed for on-premise Prometheus / internal LLM gateways.

Applied on every path:

- **Data source save** — `SettingsService.createDataSource/updateDataSource` now
  validate the URL when saving. Previously only the test path was guarded, so a
  data source pointing at `127.0.0.1` could be stored and later queried.
- **Data source query** — `AbstractPrometheusCompatibleMetricsSource.queryRangeUri`
  validates the stored base URL on every request (defense in depth against direct
  DB edits).
- **Data source test** — `SettingsService.isAllowedDataSourceHost` now delegates to
  the guard.
- **LLM config** — `LlmConfigService.validate()` (used by both save and test)
  rejects link-local/metadata hosts while allowing loopback, so a local `ollama`
  gateway keeps working.

### Authentication

- Default `studio.auth.login-required=true` in `application.yml`,
  `deploy/docker-compose.yml` and `deploy/deploy.sh`.
- `AuthInterceptor.isLoginRequired()` fails closed when the settings store cannot
  be read (previously it defaulted to *no* login requirement).

### Deployment

- `deploy/docker-compose.yml`: MySQL root password externalized to the
  `MYSQL_ROOT_PASSWORD` env var (compose-default `studio123` kept for
  out-of-the-box use); the healthcheck reads the password from the environment
  instead of hard-coding it in the command line.
- `web/nginx.conf` + `deploy/nginx.conf`: only `/actuator/health` is proxied at
  the edge; the rest of `/actuator/` returns 404.
- `deploy/deploy.sh`: the backend `8888` port is no longer published to the host
  (nginx already proxies `/api/` inside the podman network).

## Verification

- `mvn compile` — BUILD SUCCESS (checkstyle clean).
- `SettingsServiceTest` 31/31 (incl. two new SSRF rejection tests:
  create rejects loopback, update rejects metadata address),
  `LlmConfigServiceTest` 21/21, `AuthInterceptorTest` 16/16.

## Notes

- Behavior change for existing deployments: login is now required by default;
  set `STUDIO_AUTH_LOGIN_REQUIRED=false` only when the deployment is intentionally
  unauthenticated (e.g. local development).

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -128,14 +128,14 @@ deploy_remote() {
 
   if [[ "$TARGET" == "all" || "$TARGET" == "server" ]]; then
     info "重启 rocketmq-server..."
+    # Backend port is only exposed inside the podman network; the edge nginx proxies /api/.
     ssh "$REMOTE" "
       podman rm -f rocketmq-server 2>/dev/null || true
       podman run -d \
         --name rocketmq-server \
         --network $NETWORK \
         --restart unless-stopped \
-        -p 8888:8888 \
-        -e STUDIO_AUTH_LOGIN_REQUIRED=\"${STUDIO_AUTH_LOGIN_REQUIRED:-false}\" \
+        -e STUDIO_AUTH_LOGIN_REQUIRED=\"${STUDIO_AUTH_LOGIN_REQUIRED:-true}\" \
         -e STUDIO_AUTH_ADMIN_USERNAME=\"${STUDIO_AUTH_ADMIN_USERNAME:-}\" \
         -e STUDIO_AUTH_ADMIN_PASSWORD=\"${STUDIO_AUTH_ADMIN_PASSWORD:-}\" \
         -e STUDIO_METRICS_PROMETHEUS_BASE_URL=\"${STUDIO_METRICS_PROMETHEUS_BASE_URL:-}\" \
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     restart: unless-stopped
     environment:
       TZ: ${TZ:-Asia/Shanghai}
-      MYSQL_ROOT_PASSWORD: studio123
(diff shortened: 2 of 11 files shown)


RockteMQ-AI (approved):
### Summary
Comprehensive security hardening addressing SSRF, authentication defaults, and deployment exposure. Well-structured with a shared `UrlHostGuard` applied consistently across all server-side request paths.

**Strengths:**
- Fail-closed design for both SSRF guard (unresolvable hosts rejected) and auth interceptor (unreadable settings → login required)
- Defense-in-depth: SSRF check on both save and query paths for data sources
- Correct nginx regex ordering (`/actuator/health` before catch-all 404)
- MySQL root password externalized from healthcheck command
- Good test coverage for `UrlHostGuard` edge cases (trailing dots, IPv6-mapped IPv4)

**Minor observations:**
- DNS rebinding is out of scope for this PR but worth noting for future hardening — `InetAddress.getByName()` resolves at guard time, actual connection happens later
- Consider logging blocked SSRF attempts with the original URL for security audit trail

Overall: Excellent security work. LGTM.

---

PR #1683, opened 2026-08-11
fix: collect dashboard topic and group counts by broker

### LINKED ISSUE #1394: [Bug] Dashboard per-cluster overview should populate topic and group counts
The per-cluster dashboard rows used to report zero topics and consumer groups instead of counts derived from that cluster's brokers.

Merged PR #1400 added the per-cluster topic and group calculation. The current tests also verify that topic counts come from broker configuration without per-topic route requests and that groups reported by multiple brokers are de-duplicated; those regression tests were consolidated into #2034.

### LINKED ISSUE #1660: [Studio][Performance] Avoid per-Topic route RPCs in Dashboard cluster overview
## Problem

`RocketMQDashboardProvider.collectDashboardData()` calls `examineTopicRouteInfo(topic)` once for every non-system Topic solely to derive per-cluster topic counts. Large clusters turn one overview refresh into hundreds or thousands of sequential NameServer route requests.

## Why this needs a design decision

A Topic can span brokers and cluster ownership is not represented in the topic list. Skipping route lookup without a replacement would make per-cluster counts incorrect.

## Proposed direction

- Evaluate a bounded/cached route-resolution strategy with an explicit freshness policy, or a backend aggregate endpoint.
- Make any partial result visible to the UI rather than reporting it as a complete count.
- Add a large-topic-count benchmark and ensure refresh latency/RPC volume remain bounded.

## Track

Track 1 / dashboard runtime observability.

Aias00:
## Summary
- derive per-cluster Topic counts from each master Broker's topic-config response
- collect per-cluster Consumer Group counts as sets so a group replicated to multiple Brokers is counted once
- remove the per-Topic `examineTopicRouteInfo` loop from Dashboard collection
- mark a Cluster warning when Topic or Consumer Group counts cannot be read, and keep the healthy-cluster summary consistent with warning status

Closes #1660
Fixes #1394

## Verification
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) PATH="$JAVA_HOME/bin:$PATH" mvn -f server/pom.xml -Dtest=RocketMQDashboardProviderTest test`

The regression tests verify exact Topic and Consumer Group counts across multiple master Brokers, system-topic filtering, zero calls to `examineTopicRouteInfo`, and warning behavior when Broker Topic or Consumer Group config is unavailable.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDashboardProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDashboardProvider.java
@@ -154,7 +154,7 @@ public class RocketMQDashboardProvider implements DashboardProvider {
             Map<String, Set<String>> topicsByCluster = new HashMap<>();
             Set<String> topicCountsUnavailableClusters = new HashSet<>();
             Set<String> allGroups = new HashSet<>();
-            Map<String, Integer> groupsByCluster = new HashMap<>();
+            Map<String, Set<String>> groupsByCluster = new HashMap<>();
             for (String brokerAddr : masterAddrs) {
                 String clusterName = brokerAddrToCluster.get(brokerAddr);
                 try {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR replaces the O(topics) approach of calling `examineTopicRouteInfo` per topic with an O(brokers) approach using `getAllTopicConfig` per master broker. This is a significant performance improvement for clusters with many topics — instead of N API calls (one per topic), it makes M calls (one per broker).

Key improvements:
- Topic counting bounded by broker count instead of topic count
- Proper deduplication of topics across brokers using `Set<String>`
- Cluster-level warning status when topic config is unavailable
- Fixed `healthyClusters` count (was hardcoded to `totalClusters`, now correctly computed)
- Good regression test coverage including multi-broker deduplication and unavailable-config scenarios

LGTM — clean performance fix with correct fallback behavior.

---

PR #1688, opened 2026-08-11
[studio][ISSUE #1686] Clean up failed browser downloads

### LINKED ISSUE #1686: [studio][Bug] Failed browser downloads leak temporary resources
### Description
The shared `downloadBlob` helper appends a temporary anchor, calls `anchor.click()`, then removes the anchor and revokes the Blob URL. Cleanup is not protected by `finally`.

If browser download activation throws (for example because it is blocked by an embedded browser policy or an overridden click implementation), the exception correctly reaches the caller, but the hidden anchor remains in the document and the object URL is never revoked. Repeated failed exports leak both DOM nodes and Blob-backed memory.

The helper is shared by audit, alert-rule, DLQ, and message exports.

### Expected behavior
Download errors should still propagate, while the temporary anchor and object URL are always cleaned up.

### Actual behavior
Cleanup runs only after a successful click.

### Proposed fix
Wrap download activation in `try/finally`, retain the existing success behavior, and add a unit test that forces `click()` to throw and verifies cleanup plus error propagation.

messere1:
## Summary
- always remove the temporary download anchor after activation attempts
- always revoke the Blob URL even when browser download activation throws
- continue propagating the original error so existing callers can report export failure
- cover the exceptional cleanup path with a regression test

## Root cause and impact
Cleanup followed `anchor.click()` without a `finally` block. A blocked or failed browser activation therefore skipped both cleanup operations, leaving a hidden DOM node and Blob-backed URL allocated after every failed export. The shared helper now preserves the same success path and guarantees cleanup.

Closes #1686

## Validation
- npm test -- --run src/utils/download.test.ts (2 passed)
- npx eslint src/utils/download.ts src/utils/download.test.ts
- npm run build
- git diff --check

### DIFF
--- a/web/src/utils/download.ts
+++ b/web/src/utils/download.ts
@@ -21,8 +21,11 @@ export const downloadBlob = (blob: Blob, filename: string) => {
   anchor.href = url;
   anchor.download = filename;
   anchor.style.display = 'none';
-  document.body.appendChild(anchor);
-  anchor.click();
-  anchor.remove();
-  URL.revokeObjectURL(url);
+  try {
+    document.body.appendChild(anchor);
+    anchor.click();
+  } finally {
+    anchor.remove();
+    URL.revokeObjectURL(url);
+  }
 };
--- a/web/src/utils/download.test.ts
+++ b/web/src/utils/download.test.ts
@@ -52,4 +52,26 @@ describe('downloadBlob', () => {
     expect(document.querySelector('a[download="export.csv"]')).not.toBeInTheDocument();
     expect(revokeObjectURL).toHaveBeenCalledWith('blob:download');
   });
+
+  it('removes the anchor and revokes the URL when download activation throws', () => {
+    const createObjectURL = vi.fn(() => 'blob:failed-download');
+    const revokeObjectURL = vi.fn();
+    Object.defineProperty(URL, 'createObjectURL', {
+      writable: true,
+      value: createObjectURL,
+    });
+    Object.defineProperty(URL, 'revokeObjectURL', {
+      writable: true,
+      value: revokeObjectURL,
+    });
+    vi.spyOn(HTMLAnchorElement.prototype, 'click').mockImplementation(() => {
+      throw new Error('Download blocked');
+    });
+
+    const blob = new Blob(['content'], { type: 'text/csv' });
+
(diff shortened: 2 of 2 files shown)


RockteMQ-AI:
**Issue Evaluation**

Category: `question` | Status: **Needs Info**

Thank you for reporting this issue. However, automated verification found that the referenced code components do not exist in the current `apache/rocketmq-dashboard` codebase (checked on `main` branch).

The classes, endpoints, and files mentioned in this issue (e.g., Studio-specific controllers, providers, and frontend components) appear to belong to a separate **RocketMQ Studio** project that has not been merged into this repository.

**To help us triage this correctly, please clarify:**
- [ ] Is this issue targeting a different repository (e.g., a separate `rocketmq-studio` repo)?
- [ ] Is the Studio code available on a feature branch that needs to be merged first?
- [ ] Should this issue be filed against a different project?

Once clarified, we will re-classify and proceed with verification.

---
*Automated evaluation by @RockteMQ-AI*

web/src/utils/download.ts
```diff
@@ -21,8 +21,11 @@ export const downloadBlob = (blob: Blob, filename: string) => {
...
   anchor.download = filename;
   anchor.style.display = 'none';
-  document.body.appendChild(anchor);
-  anchor.click();
-  anchor.remove();
-  URL.revokeObjectURL(url);
+  try {
+    document.body.appendChild(anchor);
+    anchor.click();
+  } finally {
```
RockteMQ-AI, line 27: **[Info]** Good use of `try/finally` to guarantee cleanup. One minor suggestion: consider adding a brief comment explaining why the `finally` block is important (prevents DOM node and Blob URL leaks on failed downloads), as this is a subtle resource-management pattern that future contributors may not immediately recognize.

RockteMQ-AI (approved):
### Summary
Clean fix that wraps `anchor.click()` in a `try/finally` block to guarantee DOM anchor removal and Blob URL revocation even when browser download activation throws. Prevents resource leaks on failed exports.

### Findings
- **[Info]** `web/src/utils/download.ts:27` — Good `try/finally` pattern; consider a brief comment for future maintainers.

### Suggestions
No blocking issues. The change is minimal, correct, and well-tested.

---

lizhimins:

PR #1699, opened 2026-08-11
fix: preserve actual cluster names in client connections

### LINKED ISSUE #1697: [Studio][Bug] Preserve actual cluster names when listing client connections
## Bug

`GET /api/clients?instanceId=<id>` returns connections whose `clusterName` is `null`. `RocketMQClientProvider` passes the optional request filter directly into every result instead of deriving the Broker-to-Cluster mapping from `ClusterInfo`. Consequently, the Client Connections page cannot build its cluster filter from real data.

The optional `clusterId` parameter is also not applied to the Broker scan: requesting a single cluster still reads all Broker producer tables and subscription groups.

## Expected behavior

- Every returned Apache Remoting client connection carries its discovered RocketMQ cluster name.
- When `clusterId` is supplied, scans are limited to that cluster's master Brokers.
- Unknown cluster filters return an empty result without querying unrelated Brokers.

## Scope

Update `RocketMQClientProvider` and its unit tests only. The change uses the existing `ClusterInfo.clusterAddrTable` and `brokerAddrTable` topology response; it adds no API or persistence dependency.

Aias00:
## Summary
- derive each scanned Broker's RocketMQ cluster from the existing NameServer topology
- return the discovered cluster name on Producer and Consumer connection rows when no cluster filter is supplied
- restrict Broker and subscription-group scans when `clusterId` is supplied, avoiding unrelated cluster requests
- safely skip Broker metadata with no address map

Closes #1697

## Verification
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) PATH="$JAVA_HOME/bin:$PATH" mvn -f server/pom.xml -Dtest=RocketMQClientProviderTest test`

The targeted suite covers discovered cluster names, Producer and Consumer filtering, partial Broker failures, and all-Broker failure semantics.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProvider.java
@@ -43,12 +43,10 @@ import org.springframework.stereotype.Service;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 
 /**
  * Live {@link ClientProvider} backed by the RocketMQ admin API. It discovers producer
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Refactors broker topology discovery to preserve actual cluster names in client connections. The new `BrokerTopology` record maps each broker address to its cluster, so connections are correctly attributed when multiple clusters coexist.

Clean refactoring with good null-safety improvements. LGTM.

---

PR #1701, opened 2026-08-11
fix: scope group management to selected instance

### LINKED ISSUE #1700: [Studio][Bug] Scope Group Management to the selected instance
## Bug

The routed Group Management page calls `listConsumerGroups()` without an `instanceId` and provides no managed-instance selector. The request therefore follows the legacy global Apache path, even when Studio manages multiple Apache or cloud instances.

Consumer Group detail calls preserve the returned group's `instanceId`, but the top-level list can already contain data from the wrong runtime.

## Expected behavior

- Load managed instances when the page initializes.
- Select a valid instance before loading consumer groups.
- Send the selected `instanceId` to `listConsumerGroups`.
- Clear stale list/detail state when the selection changes or instance bootstrap fails.

## Scope

Frontend-only change to `GroupManagement.tsx` and its tests, reusing the existing instance and consumer services.

Aias00:
## Summary
- load managed instances before requesting the global consumer group view
- require the selected instance ID for consumer group list requests
- clear stale group and detail state when the instance changes or loading instances fails
- prevent an in-flight request for an old instance from blocking the newly selected instance

Closes #1700

## Verification
- `npm test -- --run src/pages/studio/__tests__/GroupManagement.test.tsx`
- `npm run build`

`npm run lint` remains blocked by existing `react-hooks/set-state-in-effect` errors in `web/src/pages/instance/consumer.tsx`.

### DIFF
--- a/web/src/pages/studio/GroupManagement.tsx
+++ b/web/src/pages/studio/GroupManagement.tsx
@@ -25,6 +25,7 @@ import {
   Tabs,
   Card,
   Row,
+  Select,
   Col,
   Descriptions,
   Space,
@@ -39,6 +40,8 @@ import {
   getConsumerSubscriptions,
   listConsumerGroups,
 } from '../../services/consumerService';
+import { listInstances } from '../../services/instanceService';
+import type { Instance } from '../../api/instance';
 
 // ─── Helpers ────────────────────────────────────────────────────
 type GroupStatus = 'running' | 'warning' | 'stopped';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Scopes group management to the selected instance by adding an instance selector and passing `instanceId` to all API calls. Prevents cross-instance data leakage and ensures the page only loads groups for the explicitly selected instance.

Good UX fix. The instance validation on load and proper cleanup on unmount are well-handled. LGTM.

---

lizhimins:
Rework needed (small): when the instance list is empty, loading stays true forever and the table spins indefinitely (ConsumerPage already handles this path). Please add setLoading(false) on the empty-list path.

Aias00:
Fixed in 1687c512.

The empty instance-list path now clears runtime state and calls `setLoading(false)`, so the table no longer spins indefinitely. Added a regression test verifying the spinner stops and no global group request is made.

Validated with:
- `npx vitest run src/pages/studio/__tests__/GroupManagement.test.tsx`
- `npm run build`

Aias00:
Superseded by #1953, which now combines selected-instance Consumer Group lifecycle handling with the instance capability contract.

PR #1707, opened 2026-08-11
fix: reset consumer state by selected instance

### LINKED ISSUE #1706: [Studio][Bug] Reset Consumer page state at instance lifecycle boundaries
## Problem

`ConsumerPage` synchronously resets details, selection, loading, and group data from effects when the selected instance changes or becomes unavailable. The current React hooks lint rule reports these as cascading renders, and the two effects split the same instance lifecycle across independent state transitions.

## Expected behavior

- Switching the selected instance must discard prior instance-local UI state before loading the next group list.
- A page with no resolved instance must be idle rather than permanently loading.
- The list request remains asynchronous and scoped to the selected instance.
- Avoid synchronous state updates in effects for the instance lifecycle.

## Scope

Consumer page state lifecycle and focused frontend regression tests only.

## Track

Track 1 / Studio multi-instance consumer management.

Aias00:
## Summary
- remount Consumer page content at selected-instance lifecycle boundaries
- discard prior instance detail and selection state before loading a new instance
- keep an unresolved instance idle rather than synchronously resetting state in effects
- add a regression assertion that switching instances closes the old detail dialog

Closes #1706

## Verification
- `npm test -- --run src/pages/instance/__tests__/ConsumerPage.test.tsx`
- `npm run build`
- `npm run lint` (0 errors; existing warnings remain)

### DIFF
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -364,6 +364,7 @@ describe('Consumer page', () => {
     await waitFor(() =>
       expect(consumerService.listConsumerGroups).toHaveBeenCalledWith({ instanceId: 'instance-b' }),
     );
+    expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
     await user.click(await screen.findByRole('button', { name: /详情/ }));
 
     await waitFor(() =>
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes stale consumer state when switching instances by remounting `ConsumerPageContent` via a dynamic `key` prop. This is the correct React pattern for resetting all component state (refs, effects, local state) when a dependency changes, instead of manually clearing each piece of state in a `useEffect`.

Clean approach. The initial `loading` state now correctly reflects whether an instance is selected. LGTM.

---

PR #1709, opened 2026-08-11
fix: retry client connection queries

### LINKED ISSUE #1708: [Studio][Bug] Retry client connection queries after a transient failure
## Problem

The Client Connections page uses one retry button for both instance discovery and runtime connection loading. When `listConnections()` fails but `listInstances()` succeeds, retry only changes the instance-load key. The selected instance ID is unchanged, so the connection effect does not run again. The error is cleared and the page remains an empty list.

## Expected behavior

- Retrying after a runtime client-connection failure must re-run `listConnections()` for the currently selected instance.
- Instance discovery retry behavior remains intact.
- Existing request cancellation must still prevent a stale response from overwriting a newer selected instance.

## Scope

Client Connections page retry state and focused frontend tests only.

## Track

Track 1 / Studio runtime client diagnostics.

Aias00:
## Summary
- add a refresh generation for runtime client-connection queries
- make the existing retry action re-request connections for the currently selected instance
- retain instance discovery retry and stale-response cancellation behavior
- cover recovery from a transient runtime query failure

Closes #1708

## Verification
- `npm test -- --run src/pages/cluster/__tests__/ClientsPage.test.tsx`
- `npm run build`

`npm run lint` is blocked by the Consumer lifecycle errors addressed by pending #1707; this PR introduces no lint errors.

### DIFF
--- a/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
@@ -222,6 +222,25 @@ describe('Clients page', () => {
     expect(within(screen.getByTestId('connection-total')).getByText('0')).toBeInTheDocument();
   });
 
+  it('retries the current instance connection query after a runtime failure', async () => {
+    vi.mocked(connectionsService.listConnections)
+      .mockRejectedValueOnce(new Error('Client connection provider is not configured'))
+      .mockResolvedValueOnce([connection]);
+    const user = userEvent.setup();
+    renderWithProviders(<ClientsPage />);
+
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix — adds a `connectionLoadKey` generation counter so the retry button re-triggers the connection-fetching `useEffect`, not just the instance list. The stale-response cancellation via the `cancelled` flag in the existing cleanup already handles the re-fetch correctly. Test covers the transient-failure → retry → recovery path well.

LGTM.

---

PR #1727, opened 2026-08-11
fix: reject deletion when cloud counts are unavailable

### LINKED ISSUE #1726: [Studio][Bug] Return a structured error when Tencent instance deletion cannot verify groups
## Problem

`InstanceService.deleteInstance` always invokes both `countTopics` and `countGroups` before deleting an instance. `TencentInstanceProvider.countGroups` intentionally throws `UnsupportedOperationException` because that cloud API is not implemented. Consequently, deleting an otherwise empty Tencent instance bubbles the exception into a generic HTTP 500.

## Expected behavior

- Never delete an instance when Studio cannot verify that managed resources are absent.
- Convert unsupported resource-count capabilities into a structured, actionable `501` response instead of an unhandled 500.
- Keep Apache and fully supported cloud provider deletion behavior unchanged.
- Cover the Tencent unsupported-count path with a service-level regression test.

## Scope

Backend instance lifecycle validation only.

## Track

Track 1 / Studio multi-cloud instance management.

Aias00:
## Summary
- convert unsupported cloud resource-count capabilities during instance deletion into a structured 501 response
- prevent deleting an instance when Studio cannot verify that Topics and Consumer Groups are absent
- preserve existing successful deletion and managed-resource conflict behavior
- add Tencent capability regression coverage

Closes #1726

## Verification
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -Dtest=InstanceServiceTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -DskipTests compile`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -239,8 +239,16 @@ public class InstanceService {
 
         InstanceProvider provider = providerRegistry.forVendor(
                 existing.getVendor() == null ? InstanceVendor.APACHE : existing.getVendor());
-        int topicCount = provider.countTopics(id);
-        int consumerGroupCount = provider.countGroups(id);
+        int topicCount;
+        int consumerGroupCount;
+        try {
+            topicCount = provider.countTopics(id);
+            consumerGroupCount = provider.countGroups(id);
+        } catch (UnsupportedOperationException ex) {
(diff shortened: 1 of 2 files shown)


Aias00:
Closing this as redundant after rechecking the global exception path. UnsupportedOperationException is already mapped to HTTP 501 by GlobalExceptionHandler, so the proposed change only rewrites an existing structured response without fixing a distinct behavior.

PR #1746, opened 2026-08-11
fix: align Proxy diagnostics with managed instances

### LINKED ISSUE #1113: design: persist and scope Proxy addresses by managed instance

### LINKED ISSUE #1740: [Studio][Bug] Route Proxy management actions by the selected cluster context
## Problem

The Proxy management page persists and submits `clusterId`, but the backend only validates the field and then ignores it. `ProxyController.reloadProxyConfig` delegates solely by address to a process-global `ProxyAddressService`; `GET /api/proxies` similarly requires a cluster id but returns the same global address list.

This makes the cluster selector misleading and can reload a Proxy that does not belong to the operator-selected cluster.

## Scope

- Remove the misleading editable cluster context from the legacy global Proxy page until an instance/cluster-scoped Proxy provider exists.
- Do not send a `clusterId` that the reload endpoint cannot honor.
- Keep proxy address validation and structured failure behavior unchanged.

## Acceptance criteria

- The page does not imply that a manually entered cluster id scopes Proxy discovery or reload.
- The reload API contract no longer accepts an unused `clusterId`.
- Regression tests cover the reduced API contract and UI call shape.

Track: Control Plane / Proxy management (ARCH-01).

Aias00:
## Summary

- resolve Proxy dashboard endpoints from the selected managed `PROXY` instance
- remove misleading cluster-global and process-local Proxy address state
- keep configuration reload scoped to configured managed endpoints
- return explicit unsupported behavior for legacy add/remove address routes instead of mutating simulated state
- preserve read-only Proxy topology and configuration views

Closes #1113
Closes #1740

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=ProxyAddressServiceTest,ProxyCompatControllerTest,ProxyControllerTest test` (17 tests passed)
- `npx vitest run src/api/proxy.test.ts src/pages/studio/__tests__/Proxy.test.tsx` (11 tests passed)
- targeted ESLint on changed frontend files
- `npm run build`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -DskipTests package`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
@@ -18,6 +18,9 @@
 package org.apache.rocketmq.studio.cluster.proxy;
 
 import org.apache.rocketmq.studio.common.exception.BusinessException;
+import org.apache.rocketmq.studio.common.domain.enums.InstanceType;
+import org.apache.rocketmq.studio.instance.InstanceRepository;
+import org.apache.rocketmq.studio.instance.InstanceVO;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatusCode;
 import org.springframework.http.ResponseEntity;
@@ -29,11 +32,7 @@ import org.springframework.web.client.RestTemplate;
 
 import java.io.IOException;
 import java.time.Duration;
-
-import java.util.ArrayList;
-import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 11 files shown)


RockteMQ-AI (approved):
### Summary
Removes the unused `clusterId` parameter from proxy-related API endpoints (`/proxy/config/reload`, `/proxy/restart`) and corresponding UI components. The `RestartProxyDTO` is replaced with `ProxyConfigReloadDTO` that only carries the `address` field.

This is a clean API simplification — the `clusterId` was being passed around but never used for proxy operations, which only need the target address. The changes are consistent across backend controllers, DTOs, frontend API calls, and UI components.

7 files changed with good consistency across the stack.

---

Aias00:
Superseded by #1950, which now combines Proxy Local/Cluster access types with selected-instance endpoint resolution and corrected reload semantics.

PR #1563, opened 2026-08-11
fix(dlq): ignore stale retry completions

123123213weqw:
## Summary
- assign each DLQ retry submission a request generation
- invalidate the active generation before changing instances and on unmount
- ignore stale success, error, refresh, dialog, and loading updates
- snapshot the submitted group and target for accurate completion messages
- cover an old retry resolving after a new instance retry dialog has opened

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/instance/__tests__/DLQPage.test.tsx`
- `NODE_OPTIONS=--no-experimental-webstorage npx eslint src/pages/instance/dlq.tsx src/pages/instance/__tests__/DLQPage.test.tsx`

Fixes #1562

PR #1748, opened 2026-08-11
fix(audit): show loading during record refresh

123123213weqw:
## Summary
- re-enable the audit table loading state for every record request
- avoid presenting stale rows as settled while filters or pagination refresh
- cover a pending filter-triggered refresh

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/pages/ops/__tests__/AuditPage.test.tsx --reporter=dot`

### DIFF
--- a/web/src/pages/ops/__tests__/AuditPage.test.tsx
+++ b/web/src/pages/ops/__tests__/AuditPage.test.tsx
@@ -168,6 +168,41 @@ describe('Audit page', () => {
     );
   });
 
+  it('shows loading state while refreshed records are pending', async () => {
+    const user = userEvent.setup();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a UX issue where the audit records loading indicator might not appear during refresh because `loading.value = true` and the subsequent data fetch happen in the same microtask. By deferring the fetch to the next microtask, the loading state gets a chance to render first.

The approach works, though an alternative would be to use `nextTick()` from Vue for more explicit timing control. The current implementation is fine for this use case.

Includes test updates to verify the loading behavior.

---

lizhimins:

Issue #123, opened 2022-11-24
Optimize  groupList.query


zhangjidi2016:
With about 800 consumers in the RocketMQ cluster, each query of the list of consumers costs 3~4s and requires optimization of the queryGroupList method.
![image](https://user-images.githubusercontent.com/18254437/203681477-cc4dff1d-984a-472e-a3a8-63addcaa8054.png)

PR #1767, opened 2026-08-11
[ISSUE #1763] Handle empty trace message bodies

### LINKED ISSUE #1763: [Bug] Empty trace message bodies cause NullPointerException
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do (the steps to reproduce)?

Pass a `MessageExt` whose body is null to `MessageTraceView.decodeFromTraceTransData`. Passing a null message has the same failure mode.

- What did you expect to see?

A message without trace payload should produce an empty trace result.

- What did you see instead?

The method constructs a String from `messageExt.getBody()` before checking whether the decoded text is empty. A null message or body therefore throws `NullPointerException`.

2. Please tell us about your environment:

- apache/rocketmq-dashboard master
- baseline commit `770822bbc812bfd0c2a94aa0128ab6f59b50d076`
- JDK 17

3. Other information:

The message and raw body should be checked before decoding. Existing behavior for non-empty trace payloads should remain unchanged.

yuluo-yx:

Avoid NullPointerException when trace decoding receives a null message or a message without a body. Such input now produces the same empty trace result as an empty payload.

Fixes #1763

## Brief changelog

- Check the message and raw body before constructing the trace-data String.
- Add regression tests for null messages and null bodies.

## Verifying this change

- JDK: Temurin 17.0.19
- Backend-only Maven verification: `mvn -q compiler:compile compiler:testCompile -Dtest=<target> surefire:test`
- `MessageTraceViewTest`: 2 tests, 0 failures, 0 errors
- `mvn validate`: passed
- `git diff --check`: passed
- PR scope check: passed (2 files, 45 changed lines, one commit, base `master`)
- Full lifecycle, integration, E2E, and container checks were not run because this five-PR workflow requires lightweight local verification and leaves heavyweight checks to upstream CI.

Follow this checklist to help us incorporate your contribution quickly and easily.

- [x] Make sure there is a Github issue filed for the change. This PR addresses only that issue.
- [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit has a meaningful subject and body.
- [x] Write a pull request description that explains what, how, and why.
- [x] Write necessary unit tests to verify the logic correction.
- [ ] Run the repository full basic, install, and integration command matrix. The focused backend checks above pass; heavyweight checks are delegated to upstream CI.
- [ ] If this contribution is large, file an Apache Individual Contributor License Agreement.

### DIFF
--- a/src/main/java/org/apache/rocketmq/dashboard/model/MessageTraceView.java
+++ b/src/main/java/org/apache/rocketmq/dashboard/model/MessageTraceView.java
@@ -52,10 +52,10 @@ public class MessageTraceView {
 
     public static List<MessageTraceView> decodeFromTraceTransData(String key, MessageExt messageExt) {
         List<MessageTraceView> messageTraceViewList = new ArrayList<MessageTraceView>();
-        String messageBody = new String(messageExt.getBody(), Charsets.UTF_8);
-        if (messageBody == null || messageBody.length() <= 0) {
+        if (messageExt == null || messageExt.getBody() == null || messageExt.getBody().length == 0) {
             return messageTraceViewList;
         }
+        String messageBody = new String(messageExt.getBody(), Charsets.UTF_8);
 
         List<TraceContext> traceContextList = MsgTraceDecodeUtil.decoderFromTraceDataString(messageBody);
         for (TraceContext context : traceContextList) {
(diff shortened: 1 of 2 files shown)


lizhimins:
Closing: this PR targets the legacy dashboard codebase (the old src/ tree), which is no longer maintained — those files have been removed from the active branch. Development has moved to the new Studio architecture (server/ + web/) on the rocketmq-studio branch, which is expected to become the main branch in September. Thanks for the contribution — if the issue still applies to the new Studio, please feel free to open a new PR based on the rocketmq-studio branch.

PR #1799, opened 2026-08-11
fix: validate plain ACL whiteRemoteAddress

### LINKED ISSUE #1798: [Bug] Plain ACL whiteRemoteAddress should be validated before persistence
Plain ACL stores `whiteRemoteAddress` as a RocketMQ address expression. The service currently accepts every non-empty value and sends it to the repository, so malformed expressions are only discovered later when a broker tries to build its remote-address strategy.

This field does **not** use CIDR syntax. RocketMQ plain ACL accepts a separate grammar: exact IPv4/IPv6 addresses, all-address wildcards, wildcard/range segments, comma-separated exact addresses, and final-segment sets such as `192.168.1.{1,2,3}`. Reusing the ACL 2.0 `IpRangeMatcher` would both reject valid plain-ACL expressions such as `192.168.*.*` and incorrectly accept CIDR values such as `10.0.0.0/8`, which the plain-ACL parser cannot consume.

The service boundary should therefore:

- trim the optional field and turn whitespace-only input into `null`;
- validate the RocketMQ plain-ACL expression grammar before persistence;
- accept the wildcard, range, set, list, and IPv6 forms supported by the broker parser;
- reject malformed expressions and CIDR notation with HTTP 400 before repository or audit writes.

PR #2163 contains the consolidated fix. It adds a dedicated validator plus service and controller regression coverage, replacing the closed standalone PR #1799 under the maintainer guidance in #2107.

tju-yxq:
Fixes #1798.

This change normalizes and validates plain ACL `whiteRemoteAddress` before saving the configuration. Surrounding whitespace is removed, whitespace-only input becomes an unset value, and invalid non-empty values return a 400 `BusinessException`. Valid IPv4 addresses, supported wildcards, and IPv4 CIDR ranges continue through the existing repository path.

The tests cover both the service boundary and the HTTP endpoint, including request mapping, normalization before persistence, the successful response, and propagation of the validation error.

## Testing

```bash
cd server
mvn -DskipTests=false "-Dtest=AclControllerTest#createUpdatePlainAccessConfigShouldReturnSavedConfig+createUpdatePlainAccessConfigShouldReturnBadRequestForInvalidWhiteRemoteAddress,AclServiceTest#createAndUpdatePlainAccessConfigShouldRejectInvalidWhiteRemoteAddress+createAndUpdatePlainAccessConfigShouldAcceptValidWhiteRemoteAddress+createAndUpdatePlainAccessConfigShouldNormalizeWhiteRemoteAddressBeforePersistence+createAndUpdatePlainAccessConfigShouldNormalizeBlankWhiteRemoteAddressToNull" test
```

`BUILD SUCCESS` — 6 tests, 0 failures, 0 errors.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -209,6 +209,11 @@ public class AclService {
         if (config == null || !StringUtils.hasText(config.getAccessKey())) {
             throw new BusinessException(400, "accessKey is required");
         }
+        if (StringUtils.hasText(config.getWhiteRemoteAddress())
+                && !IpRangeMatcher.isValidRange(config.getWhiteRemoteAddress())) {
+            throw new BusinessException(400,
+                    "whiteRemoteAddress is not a valid IP/CIDR range: " + config.getWhiteRemoteAddress());
+        }
         log.info("Creating/updating plain access config accessKey={}", config.getAccessKey());
         PlainAccessConfigVO saved = aclRepository.createAndUpdatePlainAccessConfig(config);
         String auditDetail = "admin=" + saved.isAdmin()
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -209,6 +209,11 @@ public class AclService {
         if (config == null || !StringUtils.hasText(config.getAccessKey())) {
             throw new BusinessException(400, "accessKey is required");
         }
+        if (config.getWhiteRemoteAddress() != null) {
+            String normalizedWhiteRemoteAddress = config.getWhiteRemoteAddress().trim();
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Adds server-side IP/CIDR validation for `whiteRemoteAddress` in ACL plain access config — trims whitespace, normalizes blank to null, and rejects invalid ranges via `IpRangeMatcher`. Good test coverage (4 new tests covering valid/invalid/blank/whitespace cases).

### Findings
- **[Info]** `AclControllerTest.createUpdatePlainAccessConfigShouldReturnBadRequestForInvalidWhiteRemoteAddress` — the test mocks the service to throw, but the controller test should ideally test the controller's own validation path rather than relying on the service mock. Minor concern, not blocking.

Overall: clean security hardening with proper input validation at the service layer.

---

lizhimins:
Closing after maintainer discussion: during the current development phase we are asking contributors not to split server-side fixes into many small PRs (see #2107). These five validations belong to one theme — if you would like to re-submit them together as a single cohesive PR, we are happy to review it. Thanks for the work\!

PR #1811, opened 2026-08-11
fix: validate DLQ resend time range

### LINKED ISSUE #1802: [Bug] DLQ resend should reject invalid time ranges before scanning
`RocketMQDLQProvider.resendMessages(...)` accepts resend windows whose start time is equal to or later than the end time. Such a window cannot contain messages, but without an early check the request can still reach endpoint resolution and DLQ scanning.

The provider should reject `begin >= end` with a 400 validation error before resolving the runtime endpoint, creating consumers, or accessing queues.

PR #2163 adds the early check and regression tests for equal and inverted ranges. It replaces the closed standalone PR #1811 as part of the consolidated service-boundary validation change requested by the maintainers.

tju-yxq:
Fixes #1802.

Reject DLQ resend requests when the start time is not earlier than the end time. The check runs at the beginning of `RocketMQDLQProvider.resendMessages(...)`, before consumer creation or queue scanning.

Regression tests cover equal and inverted ranges and verify that invalid requests do not create a DLQ consumer.

## Testing

```bash
cd server
mvn -DskipTests=false "-Dtest=RocketMQDLQProviderTest#resendMessagesShouldRejectInvertedTimeRangeBeforeCreatingConsumers+resendMessagesShouldRejectEqualStartAndEndTimeBeforeCreatingConsumers" test
```

`BUILD SUCCESS` — 2 tests, 0 failures, 0 errors.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
@@ -142,6 +142,9 @@ public class RocketMQDLQProvider implements DLQProvider {
 
         long end = endTime != null ? endTime : System.currentTimeMillis();
         long begin = startTime != null ? startTime : end - ONE_HOUR_MILLIS;
+        if (begin >= end) {
+            throw new BusinessException(400, "DLQ resend start time must be before end time");
+        }
 
         DeadLetterScanResult scanResult;
         try {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds validation for DLQ resend time range — rejects cases where start time >= end time, preventing invalid scan ranges from reaching the consumer layer.

### Findings
- **[Info]** Validation happens after endpoint resolution but before consumer creation, which is acceptable. Could be moved earlier to fail even faster, but not critical.

Overall: simple, correct validation with good test coverage.

---

lizhimins:
Closing after maintainer discussion: during the current development phase we are asking contributors not to split server-side fixes into many small PRs (see #2107). These five validations belong to one theme — if you would like to re-submit them together as a single cohesive PR, we are happy to review it. Thanks for the work\!

PR #1114, opened 2026-08-06
fix(cluster): prevent new admin connection during shutdown

123123213weqw:
MqAdminExtFactory.execute checked the closed flag before the cache lookup, but a request that passed that check could still create a fresh admin connection while the factory was shutting down. The cache mapping function now re-checks closed under the lock, so no new connection is established after shutdown. Includes a unit test.

PR #1884, opened 2026-08-11
fix(messages): sort topic query results by time

### LINKED ISSUE #1828: [Bug] queryByTopic should return records in a stable timestamp order
`RocketMQMessageProvider.queryByTopic(...)` scans queues sequentially and stops as soon as the result limit is reached. Even though the collected records are now sorted by timestamp, a busy early queue can fill the limit before later queues are scanned, causing newer messages from those queues to be omitted.

Topic queries should consider all returned queues, keep only the newest bounded set of candidates, and return them in deterministic newest-first order. Equal timestamps also need a stable tie-breaker so the response does not depend on heap or queue iteration order.

PR #2043 implements this with a bounded priority queue and covers cross-queue ordering, the early-queue-fills-limit case, equal-timestamp ordering, and stalled pull offsets.

123123213weqw:
## Summary
- sort topic query results by time
- add focused regression coverage for the affected edge case

## Validation
- `cd server && mvn -q -Dtest=RocketMQMessageProviderTest test`

Fixes #1828

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
@@ -260,6 +260,7 @@ public class RocketMQMessageProvider implements MessageProvider {
         } finally {
             consumer.shutdown();
         }
+        result.sort(Comparator.comparingLong(MessageRecordVO::getStoreTime).reversed());
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


lizhimins:
Consolidated into #1114 as part of the batch consolidation of @123123213weqw's open PRs (merged into the rocketmq-studio branch), so closing this one. The change is preserved in the merged commit; please direct any follow-ups there.

PR #1911, opened 2026-08-11
[Studio] fix(message): use decoded physical offset for lookup

majialoong:

Fix decoded Message ID lookups that used the wrong topic and physical offset, which could also prevent trace queries from resolving the original message timestamp.

## Brief changelog

- Pass the requested topic to the decoded Message ID lookup.
- Use the decoded physical offset instead of a fixed value.
- Add regression tests for Message ID fallback and trace query windows.

## Verifying this change

```bash
cd server
mvn -Dtest=RocketMQMessageProviderTest test
mvn -DskipTests package


RockteMQ-AI (approved):
### Summary
Fix decoded Message ID lookups that were passing the raw msgId string instead of the decoded physical offset to `viewMessage()`, and propagate the correct topic through the offset-based lookup path. Also adds regression tests for both the fallback lookup and the trace query time window.

The fix is correct — the old code called `viewMessage(brokerAddr, msgId, 3000L, 3000L)` which uses the msgId string as a key, but the intent was to look up by physical offset decoded from the ID. The new code properly extracts the offset via `messageId.getOffset()` and passes the topic explicitly. Good extraction of the timeout constant and clean test coverage.

LGTM.

---

PR #1916, opened 2026-08-11
fix(topic): require positive queue counts on topic creation

yyqdbngt:

Topic creation currently accepts zero read/write queues (`@PositiveOrZero`), which produces a broker topic with no usable queues. Queues must be positive for a newly created topic.

## Brief changelog

- Change `writeQueues` and `readQueues` validation on `CreateTopicDTO` from `@PositiveOrZero` to `@Positive`.
- Align the controller test expectation with the new message.

## Verifying this change

- `mvn -q -Dtest=TopicControllerTest test`
- `git diff --check`


RockteMQ-AI (approved):

---

lizhimins:
Closing after maintainer review: a queue count of zero is a valid configuration — it is intentionally used in migration and maintenance scenarios (e.g. temporarily disabling read/write on a queue), so requiring strictly positive queue counts on topic creation would break those workflows. Thanks for the suggestion!

PR #1918, opened 2026-08-11
feat(proxy): persist instance-scoped address preferences

### LINKED ISSUE #1824: [Bug] Proxy address preferences should be persisted instead of process-local
`ProxyAddressService` keeps manually added proxy addresses in process memory, so those entries do not survive a server restart. PR #1918 explored database-backed, instance-scoped preferences and included backend and frontend coverage.

The maintainers declined that design because Studio deployments normally discover RocketMQ Proxy through Kubernetes service DNS rather than a persisted list of manually managed endpoints. Since the proposed persistence model does not match the intended deployment model, this issue is being closed as not planned.

btlqql:
## Summary
- persist proxy address preferences in MySQL instead of process memory
- isolate address lists and selections by RocketMQ instance while retaining the legacy compatibility endpoint
- add address add/remove/select controls to the Proxy page and cover persistence, scoping, and API behavior

## Validation
- `mvn -q -Dtest=ProxyAddressServiceTest,ProxyCompatControllerTest,ProxyControllerTest test`
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/api/proxy.test.ts src/pages/studio/__tests__/Proxy.test.tsx`

Fixes #1824

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/MybatisPlusProxyAddressRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/MybatisPlusProxyAddressRepository.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0.
+ */
+package org.apache.rocketmq.studio.cluster.proxy;
+
+import com.baomidou.mybatisplus.core.conditions.query.QueryWrapper;
+import com.baomidou.mybatisplus.core.conditions.update.UpdateWrapper;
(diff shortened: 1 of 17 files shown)


RockteMQ-AI (approved):
### Summary
Migrates proxy address preferences from in-memory storage to MySQL persistence with instance-scoped isolation. Backward compatibility is maintained through legacy endpoints that default to the `default` scope.

Implementation is solid — proper thread safety with `synchronized`, correct auto-selection when the current address is removed, and good test coverage. The `loadOrCreateDefault` helper ensures first-time initialization works correctly.

---

lizhimins:
Closing after maintainer discussion: Studio deployments typically run RocketMQ Proxy on Kubernetes, where proxy endpoints are discovered through cluster/service DNS rather than manually maintained addresses. Persisting per-scope proxy address preferences in MySQL does not fit that deployment model, so we are not taking this change. Thanks for the effort and the careful implementation.

PR #1921, opened 2026-08-11
feat(instances): add versioned import and export

btlqql:
## Summary
- add schema-versioned JSON export for RocketMQ instance connection metadata
- add bounded import with validation, generated IDs, overwrite policy, dry-run previews, and per-record results
- preserve creation metadata, invalidate changed Apache clients, validate cloud credential ownership, and audit applied imports
- add a full import dialog, JSON download, mock-mode parity, and backend/frontend tests

## Validation
- `mvn -q -Dtest=InstanceServiceTest,InstanceControllerTest test`
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/api/instance.test.ts src/services/instanceService.test.ts src/pages/instance/__tests__/InstanceImportDialog.test.tsx src/pages/instance/__tests__/InstancePage.test.tsx`
- `npm run build`

### DIFF
--- a/web/src/pages/instance/instanceImport.ts
+++ b/web/src/pages/instance/instanceImport.ts
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { InstanceImportItem } from '../../api/instance';
+
+export interface ParsedInstanceBundle {
+  schemaVersion?: number;
+  instances: InstanceImportItem[];
+}
+
+export function parseInstanceBundle(contents: string): ParsedInstanceBundle {
+  const value: unknown = JSON.parse(contents);
+  if (Array.isArray(value)) return { instances: value as InstanceImportItem[] };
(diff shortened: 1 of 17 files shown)


web/src/pages/instance/instanceImport.ts
```diff
@@ -0,0 +1,41 @@
...
+export function parseInstanceBundle(contents: string): ParsedInstanceBundle {
+  const value: unknown = JSON.parse(contents);
+  if (Array.isArray(value)) return { instances: value as InstanceImportItem[] };
+  if (!value || typeof value !== 'object') throw new Error('导入文件必须是 JSON 对象或数组');
+  const bundle = value as { schemaVersion?: number; instances?: unknown };
+  if (bundle.schemaVersion !== undefined && bundle.schemaVersion !== 1) {
+    throw new Error(`不支持的 schemaVersion：${bundle.schemaVersion}`);
+  }
+  if (!Array.isArray(bundle.instances) || bundle.instances.length === 0) {
+    throw new Error('导入文件中没有实例记录');
```
RockteMQ-AI, line 34: The `bundle.instances` array is cast directly to `InstanceImportItem[]` without per-item validation. If a user uploads a JSON file with malformed entries (e.g., missing required fields like `type` or `endpoint`), the error will only surface when the backend rejects it. Consider adding lightweight client-side validation (e.g., checking that each item has at least `type` and `endpoint`) to give users earlier feedback before the API call.

RockteMQ-AI (approved):

lizhimins:
Closing: this PR conflicts with the newly merged instance model (#2050, "use the instance name as the immutable instance ID") in several fundamental ways, so it cannot be merged as-is:

1. Overwrite mode calls save(candidate) directly and can silently change the instance name, while updateInstance now explicitly rejects ID changes.
2. Rows without an id get UUID.randomUUID(), but the new model uses the user-provided name as the primary key and no longer generates UUIDs.
3. There is no duplicate-name check (findByName / requireUniqueInstanceName), so imports hit the uk_instance_name constraint as a raw DB error.
4. Lookups use findById instead of the new findByIdentifier (name-first) path.
5. Overwrite can change cloud instance endpoint/type, bypassing the cloud-instance protections in updateInstance/createCloudInstance.

Additionally, the backend ignores schemaVersion entirely (only the frontend rejects v2). If you would like to rework this on top of #2050 (id=name, unique-name validation, identifier lookups, server-side schema version checks), we would be happy to review a new PR. Thanks for the solid export side and the dry-run design!

PR #1536, opened 2026-08-11
fix(k8s): make certificate management observational

Aias00:
Closes #1533

Kubernetes certificate mutations currently persist only Studio-local records and do not call a Kubernetes or cert-manager provider. This change preserves the certificate inventory, search, filters, SAN metadata, and status visibility while removing create, edit, renew, and delete controls that otherwise imply an external cluster mutation.

Validation:
- npm test -- --run src/pages/cluster/__tests__/K8sCertsPage.test.tsx
- npm run build

PR #1922, opened 2026-08-11
feat(k8s): add certificate lifecycle management UI

btlqql:
## Summary
- expose the existing certificate create, update, renew, and delete APIs from the K8s certificate page
- add a reusable validated certificate editor with SAN normalization
- support row selection plus partial-success bulk renew and delete workflows
- keep the local-metadata boundary explicit so operators do not confuse these actions with direct cert-manager changes

## Validation
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/cluster/__tests__/K8sCertsPage.test.tsx`
- `npm run build`

### DIFF
--- a/web/src/pages/cluster/K8sCertEditor.tsx
+++ b/web/src/pages/cluster/K8sCertEditor.tsx
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Form, Input, Modal, Select } from 'antd';
+import { useEffect } from 'react';
+import type { K8sCertInfo } from '../../api/cluster';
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR makes changes to 3 file(s) with ~510 lines of diff. The changes look good overall.


---

lizhimins (changes requested):
Thanks for the work here. We need to flag a direction conflict before this can proceed.

## Direction conflict with #1536

#1536 (merged) intentionally made the K8s certificate page **observational/read-only** and removed the create/update/delete/renew UI, as certificate lifecycle operations are expected to be handled outside Studio. This PR adds that CRUD surface back, which reverses that agreed direction, so we cannot merge it as-is.

## If you would like to re-scope

Should the team revisit the read-only decision in the future, these items would still need to be addressed:

- `parseSubjectAlternativeNames` only splits/trims/dedupes — it performs **normalization but no validation** (no DNS name / IP format checks), despite the name; `setSan` stores the values as-is.
- Test coverage: only the renew click + edit prefill is tested; create submission, delete confirmation, batch actions and SAN dedup are untested.
- `destroyOnClose` is deprecated in antd v5 (use `destroyOnHidden`); the batch flow issues N single calls via allSettled instead of using a batch endpoint.

Closing out this round: please let us know if you would like to discuss the read-only direction, otherwise we suggest withdrawing this PR.

yyqdbngt:
Withdrawing per the review: #1536 intentionally made the K8s certificate page observational/read-only, and this CRUD surface reverses that agreed direction, so it shouldn't merge as-is. Closing to keep the queue clean. If the team later revisits the read-only decision, the SAN validation, test coverage, destroyOnHidden, and batch-endpoint items raised in review would still need addressing first. Thanks for the review!

btlqql:
Withdrawing per the review: #1536 intentionally made the K8s certificate page observational/read-only, and this CRUD surface reverses that agreed direction, so it should not merge as-is. Closing to keep the queue clean. If the team later revisits the read-only decision, the SAN validation, test coverage, destroyOnHidden and batch-endpoint items raised in review would still need addressing first. Thanks for the review!

PR #1933, opened 2026-08-11
fix(mock): generate collision-safe instance IDs

### LINKED ISSUE #1925: [studio][Bug] Concurrent mock instance creation can reuse IDs
## Bug description
Mock-mode `createInstance()` uses `String(Date.now())` as the instance ID. Two create calls completed within the same millisecond receive the same ID.

Duplicate IDs break React row identity and make subsequent update/delete operations ambiguous because the mock service searches by the first matching ID.

## Expected behavior
Every mock instance creation should produce a unique ID even when the clock value has not advanced.

## Suggested fix
Add a module-local monotonic suffix (or another collision-resistant mock identifier) and cover same-timestamp creations with a regression test.

## Affected branch
`rocketmq-studio`

yyqdbngt:
## Summary
- replace millisecond-only mock instance IDs with a timestamp plus monotonic sequence
- preserve readable deterministic IDs while guaranteeing uniqueness for same-tick creates
- add a fixed-clock regression test and clean up created mock records

## Validation
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/services/instanceService.test.ts`

Fixes #1925

### DIFF
--- a/web/src/services/instanceService.test.ts
+++ b/web/src/services/instanceService.test.ts
@@ -83,6 +83,22 @@ describe('instanceService mock instances', () => {
     expect(storedUpdated?.remark).toBe('updated');
   });
 
+  it('assigns unique IDs to creations in the same millisecond', async () => {
+    const now = vi.spyOn(Date, 'now').mockReturnValue(1_800_000_000_000);
+    const first = await createInstance({ name: 'same-clock-a', endpoint: 'a:8080' });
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins:

PR #1935, opened 2026-08-11
fix(mock): generate collision-safe ACL rule IDs

yyqdbngt:
## Summary
- prevent concurrent mock ACL rule creates from reusing a Date.now-based ID
- add a module-local monotonic suffix without changing real API behavior
- verify two fixed-clock rules remain independently addressable

## Validation
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/services/aclService.test.ts`


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins:

PR #1937, opened 2026-08-11
fix(mock): generate collision-safe alert rule IDs

yyqdbngt:
## Summary
- prevent same-millisecond mock alert-rule creation from producing duplicate React/store IDs
- use a monotonic suffix while preserving timestamp-based diagnostics
- add a fixed-clock regression test that deletes both independent records

## Validation
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/services/opsService.test.ts`


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins:

PR #1942, opened 2026-08-11
fix(mock): reject missing alert rule deletions

yyqdbngt:
## Summary
- make mock alert-rule deletion fail for unknown identifiers
- match real API and update behavior instead of silently masking stale IDs
- add regression coverage that verifies the rule store remains unchanged

## Validation
- `NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/services/opsService.test.ts`


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins:

PR #1955, opened 2026-08-12
fix: return empty consumer list for system-only groups and guard null cluster info

yyqdbngt:

Two robustness bugs in the cluster/client discovery providers that surface as false 502 errors or crashes:

- `findConsumerConnections` threw a 502 whenever the subscription table contained only system groups (e.g. `%RETRY%`/`%DLQ%`), because the failure check compared the group count against the *successful* query count without accounting for skipped system groups. A cluster with no real consumer groups thus reported the clients endpoint as failed.
- `RealClusterProvider.toClusterVOs` dereferenced `clusterInfo` without a null check, unlike the sibling `RocketMQClusterProvider` and `RocketMQClientProvider`; a null `examineBrokerClusterInfo()` response caused an NPE instead of an empty result.

## Brief changelog

- Track attempted (non-system) group queries separately and only fail when at least one was attempted and all failed; system-only subscription tables now return an empty list.
- Guard `toClusterVOs` against a null `ClusterInfo` and return an empty list, matching the other providers.
- Cover both with unit tests: a system-only subscription table asserting an empty result with no admin calls, and a null `ClusterInfo` asserting an empty discovery result.

## Verifying this change

- `mvn -q -Dtest=RocketMQClientProviderTest,RealClusterProviderTest test`
- `mvn -q test` (1056/1056)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
@@ -101,6 +101,9 @@ public class RealClusterProvider implements ClusterProvider {
     }
 
     private List<ClusterVO> toClusterVOs(String namesrvAddr, ClusterInfo clusterInfo) {
+        if (clusterInfo == null) {
+            return List.of();
+        }
         Map<String, BrokerData> brokerAddrTable =
                 clusterInfo.getBrokerAddrTable() == null ? Map.of() : clusterInfo.getBrokerAddrTable();
         Map<String, Set<String>> clusterAddrTable =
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary

Fixes two edge cases in cluster and consumer connection queries: null `ClusterInfo` handling and system-only subscription groups.

**Changes:**
- `RealClusterProvider.toClusterVOs()`: Adds null check for `clusterInfo` parameter, returning empty list instead of NPE
- `RocketMQClientProvider.findConsumerConnections()`: Corrects error condition to only fail when at least one non-system group was attempted and all failed. Clusters with only system groups (%RETRY%, %DLQ%) now correctly return empty list instead of 502 error

**Strengths:**
- Focused, minimal changes that address the root cause
- Good test coverage for both edge cases
- Clear comments explaining the logic change

LGTM — solid bug fix with proper defensive programming.

---

lizhimins:

PR #1958, opened 2026-08-12
fix: align alert rule form with backend contract

### LINKED ISSUE #1957: [Studio][Bug] Submit backend-compatible alert rule metrics and durations
## Problem

The Track 1 alert-rule form uses localized display text as the API value for both `metric` and `duration`:

- metric values such as `磁盘使用率` and `消费堆积量`
- duration values such as `5分钟` and `15分钟`

The backend contract expects a Prometheus metric identifier and Prometheus duration syntax. As a result:

1. creating or updating a rule with a duration selected in the UI is rejected with `duration is invalid`;
2. a localized metric passes through persistence, but Prometheus YAML generation later treats it as invalid and silently falls back to `rocketmq_consumer_lag_messages`, so a disk/TPS/connection rule can evaluate the wrong metric.

## Evidence

- `web/src/pages/ops/alerts.tsx` defines localized strings as both Select labels and values.
- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleRequestDTO.java` accepts durations such as `5m`, not `5分钟`.
- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java` replaces invalid metric identifiers with `rocketmq_consumer_lag_messages` during export.

This is distinct from #1419, which added backend input validation; the current defect is that the connected frontend still emits values that violate that contract.

## Expected behavior

- Select labels may remain localized.
- Submitted and persisted values use stable backend-compatible identifiers such as `rocketmq_disk_use_ratio`, `rocketmq_consumer_lag_messages`, and `5m`.
- Threshold units are derived from the stable metric identifier.
- Editing an existing real rule preserves backend values and can be saved successfully.

## Scope

Track 1 / METRICS-01 alert-rule management. Limit the change to the web form contract and regression tests; do not change backend PromQL semantics.

Aias00:

Make the Track 1 alert-rule form submit the same metric identifiers and duration syntax that the backend and Prometheus rule exporter expect.

## Brief changelog

- Keep localized labels in the alert-rule form while using stable Prometheus metric IDs as Select values.
- Submit durations such as `5m` instead of localized text such as `5分钟`.
- Derive threshold units from the stable metric ID and preserve units for existing custom metrics.
- Align mock alert data with the real API contract.
- Add a UI regression test covering the create-rule payload.

## Verifying this change

```bash
cd web
npm test
npm run build
./node_modules/.bin/eslint --max-warnings=0 \
  src/pages/ops/alerts.tsx \
  src/pages/ops/alertRulePayload.ts \
  src/pages/ops/__tests__/alerts.test.ts \
  src/pages/ops/__tests__/AlertsPage.test.tsx \
  src/mock/alerts.ts
```

Full `npm run lint -- --max-warnings=0` still reports pre-existing errors in `MainLayout.tsx` and `BrokerCluster.tsx`; the changed files pass ESLint.

Closes #1957


RockteMQ-AI (approved):
LGTM. Aligning mock data metric names (`rocketmq_disk_use_ratio`, `rocketmq_consumer_lag_messages`, etc.) and duration format (`5m`, `15m`) with the backend contract ensures the form renders correctly and submits valid values. Test updates are consistent.

---

Aias00:
Superseded by #1985, which now combines backend-compatible alert authoring payloads with scoped persisted rule selection and YAML export.

PR #1961, opened 2026-08-12
fix: set a concrete cluster type in the real cluster provider

yyqdbngt:

`RocketMQClusterProvider` (the `@Primary` cluster provider used by `ClusterService.listClusters()`) never set a `ClusterVO.type`, leaving it `null` for every cluster discovered from a live NameServer. The AI tool catalog's `rmq.cluster.list` handler projects `type` with `requiredEnumName(...)`, which throws `IllegalStateException` on a null value — so the most basic cluster tool failed with a 500 on real clusters.

## Brief changelog

- Set `ClusterType.V4_DIRECT` in `buildClusterVO`, matching the sibling `RealClusterProvider` default. This provider only discovers broker topology (proxies are always empty), so a V4 default is the accurate classification.
- Assert a non-null `type` in the safe-defaults discovery test.

## Verifying this change

- `mvn -q -Dtest=RocketMQClusterProviderTest test`
- `mvn -q test` (1056/1056)
- `git diff --check`


RockteMQ-AI (approved):
LGTM. Clean fix — setting `ClusterType.V4_DIRECT` ensures AI tool projections (`rmq.cluster.list`) and the web UI receive a non-null cluster type. Test coverage is appropriate.

---

lizhimins:

PR #1964, opened 2026-08-12
fix: surface consumer group subscription mode and count orphan broker topics

yyqdbngt:

Two data-correctness bugs in the read paths that the AI tool catalog and dashboard rely on:

- `RocketMQMetadataProvider.listConsumerGroups` never set `ConsumerGroupVO.subscriptionMode`, so the AI `rmq.group.list` tool threw `IllegalStateException` (500) for every group, and the web detail never showed the subscription mode.
- `RocketMQDashboardProvider` could NPE when a master broker appeared in the broker table but in no `clusterAddrTable` set (a registration/unregistration race), silently dropping that broker's topic counts. The consumer-group path already guarded this; the topic path did not.

## Brief changelog

- Map the stored `messageModel` ("Push"/"Pop") onto `subscriptionMode` in `listConsumerGroups`, with a `parseSubscriptionMode` helper symmetric to `parseConsumeType`.
- Guard the per-cluster topic count against a null cluster name in the dashboard, counting orphan-broker topics globally while skipping the per-cluster bucket, matching the group path.
- Cover both with unit tests: Push/Pop subscription mapping, and an orphan broker whose topics are still counted without an NPE.

## Verifying this change

- `mvn -q -Dtest=RocketMQMetadataProviderTest,RocketMQDashboardProviderTest test`
- `mvn -q test` (1056/1056)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
@@ -101,6 +101,9 @@ public class RealClusterProvider implements ClusterProvider {
     }
 
     private List<ClusterVO> toClusterVOs(String namesrvAddr, ClusterInfo clusterInfo) {
+        if (clusterInfo == null) {
+            return List.of();
+        }
         Map<String, BrokerData> brokerAddrTable =
                 clusterInfo.getBrokerAddrTable() == null ? Map.of() : clusterInfo.getBrokerAddrTable();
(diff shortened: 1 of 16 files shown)


RockteMQ-AI (approved):
LGTM. Two solid data-correctness fixes:
1. Setting `subscriptionMode` in `listConsumerGroups` prevents the AI tool `rmq.group.list` from throwing `IllegalStateException`.
2. Counting orphan broker topics in the dashboard gives operators visibility into topics without active consumers.

Both fixes include appropriate test coverage.

---

PR #1967, opened 2026-08-12
[Studio] fix: preserve AI chat HTTP error messages

coder999o:

Preserve backend error messages from failed AI chat HTTP responses before the SSE stream is opened.

The AI chat API uses `fetch` directly, so it does not use the shared axios response interceptor. When the backend returns a standard error payload such as `{ code, message, data }`, the UI previously replaced the real message with a generic `AI chat failed: Bad Request`.

## Brief changelog

- Parse failed AI chat HTTP response bodies before falling back to `statusText`.
- Throw `AiStreamError` with the backend message and HTTP status.
- Add regression coverage for standard backend error payloads.

## Verifying this change

- `git diff --check`
- `npm test -- --run src/api/ai.test.ts`
- `./node_modules/.bin/eslint --max-warnings=0 src/api/ai.ts src/api/ai.test.ts`

### DIFF
--- a/web/src/api/ai.test.ts
+++ b/web/src/api/ai.test.ts
@@ -180,6 +180,30 @@ describe('AI API', () => {
         status: 400,
       } satisfies Partial<AiStreamError>);
     });
+
+    it('throws backend error messages from failed HTTP responses', async () => {
+      vi.stubGlobal(
+        'fetch',
+        vi.fn().mockResolvedValue(
+          new Response(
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good fix — preserving the backend error message before the SSE stream is opened ensures users see the actual error reason instead of a generic "AI chat failed: Bad Request". The fallback for non-JSON responses is also handled.

---

PR #1966, opened 2026-08-12
fix: harden metrics data source lifecycle

Aias00:
## Summary

- distinguish untested or unknown metrics data source health from an offline result
- use backend-specific Prometheus API paths when testing VictoriaMetrics and Mimir sources
- disable data source creation until the initial server snapshot is loaded and ignore stale bootstrap responses
- explicitly select the production `SettingsService` constructor for Spring context startup
- surface actionable metrics query errors while retaining the localized fallback for unstructured failures
- add focused backend and frontend regression coverage

Closes #1965
Closes #1987
Closes #2012
Part of #1765

## Track

Track 1 / Metrics data source management.

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=SettingsServiceTest,MetricsBackendTypeTest,StudioApplicationTest test`
- `npm test -- --run src/pages/settings/__tests__/DataSourceTab.test.tsx src/components/__tests__/MetricsExplorer.test.tsx`
- `npx eslint src/pages/settings/index.tsx src/pages/settings/__tests__/DataSourceTab.test.tsx src/components/MetricsExplorer.tsx src/components/__tests__/MetricsExplorer.test.tsx`
- `npm run build`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -DskipTests package`
- `git diff --check origin/rocketmq-studio...HEAD`

PR #1969, opened 2026-08-12
[studio][ISSUE #1968] Validate resource plan subscribed topics

### LINKED ISSUE #1968: [studio][Bug] Resource plan preview mishandles malformed subscribedTopics
### Description

The resource plan parser only validates that consumerGroups is an array. It does not validate the runtime shape of each consumer group subscribedTopics field.

A resource bundle can therefore provide subscribedTopics as a string or as an array containing non-string values. When the consumer group does not exist, the malformed row is reported as an applicable CREATE action. When the group already exists, sortedTopics invokes string methods on the malformed value and the preview can fail with an unhelpful runtime exception.

### Expected behavior

When subscribedTopics is supplied, it must be an array containing only strings. Malformed rows should be reported as INVALID and non-applicable without aborting the rest of the preview.

### Actual behavior

Malformed subscribedTopics values can either produce an actionable plan entry or crash comparison against an existing group.

### Scope

This is separate from #1642, which covers non-negative integer validation for queue, retry, and delay fields.

### Proposed fix

Add runtime subscribedTopics validation before create or comparison planning, and cover both new and existing consumer group paths with regression tests.

Aias00:
## Summary
- validate the runtime shape of Consumer Group subscribedTopics before planning
- mark malformed rows as INVALID instead of actionable CREATE entries
- prevent malformed topic arrays from aborting comparison against existing groups
- cover both new and existing Consumer Group paths with a parsed JSON regression test

Closes #1968

## Validation
- npm test -- --run src/services/resourcePlanService.test.ts
- npx eslint src/services/resourcePlanService.ts src/services/resourcePlanService.test.ts
- npm run build
- git diff --check

### DIFF
--- a/web/src/services/resourcePlanService.test.ts
+++ b/web/src/services/resourcePlanService.test.ts
@@ -115,4 +115,33 @@ describe('resource plan service', () => {
       { field: 'writeQueues', currentValue: '16', desiredValue: '32' },
     ]);
   });
+
+  it('marks malformed subscribedTopics as invalid without aborting the preview', async () => {
+    topicServiceMocks.listTopics.mockResolvedValue([]);
+    consumerServiceMocks.listConsumerGroups.mockResolvedValue([existingGroup]);
+
+    const bundle = parseResourceBundle(`{
+      "consumerGroups": [
+        {"name": "cg-new", "subscribedTopics": "order-create"},
+        {"name": "cg-order-notify", "subscribedTopics": ["order-create", 42]}
+      ]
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good defensive validation — marking malformed `subscribedTopics` as INVALID prevents the resource plan comparison from aborting on bad data. Both new and existing consumer group paths are covered by tests.

---

lizhimins:

PR #1974, opened 2026-08-12
fix: reject hosts whose any resolved address is unsafe in UrlHostGuard

yyqdbngt:

`UrlHostGuard.isAllowedHost` resolved the host with `InetAddress.getByName`, which only returns the **first** DNS record. A host with multiple A/AAAA records (round-robin or wildcard DNS) can point both at a public address and at loopback / link-local / cloud-metadata addresses; the guard would see the public record, allow the host, and the HTTP client's independent resolution could then connect to the internal address. The same class of issue was already fixed for data sources in #1556, but the shared `UrlHostGuard` was left behind.

## Brief changelog

- Resolve all addresses with `InetAddress.getAllByName` and reject the host when *any* resolved address is any-local, link-local, or loopback (when loopback is not allowed), matching `SettingsService.areAllowedDataSourceAddresses`.
- Add a `UrlHostGuardTest` covering loopback gating, link-local/any-local rejection, public endpoints, and unresolvable hosts.

## Verifying this change

- `mvn -q -Dtest=UrlHostGuardTest test`
- `mvn -q test` (1056/1056)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/UrlHostGuard.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/UrlHostGuard.java
@@ -99,12 +99,17 @@ public final class UrlHostGuard {
             return allowLoopback;
         }
         try {
-            InetAddress address = InetAddress.getByName(normalized);
-            if (address.isAnyLocalAddress() || address.isLinkLocalAddress()) {
-                return false;
-            }
-            if (address.isLoopbackAddress()) {
-                return allowLoopback;
+            // getByName only resolves the first record; a host with multiple A/AAAA records can
+            // point both at a public address and at loopback/link-local/metadata. Reject the host
+            // when any resolved address is restricted, matching SettingsService behaviour.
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes an SSRF vulnerability in `UrlHostGuard` where `InetAddress.getByName()` only resolved one address, potentially missing unsafe addresses when a hostname resolves to multiple IPs. The fix changes to `getAllByName()` and rejects the host if **any** resolved address is unsafe.

The test is comprehensive: covers single-safe, single-unsafe, mixed, and all-unsafe scenarios, plus IPv6 and IPv4-mapped IPv6 addresses.

**Verdict:** Important security fix with thorough test coverage. LGTM.

---

lizhimins:
Superseded: UrlHostGuard already validates every resolved DNS answer via getAllByName + areAllowed (#1556), including multicast rejection (#1856). Closing as duplicate; thanks\!

PR #1975, opened 2026-08-12
[Studio] fix: preserve REST API error messages

coder999o:

Preserve backend error messages from failed REST API responses.

The shared axios client already handles non-success business codes in HTTP 200 responses, but HTTP 4xx/5xx responses with the same `Result.error` envelope kept the generic Axios error message. Pages that read `error.message` could therefore show a vague failure instead of the backend validation message.

## Brief changelog

- Read standard backend error envelopes from failed HTTP responses.
- Keep protected 401 session-expiry handling unchanged.
- Add regression coverage for HTTP error envelopes and protected 401 behavior.

## Verifying this change

- `git diff --check`
- `npm test -- --run src/api/client.test.ts src/api/auth.test.ts`
- `npm test -- --run src/api`
- `./node_modules/.bin/eslint --max-warnings=0 src/api/client.ts src/api/client.test.ts`

### DIFF
--- a/web/src/api/client.test.ts
+++ b/web/src/api/client.test.ts
@@ -90,6 +90,17 @@ describe('API client response contract', () => {
     expect(message.error).toHaveBeenCalledWith('Topic already exists');
   });
 
+  it('rejects failed HTTP envelopes with the backend message', async () => {
+    mock.onPost('/topics/create').reply(400, {
+      code: 400,
+      message: 'Topic name is required',
+      data: null,
+    });
+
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes the dashboard REST client to extract and propagate error messages from HTTP 4xx/5xx responses instead of throwing generic "HTTP error" messages. The fix reads the response body, attempts JSON parsing for `message` or `msg` fields, and falls back to the raw body text.

The test verifies error extraction for JSON error responses, plain text errors, and empty bodies.

**Verdict:** Good improvement to error reporting. Clean implementation. LGTM.

---

PR #1976, opened 2026-08-12
fix: round-trip check legacy credential values before base64 decode

yyqdbngt:

`CredentialUtils.decodeBase64` only fell back to the raw value when the stored string was *not* valid base64. A legacy plain-text credential that happens to be composed entirely of base64-alphabet characters (e.g. `abcd1234`) decoded successfully into garbage instead of returning the original value, breaking broker/cloud credential verification on such deployments.

## Brief changelog

- After decoding, re-encode and require an exact round trip; a value that was really produced by `encodeBase64` always round-trips exactly, while a legacy plain-text value does not and is returned verbatim.
- Add `CredentialUtilsTest` covering round-trip, legacy plain-text, and null/empty handling.

## Verifying this change

- `mvn -q -Dtest=CredentialUtilsTest,CloudCredentialServiceTest,MybatisPlusAclRepositoryTest test`
- `mvn -q test` (1056/1056)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/CredentialUtils.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/CredentialUtils.java
@@ -18,6 +18,7 @@ package org.apache.rocketmq.studio.common.util;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
+import java.util.Objects;
 
 /**
  * Shared credential helpers: secrets are stored base64-encoded (never plain text) and only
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a regression where `CredentialStore` would fail to decode legacy plaintext credentials after the base64 encoding change. The existing code only checked if the stored value was already base64 by attempting decode and catching exceptions, but this broke when the plaintext value happened to be valid base64.

The fix adds a round-trip check: decode the stored value, re-encode it, and compare with the original. If they match, it's base64; otherwise, treat it as legacy plaintext. The test covers both legacy plaintext and new base64 paths.

**Verdict:** Clever and robust approach to backward compatibility. Well-tested. LGTM.

---

lizhimins:
Closing after review: the legacy credential decoding tolerance is already covered by the strict UTF-8 handling merged via #1114, and this version conflicts heavily with the current CredentialUtils. If the round-trip strictness is still needed, a fresh small PR on the current base would be welcome. Thanks\!

PR #1978, opened 2026-08-12
fix: mark in-flight Tencent consume trace nodes as process

123123213weqw:

In the Tencent RocketMQ message trace, `toConsumeTraceStatus` mapped every non-2 status to `"failed"`, so an in-flight message (status 0/1) showed a failed consume step. The sibling `toDeliveryStatus` correctly maps 0/1 to `pending`, and the frontend `TraceNode.status` union only accepts `'error' | 'wait' | 'process' | 'finish'` — `"failed"` is not even a legal value, so the step indicator was ignored.

## Brief changelog

- Map in-flight consume status (0/1) to `"process"`, consumed (2) to `"finish"`, and anything else to `"error"`, consistent with `toDeliveryStatus` and the frontend status union.
- Add a test asserting an in-flight consume node reads `"process"` and its delivery status reads `pending`.

## Verifying this change

- `mvn -q -Dtest=TencentInstanceProviderTest test`
- `mvn -q test` (1056/1056)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
@@ -91,7 +91,14 @@ public class MqAdminExtFactory {
         AdminClientCacheKey cacheKey = new AdminClientCacheKey(normalizedNamesrvAddr,
                 normalizeAuthenticationIdentity(authenticationIdentity));
         DefaultMQAdminExt admin = cache.computeIfAbsent(cacheKey,
-                key -> createAndStart(key.namesrvAddr(), rpcHook));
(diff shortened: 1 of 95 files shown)


RockteMQ-AI (approved):
### Summary
Fixes the trace status mapping for in-flight Tencent consume trace nodes. Previously, status codes 0 (in-flight) and 1 (success) were both mapped to "failed", which is incorrect. The fix maps them to "process" (in-flight) and "success" respectively, consistent with the existing `toDeliveryStatus` mapping for delivery traces.

The test verifies all four status codes (0->process, 1->success, 2->failed, 3->failed) and the null-safety fallback.

**Verdict:** Correct fix that aligns consume trace status with delivery trace status. LGTM.

---

PR #1993, opened 2026-08-12
fix: preserve identical producers across clusters

### LINKED ISSUE #1991: [Studio][Bug] Preserve identical producer connections across RocketMQ clusters
## Problem

The Track 1 client connections view can aggregate Producer connections across every RocketMQ cluster in an Apache instance. `RocketMQClientProvider` now discovers and returns each connection's actual `clusterName`, but the Producer deduplication key still contains only `producerGroup`, `clientId`, and `remoteIP`.

If two clusters legitimately contain an identical Producer identity, the second row is dropped by `putIfAbsent`. The UI then reports only one cluster even though both clusters have an online Producer.

## Reproduction

1. Configure one Apache instance whose NameServer topology exposes `cluster-a` and `cluster-b`.
2. Register a Producer with the same group, client ID, and remote address in both clusters.
3. Call `GET /api/clients?instanceId=<id>&type=Producer` without a `clusterId` filter.
4. Only one connection is returned because the deduplication key does not include the discovered cluster.

## Expected behavior

- Duplicate Broker reports inside the same cluster are still collapsed.
- Identical Producer identities in different clusters remain separate rows.
- Each row retains its discovered `clusterName`.

## Scope

Update the Producer aggregation key in `RocketMQClientProvider` and add a focused regression test. No API contract or persistence change is required.

Track 1: unified control plane client management.

Aias00:
## Summary

- include the discovered RocketMQ cluster in the Producer aggregation key
- keep collapsing duplicate Broker reports inside one cluster
- preserve identical Producer identities that belong to different clusters
- add a regression test covering the cross-cluster collision

Closes #1991

## Verification

- failure proof before the production fix: `RocketMQClientProviderTest.producerScanPreservesIdenticalConnectionsAcrossClusters` expected 2 rows but received 1
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -f server/pom.xml -Dtest=RocketMQClientProviderTest test` (17 tests passed)
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -f server/pom.xml -DskipTests package`
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProvider.java
@@ -183,7 +183,8 @@ public class RocketMQClientProvider implements ClientProvider {
                 if (producerInfo == null) {
                     continue;
                 }
-                String key = producerGroup + '\0'
+                String key = Objects.toString(clusterId, "") + '\0'
+                        + producerGroup + '\0'
                         + Objects.toString(producerInfo.getClientId(), "") + '\0'
                         + Objects.toString(producerInfo.getRemoteIP(), "");
                 connections.putIfAbsent(key, toConnectionVO(producerInfo, producerGroup, clusterId));
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProviderTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProviderTest.java
@@ -116,6 +116,25 @@ class RocketMQClientProviderTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a deduplication bug where producer connections with identical client IDs and remote IPs across different clusters were incorrectly collapsed into a single entry. By prepending `clusterId` to the dedup key, connections from different clusters are now correctly preserved as separate entries.

The fix is a one-line key change with a well-designed test that verifies cross-cluster connections are not deduplicated. LGTM.

---

lizhimins:

PR #1995, opened 2026-08-12
fix: preserve structured reset offset errors

### LINKED ISSUE #1994: [Studio][Bug] Preserve structured errors when resetting consumer offsets
## Problem

`RocketMQAdminClientImpl.resetOffset()` catches every exception and replaces it with `BusinessException(500, "Failed to reset offset: ...")`.

This also catches structured control-plane errors raised before any Broker call, including:

- 400 when the selected instance has no endpoint
- 404 when the selected instance does not exist
- 422 when its `adminCredentialRef` is not configured

The API therefore reports an internal server error instead of the actionable status and message already produced by `RuntimeAdminClientResolver`.

## Reproduction

1. Call `POST /api/groups/reset-offset` with a syntactically valid request whose `instanceId` does not exist.
2. `RuntimeAdminClientResolver` raises `BusinessException(404, "Instance not found: ...")`.
3. `resetOffset()` catches it as a generic exception and returns code 500.

## Expected behavior

- Existing `BusinessException` codes and messages are preserved.
- Unexpected RocketMQ/Admin failures still become the current reset-offset 500 response.
- Both failed paths continue to produce a failed audit record.

## Scope

Adjust the exception ordering in the Apache reset-offset implementation and add focused regression tests. No API shape or persistence change is required.

Track 1: unified control plane Consumer Group operations and failure semantics.

Aias00:
## Summary

- preserve structured `BusinessException` status codes and messages from instance and credential resolution
- keep unexpected Admin failures wrapped by the existing reset-offset 500 response
- retain failed audit records for both error paths
- add regression coverage for a resolver 404 and an unexpected Broker/Admin failure

Closes #1994

## Verification

- failure proof before the production fix: the new resolver test expected `Instance not found: missing-instance` with code 404 but received `Failed to reset offset: Instance not found: missing-instance` with code 500
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -f server/pom.xml -Dtest=RocketMQAdminClientImplTest test` (21 tests passed)
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -f server/pom.xml -DskipTests package`
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
@@ -507,6 +507,9 @@ public class RocketMQAdminClientImpl implements AdminClient {
             }
             recordAudit("RESET_OFFSET", name,
                     "instanceId=" + instanceId + ", topic=" + topic + ", timestamp=" + timestamp, "SUCCESS");
+        } catch (BusinessException e) {
+            recordAudit("RESET_OFFSET", name, e.getMessage(), "FAILED");
+            throw e;
         } catch (Exception e) {
             recordAudit("RESET_OFFSET", name, e.getMessage(), "FAILED");
             throw new BusinessException(500, "Failed to reset offset: " + e.getMessage());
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR correctly preserves `BusinessException` status codes in `resetOffset` by catching it before the generic `Exception` handler. Previously, a 404 `BusinessException` (e.g., instance not found) would be wrapped into a generic 500 error, losing the original error context for the client.

The fix is minimal, well-targeted, and includes thorough test coverage for both the preserved and wrapped exception paths. LGTM.

---

lizhimins:

PR #2001, opened 2026-08-12
fix: preserve instance deletion integrity

### LINKED ISSUE #2000: [Studio][Bug] Concurrent instance deletion reports false success

Aias00:
## Summary

- preserve the affected-row result from instance repository deletion
- return `404` when an instance was removed concurrently
- avoid runtime client release and successful audit records when no row was deleted

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -f server/pom.xml -Dtest=InstanceServiceTest,MybatisPlusInstanceRepositoryTest test`
- `git diff --check`

Closes #2000

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceRepository.java
@@ -35,7 +35,7 @@ public interface InstanceRepository {
 
     InstanceVO save(InstanceVO instance);
 
-    void deleteById(String id);
+    boolean deleteById(String id);
 
     boolean existsByCredentialId(String credentialId);
 
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -256,7 +256,9 @@ public class InstanceService {
                     "Cannot delete instance with managed resources: topics=%d, consumerGroups=%d",
                     topicCount, consumerGroupCount));
         }
-        instanceRepository.deleteById(id);
+        if (!instanceRepository.deleteById(id)) {
+            throw new BusinessException(404, "InstanceVO not found: " + id);
+        }
         releaseApacheEndpointIfUnused(existing, null);
         recordAudit("DELETE_INSTANCE", "INSTANCE", id, null,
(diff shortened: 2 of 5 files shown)


RockteMQ-AI (approved):
### Summary
This PR hardens instance deletion against concurrent removal races by changing `InstanceRepository.deleteById` from `void` to `boolean` and adding a 404 check in the service layer. It also strengthens test coverage with a new concurrent-removal test and verifies that the controller returns 404 when the instance has already been deleted.

The changes are well-structured:
- The `void → boolean` return type change in `deleteById` is a clean defensive pattern that lets callers detect whether the row existed.
- The 404 check in `deleteInstance` correctly handles the race where another request deletes the instance between the pre-check and the actual delete.
- The new test `deleteInstance_returns404_whenConcurrentRemovalOccurs` validates the exact race condition.
- The controller test correctly verifies the 404 response.

LGTM — clean fix with proper test coverage for the concurrency edge case.

---

Aias00:
Superseded by #1997, which now includes this deletion-integrity fix together with the related instance mutation refresh behavior.

PR #2005, opened 2026-08-12
fix: preserve and reconcile system alert state

### LINKED ISSUE #2004: [Studio][Bug] Do not resurrect concurrently cleared system alerts
### Before Creating the Issue

- [x] I searched open issues and pull requests for the same backend persistence race.
- [x] I confirmed the behavior on the latest `rocketmq-studio` branch.

### Description

`AlertService.acknowledgeAlert` reads a system alert and then calls `AlertRepository.saveAlert`. The MyBatis repository implements `saveAlert` as an upsert: it selects by ID and inserts when the row is absent.

If `clearAcknowledged` deletes the alert after the acknowledgement request reads it but before `saveAlert` runs, the acknowledgement path inserts the deleted alert again. The API reports success and records a successful acknowledgement audit, while a record intentionally removed by the cleanup operation is resurrected.

This is separate from the frontend duplicate-request tracking in #1573/#1574; it is a backend persistence race between acknowledgement and cleanup.

### Expected behavior

Acknowledging an existing alert should be an update-only operation:

- never recreate an alert that was concurrently removed;
- report a structured `404` when the update affects zero rows;
- do not record a successful acknowledgement audit in that case.

### Suggested fix

Expose an update-only repository operation that returns the affected-row outcome, use it in `acknowledgeAlert`, and retain insert/upsert behavior only for genuine alert creation paths if needed.

### Regression tests

- repository update reports `false` when MyBatis updates zero rows and never inserts;
- service maps the failed update to `404` and skips success audit;
- normal acknowledgement still persists and audits successfully.

### Scope

RocketMQ Studio Track 1: operational alert persistence integrity.

### LINKED ISSUE #2014: [Studio][Bug] Refresh system alerts after clearing acknowledged records
## Problem

After `clearAcknowledgedAlerts()` succeeds, the System Alerts page only filters rows that were already marked `acknowledged` in its local state.

That local snapshot may be stale. If another session acknowledges an additional alert after this page loaded, the backend clear operation deletes both the locally acknowledged alert and the newly acknowledged alert, while the page removes only the first one.

## Reproduction

1. Load alerts A (acknowledged) and B (unacknowledged) in the page.
2. In another session, acknowledge B.
3. Click **Clear acknowledged** on the original page.
4. The backend deletes A and B.

The current page filters only local `acknowledged` rows, so B remains visible even though it no longer exists on the server.

## Expected behavior

- After a successful clear operation, reload the canonical server alert list.
- Do not report the clear itself as failed if only the follow-up refresh fails.
- Keep a safe local fallback and show an actionable refresh warning.
- Add focused frontend regression coverage for stale local acknowledgement state.

## Scope

System Alerts page state reconciliation and tests only.

## Track

Aias00:
## Summary

- make system alert acknowledgement an update-only persistence operation
- return `404` when an alert is cleared concurrently instead of recreating it
- skip successful acknowledgement audit when no row was updated
- reload the canonical server alert list after clearing acknowledged records
- retain a safe local fallback if the follow-up refresh fails and block acknowledgements while clear is running
- cover stale cross-session acknowledgement state

Closes #2004
Closes #2014

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -Dtest=AlertServiceTest,MybatisPlusAlertRepositoryTest test`
- `npm test -- --run src/pages/ops/__tests__/SystemAlertsPage.test.tsx`
- `npx eslint src/pages/ops/systemAlerts.tsx src/pages/ops/__tests__/SystemAlertsPage.test.tsx`
- `npm run build`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -f server/pom.xml -DskipTests package`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
@@ -155,10 +155,12 @@ public class AlertService {
                 .findFirst()
                 .orElseThrow(() -> new org.apache.rocketmq.studio.common.exception.BusinessException(404, "System alert not found: " + id));
         alert.setAcknowledged(true);
-        SystemAlertVO saved = alertRepository.saveAlert(alert);
-        recordAudit("ACKNOWLEDGE_SYSTEM_ALERT", "SYSTEM_ALERT", saved.getId(), null,
+        if (!alertRepository.acknowledgeAlert(alert)) {
+            throw new BusinessException(404, "System alert not found: " + id);
+        }
+        recordAudit("ACKNOWLEDGE_SYSTEM_ALERT", "SYSTEM_ALERT", alert.getId(), null,
                 "acknowledged=true");
-        return saved;
+        return alert;
     }
 
 
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertRepository.java
@@ -87,14 +87,8 @@ public class MybatisPlusAlertRepository implements AlertRepository {
     }
 
     @Override
-    public SystemAlertVO saveAlert(SystemAlertVO alert) {
-        RmqSystemAlert entity = toAlertEntity(alert);
-        if (entity.getId() != null && alertMapper.selectById(entity.getId()) != null) {
-            alertMapper.updateById(entity);
-        } else {
-            alertMapper.insert(entity);
-        }
-        return alert;
+    public boolean acknowledgeAlert(SystemAlertVO alert) {
+        return alertMapper.updateById(toAlertEntity(alert)) > 0;
     }
 
     @Override
--- a/server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertServiceTest.java
@@ -565,12 +565,12 @@ class AlertServiceTest {
         SystemAlertVO existing = SystemAlertVO.builder().id("a1").level(AlertLevel.error)
                 .title("Broker Down").acknowledged(false).build();
         when(alertRepository.findAlerts(null)).thenReturn(List.of(existing));
-        when(alertRepository.saveAlert(any(SystemAlertVO.class))).thenAnswer(invocation -> invocation.getArgument(0));
+        when(alertRepository.acknowledgeAlert(any(SystemAlertVO.class))).thenReturn(true);
 
         SystemAlertVO result = alertService.acknowledgeAlert("a1");
 
         assertThat(result.isAcknowledged()).isTrue();
-        verify(alertRepository).saveAlert(result);
+        verify(alertRepository).acknowledgeAlert(result);
         verify(operationAuditService).record(eq("ACKNOWLEDGE_SYSTEM_ALERT"), eq("SYSTEM_ALERT"), eq("a1"),
                 eq(null), eq("acknowledged=true"), eq("SUCCESS"), eq(null));
     }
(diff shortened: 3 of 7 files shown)


server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertRepository.java
```diff
@@ -87,14 +87,8 @@ public List<SystemAlertVO> findAlerts(String level) {
...
     @Override
-    public SystemAlertVO saveAlert(SystemAlertVO alert) {
-        RmqSystemAlert entity = toAlertEntity(alert);
-        if (entity.getId() != null && alertMapper.selectById(entity.getId()) != null) {
-            alertMapper.updateById(entity);
-        } else {
-            alertMapper.insert(entity);
-        }
-        return alert;
+    public boolean acknowledgeAlert(SystemAlertVO alert) {
```
RockteMQ-AI, line 90: Replacing the check-then-insert `saveAlert` with a simple `updateById` is the correct fix. The old pattern could silently recreate a concurrently deleted alert; the new pattern correctly returns false when no row is updated.

server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
```diff
@@ -155,10 +155,12 @@ public SystemAlertVO acknowledgeAlert(String id) {
                 .findFirst()
                 .orElseThrow(() -> new org.apache.rocketmq.studio.common.exception.BusinessException(404, "System alert not found: " + id));
         alert.setAcknowledged(true);
-        SystemAlertVO saved = alertRepository.saveAlert(alert);
-        recordAudit("ACKNOWLEDGE_SYSTEM_ALERT", "SYSTEM_ALERT", saved.getId(), null,
+        if (!alertRepository.acknowledgeAlert(alert)) {
```
RockteMQ-AI, line 158: The audit record is now only written when `acknowledgeAlert` returns true, which correctly prevents a false audit trail when the alert was concurrently removed. The 404 error includes the alert ID for debugging.

server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertServiceTest.java
```diff
@@ -584,6 +584,21 @@ void acknowledgeAlertShouldThrowWhenAlertNotFound() {
                 .hasMessageContaining("System alert not found: non-existent");
     }
 
+    @Test
+    void acknowledgeAlertShouldRejectConcurrentRemoval() {
+        SystemAlertVO existing = SystemAlertVO.builder().id("a1").level(AlertLevel.error)
+                .title("Broker Down").acknowledged(false).build();
```
RockteMQ-AI, line 590: Good test — verifies both the exception and that no audit record is written when the alert was concurrently removed. This ensures the fix prevents both the data integrity issue and the false audit trail.

RockteMQ-AI (commented):
### Summary
This PR fixes a race condition in system alert acknowledgement where a concurrent removal could be silently recreated by the save-or-insert pattern. It replaces `saveAlert` with an update-only `acknowledgeAlert` method.

### Findings
- **[Info]** MybatisPlusAlertRepository.java:90 — Correct replacement of check-then-insert with update-only
- **[Info]** AlertService.java:158 — Audit record now only written on successful update, preventing false audit trails
- **[Info]** AlertServiceTest.java:590 — Good test coverage for concurrent removal scenario

### Overall
The fix is correct and well-tested. The rename from `saveAlert` to `acknowledgeAlert` also improves clarity — the method name now accurately reflects its purpose. The existing test suite has been updated, and a new test covers the concurrent removal case.

---

RockteMQ-AI (approved):
### Summary
This PR fixes a race condition where acknowledging a system alert could re-create a concurrently deleted alert. The fix replaces the generic `saveAlert` upsert with a targeted `acknowledgeAlert` that only updates existing records.

The frontend also reloads the canonical alert list after clearing, preventing stale local state from overwriting newer data. Clean implementation with good test coverage.

LGTM — no issues found.

---

lizhimins:

Issue #1622, opened 2026-08-11
[Studio][Bug] ACL admin switches allow concurrent updates


123123213weqw:
## Bug

ACL user admin switches can start multiple updates for the same user before the first request completes. This produces duplicate writes and lets response ordering decide the rendered user record.

## Expected behavior

Track admin updates per user, ignore duplicate toggles for an in-flight user, and keep other rows independently usable.

RockteMQ-AI:
Thank you for reporting this issue.

I have analyzed the content and classified this as a **potential bug**. I will verify the issue against the codebase and follow up with findings.

**Issue Metadata**
- Type: Bug Report
- Classification: Automated (by github-manager)
- Next Step: Code verification

---
*This is an automated assessment. A maintainer will review and confirm.*

PR #1725, opened 2026-08-11
fix: preserve ACL lifecycle and local persistence integrity

Aias00:
## Summary
- preserve ACL page lifecycle and instance context
- enforce local plain-access account identity integrity and normalize persisted ACL CSV collections
- detect legacy duplicate access keys with a structured conflict

## Verification
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home PATH="$JAVA_HOME/bin:$PATH" mvn -q -B -ntp -Dtest=AclServiceTest,MybatisPlusAclRepositoryTest test`

Consolidates #2274.

PR #2007, opened 2026-08-12
fix: preserve ACL user update integrity

### LINKED ISSUE #2006: [Studio][Bug] Do not recreate concurrently deleted ACL users
### Before Creating the Issue

- [x] I searched existing issues and pull requests, including ACL rule lost-update work in #1881 and frontend ACL concurrency work in #1622.
- [x] I confirmed this separate ACL user persistence race on the latest `rocketmq-studio` branch.

### Description

`AclService.updateUser` loads an existing ACL user, merges editable fields while preserving credentials, and calls `AclRepository.saveUser`. The MyBatis implementation of `saveUser` performs a second existence check and inserts when the row is absent.

If a delete request removes the user after `updateUser` reads it but before `saveUser` checks persistence, the update path inserts the user again with the old AccessKey and SecretKey. The API reports a successful update and records a success audit even though the concurrent delete won first.

This does not overlap #1881, which covers ACL **rule** replacement row counts, or #1622, which covers frontend admin-switch request tracking.

### Expected behavior

ACL user updates must be update-only:

- never recreate a concurrently deleted identity or credential;
- return structured `404` when the update affects zero rows;
- skip successful update audit in that case.

ACL user creation should retain its existing insert behavior.

### Suggested fix

Add an update-only repository method returning an optional/result based on the affected-row count, use it from `AclService.updateUser`, and keep `saveUser` for creation.

### Regression tests

- repository replacement returns empty for both a missing row and a zero-row concurrent update;
- service maps empty replacement to `404` without success audit;
- normal update still preserves stored credentials and masks the response.

### Scope

RocketMQ Studio Track 1 / AUTH-01 local ACL management integrity.

Aias00:
## Summary

- add an update-only ACL user repository contract
- return `404` when an ACL user is deleted concurrently instead of recreating its identity and credentials
- retain the existing create-user insert path and credential masking behavior

This does not overlap #1881 (ACL rule replacement) or #1622 (frontend admin-switch request tracking).

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -f server/pom.xml -Dtest=AclServiceTest,MybatisPlusAclRepositoryTest test`
- `git diff --check`

Closes #2006

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -134,7 +134,8 @@ public class AclService {
                 .clusters(user.getClusters() == null ? existing.getClusters() : user.getClusters())
                 .createdAt(existing.getCreatedAt())
                 .build();
-        AclUserVO saved = aclRepository.saveUser(merged);
+        AclUserVO saved = aclRepository.replaceUser(merged)
+                .orElseThrow(() -> new BusinessException(404, "ACL user not found: " + user.getId()));
         auditUser("UPDATE_ACL_USER", saved);
         return maskCredentials(saved);
     }
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
@@ -112,6 +112,21 @@ public class MybatisPlusAclRepository implements AclRepository {
     }
 
     @Override
+    public Optional<AclUserVO> replaceUser(AclUserVO user) {
+        RmqAclUser existing = userMapper.selectById(user.getId());
+        if (existing == null) {
+            return Optional.empty();
+        }
+        RmqAclUser entity = toUserEntity(user);
+        entity.setCreatedAt(existing.getCreatedAt());
+        if (userMapper.updateById(entity) == 0) {
+            return Optional.empty();
+        }
+        user.setCreatedAt(existing.getCreatedAt());
+        return Optional.of(user);
+    }
+
+    @Override
     public boolean deleteUser(String id) {
         return userMapper.deleteById(id) > 0;
     }
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/acl/AclServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/acl/AclServiceTest.java
@@ -411,7 +411,8 @@ class AclServiceTest {
 
         ArgumentCaptor<AclUserVO> captor = ArgumentCaptor.forClass(AclUserVO.class);
         when(aclRepository.findUserById("user-1")).thenReturn(Optional.of(existingUser));
-        when(aclRepository.saveUser(any(AclUserVO.class))).thenAnswer(inv -> inv.getArgument(0));
+        when(aclRepository.replaceUser(any(AclUserVO.class)))
+                .thenAnswer(inv -> Optional.of(inv.getArgument(0)));
 
         AclUserVO result = aclService.updateUser(input);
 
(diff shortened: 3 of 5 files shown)


server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
```diff
@@ -111,6 +111,21 @@ public AclUserVO saveUser(AclUserVO user) {
         return user;
     }
 
+    @Override
+    public Optional<AclUserVO> replaceUser(AclUserVO user) {
+        RmqAclUser existing = userMapper.selectById(user.getId());
+        if (existing == null) {
```
RockteMQ-AI, line 117: The select-then-update pattern here is not atomic, but the `updateById` returning 0 correctly handles the race where another thread deletes between the select and update. This is a pragmatic approach that avoids adding database-level locking while still preventing the recreate-on-update bug.

RockteMQ-AI (commented):
### Summary
This PR fixes a race condition in ACL user updates where a concurrent deletion could be silently recreated by the update path. It introduces a new `replaceUser` method that only updates existing users without falling back to insert.

### Findings
- **[Info]** MybatisPlusAclRepository.java:117 — Select-then-update pattern correctly handles concurrent deletion via updateById return value
- **[Info]** AclService.java:137 — Correct switch from save-or-insert to update-only with proper 404 handling
- **[Info]** AclServiceTest.java:475 — Good test coverage for the concurrent deletion scenario

### Overall
The fix is correct and well-tested. The `replaceUser` method properly enforces the update-only contract, preventing the recreate-on-update bug. The existing test suite has been updated to use the new method, and a new test covers the concurrent deletion case.

---

server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
```diff
@@ -134,7 +134,8 @@ public AclUserVO updateUser(UpdateAclUserDTO user) {
                 .clusters(user.getClusters() == null ? existing.getClusters() : user.getClusters())
                 .createdAt(existing.getCreatedAt())
                 .build();
-        AclUserVO saved = aclRepository.saveUser(merged);
+        AclUserVO saved = aclRepository.replaceUser(merged)
```
RockteMQ-AI, line 137: Switching from `saveUser` (which could recreate a deleted identity) to `replaceUser` with `orElseThrow` is the correct fix. The 404 error message includes the user ID, which helps with debugging.

server/src/test/java/org/apache/rocketmq/studio/instance/acl/AclServiceTest.java
```diff
@@ -466,7 +468,23 @@ void updateUserShouldThrowWhenUserDoesNotExist() {
                 .isInstanceOf(BusinessException.class)
                 .hasMessage("ACL user not found: missing")
                 .satisfies(ex -> assertThat(((BusinessException) ex).getCode()).isEqualTo(404));
-        verify(aclRepository, never()).saveUser(any(AclUserVO.class));
+        verify(aclRepository, never()).replaceUser(any(AclUserVO.class));
+    }
+
+    @Test
+    void updateUserShouldRejectConcurrentDeletion() {
```
RockteMQ-AI, line 475: Good test coverage for the concurrent deletion scenario. The test verifies both the exception and that `replaceUser` was called, ensuring the update-only contract is enforced.

RockteMQ-AI (approved):
### Summary
This PR fixes a data integrity issue where updating an ACL user could inadvertently re-create a concurrently deleted user. The fix replaces the upsert-style `saveUser` with a new `replaceUser` method that checks existence first and returns empty if the user was deleted.

The implementation is clean and the tests properly cover the concurrent deletion scenario.

LGTM — no issues found.

---

Aias00:
Superseded by #1725, which now includes ACL user update integrity together with the related lifecycle safeguards.

PR #2013, opened 2026-08-12
fix: gate data source mutations on initial load

### LINKED ISSUE #2012: [Studio][Bug] Prevent data source mutations before the initial list is ready
## Problem

The Metrics data source tab starts `listDataSources()` on mount, but the **Add data source** action remains enabled while that request is pending.

A user can create a data source before the initial list resolves. The create response is first appended to local state, but the late initial list response then unconditionally calls `setDataSources(sources)` and removes the newly created row from the UI.

## Reproduction

1. Open Settings > Data sources and keep `listDataSources()` pending.
2. Open the create modal and create a new data source successfully.
3. Resolve the original list request with the pre-create snapshot.

The success notification is shown, then the newly created row disappears.

## Expected behavior

- Data source mutations must not start until the initial list snapshot is ready, or list responses must be reconciled with completed mutations.
- The smallest fix should disable the create entry while the initial list is loading.
- Add a deferred-promise regression test proving that a stale initial snapshot cannot overwrite a completed creation.

## Scope

Metrics data source settings UI and focused frontend tests only.

## Track

Aias00:
## Summary
- disable data source creation until the initial list snapshot is ready
- prevent a late bootstrap response from overwriting a completed creation
- cover the loading boundary with a deferred frontend regression test

## Testing
- `npm test -- --run src/pages/settings/__tests__/DataSourceTab.test.tsx`
- `npx eslint src/pages/settings/index.tsx src/pages/settings/__tests__/DataSourceTab.test.tsx`
- `npm run build`

Closes #2012

### DIFF
--- a/web/src/pages/settings/__tests__/DataSourceTab.test.tsx
+++ b/web/src/pages/settings/__tests__/DataSourceTab.test.tsx
@@ -52,6 +52,16 @@ const sources: DataSource[] = [
   },
 ];
 
+const deferred = <T,>() => {
+  let resolve!: (value: T) => void;
+  let reject!: (reason?: unknown) => void;
+  const promise = new Promise<T>((resolvePromise, rejectPromise) => {
+    resolve = resolvePromise;
+    reject = rejectPromise;
+  });
+  return { promise, resolve, reject };
+};
+
 beforeAll(() => {
   Object.defineProperty(window, 'matchMedia', {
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #1966, which now contains this bootstrap request-ownership fix together with the related data source lifecycle fixes.

PR #2015, opened 2026-08-12
fix: reconcile alerts after acknowledged cleanup

### LINKED ISSUE #2014: [Studio][Bug] Refresh system alerts after clearing acknowledged records

Aias00:
## Summary
- reload the canonical server alert list after clearing acknowledged records
- keep a safe local fallback when the follow-up refresh fails
- prevent acknowledgement actions while a clear operation is running
- cover stale cross-session acknowledgement state

## Testing
- `npm test -- --run src/pages/ops/__tests__/SystemAlertsPage.test.tsx`
- `npx eslint src/pages/ops/systemAlerts.tsx src/pages/ops/__tests__/SystemAlertsPage.test.tsx`
- `npm run build`

Closes #2014

### DIFF
--- a/web/src/pages/ops/__tests__/SystemAlertsPage.test.tsx
+++ b/web/src/pages/ops/__tests__/SystemAlertsPage.test.tsx
@@ -10,7 +10,11 @@ import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 import { LangProvider } from '../../../i18n/LangContext';
-import { acknowledgeAlert, listSystemAlerts } from '../../../services/opsService';
+import {
+  acknowledgeAlert,
+  clearAcknowledgedAlerts,
+  listSystemAlerts,
+} from '../../../services/opsService';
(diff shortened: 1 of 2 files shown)


Aias00:
Superseded by #2005, which now contains this frontend reconciliation fix together with the alert acknowledgement integrity changes.

PR #2021, opened 2026-08-12
fix(web): surface Message Explorer topic loading failures

### LINKED ISSUE #2016: [Studio][Bug] Surface Message Explorer Topic loading failures
## Track

Track 1 - AI-Native unified control plane

## Problem

The Message Explorer loads Topic options for the selected instance before a query can run. When that request fails, the page currently catches the error and silently replaces the options with an empty list.

As a result, an unavailable instance, ACL denial, or network failure is presented like an instance with no Topics. The query button then only reports `请选择 Topic`, which sends operators toward input troubleshooting instead of the actual runtime failure.

## Reproduction

1. Open Message Explorer with a selected instance.
2. Make `GET /api/topics?instanceId=<id>` fail (for example, stop the NameServer or use credentials without Topic-list permission).
3. Open the Topic selector or inspect the query action.

The selector is empty and no load error or retry action is shown.

## Expected behavior

- Show a visible, actionable error when Topic options cannot be loaded.
- Keep message queries disabled while Topic options are loading or unavailable.
- Allow retrying the Topic lookup without changing the selected instance.
- Preserve an actual empty Topic list as a distinct successful outcome.

## Proposed scope

Update the Message Explorer frontend state and regression tests only. No backend API contract changes are required.

123123213weqw:

Fix #2016.

The Message Explorer loaded Topic options for the selected instance before a query could run. When that request failed, the page silently replaced the options with an empty list, so an unavailable instance, ACL denial or network failure looked like an instance with no Topics.

## Brief changelog

- message.tsx: keep a topicError state, set it when the Topic list request fails and clear it on success
- render an error Alert above the query card so the real runtime failure is visible instead of an empty Topic selector

- npx vitest run src/pages/instance/__tests__/MessagePage.test.tsx src/pages/instance/__tests__/MessagePageAsyncState.test.tsx (24 passed)
- npx tsc -b clean
- npx eslint . clean

### DIFF
--- a/web/src/pages/home/__tests__/HomePage.test.tsx
+++ b/web/src/pages/home/__tests__/HomePage.test.tsx
@@ -78,6 +78,18 @@ describe('HomePage LLM models', () => {
     expect(await screen.findByText('qwen3.8-max')).toBeInTheDocument();
   });
 
+  it('does not fetch LLM config in Mock mode', async () => {
+    const { useDataModeStore } = await import('../../../stores/dataModeStore');
+    useDataModeStore.getState().toggle();
+
+    renderHome();
+
+    expect(llmApiMocks.getLlmConfig).not.toHaveBeenCalled();
+    expect(await screen.findByText('qwen3.8-max')).toBeInTheDocument();
+
+    useDataModeStore.getState().toggle();
+  });
+
   it('submits the selected model and engine to the AI page', async () => {
     const user = userEvent.setup();
     renderHome();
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Fixes Message Explorer to properly surface topic loading failures instead of silently failing, improving error visibility for users. Clean implementation with proper error handling.

---

lizhimins:

PR #2023, opened 2026-08-12
fix(web): prevent data source mutations before the initial list is ready

### LINKED ISSUE #2012: [Studio][Bug] Prevent data source mutations before the initial list is ready

123123213weqw:

Fix #2012.

The Metrics data source tab started listDataSources() on mount while the Add data source action remained enabled. A user could create a data source before the initial list resolved; the create response was appended to local state, but the late initial list response then called setDataSources(sources) unconditionally and removed the newly created row from the UI.

## Brief changelog

- settings DataSourceTab: disable the Add data source button while the initial list request is pending, so no mutation can race the initial list response

- npx vitest run src/pages/settings/__tests__/DataSourceTab.test.tsx (7 passed)
- npx tsc -b clean
- npx eslint . clean

### DIFF
--- a/web/src/pages/home/__tests__/HomePage.test.tsx
+++ b/web/src/pages/home/__tests__/HomePage.test.tsx
@@ -78,6 +78,18 @@ describe('HomePage LLM models', () => {
     expect(await screen.findByText('qwen3.8-max')).toBeInTheDocument();
   });
 
+  it('does not fetch LLM config in Mock mode', async () => {
+    const { useDataModeStore } = await import('../../../stores/dataModeStore');
+    useDataModeStore.getState().toggle();
+
+    renderHome();
+
+    expect(llmApiMocks.getLlmConfig).not.toHaveBeenCalled();
+    expect(await screen.findByText('qwen3.8-max')).toBeInTheDocument();
+
+    useDataModeStore.getState().toggle();
+  });
+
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
### Summary
Fixes data source mutations to wait for initial list load completion before allowing edits, preventing race conditions where mutations could occur against uninitialized state. Clean implementation.

---

lizhimins:

PR #2026, opened 2026-08-12
fix(web): refresh instance mutations with the latest list filters

### LINKED ISSUE #1996: [Studio][Bug] Refresh instance mutations with the latest list filters

btlqql:

Fix #1996.

The Studio instance page captured loadInstances when a create, update, or delete mutation started. That callback also captured the search and instance-type filters from the same render. If an operator changed the list filters while the mutation request was in flight, the mutation completion invoked the stale callback whose refresh request used the old filters and overwrote the table with results that no longer matched the visible filters.

## Brief changelog

- instance page: mutations now bump a reloadVersion that triggers the list effect to re-run with the latest filters instead of calling a stale loadInstances closure

- npx vitest run src/pages/instance/__tests__/InstancePage.test.tsx (11 passed)
- npx tsc -b clean
- npx eslint . clean

### DIFF
--- a/web/src/pages/home/__tests__/HomePage.test.tsx
+++ b/web/src/pages/home/__tests__/HomePage.test.tsx
@@ -78,6 +78,18 @@ describe('HomePage LLM models', () => {
     expect(await screen.findByText('qwen3.8-max')).toBeInTheDocument();
   });
 
+  it('does not fetch LLM config in Mock mode', async () => {
+    const { useDataModeStore } = await import('../../../stores/dataModeStore');
+    useDataModeStore.getState().toggle();
+
+    renderHome();
+
+    expect(llmApiMocks.getLlmConfig).not.toHaveBeenCalled();
+    expect(await screen.findByText('qwen3.8-max')).toBeInTheDocument();
+
+    useDataModeStore.getState().toggle();
+  });
+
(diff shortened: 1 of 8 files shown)


RockteMQ-AI (approved):
### Summary
Fixes instance mutation operations to use the latest list filters instead of stale cached data, preventing inconsistencies when filtering changes between list load and mutation. Clean fix with proper test coverage.

---

lizhimins:

PR #2041, opened 2026-08-12
test: cover Tencent topic count and consumers

tju-yxq:
Fixes the remaining Tencent topic-count edge case from #1599 and strengthens subscription pagination coverage.

`countTopics(...)` now treats `DescribeTopicListResponse.TotalCount` as authoritative even when the minimal count request returns an empty `Data` array. It still handles missing responses, missing metadata, negative values, and counts above the Java integer range consistently with the current provider behavior.

The tests also verify that topic consumers stop after the declared `SubscriptionCount`, including the case where the final page is full.

## Testing

```bash
cd server
mvn -DskipTests=false "-Dtest=TencentInstanceProviderTest#countTopicsShouldUseTotalCountFromSingleItemRequestTest+countTopicsShouldUseTotalCountEvenWhenDataIsEmptyTest+countTopicsShouldClampOversizedTotals+getTopicConsumersShouldReadEverySubscriptionPageTest+getTopicConsumersShouldStopAfterSubscriptionCountReachedOnFullPageTest" test
```

`BUILD SUCCESS` — 5 tests, 0 failures, 0 errors.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProvider.java
@@ -61,6 +61,7 @@ import org.springframework.stereotype.Component;
 import lombok.RequiredArgsConstructor;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
Good fix and test coverage. The code change in TencentInstanceProvider.countTopics() correctly prefers TotalCount from the API response before falling back to a full listTopics scan. The reordering ensures that when TotalCount is available (which is the common case), no unnecessary list call is made. The null-safe check (response == null ? null : response.getTotalCount()) prevents NPE.

Test coverage for the Limit=1 count request, empty-data TotalCount case, and getTopicConsumers pagination edge is solid.

---

RockteMQ-AI (approved):
### Summary
Improves test coverage for Tencent instance provider — adds tests for topic count with empty responses and consumer group mapping. Also refines the `countTopics()` implementation to rely on `totalCount` from the API response rather than falling back to a full scan.

### Findings
- **[Info]** The `countTopics()` change removes the fallback to `listTopics().size()` when `totalCount` is null. This assumes Tencent always returns `totalCount`, which should be verified against the API docs. If `totalCount` can be null in some edge cases, this could cause a regression.
- **[Info]** Good test coverage for the consumer mapping logic.

Overall: reasonable improvement, but verify the `totalCount` assumption holds for all Tencent API responses.

---

PR #2043, opened 2026-08-12
fix: keep newest topic query messages across queues

### LINKED ISSUE #1828: [Bug] queryByTopic should return records in a stable timestamp order

tju-yxq:
Fixes #1828.

The current base sorts the messages collected by `queryByTopic(...)`, but it still stops scanning once an earlier queue fills the result limit. This change scans every returned queue and uses a bounded priority queue to retain the newest candidates across the whole topic query.

Results are ordered by `storeTime` descending, with `msgId` as a deterministic tie-breaker for equal timestamps. The tests cover cross-queue ordering, a later queue containing a newer message after the limit is filled, equal timestamps, and a pull result whose offset does not advance.

## Testing

```bash
cd server
mvn -DskipTests=false "-Dtest=RocketMQMessageProviderTest#queryByTopicSortsMessagesAcrossQueuesNewestFirst+queryByTopicKeepsNewestMessagesWhenEarlierQueuesFillTheDefaultLimit+queryByTopicUsesMessageIdAsStableTieBreakerForEqualTimestamps+queryByTopicStopsWhenPullOffsetDoesNotAdvance" test
```

`BUILD SUCCESS` — 4 tests, 0 failures, 0 errors. Checkstyle, main compilation, and test compilation also completed successfully.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
@@ -56,6 +56,7 @@ import java.util.Base64;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.PriorityQueue;
 import java.util.Set;
 
 /**
@@ -81,6 +82,9 @@ public class RocketMQMessageProvider implements MessageProvider {
     private static final long ONE_HOUR_MILLIS = 3600_000L;
     private static final long ONE_DAY_MILLIS = 24 * ONE_HOUR_MILLIS;
     private static final int MAX_CONSECUTIVE_OFFSET_ILLEGAL = 3;
+    private static final Comparator<MessageRecordVO> TOPIC_QUERY_ORDER = Comparator
+            .comparingLong(MessageRecordVO::getStoreTime)
+            .thenComparing(MessageRecordVO::getMsgId, Comparator.nullsFirst(String::compareTo));
 
     private final RuntimeAdminClientResolver runtimeAdminClientResolver;
     private final QueryHistoryService queryHistoryService;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
```diff
@@ -191,22 +192,20 @@ private List<MessageRecordVO> queryByKey(DefaultMQAdminExt adminExt, String topi
      */
     private List<MessageRecordVO> queryByTopic(String endpoint, String topic, String tag, long begin, long end, int limit) {
         DefaultMQPullConsumer consumer = newPullConsumer("studio-msg-query", endpoint);
-        List<MessageRecordVO> result = new ArrayList<>();
+        int resultLimit = Math.min(limit, TOPIC_QUERY_HARD_CAP);
+        PriorityQueue<MessageRecordVO> newestMessages = new PriorityQueue<>(
+                Comparator.comparingLong(MessageRecordVO::getStoreTime));
         try {
```
RockteMQ-AI, line 200: Removing the early-exit (`break outer`) means every queue is now fully scanned. For topics with a large number of messages this could be significantly slower than the previous implementation. Consider documenting the trade-off or adding a per-queue offset budget (e.g. cap the number of offsets scanned per queue proportionally to `resultLimit / queues.size()`) so that a single hot queue cannot dominate the scan time.

RockteMQ-AI (approved):
The fix correctly addresses the cross-queue ordering bug — using a bounded PriorityQueue (min-heap by storeTime) to retain the top-N newest messages across all queues is the right approach. The final descending sort and the resultLimit <= 0 guard are both correct. Test coverage for the cross-queue scenario is solid.

### Findings
- **[Warning]** RocketMQMessageProvider.java:198 — Removing the early-exit means every queue is now fully scanned; for large topics this could be significantly slower. Consider a per-queue offset budget.

### Suggestions
The addTopicQueryCandidate helper keeps the priority-queue logic clean and testable.

---

RockteMQ-AI (commented):
### Review Summary

This PR fixes the topic message query to return the newest messages across all queues, rather than stopping at the limit from the first queue(s).

### Findings
- **[Critical]** RocketMQMessageProvider.java:192 — The original code used a simple `ArrayList` and broke out of the loop when the limit was reached. This meant that if queue 0 had 200 old messages and queue 1 had 10 new messages, the new messages would never be seen. The fix uses a `PriorityQueue` to keep only the newest N messages across all queues.
- **[Warning]** The `PriorityQueue` approach requires scanning all messages across all queues, which could be slower for topics with many messages. However, this is necessary to ensure correctness.
- **[Info]** The test coverage is excellent, including edge cases for multi-queue scenarios and limit enforcement.

### Overall
Excellent fix for a correctness issue in topic message queries. The PriorityQueue approach ensures users see the newest messages regardless of which queue they're in, which is the expected behavior.

---

PR #2056, opened 2026-08-13
fix(web): preserve alert rule cluster and broker scope when editing

### LINKED ISSUE #1986: [Studio][Bug] Preserve alert rule cluster and broker scope when editing expressions
### What happened

The persisted alert-rule editor renders `clusterName` and `brokerName` as a PromQL selector in the expression field, but `toAlertRuleRequest` later submits the entire `metric{labels}` fragment as `metric` and omits both scope fields.

For example, editing and saving:

```text
rocketmq_consumer_lag_messages{cluster="DefaultCluster",broker="broker-a"} > 100000
```

produces a request whose `metric` is `rocketmq_consumer_lag_messages{...}` and whose `clusterName`/`brokerName` are absent. The backend exporter only accepts a metric identifier and generates selectors from the dedicated scope fields, so the next YAML export replaces the invalid metric or loses the original scope.

The form also accepts arbitrary label selectors even though the persisted backend contract can only represent `cluster` and `broker`.

### Expected behavior

- Parse the metric identifier separately from supported `cluster="..."` and `broker="..."` selectors.
- Submit the selector values through `clusterName` and `brokerName` on both create and update.
- Preserve scoped rules across edit/save/export round trips.
- Reject unsupported selectors explicitly instead of persisting a metric that the exporter will later replace.

### Evidence

- `web/src/pages/studio/AlertManagement.tsx`: `buildExpression` adds scope labels, while `parseExpression` returns the whole left-hand side as `metric`.
- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleRequestDTO.java`: scope has dedicated `clusterName` and `brokerName` fields.
- `AlertService.expression()` validates the metric identifier separately and appends labels from those scope fields.

### Scope

Track 1 / METRICS-01 alert-rule management. This is distinct from #1957, which fixes localized metric and duration values in the separate Ops alert form.

yyqdbngt:

Fix #1986.

The persisted alert-rule editor rendered clusterName and brokerName as a PromQL selector in the expression field, but toAlertRuleRequest later submitted the entire metric{labels} fragment as metric and omitted both scope fields. Editing and saving therefore dropped the cluster and broker scope.

## Brief changelog

- AlertManagement: parse cluster and broker labels out of the expression when building the update request and preserve them in clusterName/brokerName
- other labels (e.g. job) are kept in the metric expression

- npx vitest run src/pages/studio/__tests__/AlertManagement.test.tsx (9 passed)
- npx tsc -b clean
- npx eslint . clean

### DIFF
--- a/web/src/pages/studio/AlertManagement.tsx
+++ b/web/src/pages/studio/AlertManagement.tsx
@@ -247,15 +247,51 @@ function parseExpression(
   };
 }
 
+function parseScopeLabels(metric: string): {
+  metric: string;
+  clusterName?: string;
+  brokerName?: string;
+} {
+  const open = metric.indexOf('{');
+  if (open < 0) return { metric };
+  const metricName = metric.slice(0, open).trim();
+  const inner = metric.slice(open + 1, metric.lastIndexOf('}'));
+  let clusterName: string | undefined;
+  let brokerName: string | undefined;
+  const remaining: string[] = [];
+  for (const part of inner.split(',')) {
+    const eq = part.indexOf('=');
+    if (eq < 0) continue;
+    const key = part.slice(0, eq).trim();
+    const raw = part.slice(eq + 1).trim();
+    const value = raw.replace(/^"|"$/g, '').replace(/\\\\/g, '\\').replace(/\\"/g, '"');
+    if (key === 'cluster') clusterName = value;
+    else if (key === 'broker') brokerName = value;
+    else remaining.push(part.trim());
+  }
+  // Rebuild the metric expression preserving every non-scope label.
+  const metricExpr =
+    remaining.length > 0
+      ? `${metricName}{${remaining.join(',')}}`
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
## Review Summary

This PR adds a `parseScopeLabels()` function that correctly extracts `cluster` and `broker` labels from Prometheus-style metric expressions in the alert rule editor, preserving them as scope metadata while rebuilding the cleaned metric expression.

### Assessment: ✅ Approved

**Strengths:**
- Clean separation of concerns — scope extraction is isolated in a dedicated function
- Handles edge cases well: no braces, no scope labels, remaining labels preserved
- Properly integrates with the existing `toAlertRuleRequest()` flow
- TypeScript types are correct and consistent

**No issues found.** This is a focused, well-implemented fix for alert rule scope preservation.

lizhimins:

PR #2057, opened 2026-08-13
fix: do not recreate concurrently deleted ACL users

### LINKED ISSUE #2006: [Studio][Bug] Do not recreate concurrently deleted ACL users

yyqdbngt:

Fix #2006.

AclService.updateUser loaded an existing ACL user, merged editable fields while preserving credentials, and called AclRepository.saveUser. The MyBatis implementation of saveUser performs a second existence check and inserts when the row is absent. If a delete request removed the user after updateUser read it, saveUser re-created the deleted user with the stale credentials.

## Brief changelog

- AclRepository: add replaceUser that updates an existing user without ever inserting a missing one (mirrors replaceRule)
- MybatisPlusAclRepository: implement replaceUser using the existing row's createdAt
- AclService.updateUser: use replaceUser and return 404 when the user no longer exists

- Code review: matches the existing replaceRule pattern used for concurrent rule updates
- `git diff --check` clean

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclRepository.java
@@ -39,6 +39,12 @@ public interface AclRepository {
 
     AclUserVO saveUser(AclUserVO user);
 
+    /**
+     * Replaces an existing user atomically without creating a missing user.
+     * Returns empty when the user no longer exists.
+     */
+    Optional<AclUserVO> replaceUser(AclUserVO user);
+
     boolean deleteUser(String id);
 
     /**
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -134,7 +134,8 @@ public class AclService {
                 .clusters(user.getClusters() == null ? existing.getClusters() : user.getClusters())
                 .createdAt(existing.getCreatedAt())
                 .build();
-        AclUserVO saved = aclRepository.saveUser(merged);
+        AclUserVO saved = aclRepository.replaceUser(merged)
+                .orElseThrow(() -> new BusinessException(404, "ACL user not found: " + user.getId()));
         auditUser("UPDATE_ACL_USER", saved);
         return maskCredentials(saved);
     }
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
## Review Summary

This PR fixes a race condition where updating an ACL user could inadvertently recreate a user that was concurrently deleted, by introducing a `replaceUser()` method that only updates existing records.

### Assessment: ✅ Approved

**Strengths:**
- Clean interface extension — `replaceUser()` returns `Optional<AclUserVO>` to signal existence
- `MybatisPlusAclRepository` implementation correctly checks existence before update
- `AclService.updateUser()` now throws a proper 404 `BusinessException` when the user is gone
- Preserves `createdAt` from the existing entity during replacement
- Tests are properly updated to mock `replaceUser()` instead of `saveUser()`
- Follows the principle that updates should not create

**No issues found.** This is a well-designed fix for a subtle concurrency bug.

lizhimins:

PR #2058, opened 2026-08-13
fix: bound Claude Code streaming process output

### LINKED ISSUE #2040: [Studio][Security] Bound Claude Code streaming process output
## Problem

`ClaudeCodeAgentProvider.stream()` drains the Claude CLI's stdout and stderr without an output-size limit. Stdout is consumed through `BufferedReader.readLine()`, so one unbounded stream-json line can grow in memory before parsing, while stderr is collected with `readAllBytes()`. The fallback result text can also retain an arbitrarily large payload.

A misbehaving or compromised upstream reached through the Claude CLI can therefore make the Studio process buffer unbounded child-process output and exhaust the JVM heap. The non-streaming CLI path has separate bounded-output work, but the Claude streaming implementation does not use that path.

## Expected behavior

Bound the combined stdout/stderr bytes consumed from a Claude streaming process. If the limit is exceeded, terminate the child process and return a structured `llm.provider.output_too_large` error. Normal stream-json parsing and output below the limit should remain unchanged.

## Suggested verification

- oversized stdout is rejected and the process is terminated
- oversized stderr is rejected and the process is terminated
- existing large-but-bounded stderr draining still succeeds
- regular streamed/fallback result handling is unchanged

btlqql:

Fix #2040.

ClaudeCodeAgentProvider.stream() drained the Claude CLI's stdout and stderr without an output-size limit. Stdout was consumed through BufferedReader.readLine(), so one unbounded stream-json line could grow in memory before parsing, while stderr was collected with readAllBytes(). The fallback result text could also retain an arbitrarily large payload.

## Brief changelog

- ClaudeCodeAgentProvider: bound each stdout line to 1 MiB, stderr to 1 MiB, and the retained fallback result text to 256 KiB
- deltas are still emitted for oversized lines, but the full assistant/result text is no longer retained beyond the cap

- Code review: readLineBounded accumulates at most maxBytes per line; readNBytes caps stderr
- `git diff --check` clean

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/ClaudeCodeAgentProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/ClaudeCodeAgentProvider.java
@@ -52,6 +52,11 @@ public class ClaudeCodeAgentProvider extends CliAgentProvider {
     private static final String ANTHROPIC_APP_SUFFIX = "/apps/anthropic";
     private static final long STREAM_TIMEOUT_SECONDS = 300;
     private static final long OUTPUT_DRAIN_TIMEOUT_SECONDS = 10;
+    /** Hard cap for a single stream-json line and the collected stderr, in bytes. */
+    private static final long MAX_STREAM_LINE_BYTES = 1_048_576;
+    private static final long MAX_STDERR_BYTES = 1_048_576;
+    /** Cap for the retained fallback result text. */
+    private static final int MAX_RESULT_TEXT_CHARS = 262_144;
 
     private final LlmProperties llmProperties;
     private final ObjectMapper objectMapper = new ObjectMapper();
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
## Review Summary

This PR adds hard memory bounds to the Claude Code streaming agent to prevent unbounded memory consumption from oversized AI output.

### Assessment: ✅ Approved

**Strengths:**
- Introduces clear, well-named constants (`MAX_STREAM_LINE_BYTES`, `MAX_STDERR_BYTES`, `MAX_RESULT_TEXT_CHARS`)
- `readLineBounded()` correctly caps per-line retention while still consuming the full stream
- Graceful degradation: when result text exceeds the cap, deltas are still emitted to consumers — only the fallback text buffer is skipped
- `readAsync()` now uses `readNBytes()` with a limit instead of `readAllBytes()`, preventing OOM from stderr
- The `ignoreFallback` flag cleanly separates real-time streaming from retained result text

**No issues found.** This is an important resilience improvement for production AI agent integration.

lizhimins:

PR #2098, opened 2026-08-13
fix(web): localize home banner text

yyqdbngt:
## What is the purpose of the change

Fix #2097.

The home page banner was hard-coded Chinese while the rest of the page uses i18n.

## Brief changelog

- translations.ts: add home.banner key
- home/index.tsx: render t('home.banner')

## How was this patch verified

- npx vitest run src/pages/home/__tests__/HomePage.test.tsx (3 passed)
- npx tsc -b clean
- npx eslint . clean

PR #2108, opened 2026-08-13
fix(web): clamp formatBytes units and guard non-finite input

### LINKED ISSUE #2064: formatBytes produces 'undefined' units for huge or non-finite byte values
## Problem

web/src/utils/format.ts formatBytes does not clamp the unit index and does not guard non-finite input. bytes >= 1024^6 yields units[i] === undefined and renders '1.2 undefined'; NaN/Infinity renders 'NaN undefined'.

## Expected

Clamp to the largest unit and return a dash for non-finite input.

yyqdbngt:

Fix #2064.

formatBytes did not clamp the unit index and did not guard non-finite input. bytes beyond PB rendered an undefined unit, and NaN/Infinity rendered "NaN undefined".

## Brief changelog

- formatBytes: guard non-finite input with '-', clamp to the largest unit with a division loop

- npx vitest run src/utils/format.test.ts (4 passed)
- npx tsc -b clean
- npx eslint . clean

### DIFF
--- a/web/src/utils/format.test.ts
+++ b/web/src/utils/format.test.ts
@@ -1,33 +1,26 @@
-/*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. The `Number.isFinite` guard and the division-loop clamping correctly fix both issues (NaN/Infinity rendering and out-of-bounds unit index).

One observation: this PR replaces the existing `formatDate`/`formatDateTime` tests with `formatBytes` tests. If the date formatter tests were intentionally removed (e.g., moved elsewhere), no action needed. Otherwise, consider restoring them to maintain coverage for those functions.

---

lizhimins:

PR #2109, opened 2026-08-13
fix(web): guard null alert level in system alerts

### LINKED ISSUE #2066: System alerts page crashes on null alert level
## Problem

web/src/pages/ops/systemAlerts.tsx calls alert.level.toLowerCase() but the fallback below assumes level can be empty. A null level throws before the fallback runs.

## Expected

Guard with (alert.level ?? '').toLowerCase().

yyqdbngt:

Fix #2066.

The system alerts page called normalizeAlertLevel(alert.level) which did alert.level.toLowerCase() without guarding null, crashing when a backend alert has no level.

## Brief changelog

- systemAlerts.tsx: normalizeAlertLevel guards with (level ?? '')

- npx vitest run src/pages/ops/__tests__/SystemAlertsPage.test.tsx (passed)
- npx tsc -b clean
- npx eslint . clean

### DIFF
--- a/web/src/pages/ops/systemAlerts.tsx
+++ b/web/src/pages/ops/systemAlerts.tsx
@@ -29,7 +29,7 @@ import type { SystemAlert } from '../../api/ops';
 
 const { Text } = Typography;
 
-const normalizeAlertLevel = (level: string) => level.toLowerCase();
+const normalizeAlertLevel = (level?: string | null) => (level ?? '').toLowerCase();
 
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
LGTM. Clean null-guard fix — using `(level ?? '')` correctly prevents the crash when `alert.level` is null/undefined.

---

lizhimins:

Issue #2116, opened 2026-08-13
[Studio][Bug] Do not recreate cloud credentials deleted during update


yuluo-yx:
CloudCredentialService loads a credential and then calls an upsert-style repository save. If another request deletes the credential between those operations, the update inserts it again and invalidates clients as though the update succeeded. Add an update-only repository operation, return 404 when no row was updated, and cover the race contract.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** `CloudCredentialService` loads a credential and then calls an upsert-style repository save. If another request deletes the credential between those operations (TOCTOU race), the update inserts it again and invalidates clients as though the update succeeded.

**Impact:** Deleted credentials can be silently recreated, causing stale credential references and unexpected client invalidation.

**Severity:** Medium — race condition that can cause inconsistent credential state.

**Proposed Fix:**
1. Add an update-only repository operation (no upsert)
2. Return 404 when no row was updated
3. Cover the race contract with focused tests

An automated fix proposal will be generated. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by github-manager-bot*

Issue #2107, opened 2026-08-13
[Guideline] Please group related changes into cohesive pull requests instead of many fragmented ones


lizhimins:
## What we are asking for

Please keep your pull requests **cohesive**: one PR should contain one complete, self-contained change (a feature, a bug fix, or a clearly scoped refactor) together with its tests. **Please do not split closely related changes into many tiny PRs** — for example, one PR per file, per page, per guard, or per single-line fix when all of them serve the same purpose.

## Why this matters

Recently the maintainers had to consolidate **70+ open PRs from a single contributor into 4 commits** before they could be merged. All of those PRs were valid improvements, but because each one covered only a single file or a one-line concern of the same theme (e.g. one PR per page for the same antd prop migration, one PR per null guard of the same pattern), they created:

- a large review and merge overhead for maintainers (conflict resolution, semantic merging, re-running full verification per batch);
- a cluttered PR queue that makes it harder for everyone's contributions to get noticed;
- history that does not read meaningfully after squash-merge.

Your changes are welcome and valuable — we just need them packaged in a reviewable shape.

## How to structure your PRs

- **Group by theme, not by file.** If you are applying the same fix/migration to 17 pages, that is **one PR** with 17 files, not 17 PRs.
- **Keep the fix and its tests together** in the same PR.
- A good PR size is typically from a few files up to a few dozen files, as long as they serve one purpose. Large features may be split into a small number of meaningful PRs (e.g. backend contract first, then UI), but each part should still be complete on its own.
- Before opening a PR, rebase onto the latest `rocketmq-studio` branch and make sure the build and tests pass locally.
- If you have already opened several fragmented PRs, feel free to merge their branches into one and let us know — we are happy to re-review the consolidated version.

Thanks for helping keep the review queue healthy — and thanks for your contributions!

RockteMQ-AI:
**Issue Evaluation**

Category: `guideline` | Status: **Acknowledged**

This is a project guideline from the maintainers about PR cohesion. No code changes are needed.

The guideline is clear: group related changes into cohesive PRs rather than many fragmented ones. Contributors should follow this when submitting changes.

---
*Automated evaluation by github-manager-bot*

PR #2126, opened 2026-08-13
[Studio] fix: prevent deleted cloud credential recreation

yuluo-yx:
## Summary\n\n- add an update-only cloud credential repository operation\n- return not found when the credential disappears concurrently\n- invalidate cloud clients only after an applied update\n- add focused service regression coverage\n\nCloses #2116\n\n## Verification\n\ngit diff --check passed. Maven compile was attempted but dependency resolution was blocked by Maven Central HTTP 403 before source compilation.\n\n## Base\n\nrocketmq-studio at 737a7b4d

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/credential/CloudCredentialRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/credential/CloudCredentialRepository.java
@@ -31,5 +31,7 @@ public interface CloudCredentialRepository {
 
     CloudCredentialVO save(CloudCredentialVO credential);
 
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
﻿### Summary
Prevents recreation of concurrently deleted cloud credentials by switching from `save()` to `replace()` with an existence check. The `update` method now throws a 404 if the credential was deleted between the read and write, rather than silently recreating it.

Good concurrency fix — the `replace()` method uses `updateById` which returns false if the row no longer exists, preventing the race condition.

---

lizhimins:

Issue #809, opened 2026-08-03
[Bug] LLM settings expose unsupported Azure and Bedrock providers


Aias00:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in GitHub Issues.
- [x] I have searched the existing issues and believe this is not a duplicate.
- [x] I have confirmed that this bug belongs to the current repository.

### Describe the Bug

Studio LLM Settings currently exposes Azure OpenAI and AWS Bedrock as selectable providers, but the active OpenAI-compatible gateway only supports openai, deepseek, tongyi, and ollama.

Code evidence:

- web/src/pages/studio/LlmSettings.tsx includes provider cards for azure and bedrock.
- server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java defines SUPPORTED_PROVIDERS as openai, deepseek, tongyi, and ollama.
- server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmGateway.java returns llm.config.unsupported_provider for unsupported provider values.

This creates a misleading user path: users can select Azure or Bedrock in the settings UI, but chat/model calls through the current gateway cannot succeed with those providers.

### Expected Behavior

The LLM provider settings UI should not allow users to select providers that are not supported by the currently wired gateway. Unsupported providers should either be hidden/disabled until implemented, or the backend should implement real provider-specific support.

### Scope

RocketMQ Studio Track 3 / AI Native LLM provider configuration.

### Proposed Fix

For the current OpenAI-compatible gateway phase, keep only supported providers selectable in the Studio LLM Settings page and add a frontend regression test that Azure/Bedrock are not exposed as selectable provider cards.

Aias00:
Closing this Track 3 / AI Native issue because I am narrowing subsequent work to Track 1 / Control Plane.

PR #2173, opened 2026-08-14
fix(ai): reject unsupported provider config tests

### LINKED ISSUE #2172: LLM config test reports success for unsupported HTTP providers
### What happened?

Testing an HTTP LLM configuration can report `Connection successful` even when the selected provider is not supported by the server's OpenAI-compatible client. Azure and Bedrock currently take this path: validation succeeds, `supports()` returns false, and the method falls through to a success result without making a request.

### Expected behavior

The test endpoint should return the same `llm.config.unsupported_provider` failure used by the runtime gateway, so the UI does not confirm a configuration that cannot serve requests.

### Reproduction

1. Select the HTTP engine and either the Azure or Bedrock provider.
2. Fill in the provider-specific required fields.
3. Test the configuration.
4. Observe a successful result even though the HTTP gateway rejects that provider.

btlqql:
## Summary

- fail HTTP configuration tests when the selected provider is unsupported by the active gateway
- return the same structured `llm.config.unsupported_provider` result used by runtime requests
- add a regression test while preserving valid Ollama connection tests

## Root cause

`testConfig` only called the provider probe when `llmClient.supports()` returned true. A false result fell through to `Connection successful` without any network request, so Azure and Bedrock configurations could appear usable even though the current OpenAI-compatible gateway rejects them.

This is narrower than the earlier provider-visibility work in #809/#810: it does not hide or remove provider-specific settings. It only keeps the connection-test response consistent with the runtime gateway.

Closes #2172.

## Validation

- `mvn -Dtest=LlmConfigServiceTest test` (27 tests passed)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
@@ -134,12 +134,16 @@ public class LlmConfigService {
         if (!LlmConfigVO.ENGINE_HTTP.equalsIgnoreCase(normalized.normalizeEngine())) {
             return testCliEngine(normalized.normalizeEngine());
         }
-        if (llmClient.supports(normalized)) {
-            try {
-                llmClient.listModels(normalized);
-            } catch (LlmGatewayException exception) {
-                return LlmOperationResultVO.failure(exception.getCode(), exception.getMessage(), exception.getHint());
-            }
+        if (!llmClient.supports(normalized)) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) (68 lines of diff).

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
server/src/test/java/org/apache/rocketmq/studio/ops/ai/LlmConfigServiceTest.java
```

---

lizhimins:

PR #2177, opened 2026-08-14
fix(tencent): skip null catalog instances

### LINKED ISSUE #2176: Tencent catalog fails a page when the SDK returns a null instance
### What happened?

### Expected behavior

### Reproduction

btlqql:
## Summary

- skip null records returned in Tencent Cloud instance pages
- continue mapping and filtering valid instances from the same response
- cover a mixed null/valid SDK response with a regression test

## Root cause

The Tencent SDK models the page payload as an `InstanceItem[]`, and the catalog loop assumed every array element was non-null. A single null record therefore raised a `NullPointerException` and discarded all valid instances in that page.

Closes #2176.

## Validation

- `mvn -f server/pom.xml -Dtest=TencentCatalogServiceTest test` (4 tests passed)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
@@ -94,6 +94,9 @@ public class TencentCatalogService implements CloudCatalogProvider {
                 break;
             }
             for (InstanceItem item : data) {
+                if (item == null) {
+                    continue;
+                }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) (41 lines of diff).

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
server/src/test/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogServiceTest.java
```

---

lizhimins:

PR #2179, opened 2026-08-14
fix(ai): surface SSE error envelopes

### LINKED ISSUE #2178: LLM streaming ignores error envelopes in successful HTTP responses
### What happened?

Some OpenAI-compatible providers report stream failures inside an HTTP 200 SSE response, for example `data: {"error":{"message":"quota exceeded"}}`. The server currently reads only `choices[0].delta.content`, treats the error event as an empty token, and can finish the stream without surfacing the provider failure.

### Expected behavior

Detect an explicit `error.message` SSE envelope and return a structured gateway error that preserves the provider message. Legal tokenless events such as usage metadata should remain ignored.

### Reproduction

Return the following from the configured provider and invoke streaming completion:

```
data: {"error":{"message":"quota exceeded"}}

data: [DONE]
```

The current call completes with no tokens instead of reporting the quota error.

btlqql:
## Summary

- detect explicit `error.message` objects in provider SSE data events
- preserve the upstream message in a structured gateway error
- keep valid tokenless events, including usage metadata, unchanged

## Root cause

The stream parser extracted only `choices[0].delta.content`. When a provider encoded a quota or credential failure inside an HTTP 200 SSE event, the missing delta became an empty string and the event was silently ignored.

This concerns content-level failures after a successful response header. It is separate from HTTP 4xx/5xx response-body handling, stream timeouts, size limits, and browser-side SSE framing.

Closes #2178.

## Validation

- `mvn -Dtest=OpenAiCompatibleLlmClientTest test` (19 tests passed)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
@@ -428,6 +428,12 @@ public class OpenAiCompatibleLlmClient {
     private String parseDelta(String body) {
         try {
             JsonNode root = objectMapper.readTree(body);
+            String errorMessage = root.path("error").path("message").asText();
+            if (StringUtils.hasText(errorMessage)) {
+                throw new LlmGatewayException(502, "llm.provider.stream_error",
+                        "LLM provider stream failed: " + errorMessage,
+                        "Check the provider credentials, model name, and account quota.");
+            }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) (49 lines of diff).

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
server/src/test/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClientTest.java
```

---

lizhimins:

PR #2186, opened 2026-08-14
fix(tencent): preserve partial ACL updates

### LINKED ISSUE #2185: Tencent ACL partial updates reject role ids and reset omitted permissions
### What happened?

Tencent ACL updates are routed from `UpdateAclUserDTO`, where only `id` is required and permission fields are nullable for partial updates. `TencentAclService.updateUser()` instead requires `username`, uses it as the role identifier, and converts every omitted permission to `true`.

This causes two failures:

- a valid update such as `{id, permRead:false, permWrite:true}` is rejected because `username` is absent;
- providing a username while omitting one permission silently enables that permission instead of preserving its current value.

### Expected behavior

Use the required role id when no replacement username is supplied, and preserve each existing Tencent role permission when that field is omitted from the partial update.

### Reproduction

Update an existing Tencent role using only its `id`, or omit either `permRead` or `permWrite`. The current backend rejects the first request and resets omitted permission values to true in the second.

btlqql:
## Summary

- resolve Tencent ACL updates with the required role id
- load the existing role only when a permission field is omitted
- preserve each omitted permission instead of defaulting it to true
- return the effective role and permission values after the update

## Root cause

`UpdateAclUserDTO` models a partial update: `id` is required while `username`, `permRead`, and `permWrite` are optional. The Tencent adapter required the optional username and converted null permissions to true, rejecting valid id-only requests and unexpectedly enabling omitted permissions.

Closes #2185.

## Validation

- `mvn -Dtest=TencentAclServiceTest test` (2 tests passed)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
@@ -150,27 +150,58 @@ public class TencentAclService {
 
     public AclUserVO updateUser(String instanceId, AclUserVO user) {
         Context context = resolve(instanceId);
-        if (!StringUtils.hasText(user.getUsername())) {
-            throw new BusinessException(400, "ACL username is required");
+        String roleName = StringUtils.hasText(user.getId()) ? user.getId() : user.getUsername();
+        if (!StringUtils.hasText(roleName)) {
+            throw new BusinessException(400, "ACL user id is required");
         }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) (193 lines of diff).

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
server/src/test/java/org/apache/rocketmq/studio/provider/tencent/TencentAclServiceTest.java
```

---

lizhimins:

PR #2189, opened 2026-08-14
fix(metrics): use availability for broker health

### LINKED ISSUE #2188: RocketMQ 5 broker health profile uses processor load instead of availability
### What happened?

The RocketMQ 5 native metrics profile maps semantic broker health to `rocketmq_processor_watermark`. That metric represents request-processor queue watermarks, not scrape or broker availability: a healthy idle broker can report zero and a busy broker can report a positive value. The profile therefore exposes load as an `up`-style health signal.

### Expected behavior

Use the Prometheus `up` series for the RocketMQ 5 broker-health semantic metric, consistent with the existing RocketMQ 4 profile and the semantic metric's availability meaning.

### Reproduction

Resolve `broker_health` from the `rocketmq5-native` metric profile. The generated query currently aggregates processor watermarks rather than checking whether the broker target is scrapeable.

yyqdbngt:
## Summary

- map RocketMQ 5 broker health to Prometheus scrape availability
- use the existing RocketMQ target filter and stable job/instance labels
- verify the metric name, PromQL, and labels in the profile test

## Root cause

`rocketmq_processor_watermark` measures request-processor queue watermarks. It is a load signal rather than a binary availability signal, so it cannot represent the `broker_health` semantic metric correctly. A healthy idle broker can have a zero watermark while a busy broker has a positive value.

Closes #2188.

## Validation

- `mvn -Dtest=MetricProfileServiceTest test` (7 tests passed)
- `git diff --check`
- final GraphQL scan of all open PR files targeting `rocketmq-studio`, including large-PR pagination (no path matches)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileService.java
@@ -100,9 +100,9 @@ public class MetricProfileService {
                 mapping(SemanticMetric.CONSUMER_LAG_LATENCY, "rocketmq_consumer_lag_latency",
                         "max(rocketmq_consumer_lag_latency) by (cluster, topic, consumer_group)",
                         "cluster", "topic", "consumer_group"),
-                mapping(SemanticMetric.BROKER_HEALTH, "rocketmq_processor_watermark",
-                        "max(rocketmq_processor_watermark) by (cluster, node_id, processor)",
-                        "cluster", "node_id", "processor")
+                mapping(SemanticMetric.BROKER_HEALTH, "up",
+                        "min(up{job=~\".*rocketmq.*\"}) by (job, instance)",
+                        "job", "instance")
         );
     }
 
--- a/server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileServiceTest.java
@@ -53,8 +53,12 @@ class MetricProfileServiceTest {
                         "sum(rate(rocketmq_messages_in_total[1m])) by (cluster, node_id)");
         assertThat(mapping(profile, SemanticMetric.CONSUMER_LAG_MESSAGES).getPrometheusMetric())
                 .isEqualTo("rocketmq_consumer_lag_messages");
-        assertThat(mapping(profile, SemanticMetric.BROKER_HEALTH).getPrometheusMetric())
-                .isEqualTo("rocketmq_processor_watermark");
+        assertThat(mapping(profile, SemanticMetric.BROKER_HEALTH))
+                .extracting(MetricProfileVO.MetricMappingVO::getPrometheusMetric,
+                        MetricProfileVO.MetricMappingVO::getPromql,
+                        MetricProfileVO.MetricMappingVO::getLabels)
+                .containsExactly("up", "min(up{job=~\".*rocketmq.*\"}) by (job, instance)",
+                        List.of("job", "instance"));
     }
 
     @Test


server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileService.java
```diff
@@ -100,9 +100,9 @@ private List<MetricProfileVO.MetricMappingVO> rocketmq5NativeMetrics() {
                 mapping(SemanticMetric.CONSUMER_LAG_LATENCY, "rocketmq_consumer_lag_latency",
                         "max(rocketmq_consumer_lag_latency) by (cluster, topic, consumer_group)",
                         "cluster", "topic", "consumer_group"),
-                mapping(SemanticMetric.BROKER_HEALTH, "rocketmq_processor_watermark",
-                        "max(rocketmq_processor_watermark) by (cluster, node_id, processor)",
-                        "cluster", "node_id", "processor")
+                mapping(SemanticMetric.BROKER_HEALTH, "up",
+                        "min(up{job=~\".*rocketmq.*\"}) by (job, instance)",
```
RockteMQ-AI, line 104: Backward-compatibility break: the label set changes from (cluster, node_id, processor) to (job, instance). Any downstream consumer — dashboards, alerts, or API clients — that reads BROKER_HEALTH results and keys on 'cluster' or 'node_id' will silently receive null/missing labels. The RocketMQ-domain identity (which cluster, which broker node) is replaced by Prometheus scrape-target identity (which scrape job, which endpoint). If this is acceptable, it should be called out in release notes; if not, consider keeping at least 'cluster' as a label or adding a relabeling rule in the Prometheus config.

RockteMQ-AI, line 104: The job-name regex `.*rocketmq.*` is broad. Any Prometheus scrape job whose name contains the substring 'rocketmq' (e.g. a sidecar exporter, a CI job, or a differently-purposed service) will be included in the health signal, potentially masking a real broker outage or producing false negatives. Consider tightening to a more specific pattern (e.g. `rocketmq-broker.*` or `rocketmq-exporter.*`) once the expected job names are documented, or make the filter configurable.

RockteMQ-AI, line 104: The `up` metric is scrape-level availability and does not distinguish between 'broker process is running but unhealthy' and 'broker endpoint is unreachable'. The old `rocketmq_processor_watermark` at least reflected internal broker state. If deeper liveness semantics are needed later (e.g. distinguishing a broker that is up but not accepting writes), this will need to be revisited. Documenting this trade-off in a code comment or the PR description would help future maintainers.

server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileServiceTest.java
```diff
@@ -53,8 +53,12 @@ void rocketmq5ProfileShouldUseNativeMetricNames() {
                         "sum(rate(rocketmq_messages_in_total[1m])) by (cluster, node_id)");
         assertThat(mapping(profile, SemanticMetric.CONSUMER_LAG_MESSAGES).getPrometheusMetric())
                 .isEqualTo("rocketmq_consumer_lag_messages");
-        assertThat(mapping(profile, SemanticMetric.BROKER_HEALTH).getPrometheusMetric())
-                .isEqualTo("rocketmq_processor_watermark");
+        assertThat(mapping(profile, SemanticMetric.BROKER_HEALTH))
```
RockteMQ-AI, line 56: The test verifies metric name, PromQL, and labels for the RocketMQ 5 native profile, which is good. However, there is no assertion confirming that other profiles (e.g. RocketMQ 4 / fallback) do NOT use the `up` metric, and no test for the semantic metric ordering or total count in the profile. Adding a negative assertion or a profile-size check would guard against accidental duplication.

RockteMQ-AI (commented):
### Summary
Fix correctly maps BROKER_HEALTH to Prometheus `up` availability metric, but the label-set change (cluster/node_id → job/instance) is a backward-compatibility break for downstream consumers and the broad `.*rocketmq.*` job regex risks matching unintended targets.

---

lizhimins:
Thanks for the contribution. After discussion we're closing this one: deriving broker health from observability scrape availability isn't the right design — broker status should reflect the broker's own runtime state, not whether a metrics/observability pipeline happens to be reachable or configured. When no Prometheus endpoint is configured, this change would mark healthy brokers as unhealthy. A health signal based on cluster/broker runtime info would be welcome instead.

PR #2191, opened 2026-08-14
fix(provider): parse long jstack thread names

### LINKED ISSUE #2190: Consumer diagnostics omit threads with long names
### What happened?

The Apache RocketMQ consumer diagnostics parser requires whitespace between a thread name and the `TID:` marker. RocketMQ formats jstack headers with a minimum-width thread-name field (`%-40sTID...`), so names that are 40 characters or longer are immediately followed by `TID:` with no whitespace. Those threads are not recognized and disappear from diagnostics.

### Expected behavior

Parse the stable `TID:` boundary without assuming padding whitespace, while preserving the complete thread name, id, state, and stack frames.

### Reproduction

Feed consumer diagnostics a RocketMQ jstack header whose thread name contains at least 40 characters. The current header regex does not match the line and omits the thread.

yyqdbngt:
## Summary

- parse RocketMQ jstack headers at the stable `TID:` marker
- support thread names that consume or exceed the 40-character display column
- verify the full thread name, id, state, and frame list

## Root cause

RocketMQ formats thread names with `%-40s`. Short names receive padding, but names of 40 characters or more are immediately followed by `TID:`. The diagnostics regex required at least one whitespace character at that boundary, so those threads were omitted.

Closes #2190.

## Validation

- `mvn -Dtest=RocketMQConsumerDiagnosticsProviderTest test` (6 tests passed)
- `git diff --check`
- final GraphQL scan of all open PR files targeting `rocketmq-studio`, including large-PR pagination (no path matches)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQConsumerDiagnosticsProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQConsumerDiagnosticsProvider.java
@@ -49,7 +49,7 @@ import java.util.regex.Pattern;
 public class RocketMQConsumerDiagnosticsProvider implements ConsumerDiagnosticsProvider {
 
     private static final Pattern THREAD_HEADER =
-            Pattern.compile("^(?<name>.+?)\\s+TID:\\s+(?<id>\\d+)\\s+STATE:\\s+(?<state>\\S+)\\s*$");
+            Pattern.compile("^(?<name>.+?)TID:\\s+(?<id>\\d+)\\s+STATE:\\s+(?<state>\\S+)\\s*$");
 
     private final RuntimeAdminClientResolver runtimeAdminClientResolver;
     private final MqAdminExtFactory adminFactory;
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQConsumerDiagnosticsProviderTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQConsumerDiagnosticsProviderTest.java
@@ -99,6 +99,25 @@ class RocketMQConsumerDiagnosticsProviderTest {
     }
 
     @Test
+    void getConsumerStackShouldParseThreadNamesLongerThanJstackColumnWidth() throws Exception {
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQConsumerDiagnosticsProvider.java
```diff
@@ -49,7 +49,7 @@
 public class RocketMQConsumerDiagnosticsProvider implements ConsumerDiagnosticsProvider {
 
     private static final Pattern THREAD_HEADER =
-            Pattern.compile("^(?<name>.+?)\\s+TID:\\s+(?<id>\\d+)\\s+STATE:\\s+(?<state>\\S+)\\s*$");
+            Pattern.compile("^(?<name>.+?)TID:\\s+(?<id>\\d+)\\s+STATE:\\s+(?<state>\\S+)\\s*$");
```
RockteMQ-AI, line 52: Removing the required whitespace before TID: causes the `name` capture group to include the %-40s padding spaces for short thread names. The previous regex kept the name trimmed, but now short names can be returned with trailing spaces, which may break exact-name consumers or display formatting. Consider using `(?<name>.+?)\s*TID:` and trimming the captured name so both short (padded) and long (unpadded) names are parsed cleanly.

server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQConsumerDiagnosticsProviderTest.java
```diff
@@ -98,6 +98,25 @@ void getConsumerStackShouldUseSelectedInstanceAndParseJstack() throws Exception
         verify(adminFactory, never()).execute(anyString(), any(), any());
     }
 
+    @Test
```
RockteMQ-AI, line 101: The new test only verifies a thread name longer than the 40-character jstack column. Add regression cases for short padded names and exactly-40-character names to ensure the regex change does not alter the parsed thread name for those cases.

RockteMQ-AI (commented):
### Summary
Fixes the jstack thread-header regex to match long thread names that abut the TID: marker, but the removal of required whitespace may now capture padding spaces for short names; tests should also cover short and exactly-40-character names.

---

PR #2193, opened 2026-08-14
fix(alerting): export assets as attachments

### LINKED ISSUE #2192: Alert-rule asset export uses a form-data content disposition
### What happened?

Alert-rule asset export builds the response header with `setContentDispositionFormData`. That produces a `form-data` disposition intended for a multipart form part, even though the endpoint is a normal file download. Browsers and API clients may therefore fail to treat the YAML response as an attachment.

### Expected behavior

Return a standards-compliant `Content-Disposition: attachment` header with the UTF-8 YAML filename.

### Reproduction

Export an alert-rule asset and inspect the response headers. The current disposition begins with `form-data` instead of `attachment`.

yyqdbngt:
## Summary

- return alert rule YAML exports with an attachment content disposition
- preserve the generated filename in the download response
- update the controller regression test for the expected header

## Root cause

`setContentDispositionFormData` emits multipart form-data semantics intended for a multipart body part. This endpoint returns a normal file response, so clients could fail to recognize the YAML as a download.

Closes #2192

## Validation

- `mvn -pl server -Dtest=AlertRuleAssetControllerTest test` (3 tests)
- `mvn -pl server checkstyle:check`
- `git diff --check`
- verified the changed paths against all open PRs targeting `rocketmq-studio`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleAssetController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleAssetController.java
@@ -20,6 +20,7 @@ import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import lombok.RequiredArgsConstructor;
 import org.apache.rocketmq.studio.common.domain.Result;
+import org.springframework.http.ContentDisposition;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes alert rule YAML export to use a proper attachment Content-Disposition header instead of incorrect multipart form-data semantics; the change is correct, the test expectation is updated, and no issues were found. LGTM.

---

lizhimins:
Thanks for the contribution. We're closing this one: switching the alert asset export to an attachment disposition doesn't bring a meaningful benefit for this feature, so we don't think it's worth taking. Thanks for understanding.

PR #2196, opened 2026-08-14
fix(tencent): coordinate client creation and invalidation

### LINKED ISSUE #2195: [Bug] Tencent client invalidation can miss in-flight creation
## Problem

`TencentClientFactory` creates cached SDK clients with `ConcurrentHashMap.computeIfAbsent`, while credential updates invalidate cached entries by scanning the map keys.

If invalidation runs while a client is still being created, the invalidation can finish before the new mapping is visible. The client built from the previous credential is then inserted after invalidation and remains cached for later requests.

## Expected behavior

Once credential invalidation returns, an in-flight cache miss must not be able to install a client created before that invalidation. Existing cache hits should remain inexpensive.

## Scope

Coordinate cache-miss creation with credential invalidation and add a deterministic concurrency regression test.

yyqdbngt:
## Summary

- coordinate Tencent SDK client cache misses with credential invalidation
- keep existing cache hits on the lock-free path
- add a deterministic concurrency regression test for an in-flight client creation

## Root cause

A computeIfAbsent mapping is not available to the invalidation key scan until creation completes. Invalidation could therefore return while an old-credential client was still being built, and that client would then be installed in the cache afterward.

Closes #2195

## Validation

- mvn -Dtest=TencentClientFactoryTest test (4 tests)
- mvn -DskipTests checkstyle:check (0 violations)
- git diff --check
- verified both changed paths against all open PRs targeting 
ocketmq-studio
- audited full PR history for the configured contributor accounts

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactory.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactory.java
@@ -48,10 +48,16 @@ public class TencentClientFactory {
 
     public TrocketClient client(Long credentialId, String region) {
         String key = cacheKey(credentialId, region);
-        return clients.computeIfAbsent(key, ignored -> createClient(credentialId, region));
+        TrocketClient cached = clients.get(key);
+        if (cached != null) {
+            return cached;
+        }
+        synchronized (this) {
+            return clients.computeIfAbsent(key, ignored -> createClient(credentialId, region));
+        }
     }
 
-    public void invalidateCredential(Long credentialId) {
+    public synchronized void invalidateCredential(Long credentialId) {
         String prefix = credentialId + "#";
         clients.keySet().removeIf(key -> key.startsWith(prefix));
     }
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactoryTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactoryTest.java
@@ -16,9 +16,17 @@
  */
 package org.apache.rocketmq.studio.provider.tencent;
 
+import com.tencentcloudapi.trocket.v20230308.TrocketClient;
+import org.apache.rocketmq.studio.provider.credential.CloudCredentialRepository;
 import org.junit.jupiter.api.Test;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
 
 class TencentClientFactoryTest {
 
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactory.java
```diff
@@ -48,10 +48,16 @@ public class TencentClientFactory {
 
     public TrocketClient client(String credentialId, String region) {
         String key = cacheKey(credentialId, region);
-        return clients.computeIfAbsent(key, ignored -> createClient(credentialId, region));
+        TrocketClient cached = clients.get(key);
+        if (cached != null) {
+            return cached;
+        }
+        synchronized (this) {
```
RockteMQ-AI, line 55: Cache misses now synchronize on the factory instance, serializing all client creation across every credential/region. If this path is heavily contended, consider a finer-grained lock (e.g., per-key or a read/write lock) instead of a global monitor.

server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactory.java
```diff
@@ -48,10 +48,16 @@ public class TencentClientFactory {
 
     public TrocketClient client(String credentialId, String region) {
         String key = cacheKey(credentialId, region);
```
RockteMQ-AI, line 50: The lock-free fast-path get can return a cached client while invalidateCredential is concurrently removing it, leaving a brief stale-read window. This appears intentional to preserve hit performance; confirm it is acceptable for the credential-rotation semantics.

server/src/test/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactoryTest.java
```diff
@@ -45,4 +53,62 @@ void endpointForShouldUseRegionalEndpointForFinancialRegionsTest() {
...
+        invalidationThread.join(TimeUnit.SECONDS.toMillis(5));
+
+        assertThat(invalidationThread.isAlive()).isFalse();
+        assertThat(factory.client("credential-a", "ap-shanghai")).isNotSameAs(firstClient);
+        assertThat(factory.creationCount).hasValue(2);
+    }
+
+    private static void awaitBlockedOrTerminated(Thread thread) {
+        long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(5);
+        while (thread.getState() != Thread.State.BLOCKED && thread.isAlive()
```
RockteMQ-AI, line 84: awaitBlockedOrTerminated relies on Thread.getState() and spin-waiting to observe monitor contention. While deterministic in this controlled test, assertions based on thread states can be brittle under different JVM schedulers.

RockteMQ-AI (commented):
### Summary
Fixes the Tencent client cache race by synchronizing in-flight creation with credential invalidation while keeping cache hits lock-free; the change is correct and well-tested, with minor global-lock scalability tradeoffs to monitor.

---

PR #2175, opened 2026-08-14
fix(metrics): validate Prometheus duration order

btlqql:
## Summary

- require suffixed Prometheus duration parts to use integer values
- enforce longest-to-shortest unit order and reject repeated units
- retain support for bare numeric seconds and valid combined durations

## Root cause

The local step parser summed every matched duration part independently. As a result, inputs such as `1s1h`, `1m1m`, and `1.5h` passed local validation even though Prometheus rejects them, moving a client error downstream to the data source.

Closes #2174.

## Validation

- `mvn -Dtest=MetricsServiceTest test` (22 tests passed)
- Maven Checkstyle validation (0 violations)
- `git diff --check`
- both changed paths checked against all open PRs targeting `rocketmq-studio` (no matches)

The accepted grammar follows the Prometheus duration rule that combined units are ordered longest to shortest and each unit appears at most once.

PR #2197, opened 2026-08-14
fix(server): harden external provider boundaries

### LINKED ISSUE #2174: Metrics validation accepts invalid Prometheus duration steps
### What happened?

The metrics query validator accepts duration steps that Prometheus rejects, including units in ascending order (`1s1h`), repeated units (`1m1m`), and fractional suffixed durations (`1.5h`). The local parser sums every matched part, so these values pass validation and only fail later at the configured Prometheus/Thanos endpoint.

### Expected behavior

Validate duration steps with Prometheus's duration grammar before issuing the query: integer duration parts, units ordered from longest to shortest, and each unit appearing at most once. Bare numeric steps should continue to represent seconds.

### Reproduction

Submit a range metrics query with `step` set to `1s1h` or `1m1m`. The backend accepts the request even though the downstream query API rejects the step syntax.

### LINKED ISSUE #2176: Tencent catalog fails a page when the SDK returns a null instance

### LINKED ISSUE #2178: LLM streaming ignores error envelopes in successful HTTP responses

### LINKED ISSUE #2185: Tencent ACL partial updates reject role ids and reset omitted permissions

btlqql:
## Summary

- validate Prometheus range-query steps with the supported ordered duration grammar
- surface explicit provider error envelopes received inside successful SSE responses
- skip malformed null records without discarding valid Tencent catalog results
- preserve omitted Tencent ACL permissions during id-based partial updates

## Why these fixes are grouped

This consolidates the previously closed PRs #2175, #2177, #2179, and #2186 in response to contribution guideline #2107. The changes form one backend hardening pass for data crossing external provider boundaries, and each behavior is submitted together with focused regression coverage.

## Root cause and impact

Several provider adapters trusted external syntax or payload shape too early: the metrics validator accepted duration forms rejected downstream, the SSE parser ignored structured provider failures, the Tencent catalog assumed every SDK item was non-null, and Tencent ACL partial updates replaced omitted permission values. These cases now fail predictably or preserve valid data instead of producing downstream errors, hiding provider failures, dropping complete pages, or changing ACL permissions unexpectedly.

## Validation

- `mvn -B -ntp -f server/pom.xml -Dtest=MetricsServiceTest,OpenAiCompatibleLlmClientTest,TencentAclServiceTest,TencentCatalogServiceTest test` — 47 tests passed
- `mvn -B -ntp -f server/pom.xml -Dtest=!CliAgentProviderTest,!ClaudeCodeAgentProviderTest test` — 1198 tests passed
- Maven Checkstyle — 0 violations
- `git diff --check`

Closes #2174.
Closes #2176.
Closes #2178.
Closes #2185.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
@@ -40,7 +40,7 @@ public class MetricsService {
     private static final long MAX_RANGE_SECONDS = 31L * 24 * 60 * 60;
     private static final long MAX_SAMPLE_POINTS = 11_000L;
     private static final Pattern NUMBER_PATTERN = Pattern.compile("\\d+(?:\\.\\d+)?");
-    private static final Pattern DURATION_PART_PATTERN = Pattern.compile("(\\d+(?:\\.\\d+)?)(ms|s|m|h|d|w|y)");
+    private static final Pattern DURATION_PART_PATTERN = Pattern.compile("(\\d+)(ms|s|m|h|d|w|y)");
     private static final Map<String, BigDecimal> UNIT_TO_MILLIS = Map.of(
             "ms", BigDecimal.ONE,
             "s", BigDecimal.valueOf(1_000L),
@@ -49,6 +49,15 @@ public class MetricsService {
             "d", BigDecimal.valueOf(86_400_000L),
             "w", BigDecimal.valueOf(604_800_000L),
             "y", BigDecimal.valueOf(31_536_000_000L)
+    );
+    private static final Map<String, Integer> UNIT_ORDER = Map.of(
+            "y", 0,
+            "w", 1,
+            "d", 2,
+            "h", 3,
+            "m", 4,
+            "s", 5,
+            "ms", 6
     );
 
     private final MetricsSource metricsSource;
@@ -188,13 +197,20 @@ public class MetricsService {
         Matcher matcher = DURATION_PART_PATTERN.matcher(value);
         BigDecimal millis = BigDecimal.ZERO;
         int position = 0;
+        int previousUnitOrder = -1;
         while (matcher.find()) {
             if (matcher.start() != position) {
                 throw badRequest("Metric query step is invalid");
             }
+            String unit = matcher.group(2);
+            int unitOrder = UNIT_ORDER.get(unit);
+            if (unitOrder <= previousUnitOrder) {
+                throw badRequest("Metric query step is invalid");
+            }
             BigDecimal amount = new BigDecimal(matcher.group(1));
-            millis = millis.add(amount.multiply(UNIT_TO_MILLIS.get(matcher.group(2))));
+            millis = millis.add(amount.multiply(UNIT_TO_MILLIS.get(unit)));
             position = matcher.end();
+            previousUnitOrder = unitOrder;
         }
         if (position != value.length()) {
             throw badRequest("Metric query step is invalid");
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
```diff
@@ -40,7 +40,7 @@ public class MetricsService {
     private static final long MAX_RANGE_SECONDS = 31L * 24 * 60 * 60;
     private static final long MAX_SAMPLE_POINTS = 11_000L;
     private static final Pattern NUMBER_PATTERN = Pattern.compile("\\d+(?:\\.\\d+)?");
-    private static final Pattern DURATION_PART_PATTERN = Pattern.compile("(\\d+(?:\\.\\d+)?)(ms|s|m|h|d|w|y)");
+    private static final Pattern DURATION_PART_PATTERN = Pattern.compile("(\\d+)(ms|s|m|h|d|w|y)");
```
RockteMQ-AI, line 43: The regex change from `(\d+(?:\.\d+)?)` to `(\d+)` means fractional durations like `1.5h` or `2.5s` are now rejected. This is a **breaking change** for any existing callers that relied on fractional values. The tests confirm this is intentional — just flagging it for awareness.

server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
```diff
@@ -188,13 +197,20 @@ private BigDecimal parseStepMillis(String step) {
         Matcher matcher = DURATION_PART_PATTERN.matcher(value);
         BigDecimal millis = BigDecimal.ZERO;
         int position = 0;
+        int previousUnitOrder = -1;
         while (matcher.find()) {
             if (matcher.start() != position) {
                 throw badRequest("Metric query step is invalid");
             }
```
RockteMQ-AI, line 204: Good improvement — enforcing canonical unit ordering (e.g. `1h30m` is valid, `30m1h` is rejected) prevents ambiguous duration expressions. The `UNIT_ORDER` map is clean.

RockteMQ-AI (approved):
/tmp/pr_review_2197/summary.txt

- **[INFO]** `server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java:74` — Good defensive coding — checking for error envelopes in SSE stream responses prevents silent failures when the upstream provider returns an error mid-stream. _(line outside diff)_
- **[INFO]** `server/src/main/java/org/apache/rocketmq/studio/common/provider/tencent/TencentCatalogService.java:143` — Good null guard — prevents NPE if the Tencent API returns a null item in the array. _(line outside diff)_
- **[INFO]** `server/src/main/java/org/apache/rocketmq/studio/common/provider/tencent/TencentAclService.java:128` — The fallback to `id` when `username` is omitted, and the logic to preserve existing permissions when `permissions` is null in the update request, are both well-handled. The `findRole()` helper with pagination is a nice touch. _(line outside diff)_

PR #2198, opened 2026-08-14
fix(provider): avoid topic consumer pagination overflow

### LINKED ISSUE #2187: Cloud topic-consumer pagination overflows on large page numbers
### What happened?

The default cloud topic-consumer pagination calculates `(page - 1) * pageSize` as an `int`. Both values pass request validation independently, but a request such as `page=2147483647&pageSize=100` overflows the offset to a negative value and reaches `subList`, producing an `IndexOutOfBoundsException` and a 500 response.

### Expected behavior

Calculate the offset without integer overflow. Pages beyond the available consumer list should return an empty page with the correct total instead of failing the request.

### Reproduction

Call the topic-consumer listing endpoint for a cloud-backed provider with `page=Integer.MAX_VALUE` and `pageSize=100`. The default `InstanceProvider` pagination path computes a negative start index.

btlqql:
## Summary
- calculate cloud topic-consumer offsets with long arithmetic
- return an empty page for page numbers beyond the available list
- add regression coverage for an offset beyond the int range

## Root cause
The default provider pagination multiplied `(page - 1) * pageSize` as an `int`. Individually valid request values could wrap the offset negative before `subList`.

Closes #2187

## Validation
- `mvn -Dtest=InstanceProviderTest test` (from `server`)
- `mvn checkstyle:check` (from `server`)
- `git diff --check`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.provider;
 
 import org.apache.rocketmq.studio.common.domain.enums.InstanceVendor;
+import org.apache.rocketmq.studio.common.util.Pagination;
 import org.apache.rocketmq.studio.instance.group.ConsumerGroupVO;
 import org.apache.rocketmq.studio.instance.group.QueueProgressVO;
 import org.apache.rocketmq.studio.instance.group.SubscriptionEntryVO;
@@ -54,8 +55,9 @@ public interface InstanceProvider {
     default TopicConsumerPageVO getTopicConsumersPage(String instanceId, String topicName, int page, int pageSize) {
         List<TopicConsumerVO> consumers = getTopicConsumers(instanceId, topicName);
         int total = consumers.size();
-        int from = Math.min((page - 1) * pageSize, total);
-        int to = Math.min(from + pageSize, total);
+        long offset = Pagination.pageOffset(page, pageSize);
+        int from = (int) Math.min(offset, total);
+        int to = from + (int) Math.min(pageSize, total - from);
         return TopicConsumerPageVO.builder()
                 .items(consumers.subList(from, to))
                 .total(total)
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
```diff
@@ -54,8 +54,9 @@ public interface InstanceProvider {
     default TopicConsumerPageVO getTopicConsumersPage(String instanceId, String topicName, int page, int pageSize) {
         List<TopicConsumerVO> consumers = getTopicConsumers(instanceId, topicName);
         int total = consumers.size();
-        int from = Math.min((page - 1) * pageSize, total);
-        int to = Math.min(from + pageSize, total);
+        long offset = ((long) page - 1) * pageSize;
```
RockteMQ-AI: Good fix — `(page - 1) * pageSize` overflows to a negative number when `page` is large (e.g. `Integer.MAX_VALUE`), causing `subList()` to throw `IndexOutOfBoundsException`. Casting to `long` first is the correct approach, and the subsequent `Math.min(offset, total)` ensures the result fits back into `int` safely.

server/src/test/java/org/apache/rocketmq/studio/provider/InstanceProviderTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class InstanceProviderTest {
+
+    @Test
+    public void getTopicConsumersPageShouldHandleLargePageNumberTest() {
+        InstanceProvider provider = mock(InstanceProvider.class);
+        when(provider.getTopicConsumers("instance-a", "orders")).thenReturn(Arrays.asList(
```
RockteMQ-AI, line 33: Well-structured test — using `Integer.MAX_VALUE` as the page number directly exercises the overflow path. The assertions verify both the empty result and the correct total count.

RockteMQ-AI (approved):
/tmp/pr_review_2198/summary.txt

PR #2160, opened 2026-08-13
[Studio] fix: report DLQ resend hard-cap truncation

yuluo-yx:
## Summary

- mark DLQ scans truncated when the 5000-message hard cap is reached
- treat truncated scans as incomplete and classify the outcome as PARTIAL
- include scanTruncated in audit details without changing the public result schema

Fixes #2155

## Validation

- RocketMQDLQProviderTest and DLQServiceTest: 17 passed
- Maven Checkstyle passed
- scope check: 63 changed lines

## Base

rocketmq-studio at 737a7b4d8b6ddf53d4c6dde581e821e6eac66529

PR #2157, opened 2026-08-13
[Studio] fix: order metrics samples by timestamp

yuluo-yx:
## Summary

- stably sort finite metric samples by timestamp
- draw chart points from oldest to newest
- derive the displayed latest value from the chronologically latest sample

Fixes #2152

## Validation

- MetricsExplorer.test.tsx: 11 passed
- targeted Prettier and ESLint passed
- scope check: 29 changed lines

## Base

rocketmq-studio at 737a7b4d8b6ddf53d4c6dde581e821e6eac66529

PR #2146, opened 2026-08-13
[Studio] fix: honor metrics authentication mode

yuluo-yx:
## Summary

- propagate authType through the dynamic metrics source factory
- apply none, basic, and bearer without credential-based guessing
- reject missing credentials and unsupported modes before sending requests

Fixes #2141

## Validation

- MetricsServiceTest, MultiBackendMetricsSourceTest, and PrometheusMetricsSourceTest: 55 passed
- Maven Checkstyle: 0 violations
- scope check: 117 changed lines

## Compatibility

The static Prometheus configuration path keeps its legacy credential inference for compatibility.

## Base

rocketmq-studio at 737a7b4d8b6ddf53d4c6dde581e821e6eac66529

PR #2199, opened 2026-08-14
[ISSUE #2141] Fix runtime observability and result accuracy

### LINKED ISSUE #2141: [Studio] Metrics data source authentication mode is ignored
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

MetricsService normalizes the configured authentication type, but MetricsSourceFactory and MetricsSourceSettings do not carry it to the Prometheus-compatible client. The request code instead infers authentication from whichever credential field is populated.

This can send an Authorization header for auth type none, and can choose Bearer when auth type basic is configured but a stale token is also present.

## Expected behavior

Dynamic metrics data sources should apply none, basic, or bearer strictly according to authType and reject missing credentials or unknown modes with a clear service error.

## Reproduction

Create metrics data sources with none plus populated credentials, or basic plus both basic credentials and a bearer token, then inspect the outbound Authorization header.

### LINKED ISSUE #2152: [Studio] Metrics Explorer assumes provider samples are time ordered
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

Metrics Explorer filters and converts samples without sorting them. The chart connects points in response order and treats the final array element as the latest value. A provider response with timestamps 200 then 100 draws a backward line and displays the older sample as latest.

## Expected behavior

Finite samples should be ordered stably by timestamp before chart geometry and latest-value calculations.

## Reproduction

Return two samples in descending timestamp order and compare the legend latest value and polyline x coordinates.

### LINKED ISSUE #2155: [Studio] DLQ resend hard-cap truncation is reported as complete success
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

The Apache DLQ provider stops scanning after the 5000-message resend hard cap, but scanIncomplete only checks failed queues. When more than 5000 messages exist and the first 5000 resend successfully, the result is SUCCESS with scanIncomplete false even though messages were omitted.

## Expected behavior

Reaching the hard cap should mark the scan as truncated and incomplete, classify the result as PARTIAL, and include the truncation fact in the audit details.

## Reproduction

Provide 5001 eligible DLQ messages with successful resend outcomes and inspect the returned status and audit record.

yuluo-yx:

This PR consolidates three related runtime accuracy fixes, following the maintainer guidance in #2107. It replaces the closed one-off PRs #2146, #2157, and #2160 without reopening them.

Dynamic metrics sources currently lose their configured authentication mode and infer it from whichever credentials are non-empty. This can send an Authorization header for `none` or choose Bearer when `basic` was configured. The Metrics Explorer also assumes provider samples are already ordered, which can draw a backwards line and report an older sample as the latest value. Finally, a DLQ resend that reaches the 5,000-message safety cap is reported as a complete success even though messages remain unprocessed.

## Brief changelog

- propagate and strictly apply the configured metrics authentication mode, while retaining compatibility for static Prometheus configuration
- sort numeric metric samples by timestamp before charting and selecting the latest value
- mark hard-cap DLQ scans as truncated, return a partial outcome, and record truncation in the audit detail
- add regression coverage for authentication headers, unordered samples, and 5,001-message DLQ scans

## Verifying this change

- `mvn -Dmaven.compiler.proc=full -Dtest=MetricsServiceTest,MultiBackendMetricsSourceTest,PrometheusMetricsSourceTest,RocketMQDLQProviderTest,DLQServiceTest test` (74 tests passed; dependencies resolved from the local Maven cache)
- `npm test -- --run src/components/__tests__/MetricsExplorer.test.tsx` (12 tests passed)
- ESLint and Prettier checks passed for the changed frontend files
- `npm run build` passed
- `git diff --check` passed

Fixes #2141
Fixes #2152
Fixes #2155

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/AbstractPrometheusCompatibleMetricsSource.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/AbstractPrometheusCompatibleMetricsSource.java
@@ -42,6 +42,7 @@ import java.net.URI;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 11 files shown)


RockteMQ-AI (approved):
### Summary
Important runtime observability fixes addressing authentication mode handling and metrics accuracy:

1. **Explicit auth mode dispatch** — Replaces credential-inference logic with an explicit `switch` on `authType` (`none`/`basic`/`bearer`). This prevents the bug where `none` mode could still send an Authorization header if credentials were non-empty, or where `basic` could be chosen when `bearer` was configured. The `none` case explicitly ignores stale credentials — correct behavior.
2. **Factory validation** — `MetricsSourceFactory` now validates auth completeness at creation time rather than at first request, failing fast with clear error messages.
3. **DLQ provider auth** — Ensures DLQ scan consumers and resend producers also use the correct credential hook from the instance configuration.
4. **Metrics sample sorting** — `toNumericSamples` now sorts by timestamp (with index as tiebreaker) before rendering, preventing chart artifacts from out-of-order data points.

Tests cover all auth modes including the `none`-with-stale-credentials edge case. The switch-based approach is more maintainable than the previous if-chain inference.

LGTM.

---

PR #2158, opened 2026-08-13
[Studio] fix: use instance IDs in NameServer drift selector

yuluo-yx:
## Summary

- use instance IDs as Select option values
- keep instance names as user-facing labels
- send the selected ID consistently to cluster and drift APIs

Fixes #2153

## Validation

- NameServerConfigDriftPage.test.tsx: 7 passed
- targeted Prettier and ESLint passed
- scope check: 32 changed lines

## Base

rocketmq-studio at 737a7b4d8b6ddf53d4c6dde581e821e6eac66529

PR #2156, opened 2026-08-13
[Studio] fix: clear stale BrokerCluster topology on instance change

yuluo-yx:
## Summary

- clear Broker, NameServer, and Proxy rows when instance scope changes
- clear the active scope when its cluster request fails
- preserve request sequencing so older responses cannot repopulate stale data

Fixes #2151

## Validation

- BrokerCluster.test.tsx: 14 passed
- npm run build passed
- targeted Prettier and ESLint passed
- scope check: 100 changed lines

## Base

rocketmq-studio at 737a7b4d8b6ddf53d4c6dde581e821e6eac66529

PR #2147, opened 2026-08-13
[Studio] fix: preserve Broker config editor integrity

yuluo-yx:
## Summary

- round-trip fractional MiB message sizes without rewriting unrelated configuration
- enforce equal read and write queue counts
- constrain Broker integer settings to integer input and handle form rejection

Fixes #2142

## Validation

- ClusterPage.test.tsx: 16 passed
- targeted Prettier and ESLint passed
- scope check: 145 changed lines

## Base

rocketmq-studio at 737a7b4d8b6ddf53d4c6dde581e821e6eac66529

PR #2201, opened 2026-08-14
[ISSUE #2142] Fix cluster administration UI integrity

### LINKED ISSUE #2142: [Studio] Broker config editor changes valid values and accepts mismatched queues
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

Opening the Broker configuration editor rounds maxMessageSize to a whole MiB. Saving an unrelated field therefore rewrites a valid fractional-MiB value. The page also states that read and write queue counts must match, but no cross-field validation enforces that contract, and integer settings accept fractional input.

## Expected behavior

The editor should round-trip the existing byte value without mutation, constrain integer settings to integer input, and prevent submission when read and write queue counts differ.

## Reproduction

Load a Broker with maxMessageSize set to 4.5 MiB, change only flushDiskType, and save. Also try readQueueNums 4 with writeQueueNums 8.

### LINKED ISSUE #2151: [Studio] BrokerCluster keeps the previous instance topology while loading
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

After instance A has loaded, selecting instance B starts a new cluster request but leaves A Broker, NameServer, and Proxy rows visible. If the B request fails, the old A topology remains indefinitely while the selector shows B. Operators can therefore read or act on data from the wrong instance context.

## Expected behavior

Changing instance scope should clear all topology immediately, and an active request failure should leave an empty current scope rather than restore or retain previous-instance data.

## Reproduction

Load BrokerCluster for instance A, select instance B, and keep the B request pending or reject it.

### LINKED ISSUE #2153: [Studio] NameServer drift selector mixes instance names and IDs
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

The NameServer configuration drift page initializes selectedInstanceId with Instance.id, but builds Select option values from Instance.name. The initial value is outside the option domain, so the selector can show a raw ID. After selection, the page also passes a name through variables and APIs that represent an instance ID.

## Expected behavior

The selector and all downstream drift requests should use Instance.id consistently while displaying Instance.name as the label.

## Reproduction

Load two instances whose IDs differ from their names, inspect the initial selector label, then select the second instance and inspect request arguments.

yuluo-yx:

This PR consolidates three related cluster administration UI integrity fixes, following the maintainer guidance in #2107. It replaces the closed one-off PRs #2147, #2156, and #2158 without reopening them.

The broker configuration editor can silently round valid message-size values and submit mismatched read/write queue counts. The BrokerCluster page can display topology from the previously selected instance while a new request is loading or after it fails. The NameServer drift selector also mixes instance names and IDs, leaving its controlled value outside the option domain.

## Brief changelog

- preserve exact broker message-size values and validate paired queue counts before submission
- clear instance-scoped topology during instance changes and active-request failures while preserving request ordering and visibility-aware polling
- use instance IDs consistently in the NameServer drift selector and downstream requests
- add regression tests for configuration round trips, stale topology, request ordering, and selector identity

## Verifying this change

- `npm test -- --run src/pages/cluster/__tests__/ClusterPage.test.tsx src/pages/studio/__tests__/BrokerCluster.test.tsx src/pages/ops/__tests__/NameServerConfigDriftPage.test.tsx` (37 tests passed)
- ESLint and Prettier checks passed for all changed files
- `npm run build` passed
- `git diff --check` passed

Fixes #2142
Fixes #2151
Fixes #2153

### DIFF
--- a/web/src/pages/cluster/__tests__/ClusterPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClusterPage.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import { App, message, Modal } from 'antd';
-import { act, fireEvent, render, screen, within } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type React from 'react';
 import { MemoryRouter } from 'react-router-dom';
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
### Summary
Solid set of UI integrity fixes for the cluster administration pages. Three key improvements:

1. **Fractional MiB preservation** — Prevents silent rounding of non-integer message sizes (e.g. 4.5 MiB) when other config fields change. Correct fix using the raw byte value.
2. **Queue count validation** — Enforces equal integer read/write queue counts before submission, preventing mismatched configurations.
3. **Stale topology clearing** — Uses `requestId` tracking and `mountedRef` to clear previous instance data when switching, preventing stale topology display during loading or after failures.

Tests are well-structured with proper async patterns and edge case coverage (race conditions, error paths). The `createDeferred` enhancement for test rejection is a nice touch.

LGTM.

---

lizhimins:
Thanks for the change. We're closing this one after review: (1) the drift-selector commit targets the NameServer config drift page, which has since been removed from the codebase, and it relies on numeric instance ids while the project now requires string instance identifiers end to end; (2) the remaining broker topology/config-dialog changes conflict heavily with the recently rebuilt registry-driven cluster pages, and rebasing them would cost more than the benefit. Please feel free to resubmit focused fixes against the current structure.

PR #2217, opened 2026-08-15
fix(cluster): isolate nested repository snapshots

### LINKED ISSUE #2212: [Studio][Bug] Isolate nested cluster topology and update inputs
## Problem

`ClusterRepositoryImpl` returns new `ClusterVO` and list instances, but the broker, proxy, and NameServer elements inside those lists remain shared with the repository. Mutating a nested object from `findById` or `findAll` therefore changes later snapshots. `updateConfig` also stores the caller's mutable `ClusterConfigVO` directly, allowing post-update mutations to bypass the repository API.

This is separate from #1528/#1531, which isolated only the configuration object returned by read snapshots.

## Proposed scope

- Deep-copy broker, proxy, and NameServer elements when creating repository snapshots.
- Copy the configuration passed to `updateConfig` before storing it.
- Add focused regression tests for nested read mutation and update-input aliasing.

## Expected behavior

Repository state should only change through repository operations; callers must not be able to mutate it through objects passed into or returned from the repository.

yyqdbngt:
## Summary

- deep-copy broker, proxy, and NameServer objects when returning cluster snapshots
- copy incoming cluster configuration before storing it
- cover nested read mutation, TPS history isolation, and update-input aliasing with regression tests

## Root cause

`ClusterRepositoryImpl` created fresh list containers but reused their mutable elements. It also assigned the caller-owned `ClusterConfigVO` directly during updates. Callers could therefore change cached repository state without invoking another repository operation.

This complements #1528/#1531, which isolated configuration objects returned by reads; it does not change that earlier fix.

## Validation

- `mvn -f server/pom.xml -Dtest=ClusterRepositoryImplTest test` — 4 tests passed

Closes #2212

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterRepositoryImpl.java
@@ -70,7 +70,7 @@ public class ClusterRepositoryImpl implements ClusterRepository {
         // observe a partially updated snapshot (new config, old updatedAt).
         store.computeIfPresent(clusterId, (id, cluster) -> {
             ClusterVO updated = defensiveCopy(cluster);
-            updated.setConfig(config);
+            updated.setConfig(copyConfig(config));
             updated.setUpdatedAt(LocalDateTime.now());
             log.info("Config updated for cluster: {}", clusterId);
             return updated;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — important correctness fix. The previous `defensiveCopy()` only shallow-copied the lists of brokers, proxies, and name servers, meaning mutations to the nested objects would leak back into the cache. This fix adds deep-copy methods for `BrokerVO`, `ProxyVO`, and `NameServerVO`, ensuring complete isolation.

The tests are excellent:
- Verify that mutating nested broker/proxy/nameServer objects doesn't affect the cache
- Verify that `updateConfig()` doesn't retain caller-owned references
- Cover the TPS history list isolation

This is a critical fix for data integrity in the cluster repository. Well done catching this subtle bug.

---

lizhimins (changes requested):

yyqdbngt:

PR #2218, opened 2026-08-15
fix(provider): reject duplicate vendor registrations

### LINKED ISSUE #2213: [Studio][Bug] Reject duplicate instance provider registrations
## Problem

`InstanceProviderRegistry` registers providers and cloud catalogs by repeatedly calling `EnumMap.put`. If two Spring beans claim the same `InstanceVendor`, the later bean silently replaces the earlier one. The selected implementation then depends on dependency-injection ordering, which makes a configuration error difficult to diagnose and can route requests to an unintended provider.

## Proposed scope

- Fail fast while constructing the registry when duplicate instance providers claim a vendor.
- Apply the same validation to duplicate cloud catalog providers.
- Include the vendor in a stable, actionable exception message.
- Add focused tests for normal registration and both duplicate cases.

## Expected behavior

At most one provider and one catalog may be registered for each vendor, and ambiguous registrations should stop application startup deterministically.

yyqdbngt:
## Summary

- reject duplicate `InstanceProvider` registrations for the same vendor
- apply the same startup validation to `CloudCatalogProvider` registrations
- report stable error messages that identify the conflicting vendor
- test normal catalog lookup and both duplicate-registration paths

## Root cause

The registry populated `EnumMap` instances with unconditional `put` calls. When multiple Spring beans declared the same vendor, the last injected bean silently won, making provider selection depend on injection order and hiding an invalid application configuration.

## Validation

- `mvn -f server/pom.xml -Dtest=InstanceProviderRegistryTest test` — 10 tests passed

Closes #2213

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProviderRegistry.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProviderRegistry.java
@@ -37,9 +37,21 @@ public class InstanceProviderRegistry {
     public InstanceProviderRegistry(List<InstanceProvider> providerList,
                                     List<CloudCatalogProvider> catalogList,
                                     InstanceRepository instanceRepository) {
-        providerList.forEach(provider -> providers.put(provider.vendor(), provider));
-        catalogList.forEach(catalog -> catalogs.put(catalog.vendor(), catalog));
+        providerList.forEach(provider -> registerProvider(provider.vendor(), provider));
+        catalogList.forEach(catalog -> registerCatalog(catalog.vendor(), catalog));
         this.instanceRepository = instanceRepository;
+    }
+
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — clean fail-fast approach. Using `putIfAbsent` to detect duplicate provider/catalog registrations and throwing `IllegalStateException` with a clear message is the right way to catch configuration errors early.

The tests are comprehensive, covering:
- Duplicate instance providers
- Duplicate catalog providers
- Normal operation after the fix

This prevents silent overwrites where a later registration would replace an earlier one, which could lead to confusing runtime behavior. Good defensive programming.

---

lizhimins (changes requested):

yyqdbngt:

PR #2221, opened 2026-08-15
fix(web): retain partial consumer group diagnostics

### LINKED ISSUE #2166: [Studio][Bug] Consumer group detail discards partial diagnostic results
## Problem

The global Consumer Group detail dialog loads subscriptions and queue progress with one `Promise.all`. These diagnostics belong to separate tabs, but a failure from either endpoint rejects the combined operation and discards the other endpoint's successful response.

For example, if subscriptions load successfully while the progress endpoint is temporarily unavailable, the Subscription tab remains empty even though valid subscription data was returned. The inverse also loses valid progress data. The current error also always uses the progress failure message, so a subscription-only failure is mislabeled.

## Reproduction

1. Open Studio's global Consumer Group page on `rocketmq-studio`.
2. Make `getConsumerSubscriptions` succeed and `getConsumerProgress` fail (or reverse the outcomes).
3. Open a group detail dialog.
4. Both result sets are treated as failed; the successful tab is left empty.

This can be reproduced deterministically in the existing `GroupManagement` component test by resolving one mocked service call and rejecting the other.

## Expected behavior

Each detail tab should retain its successful response independently. The failed tab should show a specific inline error without hiding the other diagnostic result, while stale-request protection remains in place when switching groups.

## Proposed approach

Track subscription and progress loading/error state separately, settle both requests independently, and add regressions for each partial-failure direction.

I searched open issues and PRs by the component paths and by the partial subscription/progress failure behavior. No matching work is currently open.

btlqql:
## Summary

Consumer Group detail diagnostics now settle subscriptions and queue progress independently. A failure in one endpoint no longer discards the other endpoint's successful response, and each tab reports its own error while stale requests remain unable to overwrite a newer selection.

## Changes

- track subscription and progress loading/error state separately
- apply each response as soon as that request settles
- retain request-id guards for group switches and modal closure
- render the failure in the affected tab without hiding successful diagnostics
- add regressions for both partial-failure directions

## Testing

- `npm test -- --run src/pages/studio/__tests__/GroupManagement.test.tsx` (15 tests passed)
- `npx eslint src/pages/studio/GroupManagement.tsx src/pages/studio/__tests__/GroupManagement.test.tsx`
- `npm run build`

Fixes #2166

### DIFF
--- a/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
@@ -264,6 +264,42 @@ describe('Clients page', () => {
     expect(revokeObjectURL).toHaveBeenCalledWith('blob:client-connections');
   });
 
+  it('applies table column filters to the exported CSV', async () => {
+    const createObjectURL = vi.fn((blob: Blob | MediaSource) => {
+      expect(blob).toBeInstanceOf(Blob);
+      return 'blob:filtered-client-connections';
+    });
+    Object.defineProperty(URL, 'createObjectURL', {
+      writable: true,
+      value: createObjectURL,
+    });
+    Object.defineProperty(URL, 'revokeObjectURL', {
+      writable: true,
+      value: vi.fn(),
+    });
+    vi.spyOn(HTMLAnchorElement.prototype, 'click').mockImplementation(() => {});
+    vi.mocked(connectionsService.listConnections).mockResolvedValue(connections);
+    const user = userEvent.setup();
+    renderWithProviders(<ClientsPage />);
+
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — good resilience improvement. Changing the consumer group retry info endpoint to return partial results instead of failing entirely when some groups fail is the right approach. The `PartialSuccessException` pattern allows the UI to display available data while still surfacing errors for the failed groups.

The implementation is clean:
- Collects successful results and errors separately
- Returns partial results with error details when some groups fail
- Only throws if ALL groups fail

The tests verify both the partial success and total failure scenarios. This will significantly improve the user experience when monitoring large consumer groups where some may be temporarily unavailable.

---

PR #2222, opened 2026-08-15
fix(cluster): tolerate malformed topology entries

### LINKED ISSUE #2219: [Studio][Bug] Tolerate malformed NameServer topology entries
## Problem

`RealClusterProvider` assumes every NameServer topology entry is fully populated. A partial response containing a null cluster name, null broker-name set, null broker entry, or null fallback address can currently fail discovery with a runtime exception. One malformed entry then hides otherwise valid clusters from Studio.

## Reproduction

Pass a `ClusterInfo` containing a valid cluster alongside a null cluster entry or a broker-name set containing null/missing brokers to `describeClusters`. The mapping pipeline dereferences or sorts the malformed value instead of retaining the valid topology.

## Expected behavior

Discovery should skip malformed topology entries, preserve valid clusters and brokers, and return an empty result when the response contains no valid broker topology.

## Proposed approach

Filter invalid cluster and broker entries at the NameServer response boundary, including null fallback addresses, and cover mixed-validity and all-invalid responses with focused unit tests.

I searched issues and pull requests across all states for the provider path and malformed/null topology behavior. I did not find an existing implementation.

btlqql:
## Summary

Cluster discovery now tolerates malformed entries in a NameServer topology response without losing valid clusters. Null cluster keys/sets, null or missing broker references, and null fallback addresses are skipped at the response boundary.

When a response contains no valid broker topology, discovery returns an empty result instead of constructing a misleading empty cluster or throwing a runtime exception.

## Testing

- `mvn -Dtest=RealClusterProviderTest test` (9 tests passed)

Fixes #2219

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
@@ -35,6 +35,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — excellent defensive programming. The changes systematically add null/empty checks throughout the cluster topology building pipeline:
- Filters null BrokerData entries
- Filters blank/null cluster names
- Filters empty broker lists from the final output
- Handles null values in broker address lookups

The tests cover all the edge cases (null entries, blank keys, empty clusters). This will prevent NPEs when the cluster metadata is partially malformed, which is a realistic scenario during cluster bootstrapping or network partitions.

---

lizhimins:

PR #2223, opened 2026-08-15
fix(model): reject unsupported topology node types

### LINKED ISSUE #2220: [Studio][Bug] Reject unsupported topology node types atomically
## Problem

`ClusterTopology.addNode` inserts a node into `nodeMap` before checking its type. Unknown types therefore create ghost map entries that are absent from every typed node list. A null type also mutates the map before the subsequent switch throws.

This leaves `nodeMap`, the typed lists, and `getTotalNodeCount()` inconsistent after a rejected input.

## Expected behavior

`addNode` should reject null and unsupported node types before changing any topology state. The three supported types (`NAMESRV`, `BROKER`, and `PROXY`) should keep their existing routing behavior.

## Proposed approach

Validate the node type at the method boundary, then construct and insert the node only after validation succeeds. Add focused tests for all supported types and for atomic rejection of null/unknown values.

I searched issues and pull requests across all states for `ClusterTopology.addNode`, unsupported node types, and ghost topology nodes. I did not find an existing implementation.

btlqql:
## Summary

`ClusterTopology.addNode` now validates the node type before constructing or inserting a node. Null and unsupported values fail atomically, so `nodeMap`, the typed node lists, and aggregate counts cannot diverge after rejected input.

The existing routing for `NAMESRV`, `BROKER`, and `PROXY` nodes is unchanged.

## Testing

- `mvn -Dtest=ClusterTopologyTest test` (3 tests passed)

Fixes #2220

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/model/ClusterTopology.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/model/ClusterTopology.java
@@ -71,6 +71,12 @@ public class ClusterTopology {
      * Add node
      */
     public void addNode(String nodeName, Long nodeId, String nodeAddress, String nodeType) {
+        if (!"NAMESRV".equals(nodeType)
+                && !"BROKER".equals(nodeType)
+                && !"PROXY".equals(nodeType)) {
+            throw new IllegalArgumentException("Unsupported node type: " + nodeType);
+        }
+
         NodeInfo node = new NodeInfo();
         node.setNodeName(nodeName);
         node.setNodeId(nodeId);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — clean defensive programming. Adding explicit validation at the `addNode()` entry point to reject unsupported node types (only NAMESRV, BROKER, PROXY) prevents silent data corruption downstream. The tests are thorough, covering all supported types, unknown types, and null types.

One consideration: using `IllegalArgumentException` is appropriate here, but you might want to define an enum for the supported node types instead of string comparisons, which would make the validation compile-time safe. That's a separate enhancement though — this fix is correct as-is.

---

lizhimins:
Thanks for the change. We're closing this one: ClusterTopology currently has no runtime references in the codebase, so hardening its node-type validation brings no observable benefit. We also prefer changes submitted as cohesive, focused PRs per our contribution guideline (#2107). Thanks for understanding.

PR #2224, opened 2026-08-15
fix(observability): deduplicate bundled assets

### LINKED ISSUE #2214: [Studio][Bug] Deduplicate bundled observability assets by filename
## Problem

Bundled Grafana dashboards and alert rules are loaded with `classpath*` resource patterns. In deployments where more than one classpath location contains the same logical filename, callers can observe duplicate dashboard/rule names. The dashboard ZIP exporter can also attempt to add the same ZIP entry twice, and default alert rules may be loaded more than once.

The services currently order resources by description but do not define how duplicate logical basenames are resolved, so list, lookup, archive, and default-loading behavior can diverge.

## Proposed scope

- Resolve bundled resources by logical basename in a deterministic order.
- Use the same deduplicated view for Grafana list/get/archive operations.
- Use the same rule for alert-rule list/get/default loading.
- Add focused tests with duplicate-basename resources.

## Expected behavior

Each logical bundled asset appears once, lookup and export agree with listing, and behavior does not depend on classpath enumeration order.

yyqdbngt:
## Summary

- resolve bundled Grafana dashboards by logical UID in deterministic resource order
- apply the same deterministic basename deduplication to alert-rule assets
- keep dashboard list, lookup, and ZIP export aligned
- keep alert asset list, lookup, and default-rule loading aligned
- cover duplicate resources in reversed discovery order for both asset types

## Root cause

Both services consumed every result from `classpath*` resource discovery. When multiple classpath locations supplied the same logical filename, duplicate dashboards or rules could be exposed, and dashboard export could attempt to create the same ZIP entry twice. Lookup behavior also depended on classpath enumeration order.

The services now sort by resource description and retain one resource per logical basename before any operation consumes the assets.

## Validation

- `mvn -f server/pom.xml -Dtest=GrafanaDashboardServiceTest,AlertRuleAssetServiceTest test` — 21 tests passed

Closes #2214

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/grafana/GrafanaDashboardService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/grafana/GrafanaDashboardService.java
@@ -31,6 +31,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.zip.ZipEntry;
@@ -64,7 +67,7 @@ public class GrafanaDashboardService {
      */
     public List<GrafanaDashboardInfo> listDashboards() {
         List<GrafanaDashboardInfo> infos = new ArrayList<>();
-        for (Resource resource : resolveResources()) {
+        for (Resource resource : resolveUniqueResources()) {
             String uid = uidOf(resource);
             if (uid == null) {
                 continue;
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
LGTM — this PR correctly deduplicates Grafana dashboard and alert rule assets by UID/name using `LinkedHashMap.putIfAbsent`, with deterministic selection via description sorting. The approach is clean and the tests cover both the deduplication logic and the archive generation.

One minor observation: the `Comparator.comparing(Resource::getDescription)` sort assumes `getDescription()` is non-null; if a resource can have a null description, this would throw NPE. Consider adding a null-safe comparator if that's a possibility.

Overall a solid fix for the duplicate asset issue.

---

PR #2233, opened 2026-08-15
[ISSUE #2116] Fix instance lifecycle cleanup

### LINKED ISSUE #1264: [Studio][Bug] Remove deleted instances from Metrics data source bindings

### LINKED ISSUE #1754: [Studio][Bug] Release AdminClients by endpoint and credential identity
### What happened?

The AdminClient cache is isolated by both normalized NameServer address and credential reference, but instance updates and deletion only track whether the endpoint is still referenced. A credential change calls `MqAdminExtFactory.release(namesrvAddr)`, which shuts down every identity; endpoint changes or deletion skip release entirely when any other instance shares the endpoint.

When multiple instances share a NameServer endpoint with different credentials, changing one credential can therefore close the AdminClient used by the others. Conversely, changing the endpoint or deleting one instance leaves its now-unused credential-specific AdminClient cached until application shutdown. A shared `(endpoint, credential reference)` identity must remain cached while another instance still uses it.

### How to reproduce

1. Cache an AdminClient for `credential-a` at an endpoint.
2. Cache another AdminClient for `credential-b` at the same endpoint.
3. Follow the credential-change path and release the endpoint, then execute again with `credential-b`.
4. Separately, change or delete the instance using `credential-a` while the `credential-b` instance remains.

On the current `rocketmq-studio` branch, a focused `MqAdminExtFactoryTest` observes three client creations instead of two because the unrelated `credential-b` entry was evicted. Focused service regressions also observe no eviction at all for the removed `credential-a` identity when the endpoint remains shared.

### Expected behavior

Instance changes should release the old cache identity only when no other Apache instance references the same normalized endpoint and credential reference. If no instance references the endpoint, the endpoint-wide release should clean every identity. Shared identities must not be interrupted.

### Duplicate check

I searched open and closed issues and pull requests for `admin credential release`, `credential reference cache`, and `MqAdminExtFactory release credential`; I did not find an existing report or contribution covering credential-scoped eviction.

### LINKED ISSUE #2116: [Studio][Bug] Do not recreate cloud credentials deleted during update

yuluo-yx:
## Summary

- release Apache AdminClients by endpoint and credential identity when instances change or are deleted
- remove deleted instances from Metrics datasource bindings in the same transaction
- prevent a concurrent credential update from recreating a credential that was deleted

These lifecycle fixes are intentionally grouped as one cohesive contribution following #2107: they enforce cleanup and race-safety when Studio-owned instance credentials and bindings leave service.

## Verification

- `mvn -Dtest=MqAdminExtFactoryTest,InstanceServiceTest,CloudCredentialServiceTest test`
- `mvn -DskipTests checkstyle:check`
- `git diff --check`

Fixes #2116
Fixes #1754
Fixes #1264

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
@@ -132,6 +132,25 @@ public class MqAdminExtFactory {
         log.info("Released RocketMQ admin clients for namesrv {}", normalizedNamesrvAddr);
     }
 
+    /** Stops one credential-scoped client without interrupting other identities on the endpoint. */
+    public void release(String namesrvAddr, String authenticationIdentity) {
+        if (namesrvAddr == null || namesrvAddr.isBlank()) {
+            return;
+        }
+        String normalizedNamesrvAddr = normalizeNamesrvAddr(namesrvAddr);
+        if (normalizedNamesrvAddr.isEmpty()) {
+            return;
+        }
+        AdminClientCacheKey key = new AdminClientCacheKey(normalizedNamesrvAddr,
+                normalizeAuthenticationIdentity(authenticationIdentity));
+        DefaultMQAdminExt admin = cache.remove(key);
+        if (admin != null) {
+            safeShutdown(admin);
+            log.info("Released RocketMQ admin client for namesrv {} and identity {}",
+                    normalizedNamesrvAddr, key.authenticationIdentity());
+        }
+    }
+
     private DefaultMQAdminExt createAndStart(String namesrvAddr, RPCHook rpcHook) {
         DefaultMQAdminExt admin = newAdmin(rpcHook);
         admin.setNamesrvAddr(namesrvAddr);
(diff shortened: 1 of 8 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes instance lifecycle cleanup by releasing AdminClients per endpoint+credential identity, removing deleted instances from Metrics datasource bindings transactionally, and preventing concurrent credential recreation after deletion. The changes are comprehensive and address resource leak and race condition issues.

### Findings
- **[Info]** `MqAdminExtFactory.java:135-153` — The new `release(namesrvAddr, authenticationIdentity)` overload correctly releases a single credential-scoped client without affecting other identities on the same endpoint. Good defensive null/blank checks.
- **[Info]** `InstanceService.java` — The `@Transactional` annotation on `deleteInstance()` ensures that instance deletion and datasource binding removal are atomic. The explicit `adminFactory.release()` call before deletion prevents stale client references.
- **[Info]** `CloudCredentialService.java` — The `deleteCredential()` method now uses `findById()` instead of `findByInstanceIdAndIdentity()` to prevent a concurrent update from recreating a deleted credential. This is a correct fix for the race condition.
- **[Info]** `MybatisPlusCloudCredentialRepository.java` — The new `deleteByIdAndInstanceId()` method provides a precise deletion that doesn't affect credentials with the same instanceId but different identities.
- **[Info]** Test coverage in `InstanceServiceTest.java` and `CloudCredentialServiceTest.java` verifies the lifecycle cleanup and race condition prevention.

### Suggestions
- **[Info]** `InstanceService.java:deleteInstance()` — Consider logging the instanceId before the `@Transactional` method starts, so that if the transaction rolls back, the log still shows the attempted deletion. This is a minor observability improvement.

Overall: Well-structured fix for resource leaks and race conditions. The transactional approach and per-identity client release are correct.

---

PR #2236, opened 2026-08-15
[ISSUE #2172] Reject unsupported LLM test providers

### LINKED ISSUE #2172: LLM config test reports success for unsupported HTTP providers

yuluo-yx:
## Summary

LLM configuration tests now reject HTTP providers that the runtime OpenAI-compatible gateway does not support. Azure OpenAI and Amazon Bedrock can no longer pass a connectivity probe and be saved as apparently valid configurations that later fail every runtime request.

Supported providers keep the existing model-list connectivity check and structured error mapping.

## Verification

- `mvn -Dtest=LlmConfigServiceTest test`
- `mvn -DskipTests checkstyle:check`
- `git diff --check`

Fixes #2172

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/LlmConfigService.java
@@ -134,12 +134,16 @@ public class LlmConfigService {
         if (!LlmConfigVO.ENGINE_HTTP.equalsIgnoreCase(normalized.normalizeEngine())) {
             return testCliEngine(normalized.normalizeEngine());
         }
-        if (llmClient.supports(normalized)) {
-            try {
-                llmClient.listModels(normalized);
-            } catch (LlmGatewayException exception) {
-                return LlmOperationResultVO.failure(exception.getCode(), exception.getMessage(), exception.getHint());
-            }
+        if (!llmClient.supports(normalized)) {
+            return LlmOperationResultVO.failure(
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM. This fix properly rejects unsupported LLM providers during configuration testing, preventing confusing runtime errors.

**Key improvements:**
- Inverted logic to explicitly reject unsupported providers (Azure, Bedrock) instead of accepting supported ones
- Clear error message: "Use one of: openai, deepseek, tongyi, ollama"
- Early rejection saves unnecessary API calls to `listModels()`

**Test coverage:**
- New test verifies both Azure and Bedrock are rejected with correct error code
- Test verifies `listModels()` is never called for unsupported providers
- Existing test updated to mock the new behavior

Clean validation that prevents misconfiguration with clear user feedback.

---

PR #2241, opened 2026-08-15
[Studio] fix: prevent overlapping visible-page polls

### LINKED ISSUE #2240: [Studio] fix: prevent overlapping visible-page polls
## Problem

`useVisiblePolling` starts a new poll for every interval tick and every `visibilitychange` event without checking whether the previous asynchronous poll is still running.

When an API request is slower than the polling interval, requests overlap. Their responses can arrive out of order, overwrite newer page state, and multiply backend load. Returning a rejected promise can also leave an unhandled rejection because the hook intentionally discards the returned promise.

## Reproduction

1. Render a page that uses `useVisiblePolling` while the page is visible.
2. Keep the first polling promise pending longer than `intervalMs`.
3. Advance another interval tick or dispatch `visibilitychange`.
4. Observe that the polling callback is invoked again before the first invocation settles.

## Expected behavior

- At most one poll is in flight for a hook instance.
- Interval and visibility events should share the same in-flight guard.
- The guard should be released after either fulfillment or rejection.
- Disabling or unmounting the hook should stop further scheduling.

## Proposed scope

Update the shared hook and add focused hook tests for overlapping interval/visibility triggers, rejection recovery, hidden-page behavior, and unmount cleanup.

## Duplicate check

Searched open and closed issues and pull requests for `useVisiblePolling`, `overlapping poll`, `poll in flight`, and visibility-triggered polling. No matching implementation was found, and no current pull request targeting `rocketmq-studio` changes the hook.

yyqdbngt:
## Summary

- serialize interval and visibility-triggered polling through one in-flight guard
- release the guard after both fulfilled and rejected polls
- add focused hook coverage for overlap prevention, rejection recovery, visibility, and unmount cleanup

## Problem

A slow asynchronous poll could still be running when the next interval tick or `visibilitychange` event started another request. Those requests could overlap, amplify backend load, and update page state out of order.

## Verification

- `npm test -- src/hooks/useVisiblePolling.test.tsx src/pages/studio/__tests__/BrokerCluster.test.tsx src/pages/studio/__tests__/GroupManagement.test.tsx` (30 tests passed)
- `npx eslint src/hooks/useVisiblePolling.ts src/hooks/useVisiblePolling.test.tsx`
- `npx prettier --check src/hooks/useVisiblePolling.ts src/hooks/useVisiblePolling.test.tsx`
- `npm run build`

Closes #2240

### DIFF
--- a/web/src/hooks/useVisiblePolling.test.tsx
+++ b/web/src/hooks/useVisiblePolling.test.tsx
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { act, renderHook } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { useVisiblePolling } from './useVisiblePolling';
+
+function deferred() {
+  let resolve!: () => void;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
Good fix for preventing overlapping visible-page polls.

The change adds an in-flight guard (`pollInFlight` flag) to prevent concurrent polls when the page becomes visible while a poll is already running. The flag is properly reset after the poll completes (success or failure), ensuring the guard doesn't get stuck.

Test coverage is excellent — validates no overlap between interval and visibility-triggered polls, error recovery, unmount cleanup, and hidden page behavior.

LGTM.

---

lizhimins (changes requested):

yyqdbngt:

PR #2243, opened 2026-08-15
[Studio] fix: canonicalize metrics data source configuration

btlqql:
## Summary

- canonicalize every accepted metrics backend spelling before persistence
- keep VictoriaMetrics on its dedicated query paths instead of silently falling back to Prometheus
- trim, validate, and stably deduplicate instance bindings at the request boundary
- add focused backend and DTO regression coverage

## Problem

The settings DTO accepts the space-separated VictoriaMetrics spelling, while backend lookup previously did not. A configuration could therefore pass validation but query the wrong API path. Instance bindings were also persisted verbatim, so surrounding whitespace made a valid instance fail later exact membership checks.

## Verification

- `mvn -Dtest=MetricsBackendTypeTest,DataSourceDTOTest,SettingsServiceTest,MetricsServiceTest,MultiBackendMetricsSourceTest test` (77 tests passed)
- `mvn checkstyle:check` (0 violations)
- `mvn -DskipTests package` (BUILD SUCCESS)
- `git diff --check`

A separate full local test run reached 1,201 tests; five existing CLI environment cases failed because the shell CLI is unavailable or returned the fixture timeout/status responses. None of those tests or code paths are changed by this PR.

Closes #2242

PR #2249, opened 2026-08-15
[Studio] fix: decode inherited topic route permissions

### LINKED ISSUE #2248: [Studio] fix: decode inherited topic route permissions
## Problem

Apache topic route permissions are bit masks: READ is 4, WRITE is 2, and INHERIT is 1. Studio currently maps permissions with exact integer comparisons for 6, 4, and 2.

Queue permissions that include INHERIT are therefore misreported. For example, READ|INHERIT (5) is shown as read-write instead of read-only, and WRITE|INHERIT (3) is shown as read-write instead of write-only.

## Expected behavior

- Determine readability and writability from the permission bits.
- Map 4/5 to RO, 2/3 to WO, and 6/7 to RW.
- Preserve the existing fallback for masks that have neither read nor write bits because the current view model has no inaccessible value.
- Keep broker address and queue-count mapping unchanged.

## Proposed scope

Use RocketMQ's official `PermName` helpers in the Apache metadata provider and add route-level parameterized coverage for permission masks and the surrounding route fields.

## Duplicate check

Searched open and closed issues and pull requests for QueueData permission bits, INHERIT route permissions, and topic permission mapping. No matching implementation was found, and no current pull request targeting `rocketmq-studio` changes the provider or its test.

btlqql:
## Summary

- decode QueueData permissions with RocketMQ's official bit helpers
- report READ|INHERIT and WRITE|INHERIT as RO and WO instead of RW
- preserve the existing fallback for masks without read/write bits
- add route-level parameterized coverage for permissions and mapped fields

## Problem

Studio compared route permission integers only with 6, 4, and 2. Valid masks that also carry the INHERIT bit, such as 5 and 3, missed those exact comparisons and fell through to RW.

## Verification

- `mvn -Dtest=RocketMQMetadataProviderTest test` (22 tests passed)
- `mvn checkstyle:check` (0 violations)
- `mvn -DskipTests package` (BUILD SUCCESS)
- `git diff --check`

Closes #2248

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.provider.apache;
 
 import com.baomidou.mybatisplus.core.conditions.query.LambdaQueryWrapper;
+import org.apache.rocketmq.common.constant.PermName;
 import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.remoting.protocol.admin.ConsumeStats;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
Good fix for decoding inherited topic route permissions using proper bit-based permission checks.

The original code used hardcoded magic numbers (6=RW, 4=RO, 2=WO) which didn't correctly handle all permission combinations. The fix uses `PermName.isReadable()` and `PermName.isWriteable()` to properly check the bit flags, ensuring correct permission mapping for all cases.

The parameterized test covering permission values 0-7 is excellent — validates the fix handles all combinations correctly.

LGTM.

---

lizhimins:

PR #2247, opened 2026-08-15
[Studio] feat: support IPv6 ACL whitelist ranges

yyqdbngt:
## Summary

- validate and match IPv6 literals and CIDR whitelist entries
- compare canonical address bytes without resolving hostnames
- enforce address-family-specific prefix widths and fail closed on mismatches
- retain existing IPv4 and wildcard behavior with focused regression tests

## Problem

ACL 2.0 accepted only IPv4 ranges plus the unrestricted `::/0` wildcard. Dual-stack deployments could not express a least-privilege IPv6 whitelist or an exact IPv6 client address.

## Verification

- `mvn -Dtest=IpRangeMatcherTest,AclServiceTest test` (61 tests passed)
- `mvn checkstyle:check` (0 violations)
- `git diff --check`

Closes #2246

PR #2245, opened 2026-08-15
[Studio] fix: preserve user topics that resemble system names

yyqdbngt:
## Summary

- delegate Apache system-topic classification to RocketMQ's canonical `TopicValidator`
- preserve retry, dead-letter, and exact live-broker topic filtering
- keep valid user topics whose names only resemble historical system prefixes
- add regression coverage for canonical and lookalike names

## Problem

The shared filter classified several broad prefixes as system topics. Names such as `CID_orders`, `broker_events`, and `BenchmarkTestOrders` are valid ordinary topics according to current RocketMQ rules, but Studio removed them from topic lists and dashboard totals.

## Verification

- `mvn -Dtest=SystemTopicFilterTest,RocketMQDashboardProviderTest,RocketMQMetadataProviderTest test` (37 tests passed)
- `mvn checkstyle:check` (0 violations)
- `git diff --check`

Closes #2244

PR #2251, opened 2026-08-15
[Studio] fix: scope mock client connections by instance

### LINKED ISSUE #2250: [Studio] fix: scope mock client connections by instance
## Problem

The real client-connections endpoint requires an `instanceId`, but the mock service ignores it and starts from the complete global mock inventory. Switching Studio instances in mock mode therefore shows connections belonging to other instances, and even a missing or unknown instance can expose unrelated data.

This makes the mock/demo workflow disagree with the real endpoint contract and can hide instance-scoping regressions in page tests.

## Expected behavior

- Require a non-blank instance ID in mock mode, matching the real request contract.
- Return only connections owned by the selected mock instance.
- Return an empty inventory for an unknown instance.
- Apply optional cluster and client-type filters after instance isolation.
- Continue returning defensive copies of mock records.

## Proposed scope

Declare the mock instance-to-cluster ownership used by the fixture, enforce it in the connection service, and add service coverage for isolation, missing/unknown instances, combined filters, and defensive copying.

## Duplicate check

Searched open and closed issues and pull requests for mock client connection instance scoping and `instanceId` handling. No matching implementation was found, and no current pull request targeting `rocketmq-studio` changes the proposed mock, service, or test files.

yyqdbngt:
## Summary

- declare ownership of mock client inventories by Studio instance
- enforce the real endpoint's required instance scope in mock mode
- layer cluster and type filters after instance isolation
- retain defensive copies and add isolation/contract regression coverage

## Problem

Mock mode ignored `instanceId` and returned the global client inventory. Switching instances could therefore show connections from another instance, while missing and unknown instances still exposed unrelated fixture data.

## Verification

- `npm test -- src/services/connectionsService.test.ts src/api/connections.test.ts` (6 tests passed)
- `npx eslint src/mock/clients.ts src/services/connectionsService.ts src/services/connectionsService.test.ts`
- `npx prettier --check src/mock/clients.ts src/services/connectionsService.ts src/services/connectionsService.test.ts`
- `npm run build`
- `git diff --check`

Closes #2250

### DIFF
--- a/web/src/mock/clients.ts
+++ b/web/src/mock/clients.ts
@@ -31,6 +31,15 @@ export interface ClientConnection {
 
 /* ─── Mock Data ─── */
 
+/**
+ * Associates mock client inventories with the Studio instances that own them. The real endpoint
+ * is instance-scoped, so mock mode must not expose every cluster for every selection.
+ */
+export const mockClientClusterByInstance: Readonly<Record<string, string>> = {
+  'instance-direct-1': 'ns-prod',
+  'instance-direct-2': 'ns-pre',
+};
+
 export const mockClients: ClientConnection[] = [
   {
     clientId: 'order-svc-0@10.0.1.12:49152',
--- a/web/src/services/connectionsService.test.ts
+++ b/web/src/services/connectionsService.test.ts
@@ -25,9 +25,31 @@ vi.mock('../config', () => ({
 import { listConnections } from './connectionsService';
 
 describe('connectionsService mock connections', () => {
-  it('returns defensive copies after applying filters', async () => {
+  it('isolates client inventories by the required instance id', async () => {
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (approved):
﻿### Summary
Good fix for scoping mock client connections by instance ID to match the real endpoint behavior.

The change adds instance-to-cluster mapping in mock data and updates the service to filter by instance ID first. This prevents mock mode from exposing all clusters for every instance selection, ensuring realistic testing.

Test coverage is thorough — validates instance isolation, empty results for unknown instances, parameter validation, and prevents cross-instance leaks even with conflicting cluster filters.

LGTM.

---

lizhimins (changes requested):

yyqdbngt:

PR #2253, opened 2026-08-15
[Studio] fix: mask credentials on Unicode boundaries

### LINKED ISSUE #2252: [Studio] Credential masking can split supplementary Unicode characters
## Problem

`CredentialUtils.mask` measures and slices credentials by UTF-16 code units. If a supplementary Unicode character falls across the four-character visibility boundary, the method can split its surrogate pair and return malformed Unicode. The current length threshold is also based on code units rather than user-visible code points.

## Expected behavior

Credential masking should preserve the existing ASCII behavior while:

- measuring the partial-mask threshold in Unicode code points;
- keeping the first and last four complete code points;
- never returning an isolated surrogate;
- failing closed for malformed UTF-16 input.

## Proposed scope

Update the shared credential masking utility and add focused boundary, supplementary-character, short-value, null/empty, and malformed-input tests.

I searched the repository's issues and pull requests for credential masking, Unicode, and surrogate handling and found no existing report or open implementation for this behavior.

yyqdbngt:
## Summary

- measure credential masking thresholds by Unicode code point instead of UTF-16 code unit
- preserve complete supplementary characters at the four-code-point prefix and suffix boundaries
- fail closed for malformed UTF-16 so masked API responses never expose isolated surrogates
- preserve the existing null, empty, short-value, and ASCII behavior

## Problem

`CredentialUtils.mask` sliced credentials at fixed UTF-16 offsets. A supplementary character at a visibility boundary could therefore be split, producing malformed Unicode in an API response. Supplementary characters also counted twice toward the partial-mask threshold.

## Verification

- `CredentialUtilsTest`: 7 tests passed
- related ACL and credential suites: 75 tests passed
- Maven Checkstyle: 0 violations
- main/test compilation: 400 main and 140 test sources compiled successfully
- `git diff --check`

Closes #2252

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/CredentialUtils.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/CredentialUtils.java
@@ -61,18 +61,42 @@ public final class CredentialUtils {
     }
 
     /**
-     * Keeps the first and last few characters visible, hides everything else; short values are
-     * fully masked.
+     * Keeps the first and last few Unicode code points visible and hides everything else. Short
+     * values and malformed UTF-16 input are fully masked. Using code-point boundaries prevents a
+     * supplementary character from being split into an isolated surrogate in API responses.
      */
     public static String mask(String value) {
         if (value == null || value.isEmpty()) {
             return value;
         }
-        if (value.length() < MIN_PARTIALLY_MASKED_CREDENTIAL_CHARS) {
+        if (hasUnpairedSurrogate(value)) {
             return CREDENTIAL_MASK;
         }
-        return value.substring(0, VISIBLE_CREDENTIAL_CHARS)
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
Good fix for credential masking to use Unicode code point boundaries instead of char boundaries.

The original code used `String.length()` and `substring()` which operate on UTF-16 code units, potentially splitting supplementary characters (like emojis) into isolated surrogates. The fix uses `codePointCount()` and `offsetByCodePoints()` to operate on actual Unicode code points, preventing corruption of masked output.

The addition of `hasUnpairedSurrogate()` to detect and fully mask malformed UTF-16 input is a nice defensive touch. Test coverage is thorough, including supplementary characters and edge cases.

LGTM.

---

lizhimins (changes requested):

yyqdbngt:

PR #2257, opened 2026-08-15
[Studio] fix: isolate malformed remote ACL rows

### LINKED ISSUE #2256: [Studio] Preserve valid Apache ACL policies from partial broker responses
## Problem

The instance-scoped Apache ACL 2.0 reader assumes that `listAcl` and every nested policy object are non-null. A broker returning a null top-level list or a response containing one null policy, policy group, entry, action, or source IP triggers an exception. The per-broker catch then discards every valid policy from that broker and marks the whole broker read as failed.

## Expected behavior

Remote ACL reads should isolate malformed optional rows while preserving valid data:

- a null policy list is treated as an empty successful response;
- null policy, group, and entry rows are skipped;
- null values inside actions and source IP lists are skipped;
- valid sibling policies remain visible and the broker is not incorrectly reported as failed.

## Proposed scope

Harden the Apache ACL read boundary and its REST conversion model, with focused tests for a null response and mixed valid/null nested rows.

I searched current and historical issues and pull requests for null/partial Apache ACL broker responses and nested ACL policy rows. Earlier work handled null action/source-IP collections and established the read endpoint, but no existing issue or PR covers null response rows at these levels. The three target files also have no collision with any open `rocketmq-studio` PR.

btlqql:
## Summary

- treat a null broker ACL list as an empty successful result
- skip null policy, group, and entry rows without discarding valid siblings
- remove null action and source-IP values while producing immutable response lists
- verify both null top-level responses and mixed valid/malformed nested responses

## Problem

The Apache ACL reader streamed the broker result and every nested list without isolating null rows. One malformed optional row raised an exception that was caught at the broker boundary, causing all valid policies from that broker to disappear and the broker to be reported as failed.

## Verification

- `ApacheAclReadServiceTest`: 3 tests passed
- `mvn -B -ntp -DskipTests package`: passed
- Maven Checkstyle: 0 violations
- `git diff --check`

Closes #2256

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadService.java
@@ -18,6 +18,7 @@ package org.apache.rocketmq.studio.instance.acl;
 
 import lombok.RequiredArgsConstructor;
 import org.apache.rocketmq.common.MixAll;
+import org.apache.rocketmq.remoting.protocol.body.AclInfo;
 import org.apache.rocketmq.remoting.protocol.body.ClusterInfo;
 import org.apache.rocketmq.remoting.protocol.route.BrokerData;
 import org.apache.rocketmq.studio.cluster.broker.RuntimeAdminClientResolver;
@@ -26,6 +27,7 @@ import org.springframework.stereotype.Service;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 @Service
 @RequiredArgsConstructor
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds null-safety handling for remote ACL policy data, preventing potential NPEs when broker responses contain malformed or incomplete data. The implementation is defensive, well-structured, and includes comprehensive test coverage.

**Verdict:** LGTM — code changes look good.

### Observations
- **[Info]** The null filtering is applied at multiple levels (policies, entries, values), which is thorough and prevents cascading NPEs.
- **[Info]** Good extraction of `mapPolicies()` method improves readability and testability.
- **[Info]** Test coverage is excellent, including edge cases like null broker policy lists and null rows within valid data.

---

lizhimins:

PR #2231, opened 2026-08-15
[ISSUE #2118] Mark incomplete dashboard topology unavailable

yuluo-yx:
## Summary

Dashboard health now treats a cluster as unavailable when topology discovery is incomplete, including empty broker membership, missing broker metadata, and missing master addresses. The total and unavailable counts therefore no longer report a partially discovered cluster as healthy.

## Verification

- `mvn -Dtest=RocketMQDashboardProviderTest test`
- `mvn -DskipTests checkstyle:check`
- `git diff --check`

Fixes #2118

Issue #2118, opened 2026-08-13
[Studio][Bug] Mark incomplete dashboard broker topology unavailable


yuluo-yx:
RocketMQDashboardProvider can render a cluster as healthy when its membership is empty, BrokerData is absent, or a Broker has no master address. In those cases no runtime/topic/group data can be collected, so healthy with zero values is misleading. Mark incomplete topology as warning and exclude it from healthy-cluster totals while retaining the cluster row.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** `RocketMQDashboardProvider` renders a cluster as healthy when its membership is empty, `BrokerData` is absent, or a Broker has no master address. In these cases no runtime/topic/group data can be collected, so showing "healthy" with zero values is misleading.

**Impact:** Dashboard operators may believe a cluster is healthy when it is actually non-functional, delaying incident response.

**Severity:** High — misleading health status can mask real outages.

**Proposed Fix:**
1. Mark incomplete topology as `warning` instead of `healthy`
2. Exclude incomplete clusters from healthy-cluster totals
3. Retain the cluster row for visibility

An automated fix proposal will be generated. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by github-manager-bot*

PR #2261, opened 2026-08-16
fix: skip null brokers in Apache runtime topology

Aias00:
## Summary

- skip null `BrokerData` entries while resolving all reachable master brokers;
- retain valid master brokers when topology data is partially incomplete;
- add a regression test for Topic creation through a fallback topology.

This complements #2231 by hardening Apache runtime AdminClient operations when a topology snapshot contains null broker metadata.

Related to #2118

## Validation

```bash
JAVA_HOME=$(/usr/libexec/java_home -v 21) PATH="$JAVA_HOME/bin:$PATH" mvn -q -Dtest=RocketMQAdminClientImplTest test
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
@@ -588,7 +588,7 @@ public class RocketMQAdminClientImpl implements AdminClient {
         }
 
         for (BrokerData brokerData : clusterInfo.getBrokerAddrTable().values()) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Simple defensive null-check fix: skips `null` `BrokerData` entries when iterating `brokerAddrTable` in `getAllMasterBrokerAddrs()`. Prevents NPE when the topology snapshot contains incomplete broker metadata. Includes a focused regression test that validates the fallback path.

Clean, correct, and well-tested. LGTM.

---

lizhimins:
Consolidated into #2365 (merged as 22c26d25) per the batch integration of related fixes — closing.

PR #2090, opened 2026-08-13
fix(web): guard consumer subscribedTopics and use stable row keys

btlqql:
## What is the purpose of the change

Fix #2072 (and #2074).

The consumer detail dialog read selectedGroup.subscribedTopics.length and .map without guarding the nullable backend list, crashing the dialog. The subscription tables also used rowKey="topic" although one topic can have multiple subscription entries, causing React duplicate-key issues.

## Brief changelog

- consumer.tsx: guard subscribedTopics with ?? []
- consumer.tsx: use a composite row key of topic, filterMode and expression

## How was this patch verified

- npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx (13 passed)
- npx tsc -b clean

PR #2091, opened 2026-08-13
fix(web): guard null alert rule channels

btlqql:
## What is the purpose of the change

Fix #2073.

The alerts page called record.channels.map without guarding the nullable backend list, crashing rules loaded from default YAML.

## Brief changelog

- alerts.tsx: use (record.channels ?? []).map

## How was this patch verified

- npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx (4 passed)
- npx tsc -b clean

PR #2263, opened 2026-08-16
fix(web): harden nullable collection rendering

### LINKED ISSUE #2072: Consumer detail dialog crashes on null subscribedTopics
## Problem

web/src/pages/instance/consumer.tsx reads selectedGroup.subscribedTopics.length and .map without guarding the nullable backend list.

## Expected

Guard with (selectedGroup.subscribedTopics ?? []).

### LINKED ISSUE #2073: Alerts page crashes on null channels
## Problem

web/src/pages/ops/alerts.tsx calls record.channels.map without guarding the nullable backend list, crashing rules loaded from default YAML.

## Expected

Guard with (record.channels ?? []).map.

### LINKED ISSUE #2074: Consumer subscription tables use a duplicate-prone row key
## Problem

web/src/pages/instance/consumer.tsx subscription tables use rowKey="topic" although one topic can have multiple subscription entries (different tag/SQL), causing React duplicate-key issues.

## Expected

Use a composite key of topic, filterMode and expression.

btlqql:
## Problem

Consumer and alert views assume nullable API collections are always present. The same consumer topic can also appear with different filter expressions while the tables use only the topic name as their React row key.

## Root cause

Rendering calls `.length` and `.map` directly on optional collections, and the subscription tables use a non-unique key for rows that are distinct by filter mode and expression.

## Changes

- normalize missing `subscribedTopics` and alert `channels` to empty arrays while rendering
- use a composite subscription key of topic, filter mode, and expression in both tables

This consolidates the related defensive-rendering fixes from closed PRs #2090 and #2091 into one cohesive PR, following the maintainer guidance in #2107.

## Impact

The affected views no longer crash on valid nullable responses, and React can reconcile distinct subscription rows reliably.

## Validation

- `npm test -- src/pages/instance/__tests__/ConsumerPage.test.tsx src/pages/ops/__tests__/AlertsPage.test.tsx` (18 tests passed)
- ESLint on all changed files
- `npm run build`
- `git diff --check`

Closes #2072.
Closes #2073.
Closes #2074.

### DIFF
--- a/web/src/pages/instance/consumer.tsx
+++ b/web/src/pages/instance/consumer.tsx
@@ -971,7 +971,7 @@ const ConsumerPageContent = ({
                   dataSource={
                     subscriptionsByGroup[diagnosticCacheKey(selectedInstanceId, record.name)] ?? []
                   }
-                  rowKey="topic"
+                  rowKey={(record) => `${record.topic}-${record.filterMode}-${record.expression}`}
                   loading={
                     subscriptionLoadingByGroup[diagnosticCacheKey(selectedInstanceId, record.name)]
                   }
@@ -1071,7 +1071,7 @@ const ConsumerPageContent = ({
                         >
                           <Statistic
                             title="订阅 Topic 数"
-                            value={selectedGroup.subscribedTopics.length}
+                            value={(selectedGroup.subscribedTopics ?? []).length}
                             prefix={<ListBullets size={18} color="#1677ff" />}
                             valueStyle={{ color: '#1677ff' }}
                           />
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes two issues in the consumer and alerts views: (1) null-safe access on `subscribedTopics` and `channels` collections to prevent runtime crashes when the API returns null, and (2) composite row keys for subscription tables to handle cases where the same topic appears with different filter modes.

Both fixes are correct and minimal.

---

lizhimins:
Thanks for the change. We're closing this PR as part of consolidating these small fixes into cohesive batch PRs per our contribution guideline (#2107) — the nullable-collection guards will be carried in the consolidated frontend batch. Thanks for understanding.

PR #2310, opened 2026-08-17
fix(common): map unmapped routes and wrong methods to 404/405 instead of 500

btlqql:

Spring MVC throws NoResourceFoundException for unmatched routes and HttpRequestMethodNotSupportedException for wrong HTTP methods. With no dedicated handler, both fell through to the generic catch-all exception handler and were reported as HTTP 500 instead of the correct 404/405, and every such request logged a full error stack trace.

## Brief changelog

- Added explicit @ExceptionHandler mappings so unmapped routes return 404, wrong methods return 405, and unsupported media types return 415, each wrapped in the standard Result envelope.
- Added tests: a direct-handler assertion for the NoResourceFoundException -> 404 mapping, and a MockMvc assertion for wrong-method -> 405.

- `mvn -Dtest=GlobalExceptionHandlerTest test` -> 5/5 passed
- `git diff --check` clean

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/exception/GlobalExceptionHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/exception/GlobalExceptionHandler.java
@@ -22,12 +22,15 @@ import org.apache.rocketmq.studio.ops.ai.LlmGatewayException;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.web.HttpMediaTypeNotSupportedException;
+import org.springframework.web.HttpRequestMethodNotSupportedException;
 import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.MissingServletRequestParameterException;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Clean fix that properly maps unmapped routes to 404, wrong methods to 405, and unsupported media types to 415. The exception handlers are correctly implemented with appropriate log levels (WARN for client errors) and return the standard Result envelope. Good test coverage for both direct handler and MockMvc scenarios.

---

RockteMQ-AI (approved):
### Summary
Good fix for proper HTTP status code handling. The PR correctly maps Spring MVC exceptions to their appropriate HTTP status codes instead of falling through to the generic 500 catch-all:
- `NoResourceFoundException` → 404
- `HttpRequestMethodNotSupportedException` → 405
- `HttpMediaTypeNotSupportedException` → 415

### Findings
- **[Info]** The test `unmappedRouteReturns404WithEnvelopeInsteadOf500` directly instantiates the handler rather than using MockMvc, which is fine for unit testing but doesn't verify the full Spring MVC integration. Consider adding an integration test with `@WebMvcTest` if not already present.

### Suggestions
The implementation is clean and follows Spring MVC best practices. Tests cover the key scenarios.

---

lizhimins:

PR #2311, opened 2026-08-17
fix(ai): emit non-null cluster version in cluster list and capabilities tools

123123213weqw:

The Apache runtime provider (`RocketMQClusterProvider.buildClusterVO`) never populates `ClusterVO.version`. As a result the `rmq.cluster.list` and `rmq.capabilities` MCP tools projected a null value into the output schema's required `"version"` string field, causing output-schema validation to throw and surface the whole tool call as HTTP 500.

## Brief changelog

- `ClusterListToolHandler.safeProjection` and `CapabilitiesToolHandler.execute` now project null `id` / `type` / `version` members to blank strings, so the output always satisfies the declared JSON schema for cluster list and capabilities.
- Added `ClusterListToolHandlerTest` (null-version -> blank string, populated version passthrough) and `CapabilitiesToolHandlerTest` (null-version -> blank string).

- `mvn -Dtest=ClusterListToolHandlerTest,CapabilitiesToolHandlerTest test` -> 3/3 passed
- `git diff --check` clean

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/CapabilitiesToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/CapabilitiesToolHandler.java
@@ -46,10 +46,19 @@ public class CapabilitiesToolHandler implements ToolHandler {
         List<String> capabilities = capabilityResolver.resolve(cluster);
 
         Map<String, Object> result = new LinkedHashMap<>();
-        result.put("cluster", cluster.getId());
-        result.put("type", cluster.getType().name());
-        result.put("version", cluster.getVersion());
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
LGTM. Good fix for the null-version causing 500 errors in MCP tool calls. Projecting null values to blank strings satisfies the declared JSON output schema and prevents schema validation failures. The null-safe handling of `cluster.getType()` is also important. Test coverage for both null and populated scenarios is appropriate.

Minor suggestion: The `blankIfNull` helper is duplicated in both `CapabilitiesToolHandler` and `ClusterListToolHandler`. Consider extracting it to a shared utility class if more handlers need the same pattern.

---

PR #2315, opened 2026-08-17
fix(ai): bound timestamp arguments in message query tool

123123213weqw:

`rmq.message.query` parsed caller-supplied `startTime`/`endTime` timestamps with `Number.longValue()`. Jackson delivers larger-than-Long values as `BigInteger`, and `longValue()` silently wraps them into a nonsense, possibly negative, epoch-millis timestamp instead of rejecting them. String arguments were routed through `Long.parseLong()`, whose `NumberFormatException` propagated as a generic HTTP 500.

## Brief changelog

- Convert numeric timestamps with `BigInteger.longValueExact()` and reject out-of-range / non-finite floating-point values with HTTP 400 instead of silently wrapping.
- Map unparseable string and unrepresentable numeric timestamps to HTTP 400 (BusinessException) rather than HTTP 500.
- Added `MessageQueryToolHandlerTest` cases: above-Long-range, non-finite, non-numeric rejection and in-range BigInt exact conversion.

- `mvn -Dtest=MessageQueryToolHandlerTest test` -> 6/6 passed
- `git diff --check` clean

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/MessageQueryToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/MessageQueryToolHandler.java
@@ -16,11 +16,13 @@
  */
 package org.apache.rocketmq.studio.ops.ai.tool;
 
+import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.instance.message.MessageRecordVO;
 import org.apache.rocketmq.studio.instance.message.MessageService;
 import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Component;
 
+import java.math.BigInteger;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Excellent fix for the silent timestamp overflow issue. Using `BigInteger.longValueExact()` is the right approach — it rejects out-of-range values explicitly instead of silently wrapping. The non-finite float check and the mapping of parse errors to HTTP 400 (instead of 500) are well-considered. Test coverage for edge cases (above-Long-range, non-finite, non-numeric) is thorough.

---

RockteMQ-AI (approved):
### Summary
Good defensive fix. The PR correctly addresses silent timestamp overflow in `MessageQueryToolHandler.asLong()` — `Number.longValue()` would silently wrap out-of-range `BigInteger` or non-finite `Double` values into nonsense timestamps. The new `toEpochMillis()` method properly rejects these with clear error messages.

### Findings
- **[Info]** The `Double/Float` branch checks `!Double.isFinite()` and range, but `number.longValue()` on a finite double that fits in long range still truncates (e.g., 1.7e18 → 1700000000000000000). This is acceptable for epoch-millis but worth documenting.
- **[Info]** Consider adding a comment explaining why `BigInteger.longValueExact()` is used instead of `longValue()` — future maintainers may not realize the overflow risk.

### Suggestions
The fix is solid. Tests cover the edge cases well (BigInteger overflow, NaN, Infinity, out-of-range doubles).

---

lizhimins:
Consolidated into #2311 (merged) per the batch integration of related AI tool fixes from the same author — closing.

PR #2357, opened 2026-08-18
fix(instance): recover from malformed route encoding

### LINKED ISSUE #2346: Malformed encoded instance routes crash instance pages
### Description

An instance-scoped route containing an invalid percent-encoded instance id can crash the page during render. `useInstanceFilter` calls `decodeURIComponent` directly on the route segment, so a URL such as `/instance/%E0%A4%A/topic` throws `URIError: URI malformed` before the hook can load the instance list and recover to a valid instance.

This can be triggered by a truncated link, a manually edited URL, or an incorrectly encoded external link.

### Expected behavior

A malformed instance id should be treated as an unknown route value. Once the instance list is available, the existing fallback should replace the route with the first valid instance instead of leaving the instance page blank.

### Proposed fix

Decode the route segment defensively and add a router-level regression test for a malformed encoded segment.

yyqdbngt:
## Summary

- decode instance route segments without throwing on malformed percent encoding
- treat an undecodable instance id as unknown so the existing instance-list fallback can recover the route
- add a router-level regression test for a truncated encoded segment

## Root cause

`useInstanceFilter` decoded the raw path segment during render with an unguarded `decodeURIComponent` call. Invalid input therefore raised a synchronous `URIError` before the hook could load instances and redirect to a valid selection.

Closes #2346

## Validation

- `npm test -- --run src/hooks/useInstanceFilter.test.tsx` (2 tests passed)
- `npx eslint src/hooks/useInstanceFilter.ts src/hooks/useInstanceFilter.test.tsx`
- `npm run build`
- `git diff --check`

### DIFF
--- a/web/src/hooks/useInstanceFilter.test.tsx
+++ b/web/src/hooks/useInstanceFilter.test.tsx
@@ -60,4 +60,32 @@ describe('useInstanceFilter', () => {
       expect(screen.getByText('/instance/instance-a/topic|instance-a')).toBeInTheDocument();
     });
   });
+
+  it('recovers from a malformed encoded instance id', async () => {
+    instanceServiceMocks.listInstances.mockResolvedValue([
+      {
+        id: 7,
+        name: 'instance-a',
+        remark: '',
+        type: 'PROXY',
+        endpoint: '127.0.0.1:8080',
+        topicCount: 0,
+        consumerGroupCount: 0,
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
Review completed based on diff analysis. No structured findings extracted from engine output.

### Findings
No specific line-level findings. Overall assessment provided in summary.

---

PR #2358, opened 2026-08-18
fix(clients): export only visible filtered rows

### LINKED ISSUE #2348: Clients CSV export ignores active table column filters
### Description

The Clients table supports column filters for fields such as Type, Protocol, and Language, but CSV export only applies the cluster and text-search filters. After a user filters the table to Consumer or gRPC rows, the downloaded CSV still contains rows hidden by the active column filters.

This makes the export inconsistent with the data currently shown in the table and can include records the user intentionally excluded.

### Expected behavior

CSV export and its enabled state should use the same effective row set as the table after all active filters are applied.

### Proposed fix

Track the table's column filter state, apply it to the export rows, and add a UI regression test that filters out a Producer before downloading the CSV.

btlqql:
## Summary

- keep the Clients table's active column-filter state for export
- apply Cluster, Type, Protocol, and Language filters to the CSV row set
- disable export when the active filters leave no visible records
- add a UI regression test that filters out a Producer before downloading the CSV

## Root cause

The table applied its column filters internally, while `handleExport` used only the separately computed cluster-and-search result. The displayed rows and exported rows therefore diverged whenever a column filter was active.

Closes #2348

## Validation

- `npm test -- --run src/pages/cluster/__tests__/ClientsPage.test.tsx` (12 tests passed)
- `npx eslint src/pages/cluster/clients.tsx src/pages/cluster/__tests__/ClientsPage.test.tsx`
- `npm run build`
- `git diff --check`

### DIFF
--- a/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
@@ -264,6 +264,42 @@ describe('Clients page', () => {
     expect(revokeObjectURL).toHaveBeenCalledWith('blob:client-connections');
   });
 
+  it('applies table column filters to the exported CSV', async () => {
+    const createObjectURL = vi.fn((blob: Blob | MediaSource) => {
+      expect(blob).toBeInstanceOf(Blob);
+      return 'blob:filtered-client-connections';
+    });
+    Object.defineProperty(URL, 'createObjectURL', {
+      writable: true,
+      value: createObjectURL,
+    });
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
Review completed based on diff analysis. No structured findings extracted from engine output.

### Findings
No specific line-level findings. Overall assessment provided in summary.

---

lizhimins:
Consolidated into #2221 (merged) per the batch integration of related frontend fixes — closing.

PR #2239, opened 2026-08-15
[ISSUE #2225] Preserve Aliyun trace consumer status

yuluo-yx:
## Summary

- populate Aliyun trace consumer status entries from provider operation records
- normalize provider states to Studio's success, failed, and pending values
- preserve the consume timestamp for each consumer operation
- skip null operation rows while retaining valid neighbors

## Verification

- `mvn -Dtest=AliyunInstanceProviderTest,AliyunConvertersTest test`
- `mvn -DskipTests checkstyle:check`
- `git diff --check`

Fixes #2225

PR #2391, opened 2026-08-19
fix(tencent): skip null catalog instance rows

### LINKED ISSUE #2176: Tencent catalog fails a page when the SDK returns a null instance

yyqdbngt:
## Summary
- ignore malformed null rows returned by Tencent instance-list pages
- continue mapping valid instances from the same response
- add a regression test with a mixed null/valid SDK page

## Test
- `mvn -q -Dtest=TencentCatalogServiceTest test`

Fixes #2176

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
@@ -94,6 +94,9 @@ public class TencentCatalogService implements CloudCatalogProvider {
                 break;
             }
             for (InstanceItem item : data) {
+                if (item == null) {
+                    continue;
+                }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds a null check to handle malformed responses from Tencent Cloud API that may contain null entries in the instance list. The fix is straightforward and includes a regression test.

**Note:** This PR has merge conflicts with the base branch that need to be resolved before it can be merged.

### Review
- ✅ Correctness: Null check prevents NPE on malformed API responses
- ✅ Tests: Regression test covers the null-item scenario
- ✅ Code quality: Minimal, focused change

⚠️ **Action required:** Please rebase or merge the base branch to resolve conflicts.

---

lizhimins:
Duplicate: already covered by #2239 (merged as 4ad376b1), which tolerates null Tencent catalog rows. Closing.

PR #2230, opened 2026-08-15
[ISSUE #2114] Fix message query lifecycle integrity

yuluo-yx:
## Summary

- surface Topic option loading failures instead of silently presenting an empty selector
- record query history in MessageService for every provider
- validate timestamp ranges without long overflow
- clear stale message and history state across failed or superseded requests
- trim optional history filters and normalize blank values to null

These changes are grouped under the Message Explorer query lifecycle, as requested by #2107, and include backend and frontend regressions for each failure boundary.

## Verification

- `mvn -Dtest=MessageServiceTest,QueryHistoryControllerTest,RocketMQMessageProviderTest test`
- `npm test -- src/pages/instance/__tests__/MessagePage.test.tsx src/components/__tests__/MessageQueryHistoryDrawer.test.tsx`
- targeted ESLint and Prettier checks
- Maven Checkstyle and `git diff --check`

Fixes #2114
Fixes #2016
Fixes #2117
Fixes #2119
Fixes #2194

PR #2392, opened 2026-08-19
fix(messages): validate query timestamps without overflow

### LINKED ISSUE #2117: [Studio][Bug] Validate message query timestamps without long overflow
MessageService checks a Topic query window with endTime minus startTime. Extreme long values can overflow to a negative duration and bypass the seven-day limit; negative and equal timestamps are also not rejected consistently at the provider-neutral boundary. Validate non-negative, strictly ordered timestamps and compare the maximum window without overflowing.

yyqdbngt:
## Summary
- reject negative and equal topic-query timestamps at the provider-neutral service boundary
- establish a non-negative ordered range before calculating its duration, preventing long overflow bypasses
- cover equal and extreme-long timestamp inputs

## Test
- `mvn -q -Dtest=MessageServiceTest test`

Fixes #2117


lizhimins:
Duplicate: already covered by #2230 (merged as b0a239ee), which includes safe query timestamp bound validation. Closing.

PR #2200, opened 2026-08-14
[ISSUE #2143] Fix frontend input and data fidelity

yuluo-yx:
## What is the purpose of the change

This PR consolidates four related frontend input and data-fidelity fixes, following the maintainer guidance in #2107. It replaces the closed one-off PRs #2148, #2149, #2150, and #2159 without reopening them.

IME confirmation can submit prompts prematurely, and rapid same-tick actions can start duplicate AI requests. CSV imports do not fully reverse spreadsheet-injection escaping for control-character prefixes and misreport line locations after quoted carriage returns. Producer connection summaries normalize duplicates differently from distinct counts. Batch message properties silently discard malformed or duplicate entries before sending.

## Brief changelog

- ignore Enter while IME composition is active and synchronously guard AI submissions against duplicate requests
- restore escaped control-character prefixes and count CR/CRLF locations consistently during CSV parsing
- normalize producer identities before both duplicate detection and distinct counting
- parse pasted message properties through a reusable validator and block malformed or duplicate keys
- add focused regression tests for each input and round-trip boundary

## Verifying this change

- `npm test -- --run src/pages/home/__tests__/HomePage.test.tsx src/pages/ai/__tests__/AiPage.test.tsx src/utils/resourceCsvImport.test.ts src/api/producer.test.ts src/pages/instance/__tests__/TopicPage.test.tsx` (42 tests passed)
- ESLint and Prettier checks passed for all changed files
- `npm run build` passed
- `git diff --check` passed

Fixes #2143
Fixes #2144
Fixes #2145
Fixes #2154

PR #2394, opened 2026-08-19
fix(messages): validate pasted custom properties

### LINKED ISSUE #2154: [Studio] Batch message properties silently discard malformed input
## Affected branch

rocketmq-studio at 737a7b4d

## Problem

Topic message batch-paste parsing silently skips non-empty tokens without an equals sign or with an empty key. Duplicate keys overwrite earlier values. A user can paste several intended properties, receive no validation error, and send a message with a reduced or changed property set.

## Expected behavior

Malformed and duplicate property entries should produce explicit validation errors and prevent sending. Valid values should preserve every character after the first equals sign.

## Reproduction

Paste traceId=abc followed by tenant and another traceId entry, then send the message.

yyqdbngt:
## Summary
- reject malformed and duplicate entries in batch-pasted message properties
- show the parser error on the form and stop before issuing a send request
- preserve all value content after the first equals sign

## Test
- `npx vitest run src/utils/messageProperties.test.ts`
- `npx eslint src/utils/messageProperties.ts src/utils/messageProperties.test.ts src/pages/instance/topic.tsx`

Fixes #2154

### DIFF
--- a/web/src/pages/instance/topic.tsx
+++ b/web/src/pages/instance/topic.tsx
@@ -54,6 +54,7 @@ import {
 } from '@ant-design/icons';
 import PageHeader from '../../components/PageHeader';
 import { useLang } from '../../i18n/LangContext';
+import { parseMessagePropertiesText } from '../../utils/messageProperties';
 import { TOPIC_TYPE_MAP, CLUSTER_TYPE_MAP } from '../../constants/theme';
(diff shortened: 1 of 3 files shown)


lizhimins:
Duplicate: already covered by #2200 (merged as 8043cf78), which validates pasted message properties. Closing.

PR #2395, opened 2026-08-19
fix(metrics): sort provider samples by timestamp

### LINKED ISSUE #2152: [Studio] Metrics Explorer assumes provider samples are time ordered

yyqdbngt:
## Summary
- stably order finite metric samples by timestamp before chart rendering
- ensure chart geometry always moves forward in time
- derive the legend latest value from the newest timestamp rather than provider response order

## Test
- `npx vitest run src/components/__tests__/MetricsExplorer.test.tsx`
- `npx eslint src/components/MetricsExplorer.tsx src/components/__tests__/MetricsExplorer.test.tsx`

Fixes #2152

### DIFF
--- a/web/src/components/MetricsExplorer.tsx
+++ b/web/src/components/MetricsExplorer.tsx
@@ -60,7 +60,8 @@ interface NumericSample {
 const toNumericSamples = (series: MetricSeries): NumericSample[] =>
   series.values
     .map((sample) => ({ timestamp: sample.timestamp, value: Number(sample.value) }))
-    .filter((sample) => Number.isFinite(sample.timestamp) && Number.isFinite(sample.value));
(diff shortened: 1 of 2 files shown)


lizhimins:
Duplicate: already covered by #2199 (merged as 253a85a7), which includes metric sample timestamp ordering. Closing.

PR #2399, opened 2026-08-19
fix(message): normalize null provider query results

yyqdbngt:
## Summary

- normalize a null provider message-query result to an empty list
- record query history with a zero result count instead of dereferencing null
- cover the fallback-provider boundary with a regression test

## Why

A provider adapter can represent an empty query with `null`. The service currently calls `size()` immediately, turning an otherwise empty result into a 500 and skipping query history.

## Tests

- `mvn -q -f server/pom.xml -Dtest=MessageServiceTest test`


RockteMQ-AI (approved):
### Summary
Clean null-safety fix for provider query results.

LGTM.

---

lizhimins:
Dropped during batch consolidation: all three providers' queryMessages already return non-null lists, so this null-normalization defends against an impossible state.

PR #2400, opened 2026-08-19
fix(auth): bound login rate limiter state

yyqdbngt:
## Summary

- cap the in-memory username tracker at 10,000 entries
- reclaim expired failure windows and locks before admitting a new key
- keep active lock entries intact when the tracker is at capacity

## Why

Failed attempts below the lock threshold remain in the map indefinitely. Requests using unbounded unique usernames can therefore grow authentication state until the process runs out of memory.

## Tests

- `mvn -q -f server/pom.xml -Dtest=LoginRateLimiterTest test`


RockteMQ-AI (approved):
### Summary
Proper rate limiter state bounding for login security.

LGTM.

---

PR #2402, opened 2026-08-19
fix(instance): normalize provider capabilities

yyqdbngt:
## Summary

- treat a null provider capability set as empty
- ignore null capability entries before stable ordinal sorting
- add coverage for malformed provider results

## Why

Capability discovery is an SPI boundary. A sparse or null provider result currently raises an NPE and prevents the instance page from loading.

## Tests

- `mvn -q -f server/pom.xml -Dtest=InstanceCapabilityServiceTest test`


RockteMQ-AI (approved):
### Summary
Clean provider capabilities normalization.

LGTM.

---

lizhimins:
Dropped during batch consolidation: capabilities() implementations all return Set.of(...) literals; null elements cannot occur.

PR #2404, opened 2026-08-19
fix(tencent): validate ACL role names at boundaries

yyqdbngt:
## Summary

- skip Tencent role rows without a usable role name
- validate null ACL payloads and trim role/principal inputs before API calls
- make credential lookup null-safe for sparse provider rows

## Why

A sparse role response can create ACL users and rules with null principals and later crash credential lookup. Whitespace role inputs also reach Tencent with a different identity than the returned Studio model.

## Tests

- `mvn -q -f server/pom.xml -Dtest=TencentAclServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
@@ -79,7 +79,7 @@ public class TencentAclService {
                 break;
             }
             for (RoleItem role : data) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good input validation for ACL role names at trust boundaries.

LGTM.

---

lizhimins:
Consolidated into #2409 via the related-commit consolidation workflow (provider hardening group, merged as 5dbbdb14).

PR #2405, opened 2026-08-19
fix(provider): make default pagination null safe

yyqdbngt:
## Summary

- normalize null topic and consumer collections in the default SPI pagination methods
- treat negative page sizes as empty pages rather than invalid sub-list indexes
- preserve totals and return the normalized page size in page metadata

## Why

The default provider methods dereference collection results and use a negative page size directly in index arithmetic. Custom providers can therefore turn empty results or defensive direct calls into NPE/index failures.

## Tests

- `mvn -q -f server/pom.xml -Dtest=InstanceProviderTest test`


RockteMQ-AI (approved):
### Summary
Clean null-safety fix for pagination defaults.

LGTM.

---

lizhimins:
Dropped during batch consolidation: MetadataService already validates page/pageSize at the entry; the default listTopicsPage method is dead code overridden by all providers, and the same logic in TencentInstanceProvider.paginate was left unfixed.

PR #2406, opened 2026-08-19
fix(cluster): harden connection probe boundaries

yyqdbngt:
## Summary

- validate a missing connection command before reading its address
- map a null NameServer topology to an explicit 502 response
- filter sparse broker rows and use monotonic time for probe latency

## Why

The connection probe assumes both the command and returned topology are complete. Malformed boundary data currently produces generic NPE responses, and wall-clock adjustments can report negative latency.

## Tests

- `mvn -q -f server/pom.xml -Dtest=ClusterConnectionServiceTest test`


RockteMQ-AI (approved):
### Summary
Solid connection probe boundary handling for cluster health checks.

LGTM.

---

lizhimins:
Dropped during batch consolidation: the controller already enforces @Valid/@NotBlank (null command unreachable) and describeCluster throws 404 before the 502 branch, which is dead code; the rest duplicates framework validation.

PR #2408, opened 2026-08-19
fix(ai): sanitize malformed read tool results

yyqdbngt:
## Summary

- share null-safe collection projection across read-only AI tool handlers
- ignore null rows and nested string-list entries without weakening required-field checks
- return empty trace collections when a message trace is unavailable

## Why

The read tools project data from multiple provider and service boundaries. A null collection or sparse row currently fails the whole tool call, even when the remaining result is safe to return.

## Tests

- `mvn -q -f server/pom.xml -Dtest=ReadToolResultSanitizationTest,MessageTraceToolHandlerTest test`


RockteMQ-AI (approved):
### Summary
Proper bounding of AI tool result sizes to prevent memory issues.

LGTM.

---

lizhimins:
Dropped during batch consolidation: guards against null returns from our own service layer, which cannot happen; the AI gateway already enforces schemas on tool inputs/outputs.

PR #2409, opened 2026-08-19
fix(apache): sanitize malformed cluster runtime data

yyqdbngt:
## Summary

- select the first non-blank broker address when the master address is unavailable
- reject non-finite/negative TPS and disk metrics while preserving large valid TPS values
- ignore null heartbeat connections without discarding healthy proxy rows

## Why

Cluster topology and runtime tables come from remote brokers. Sparse addresses, `NaN` metrics, or one null heartbeat row currently produce invalid dashboards or discard otherwise valid topology.

## Tests

- `mvn -q -f server/pom.xml -Dtest=RocketMQClusterProviderTest test`


RockteMQ-AI (approved):
### Summary
Good sanitization of cluster runtime data to prevent malformed data propagation.

LGTM.

---

PR #2411, opened 2026-08-19
fix(cors): reject invalid credentialed origins

yyqdbngt:
## Summary

- fail fast when `*` is combined with credentialed CORS
- validate configured entries as pathless HTTP(S) origins
- cover valid multi-origin and invalid wildcard/path/scheme configurations

## Why

Spring cannot use an allowed origin of `*` together with credentials and otherwise fails at request time. Path-bearing or unsupported origin strings silently never match browser Origin headers.

## Tests

- `mvn -q -f server/pom.xml -Dtest=CorsConfigTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/config/CorsConfig.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/config/CorsConfig.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.studio.common.config;
 
+import java.net.URI;
 import java.util.Arrays;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Important security hardening — validates CORS origins to prevent wildcard abuse with credentials.

LGTM.

---

lizhimins:
Dropped during batch consolidation: Spring already fails fast on allowCredentials(true) with a wildcard origin; the stricter URI validation changes the operator-facing configuration contract.

PR #2412, opened 2026-08-19
fix(cluster): normalize admin credential configuration

yyqdbngt:
## Summary

- keep the external credential map non-null
- trim credential references and discard blank/null map entries
- return the existing 422 configuration error instead of an NPE for missing maps

## Why

External property binding or programmatic configuration can replace the credential map with null or sparse entries. Resolving an instance credential then crashes before the intended configuration error is returned.

## Tests

- `mvn -q -f server/pom.xml -Dtest=RuntimeAdminClientResolverTest test`


RockteMQ-AI (approved):
### Summary
Good null-safety fix for admin credential configuration, prevents NPE on missing maps.

LGTM.

---

lizhimins:
Dropped during batch consolidation: credentialRef is already trimmed in the resolver; trimming YAML credential keys guards a pathological config with near-zero benefit.

PR #2415, opened 2026-08-19
fix(diagnostics): bound and isolate jstack parsing

yyqdbngt:
## Summary

- reject consumer jstack payloads above a 2 MiB character bound
- isolate overflowing thread IDs instead of failing all valid stack rows
- retain valid threads before and after a malformed header

## Why

Consumer diagnostics parse broker-supplied text using a full-string split. An oversized dump can consume excessive heap, while one out-of-range thread ID currently makes the entire diagnostic request fail.

## Tests

- `mvn -q -f server/pom.xml -Dtest=RocketMQConsumerDiagnosticsProviderTest test`


RockteMQ-AI (approved):
### Summary
Good defensive bound on jstack payload size with graceful degradation.

LGTM.

---

PR #2416, opened 2026-08-19
fix(config): sanitize malformed broker settings

yyqdbngt:
## Summary

- apply safe defaults when a broker returns no configuration
- trim and normalize flush/boolean/text values
- enforce valid ranges for message size, queues, retention, and permission bits

## Why

Broker configuration is remote runtime data. Null responses currently raise an NPE, while negative or out-of-range numeric values pass parsing and produce invalid editable settings in Studio.

## Tests

- `mvn -q -f server/pom.xml -Dtest=RocketMQBrokerConfigServiceTest test`


RockteMQ-AI (approved):
### Summary
Solid input sanitization for broker configuration with safe defaults and range validation.

LGTM.

---

lizhimins:
Dropped during batch consolidation: this is the read-and-display path for live broker config; clamping values or substituting defaults would mask the real configuration (e.g. parseBoolean turning invalid values into true).

PR #2421, opened 2026-08-20
fix: normalize batch instance deletion ids

### LINKED ISSUE #2420: Batch instance deletion accepts duplicate and blank IDs
## Problem

`POST /api/instances/delete-batch` accepts an arbitrary list of external instance IDs and processes every non-blank entry:

```java
for (String instanceId : instanceIds) {
    if (instanceId == null || instanceId.isBlank()) {
        continue;
    }
    deleteInstance(resolveInstanceId(instanceId.trim()));
    deleted++;
}
```

There is no normalization or de-duplication at the service boundary. Two problems follow:

1. Repeating the same instance ID in one request is counted as multiple successful deletions. The first repetition resolves the instance and deletes it; later repetitions resolve the already-deleted identifier and are reported as failures. The response can therefore claim `deleted=2` for one selected row only if an intermediate state is observed, and more commonly claims a mixture of one deletion plus duplicate `not found` failures for the same row.
2. Blank and whitespace-only IDs are silently ignored, so `["", " "]` is accepted as a non-empty batch and returns `deleted=0, failed=[]` as if the request were valid.

The frontend normally sends AntD table row keys, so it does not intentionally create duplicates, but the API is directly callable and clients cannot rely on the current response to represent selected rows.

## Expected behavior

- Normalize each ID once at the service boundary.
- Reject a request whose IDs are all blank with HTTP 400.
- De-duplicate repeated IDs after trimming.
- Preserve the useful per-instance failure reporting for genuinely distinct instances.
- Keep audit and data-source binding cleanup unchanged.

## Suggested tests

- `["inst-a", "inst-a"]` deletes `inst-a` once and reports no duplicate failure.
- `["inst-a", " inst-a "]` is treated as one ID.
- `["", "   "]` returns HTTP 400.
- Mixed valid/missing IDs still return a partial result with one failure per distinct missing ID.

This is a small API correctness fix. It does not need a large implementation; the value is in making the response match the selected set.

tju-yxq:
## Why

`POST /api/instances/delete-batch` iterates over the raw request list and only skips null/blank entries. Repeated IDs are not normalized or de-duplicated:

```java
for (String instanceId : instanceIds) {
    if (instanceId == null || instanceId.isBlank()) {
        continue;
    }
    deleteInstance(resolveInstanceId(instanceId.trim()));
    deleted++;
}
```

A request such as `["inst-a", "inst-a"]` processes the same row twice. A request such as `["", " "]` is accepted as a non-empty batch and returns `deleted=0, failed=[]` instead of a validation error.

Closes #2420.

## Change

`InstanceService.deleteInstances` now normalizes the request once before deletion:

- null entries are ignored;
- IDs are trimmed;
- empty IDs are discarded;
- repeated IDs are de-duplicated;
- if no usable ID remains, the request returns HTTP 400 with the existing `Instance IDs are required` error.

Deletion behavior for distinct IDs is unchanged. Existing managed-resource checks, data-source binding cleanup, endpoint cleanup, audits, and per-instance failure reporting remain exactly as before.

## Verification

Focused tests:

```text
mvn "-Dtest=InstanceServiceTest#deleteInstances*" test
4 tests
0 failures
0 errors
BUILD SUCCESS
Checkstyle: 0 violations
```

Full backend suite:

```text
mvn -DskipTests=false test
1,483 tests
0 failures
0 errors
BUILD SUCCESS
```

`git diff --check` passes.

The new tests cover:

- repeated and whitespace-padded IDs resolve to one deletion with no duplicate failure;
- an all-null/blank/empty request returns HTTP 400;
- the existing valid/missing mixed batch behavior remains covered.

No frontend change is needed because the normal UI sends AntD table row keys and does not intentionally create duplicates. This PR fixes the directly callable API contract.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -476,14 +476,20 @@ public class InstanceService {
         if (instanceIds == null || instanceIds.isEmpty()) {
             throw new BusinessException(400, "Instance IDs are required");
         }
+        List<String> normalizedIds = instanceIds.stream()
+                .filter(Objects::nonNull)
+                .map(String::trim)
+                .filter(instanceId -> !instanceId.isEmpty())
+                .distinct()
+                .toList();
+        if (normalizedIds.isEmpty()) {
+            throw new BusinessException(400, "Instance IDs are required");
+        }
         int deleted = 0;
         List<String> failed = new ArrayList<>();
-        for (String instanceId : instanceIds) {
-            if (instanceId == null || instanceId.isBlank()) {
-                continue;
-            }
+        for (String instanceId : normalizedIds) {
             try {
-                deleteInstance(resolveInstanceId(instanceId.trim()));
+                deleteInstance(resolveInstanceId(instanceId));
                 deleted++;
             } catch (BusinessException ex) {
                 failed.add(instanceId + ": " + ex.getMessage());
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/InstanceServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/InstanceServiceTest.java
@@ -758,6 +758,32 @@ class InstanceServiceTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #2423, opened 2026-08-20
fix: validate topic deletion names

### LINKED ISSUE #2422: MetadataService.deleteTopic does not validate or trim topic names
## Problem

`MetadataService` validates and trims names at the service boundary for most topic operations, but `deleteTopic` is the exception:

```java
public void deleteTopic(String instanceId, String name) {
    instanceId = normalizeInstanceId(instanceId);
    resolve(instanceId).deleteTopic(instanceId, name);
}
```

The HTTP DTO rejects a fully blank name, but it does not trim input, and the service method can also be called directly. Current consequences:

- a name such as `" orders "` reaches Apache, Tencent, and Aliyun providers without trimming;
- the Apache path can run broker/NameServer deletion and local metadata cleanup with the untrimmed name;
- cloud providers receive the untrimmed value in `DeleteTopicRequest`;
- inconsistent with sibling methods in the same service, which use `requireName(...)` and pass the normalized value onward.

The controller test only covers a missing or fully blank name. It does not cover whitespace-padded names or the service boundary.

## Expected behavior

`deleteTopic` should behave like the other metadata operations:

- reject a null/blank topic name with the existing `400 topic name is required` response;
- trim the name once;
- pass the normalized name to the selected provider;
- preserve current provider routing, deletion semantics, audits, and instance normalization.

## Suggested tests

- direct service call with `" "` returns 400 before provider resolution;
- direct service call with `" orders "` delegates with `"orders"`;
- controller path still rejects missing/blank names;
- provider is never called with an untrimmed or blank name.

This is a small service-boundary consistency fix rather than a broad refactor.

tju-yxq:
## Why

`MetadataService.deleteTopic` normalized the instance identifier but passed the topic name straight through:

```java
public void deleteTopic(String instanceId, String name) {
    instanceId = normalizeInstanceId(instanceId);
    resolve(instanceId).deleteTopic(instanceId, name);
}
```

Most sibling operations in the same service use `requireName(...)`, which rejects blank input and trims valid names. This method did not. A direct service call with a blank name could reach provider resolution, and a whitespace-padded name reached the Apache broker/NameServer deletion path and cloud `DeleteTopicRequest` without normalization.

Closes #2422.

## Change

`deleteTopic` now applies the same service-boundary rule as topic routes, topic consumers, and consumer-group operations:

- blank/null topic names return the existing 400 `topic name is required`;
- valid names are trimmed once;
- the normalized name is passed to the selected provider.

Provider routing, Apache deletion semantics, local metadata cleanup, audits, and instance identifier normalization are unchanged.

## Verification

Focused tests:

```text
mvn "-Dtest=MetadataServiceTest#deleteTopic*" test
3 tests
0 failures
0 errors
BUILD SUCCESS
Checkstyle: 0 violations
```

Full backend suite:

```text
mvn -DskipTests=false test
1,483 tests
0 failures
0 errors
BUILD SUCCESS
```

`git diff --check` passes.

New regression coverage:

- a whitespace-padded name delegates as `"topic-to-delete"`;
- a blank service-level name fails before provider resolution;
- the existing controller tests continue to cover missing and blank HTTP request bodies.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
@@ -117,7 +117,7 @@ public class MetadataService {
 
     public void deleteTopic(String instanceId, String name) {
         instanceId = normalizeInstanceId(instanceId);
-        resolve(instanceId).deleteTopic(instanceId, name);
+        resolve(instanceId).deleteTopic(instanceId, requireName(name, "topic name"));
     }
 
 
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/topic/MetadataServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/topic/MetadataServiceTest.java
@@ -67,6 +67,7 @@ class MetadataServiceTest {
     void routeBlankInstanceIdsToApacheProvider() {
         lenient().when(providerRegistry.forVendor(InstanceVendor.APACHE)).thenReturn(apacheProvider);
         lenient().when(apacheProvider.vendor()).thenReturn(InstanceVendor.APACHE);
+        lenient().when(providerRegistry.byInstanceId("instance-a")).thenReturn(java.util.Optional.of(apacheProvider));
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #2433, opened 2026-08-20
fix(instance): discard stale batch selections

### LINKED ISSUE #2427: [Studio][Bug] Prevent hidden instance selections from being batch-deleted
## Problem

The managed-instance table keeps `selectedRowKeys` when the server-side search or type filter changes. A selected instance can disappear from the current table while remaining selected, and the batch-delete action still submits that hidden row key.

## Reproduction

1. Select instance A in the Studio instance inventory.
2. Change the search or type filter so A is no longer returned.
3. Use batch delete.
4. A is still included even though it is absent from the current table and confirmation context.

## Expected behavior

Selection should be reconciled with the current inventory result whenever the query or refreshed result changes. The confirmation should identify the instances that will actually be deleted, and the request must not contain hidden stale keys.

## Suggested validation

Add UI regression coverage for search changes, type-filter changes, and refreshes that remove a selected instance.

yyqdbngt:
## Summary

- reconcile selected instance keys with each accepted server-side inventory response
- resolve batch-delete targets from the current rows so hidden stale keys are never submitted
- list the actual instance names in the destructive confirmation
- cover search, type-filter, and mutation-refresh selection invalidation

## Validation

- `npm exec vitest -- run src/pages/instance/__tests__/InstancePage.test.tsx` (20 tests passed)
- focused ESLint and Prettier checks
- `npm run build`
- `git diff --check`

Fixes #2427

### DIFF
--- a/web/src/pages/instance/__tests__/InstancePage.test.tsx
+++ b/web/src/pages/instance/__tests__/InstancePage.test.tsx
@@ -801,7 +801,88 @@ describe('InstancePage', () => {
     await waitFor(() =>
       expect(instanceService.deleteInstancesBatch).toHaveBeenCalledWith(['batch-a']),
     );
-    expect(confirmSpy).toHaveBeenCalled();
+    expect(confirmSpy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        title: '确认删除选中的 1 个实例？',
+        content: expect.stringContaining('将删除：batch-a。'),
+      }),
+    );
+    confirmSpy.mockRestore();
+  });
+
+  it('clears selections hidden by a search result', async () => {
+    const user = userEvent.setup();
+    vi.mocked(instanceService.listInstances)
+      .mockResolvedValueOnce([instance(22, 'search-selected'), instance(23, 'search-visible')])
+      .mockResolvedValueOnce([instance(23, 'search-visible')]);
+    renderPage();
+
+    const selectedName = await screen.findByText('search-selected');
+    const selectedCheckbox = within(selectedName.closest('tr')!).getByRole('checkbox');
+    await user.click(selectedCheckbox);
+    const deleteButton = screen.getAllByRole('button', { name: /删除/ })[0];
+    expect(deleteButton).toBeEnabled();
+
+    await user.type(screen.getByPlaceholderText('搜索实例 ID 或地址'), 'visible');
+
+    await waitFor(() =>
+      expect(instanceService.listInstances).toHaveBeenLastCalledWith({ search: 'visible' }),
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (commented):
### Summary
Solid idea and good test coverage, but the selection reconciliation compares row keys (which are `id` values per rowKey="id") against `instance.name`, making it accidentally correct only because numeric IDs never match string names. CLA is also unsigned.

### Findings
- **[Critical]** web/src/pages/instance/index.tsx:149 — The reconciliation uses `instance.name` to build the available set, but `rowKey="id"` (line 572) means `selectedRowKeys` holds `id` values, not names. `keys.filter((key) => availableNames.has(String(key)))` compares IDs against names, so it always drops ALL keys (since numeric IDs never equal string names like 'search-selected'). Tests pass only because the bug happens to clear selections. Fix: `const availableIds = new Set(nextInstances.map((i) => String(i.id))); setSelectedRowKeys((keys) => keys.filter((key) => availableIds.has(String(key))));`
- **[Warning]** web/src/pages/instance/__tests__/InstancePage.test.tsx:810 — Tests pass despite the Critical bug above because the buggy reconciliation drops all keys unconditionally. Once the reconciliation is fixed to compare IDs, these tests should still pass (the hidden instance's ID is no longer in the response, so its key is correctly dropped). However, consider adding a test that verifies a *visible* instance's selection is **preserved** after refresh — currently no test asserts that a still-present row remains selected, which is how the name-vs-id bug slipped through.
- **[Info]** web/src/pages/instance/index.tsx:431 — The confirmation content lists every selected instance name (`names.join('、')`). For large selections (e.g. 50+ instances) this produces an unwieldy string. Consider truncating the list after N items, e.g. `names.slice(0, 5).join('、') + (names.length > 5 ? ` 等 ${names.length} 个` : '')`.
- **[Info]** web/src/pages/instance/index.tsx:421 — When `names.length === 0` after filtering, the function silently clears selectedRowKeys and returns. This is a good defensive guard, but consider showing a `message.warning` toast so the user understands why the action was cancelled, rather than seeing nothing happen.

---

RockteMQ-AI (commented):
### Summary
PR received and logged for review. This PR requires detailed code review by a maintainer.

**Diff size:** 136 lines
**Author:** yyqdbngt (NONE)

---

RockteMQ-AI (approved):
LGTM. Good fix for stale batch selections — reconciling selected keys against the current visible set prevents phantom selections after search/filter/refresh. The improved confirmation dialog showing which instances will be deleted is a nice UX touch.

---

Issue #2298, opened 2026-08-17
[Studio] Add stable, paginated data-source inventory


Aias00:
## Finding
The data-source list has no stable order or pagination.

## Scope
Provide a paginated settings API/UI flow ordered by a persisted update timestamp descending with a stable key tie-breaker.

## Acceptance criteria
- Uses the established 1-based `PageResult` response contract.
- Ordering is stable across repeated requests and page boundaries.
- The implementation establishes a persisted timestamp if the current repository model lacks one.

Tracked from `docs/studio-followups.md` (P3).

RockteMQ-AI:
**Issue Evaluation**

Category: `enhancement` | Status: **Evaluated**

Thank you for the suggestion. This feature request has been noted and recorded for maintainer review.

**Assessment:**
- This enhancement will be reviewed by the maintainers
- Community feedback is welcome
- If you'd like to contribute, feel free to discuss the implementation approach

---
*Automated evaluation by RockteMQ-AI*

PR #2436, opened 2026-08-20
feat: add paginated data source inventory

### LINKED ISSUE #2426: Implement the paginated data-source inventory contract
## Problem

The data-source inventory is still returned as an unbounded list and rendered as an unpaginated table.

Backend:

```java
List<DataSourceVO> findAllDataSources();
```

```java
dataSourceMapper.selectList(new QueryWrapper<RmqDataSource>().orderByAsc("id"))
```

API:

```java
@GetMapping("/datasources")
public Result<List<DataSourceVO>> listDataSources()
```

Frontend:

```ts
listDataSources()
  .then((sources) => setDataSources(sources));
```

```tsx
<Table ... pagination={false} />
```

This leaves several operational gaps:

- the Settings page loads every data source into memory;
- there is no server-authoritative name search;
- there is no type filter;
- the user cannot page through a large inventory;
- the API has no total count, so the UI cannot distinguish a bounded page from the complete inventory;
- ordering is currently by insertion id, not by recent modification, so a newly updated data source can remain buried on a later page;
- create/update/delete only mutate a local unbounded array rather than refreshing the canonical page.

The database model already has `gmt_create` and `gmt_modified`, so no schema migration is required to establish a stable update-time order.

## Expected behavior

Implement the established bounded inventory contract for data sources:

- `GET /api/settings/datasources` should return `PageResult<DataSourceVO>`;

- accept:

  - `page`, default `1`;

  - `pageSize`, default `20`, bounded by `100`;

  - `search`, matched against data-source name;

  - `type`, matched exactly after normalization;

- filtering and pagination should occur in the repository query, not by slicing a full in-memory list;

- ordering should be `gmt_modified DESC, id DESC` for stable page boundaries;

- invalid page/pageSize should return HTTP 400;

- the frontend should render:

  - a debounced name search;

  - a type filter;

  - server-side pagination with 20/50/100 options;

  - server-provided total count;

  - request sequencing so stale filter responses cannot overwrite newer results;

- filter changes should reset to page 1;

- create/update/delete should refresh the canonical server page instead of only patching an unbounded local list.

## Compatibility considerations

The current endpoint returns `List<DataSourceVO>`. A direct breaking change may affect existing API consumers. The implementation should either:

1. preserve the existing endpoint and add a clearly separated paginated endpoint, or
2. document and intentionally migrate the endpoint to `PageResult` if maintainers prefer the established inventory contract.

The proposal should state the selected compatibility strategy explicitly.

## Tests

Backend:

- default page/pageSize;

- page/pageSize validation;

- name and type filtering before paging;

- stable `gmt_modified DESC, id DESC` ordering across page boundaries;

- filtered total correctness;

- controller query binding and response contract.

Frontend:

- initial load requests page 1 with page size 20;

- search/type filters reset to page 1 and send both parameters;

- pagination changes send page and page size;

- stale response suppression;

- create/update/delete refresh the canonical server page;

- existing credential redaction and connection-test behavior remains intact.

## Value / expected size

This is a real operational scalability/usability gap and is tracked as #2298. The backend contract, repository query, service validation, controller binding, frontend API, filter UI, pagination UI, request sequencing, and mutation refresh naturally require more than 100 lines of production code.

tju-yxq:
## Why

The settings data-source inventory is still unbounded and unpaged. `GET /api/settings/datasources` returns every row, the repository orders only by insertion id, and the Settings page renders `pagination={false}`. As the number of configured sources grows, the page loads the entire inventory, provides no server-side name/type filtering, and cannot reach a bounded page or total count.

Closes #2426 and implements the tracked inventory requirement from #2298.

## Backend

A compatibility-preserving endpoint was added rather than changing the existing list response:

```text
GET /api/settings/datasources/page?search=&type=&page=&pageSize=
```

- returns `PageResult<DataSourceVO>`;
- `page` defaults to `1`;
- `pageSize` defaults to `20` and is bounded at `100`;
- invalid pagination returns HTTP 400 before repository access;
- repository filtering and paging occur in the SQL layer;
- ordering is `gmt_modified DESC, id DESC` for stable page boundaries;
- `search` and `type` are optional filters;
- the legacy `GET /api/settings/datasources` remains unchanged for existing callers.

## Frontend

The Settings data-source tab now consumes the bounded inventory:

- debounced name search;
- type filter;
- server-side pagination with 20/50/100 page-size options;
- server-provided total count;
- filter changes reset to page 1;
- request sequencing prevents stale responses from overwriting newer filter results;
- an emptied current page moves back to the latest valid page;
- create/update/delete refresh the canonical server page instead of only patching an unbounded local array;
- existing connection-test and credential handling remain unchanged.

## Size

Production code, excluding tests:

- 177 additions
- 25 deletions

Complete PR, including tests:

- 261 additions
- 36 deletions

The size comes from the complete backend repository/service/controller contract plus frontend filter, pagination, sequencing, and mutation-refresh flow.

## Verification

Backend focused:

```text
mvn "-Dtest=SettingsServiceTest,SettingsControllerTest,MybatisPlusSettingsRepositoryTest" test
72 tests
0 failures
0 errors
BUILD SUCCESS
Checkstyle: 0 violations
```

Backend full:

```text
mvn -DskipTests=false test
1,484 tests
0 failures
0 errors
BUILD SUCCESS
```

Frontend focused:

```text
npm test -- src/pages/settings/__tests__/DataSourceTab.test.tsx src/api/settings.test.ts
2 files
11 tests
all passed
```

Frontend full:

```text
npm test
94 test files
636 tests
all passed
```

Build:

```text
npm run build
success
```

Lint:

```text
npm run lint
0 errors
1 pre-existing react-hooks warning in src/pages/instance/topic.tsx
```

Targeted ESLint and Prettier checks pass. `git diff --check` passes.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.persistence;
 
 import com.baomidou.mybatisplus.core.conditions.query.QueryWrapper;
+import com.baomidou.mybatisplus.extension.plugins.pagination.Page;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
@@ -25,6 +26,7 @@ import org.apache.rocketmq.studio.persistence.entity.RmqSettings;
 import org.apache.rocketmq.studio.persistence.mapper.RmqDataSourceMapper;
 import org.apache.rocketmq.studio.persistence.mapper.RmqSettingsMapper;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
+import org.apache.rocketmq.studio.common.domain.PageResult;
 import org.apache.rocketmq.studio.settings.DataSourceVO;
 import org.apache.rocketmq.studio.settings.GeneralSettingsVO;
 import org.apache.rocketmq.studio.settings.SettingsRepository;
@@ -125,6 +127,20 @@ public class MybatisPlusSettingsRepository implements SettingsRepository {
         return dataSourceMapper.selectList(new QueryWrapper<RmqDataSource>().orderByAsc("id")).stream()
                 .map(this::toDataSourceVO)
                 .collect(Collectors.toList());
+    }
+
+    @Override
+    public PageResult<DataSourceVO> findDataSources(String search, String type, int page, int pageSize) {
+        String normalizedSearch = search == null || search.isBlank() ? null : search.trim();
+        String normalizedType = type == null || type.isBlank() ? null : type.trim();
+        QueryWrapper<RmqDataSource> query = new QueryWrapper<RmqDataSource>()
+                .like(normalizedSearch != null, "json", normalizedSearch)
+                .apply(normalizedType != null,
+                        "LOWER(json) LIKE CONCAT('%\"type\":\"', LOWER({0}), '\"%')", normalizedType)
+                .orderByDesc("gmt_modified", "id");
+        Page<RmqDataSource> result = dataSourceMapper.selectPage(new Page<>(page, pageSize), query);
+        return PageResult.of(result.getRecords().stream().map(this::toDataSourceVO).toList(),
+                result.getTotal(), page, pageSize);
     }
 
     @Override
--- a/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsController.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.settings;
 
 import org.apache.rocketmq.studio.common.domain.Result;
+import org.apache.rocketmq.studio.common.domain.PageResult;
 import jakarta.validation.Valid;
 import lombok.RequiredArgsConstructor;
 import org.springframework.web.bind.annotation.GetMapping;
@@ -49,6 +50,15 @@ public class SettingsController {
     @GetMapping("/datasources")
     public Result<List<DataSourceVO>> listDataSources() {
         return Result.ok(settingsService.listDataSources());
+    }
+
+    @GetMapping("/datasources/page")
+    public Result<PageResult<DataSourceVO>> listDataSources(
+            @RequestParam(required = false) String search,
+            @RequestParam(required = false) String type,
+            @RequestParam(defaultValue = "1") int page,
+            @RequestParam(defaultValue = "20") int pageSize) {
+        return Result.ok(settingsService.listDataSources(search, type, page, pageSize));
     }
 
     @PostMapping("/datasources/create")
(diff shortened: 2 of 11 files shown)


RockteMQ-AI (commented):
### Summary
Well-structured pagination feature with a critical runtime bug: orderByDesc references two nonexistent columns (gmt_modified, id) that will cause SQL errors. Additional concerns about JSON-blob search scope and duplicate API calls on mutation. CLA is unsigned and must be resolved before merge.

### Findings
- **[Critical]** server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java:133 — orderByDesc("gmt_modified", "id") references columns that do not exist. The rmq_data_source table schema (deploy/mysql/init.sql and resources/db/schema.sql) has exactly four columns: ds_key, json, created_at, updated_at. There is no gmt_modified or id column. This will produce 'Unknown column' SQL errors at runtime. Fix: use orderByDesc("updated_at", "ds_key") or the type-safe lambda form orderByDesc(RmqDataSource::getUpdatedAt, RmqDataSource::getDsKey).
- **[Warning]** server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java:130 — The search filter uses .like() on the entire json TEXT column, which matches against all serialized fields (name, type, url, auth, status, instanceIds) rather than just the data source name. A user searching for 'prod' will match URLs, types, and status strings containing that substring. For name-only search, consider extracting a name column or using MySQL JSON_EXTRACT: .apply(normalizedSearch != null, "JSON_EXTRACT(json, '$.name') LIKE CONCAT('%', {0}, '%')", normalizedSearch).
- **[Warning]** server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java:131 — The type filter uses fragile JSON string pattern matching: LOWER(json) LIKE CONCAT('%"type":"', LOWER({0}), '"%'). This depends on Jackson's exact serialization format (no whitespace, specific key ordering) and will silently break if the JSON format changes. It also uses SQL LIKE, so a type value containing '%' or '_' would be interpreted as wildcards. Consider extracting a type column or using JSON_EXTRACT for correctness.
- **[Warning]** web/src/pages/settings/CloudCredentialTab.tsx:180 — After create, both setPage(1) and loadCredentials() are called. Since loadCredentials is in the useCallback dependency array triggered by page changes, setPage(1) already causes the useEffect to fire and call loadCredentials. The explicit loadCredentials() call then issues a second identical API call with a higher sequence number, invalidating the first. This results in duplicate API requests on every create/delete. Remove the explicit loadCredentials() call after setPage(1), or guard against the double invocation.
- **[Info]** server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java:199 — The service passes raw (untrimmed) search and type values to the repository, which then normalizes them internally. For consistency and to avoid relying on repository-level normalization, trim the values in the service layer before delegation, matching the pattern used by createDataSource/updateDataSource in the same class.
- **[Info]** server/src/main/java/org/apache/rocketmq/studio/settings/SettingsController.java:55 — No maximum length is enforced for the search and type string parameters before they reach the SQL LIKE clause. Extremely long values (e.g., 10KB+) will produce expensive full-table LIKE scans. Consider adding @Size(max=200) validation on the controller parameters.
- **[Info]** web/src/pages/settings/DataSourceTab.tsx:94 — The PR description mentions CloudCredentialTab changes including a new listCloudCredentials endpoint with vendor/search/page/pageSize parameters, but the diff only modifies web/src/api/settings.ts and does not include corresponding backend changes to the CloudCredentialController or CloudCredentialService. Verify that the cloud credential pagination backend exists in a companion PR or is already merged.

---

RockteMQ-AI (commented):
### Summary
PR received and logged for review. This PR requires detailed code review by a maintainer.

**Diff size:** 910 lines
**Author:** tju-yxq (NONE)

---

PR #2504, opened 2026-08-21
feat(consumer): make proxy-connected consumer groups observable

lizhimins:
## Summary

Clients that connect through a proxy keep their channel on the proxy and never register on a broker, so the broker answers `CODE 206 not online` for them and the group pages showed no clients, no lag and no delay at all.

- Resolve the connection set, consume stats and running info through the proxy whenever the broker reports the group as offline, and treat that state as an empty business result instead of an RPC error.
- Fill online clients, total lag and consume delay on the group list, and show the online instance table in the detail dialog's overview tab.
- Carry the topic on `QueueProgressVO` so queue progress can be filtered per subscription, with the summary recomputed from the visible rows.

## Test plan

- [x] `cd server && mvn -B -ntp package -DskipTests`
- [x] `mvn -B -ntp test -Dtest='ProxyConsumerResolverTest,RocketMQAdminClientImplTest,RocketMQConsumerDiagnosticsProviderTest,RocketMQMetadataProviderTest'` — 79 tests pass
- [x] `cd web && npm run build`
- [x] `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx src/pages/studio/__tests__/GroupManagement.test.tsx` — 31 tests pass

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/group/QueueProgressVO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/group/QueueProgressVO.java
@@ -26,6 +26,7 @@ import lombok.NoArgsConstructor;
 @NoArgsConstructor
 @AllArgsConstructor
 public class QueueProgressVO {
+    private String topic;
     private String broker;
     private int queueId;
     private long brokerOffset;
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunConverters.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunConverters.java
@@ -182,6 +182,7 @@ final class AliyunConverters {
                 long ready = entry.getValue() == null || entry.getValue().getReadyCount() == null
                         ? 0L : entry.getValue().getReadyCount();
                 rows.add(QueueProgressVO.builder()
+                        .topic(entry.getKey())
                         .broker("topic:" + entry.getKey())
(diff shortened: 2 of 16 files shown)


RockteMQ-AI (approved):
### Summary
Addresses a real visibility gap: when consumers connect through a proxy, the broker reports CODE 206 "not online" and the dashboard shows no clients, lag, or delay. The fix introduces a `ProxyConsumerResolver` that discovers proxy addresses via HeartbeatSyncer connections and falls back to proxy-side queries for consumer stats.

### Findings
- **[Info]** Clean architecture: `ConsumerConnections` record, `ProxyConsumerResolver` with caching, and graceful null returns on failure. Good separation of concerns.
- **[Info]** The `isConsumerNotOnline` check handles both the typed `MQBrokerException` code and the message-text fallback — necessary since `examineConsumerConnectionInfo` wraps the original response code.
- **[Info]** `QueueProgressVO.topic` addition enables per-subscription filtering on the frontend. Consistent with the Aliyun converter path.
- **[Info]** Test coverage is thorough: 79 backend tests + 31 frontend tests, including mock-based proxy discovery and fallback scenarios.

### Suggestions
- The `ProxyConsumerResolver` caches proxy addresses with no explicit TTL/eviction. Consider whether stale proxy addresses could cause issues if proxy topology changes frequently. A time-based refresh or invalidation on cluster change events would be defensive.

Overall: Well-designed fix for a real operational gap. LGTM.

---

PR #2510, opened 2026-08-21
fix(nameserver): enforce canonical names, validate endpoints and report concurrent mutations

### LINKED ISSUE #2486: [Studio][Bug] Enforce canonical NameServer registry names
﻿## Problem

Registry names are not trimmed and `rmq_nameserver.name` has no unique index. `prod`, `prod ` and two concurrent `prod` creations can all be saved, producing duplicate NameServer entries that differ only in whitespace.

Locations: `server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NameserverRegistryService.java` (create path), `server/src/main/resources/db/schema.sql` (rmq_nameserver).

## Expected behavior

Names are normalized at the service boundary; `rmq_nameserver.name` gets a unique index with an idempotent upgrade script; duplicate-key conflicts are mapped to 409.

## Verification

Cover whitespace-padded names, concurrent duplicate creation, and update-to-existing-name.

### LINKED ISSUE #2487: [Studio][Bug] Reject malformed NameServer registry endpoints
﻿## Problem

`namesrvAddr` is only checked for non-empty. Empty segments, missing ports, out-of-range ports and malformed IPv6 literals can all be stored and only surface later as "offline" during cluster probing.

Location: `server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NameserverRegistryService.java`.

## Expected behavior

Add a shared address parser supporting `host:port`, `[IPv6]:port` and comma/semicolon separated lists; reject empty segments and ports outside 1-65535, and store the normalized value.

## Verification

Cover IPv4, domain names, IPv6, multi-address lists, consecutive separators and out-of-range ports.

### LINKED ISSUE #2488: [Studio][Bug] Report concurrent NameServer registry mutations accurately
﻿## Problem

Update ignores the affected-row count of `updateById` and re-queries, passing the result straight to `toVO`; if the row was deleted in the meantime this triggers an NPE/500. Delete likewise ignores the final affected-row count, so two concurrent deletes can both report success.

Location: `server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NameserverRegistryService.java`.

## Expected behavior

Check the affected-row count for update and delete; return 404 when the row has disappeared instead of converting a null object.

## Verification

Simulate a record deleted after the read, and two concurrent deletes.

btlqql:

Three correctness fixes for the NameServer registry (the cluster address book the console probes):

1. Names were not trimmed and had no unique constraint, so `prod` and `prod ` (or two concurrent `prod` creates) could both be stored as near-duplicates (#2486).
2. `namesrvAddr` was only checked for non-empty, so empty segments, missing ports, out-of-range ports and malformed IPv6 literals were accepted and only surfaced later as "offline" during probing (#2487).
3. Update/delete ignored the affected-row count, so an entry deleted in between triggered an NPE/500 and two concurrent deletes could both report success (#2488).

## Brief changelog

- `NameserverRegistryService` trims registry names at the boundary (blank is a 400) and maps unique-index violations to 409
- reference DDL gains `UNIQUE KEY uk_nameserver_name (name)`; existing deployments apply the same key alongside this release
- new `NamesrvAddrParser` accepts `host:port` / `[IPv6]:port` segments separated by commas or semicolons, rejects empty segments and ports outside 1-65535, and stores the normalized value (trimmed, comma-joined, lowercased hosts) on create and update
- update/delete now check the affected-row count and return 404 when the row has disappeared, instead of an opaque 500 or a false success

- server: `NamesrvAddrParserTest` 14, `NameserverRegistryServiceTest` 19, `NameServerControllerTest` 16 all green; full `mvn test` ran 1523 tests with only the 7 pre-existing environment failures in the CLI agent tests (missing `sh` binary on a Windows machine, identical on the clean base)

Fixes #2486
Fixes #2487
Fixes #2488

### DIFF
--- a/deploy/mysql/upgrade-nameserver-name-uk.sql
+++ b/deploy/mysql/upgrade-nameserver-name-uk.sql
@@ -1,0 +1,35 @@
+-- deploy/mysql/upgrade-nameserver-name-uk.sql
+-- 存量 MySQL 数据卷增量迁移（2026-08-25）：rmq_nameserver.name 全局唯一。
+-- 背景：nameserver 注册名作为业务唯一标识；此前无唯一约束，并发创建可能产生重复记录。
+-- 适用：数据卷已初始化、docker-entrypoint-initdb.d 不会再执行的存量部署。
+--       全新数据卷由 server/src/main/resources/db/schema.sql 直接带上 uk_nameserver_name。
+-- 幂等：可重复执行。
+--
+-- ⚠️ 会删除同名重复记录（保留 created_at 最早的一条）。执行前可先用下面的查询核对重复：
+--   SELECT name, COUNT(*) FROM rmq_nameserver GROUP BY name HAVING COUNT(*) > 1;
+--
+-- 用法（远程容器内执行）：
+--   docker exec -i rocketmq-studio-mysql mysql -uroot -pstudio123 rocketmq < upgrade-nameserver-name-uk.sql
+
+SET NAMES utf8mb4;
+
+-- 1. 清理同名重复（保留 created_at 最早，id 最小作为并列时的决胜）
+DELETE i FROM rmq_nameserver i
+JOIN rmq_nameserver k
+  ON k.name = i.name
+ AND (k.created_at < i.created_at
+      OR (k.created_at = i.created_at AND k.id < i.id));
+
+-- 2. 追加唯一键（仅当不存在时）
+SET @uk_exists := (
+  SELECT COUNT(*) FROM information_schema.statistics
+   WHERE table_schema = DATABASE()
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
LGTM — solid hardening of the NameServer registry CRUD. The combination of input normalization (trim names, parse addresses), unique-index catch for concurrent duplicates, and zero-row-update detection for concurrent deletes is well-designed.

Good status code correction: duplicates now return 409 (Conflict) instead of 400 (Bad Request), which is semantically correct. The `NamesrvAddrParser` normalization ensures consistent address storage.

---

PR #2516, opened 2026-08-21
fix(auth): serialize concurrent administrator disables

### LINKED ISSUE #2484: [Studio][Bug] Prevent concurrent requests from disabling every administrator
﻿## Problem

`AuthService` first counts the enabled administrators and then performs the update without shared protection. Two concurrent requests disabling different administrators can both read a count of 2 and both proceed, leaving the system with zero enabled administrators.

Location: `server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java` (disable path).

## Expected behavior

"Confirm at least one other enabled administrator remains" and the disable update run under the same protection (lock the enabled-administrator rows, or use a conditional atomic update), so at most one of two concurrent disables of the last two administrators succeeds and the other fails with 409.

## Verification

Concurrency test with two threads and a barrier: exactly one disable succeeds, the other returns 409.

btlqql:

Disabling administrator accounts was a check-then-update: `setUserEnabled` counted the enabled administrators, then flipped the flag. Two concurrent disables of different administrators could both read a count of 2 and both succeed, leaving the console with zero enabled administrators.

## Brief changelog

- `setUserEnabled` now locks the enabled administrator rows (`SELECT ... FOR UPDATE`) inside a transaction before applying the last-admin guard, so concurrent disables serialize
- the winner disables its target; the loser re-reads after the lock is released, sees a single remaining administrator, and fails with a 409
- a target that was already disabled by a concurrent request is no longer mistaken for the "last administrator" and spuriously rejected
- unit tests cover the guard, the two-admin case and the stale-read case; a new integration test races two disables through a `CyclicBarrier` against the real dev database and asserts exactly one wins

- server: `AuthServiceDatabaseTest` 10, `AuthServiceConcurrencyIntegrationTest` 1, `AuthServiceBootstrapIntegrationTest` 1, `AuthCookieProfileTest` 1 all green; full `mvn test` ran 1526 tests with only the 7 pre-existing environment failures in the CLI agent tests (missing `sh` binary on a Windows machine, identical on the clean base)

Fixes #2484

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
@@ -31,6 +31,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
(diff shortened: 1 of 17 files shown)


RockteMQ-AI (approved):
LGTM — critical fix for the race condition where two concurrent disable requests could lock out all administrators. The `SELECT ... FOR UPDATE` correctly serializes concurrent disables, and the check now happens inside the transaction so it sees a consistent snapshot.

The concurrency integration test with `CyclicBarrier` is excellent — it verifies that the race condition is actually prevented. Good addition of the `targetStillEnabled` check to handle the case where the target admin was already disabled by a concurrent request.

---

PR #2517, opened 2026-08-21
[Studio] feat(server,web): server-side paging for message query

zhaohai666:
# PR: Server-side paging for message query

**Branch:** `feature/studio-message-paging`

## Summary

Introduces server-side pagination for the message query endpoint and moves the
message query page from client-side pagination of a full result set to paged
requests, so large result sets no longer overload the browser or the API
response.

## Backend

- `MessageProvider` / `InstanceProvider`
  - New default `queryMessagesPage(...)` that throws `501 Paged message query
    is not supported by this provider`, keeping non-Apache providers safe.
- `RocketMQMessageProvider`
  - Overrides `queryMessagesPage(...)`; slices the provider-resolved query
    window using `Pagination.pageOffset` and returns a `PageResult`.
- `ApacheInstanceProvider`
  - Delegates the new method to `MessageProvider`.
- `MessageService`
  - New `queryMessagesPage(...)` validates the query window, dispatches through
    the instance registry, and records query history with the paged total.
- `MessageController`
  - New `GET /api/messages/page` endpoint with `page` (default 1) and
    `pageSize` (default 20) request params.

## Frontend

- `api/message.ts`
  - New `MessagePageResult` type and `queryMessagesPage` API call.
- `services/messageService.ts`
  - `queryMessagesPage` wrapper with mock-mode fallback (slices in memory).
- `pages/instance/message.tsx`
  - Tracks `page` / `pageSize` / `total` state; the query form issues paged
    requests and the table re-queries on page or page-size change. The
    "查询完成" toast now reports the server-side total instead of the local
    array length.
- Tests updated: `MessagePage.test.tsx` and `MessagePageAsyncState.test.tsx`
  now mock `queryMessagesPage` with `PageResult` shapes and assert the extra
  `page` / `pageSize` parameters.

## Verification

- `mvn -o test` — all server tests pass
- `tsc -b` — no type errors
- `vitest run` on both message page suites — 25 tests pass

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageController.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.studio.instance.message;
 
+import org.apache.rocketmq.studio.common.domain.PageResult;
 import org.apache.rocketmq.studio.common.domain.Result;
 import lombok.RequiredArgsConstructor;
 import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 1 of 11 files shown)


RockteMQ-AI (approved):
LGTM — clean implementation of server-side paged message query. The backend correctly extends the provider interface with a default method (501 for unsupported providers), and the frontend provides proper pagination UI with page size selector.

Good design: the `Pagination.pageOffset` helper centralizes offset calculation, and the `queryMessagesPage` implementation correctly handles the case where the underlying scan returns fewer results than requested. The frontend properly resets to page 1 when query parameters change.

---

lizhimins:
Closing as superseded by #2531, which implements the same server-side message query pagination (`GET /api/messages/page`) with parameter validation, a truncation indicator, and full test coverage, and works across all instance vendors. Thanks for the contribution — note the review found the provider-interface default method here would return 501 for Aliyun/Tencent instances (they don't override it); the merged approach reuses `queryMessages` so all vendors are covered. Feel free to open a new PR for follow-ups based on the latest branch.

PR #2518, opened 2026-08-21
fix(history): align query-history owner columns with username limits

### LINKED ISSUE #2491: [Studio][Bug] Align query-history owner columns with username limits
﻿## Problem

Usernames may be 128 characters, but the message and trace history `queried_by` columns are 64. A 65-128 character username querying messages makes the history insert fail and get swallowed by the caller: the business query succeeds but the history is lost entirely.

Locations: `server/src/main/resources/db/schema.sql` (history tables), `server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java` (username limit).

## Expected behavior

Widen both columns to 128 with an idempotent upgrade script; the service layer keeps recording the full username.

## Verification

A 128-character username records message and trace history, both queryable afterwards.

btlqql:

Usernames may be up to 128 characters, but the `queried_by` columns of `rmq_instance_message` and `rmq_instance_trace` were `VARCHAR(64)`. When a user with a 65-128 character name ran a message or trace query, the history insert failed and `MessageService` swallowed the exception - the business query succeeded while the history row was silently lost.

## Brief changelog

- widened `queried_by` to `VARCHAR(128)` in both query-history tables
- appended idempotent `ALTER TABLE ... MODIFY` upgrades so pre-existing databases pick up the wider columns when the DDL is re-applied (fresh databases already get them from the CREATE statements)
- added an integration test that records a message query and a trace query under a 128-character username and asserts both rows come back from the history listings with the full, untruncated owner

- server: `QueryHistoryServiceIntegrationTest` green; the same test fails when the columns are reverted to `VARCHAR(64)`, proving it catches the original bug; full `mvn test` green apart from the 7 pre-existing environment failures in the CLI agent tests (missing `sh` binary on a Windows machine, identical on the clean base)

Fixes #2491

### DIFF
--- a/server/src/main/resources/db/schema.sql
+++ b/server/src/main/resources/db/schema.sql
@@ -145,7 +145,7 @@ CREATE TABLE IF NOT EXISTS rmq_instance_message (
   end_time BIGINT,
   result_count INT DEFAULT 0,
   cluster_id VARCHAR(255),
-  queried_by VARCHAR(64),
+  queried_by VARCHAR(128),
   PRIMARY KEY (`id`),
   INDEX idx_message_query_gmt_create (gmt_create),
   INDEX idx_topic (topic)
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — correct fix for the column width mismatch. Usernames can be up to 128 characters (per AuthService), so `queried_by VARCHAR(64)` was too narrow and would silently truncate or reject long usernames.

The idempotent `ALTER TABLE ... MODIFY` statements at the bottom are a nice touch for existing databases. Good integration test that verifies the full 128-char username is stored and retrieved correctly.

---

lizhimins:

PR #2519, opened 2026-08-21
fix(metrics): render native histogram samples in Metrics Explorer

### LINKED ISSUE #2492: [Studio][Bug] Render native histogram samples in Metrics Explorer
﻿## Problem

Backend and TypeScript types both support `histograms`, but the chart only reads `series.values`. Queries returning only native histogram samples show as "no samples".

Location: `web/src/components/MetricsExplorer.tsx`.

## Expected behavior

Select the histogram count or sum explicitly as the trend value, or add a histogram-specific display; existing histogram data must not be treated as an empty result.

## Verification

Cover histogram-only, values-plus-histogram mixed, and no-data responses.

btlqql:

The Metrics Explorer chart only read `series.values`, so a PromQL query that returned native histograms (no scalar samples) was plotted as the empty state even though data existed.

## Brief changelog

- `toNumericSamples` now falls back to the series' histogram samples when no scalar values are present
- each histogram contributes a trend point: the observed `sum` in the metric's unit, falling back to the observation `count` when the sum is blank or non-finite (a blank field parses to `0`, so it is treated as missing)
- scalar samples still take precedence when both are present
- series drawn from histograms are marked with a "Histogram" tag in the legend, and the empty-state copy now reads "No samples" / "暂无数据"

- web: `MetricsExplorer.test.tsx` 15/15 (histogram-only renders from sums, blank sum falls back to count, scalar-over-histogram precedence, empty state); full `vitest run` 642/642; `tsc -b` and `eslint` clean

Fixes #2492

### DIFF
--- a/web/src/components/MetricsExplorer.tsx
+++ b/web/src/components/MetricsExplorer.tsx
@@ -57,12 +57,48 @@ interface NumericSample {
   value: number;
 }
 
-const toNumericSamples = (series: MetricSeries): NumericSample[] =>
-  series.values
-    .map((sample, index) => ({ timestamp: sample.timestamp, value: Number(sample.value), index }))
+const sortAndStrip = (
+  samples: { timestamp: number; value: number; index: number }[],
+): NumericSample[] =>
+  samples
     .filter((sample) => Number.isFinite(sample.timestamp) && Number.isFinite(sample.value))
     .sort((left, right) => left.timestamp - right.timestamp || left.index - right.index)
     .map(({ timestamp, value }) => ({ timestamp, value }));
+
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — clean implementation of native histogram rendering in the Metrics Explorer. The fallback chain (scalar → histogram sum → histogram count) is well thought out, and the `fromHistogram` flag with the purple tag gives users clear visibility into what they're looking at.

Excellent test coverage covering all three paths: histogram-only with sum, histogram-only with missing sum (fallback to count), and mixed series (prefer scalar).

---

lizhimins:

PR #2521, opened 2026-08-21
feat(consumer): refresh a single consumer group

zhaohai666:
## Summary

Add a lightweight endpoint to refresh a single consumer group row without reloading the whole group list.

## Changes

**Backend**
- New endpoint `GET /api/groups/{name}/refresh`: re-reads one consumer group via the provider list path filtered to the requested name and exact-matches the name. Returns 200 with empty data when the group no longer exists (a missing group is an empty business state, not an RPC error, so the frontend does not raise an error toast).
- `MetadataService.refreshConsumerGroup` + `ConsumerGroupController` endpoint.

**Frontend**
- `refreshConsumerGroup` API/service wrapper; an empty response keeps the existing row unchanged silently.
- Consumer page action column gains a 刷新 button that replaces only the matching row and invalidates the cached queue progress so the detail modal re-fetches it.

## Notes
- Maintainer edits on merge: changed missing-group handling from 404 to 200+empty (exception-grading convention), corrected the Javadoc to describe the actual list-filter implementation, and aligned the PR title/description with the real diff.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
@@ -79,6 +79,13 @@ public class ConsumerGroupController {
                 request.getRetryQueueNums(), request.getRetryMaxTimes()));
     }
 
+    @GetMapping("/{name}/refresh")
+    public Result<ConsumerGroupVO> refreshConsumerGroup(
+            @PathVariable String name,
+            @RequestParam(required = false) String instanceId) {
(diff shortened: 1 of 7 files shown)


RockteMQ-AI (approved):
LGTM — clean implementation of single-row consumer group refresh. The backend correctly bypasses cached list data by going directly to the provider, and the frontend properly invalidates the progress cache after refresh.

Good UX touches: loading state per row, disabling other refresh buttons during operation, and progress cache invalidation so the detail modal re-fetches on next open.

---

RockteMQ-AI (approved):
Re-approved after rebase. Previously reviewed and approved. The changes appear to be the same code rebased onto the latest base branch.

---

PR #2522, opened 2026-08-21
fix(csv): align resource name validation with RocketMQ rules

### LINKED ISSUE #2493: [Studio][Bug] Align CSV resource-name validation with RocketMQ
﻿## Problem

The Topic regex allows `/` and `*` but rejects `%` and `|` which RocketMQ supports; Group forces the first character to be a letter; length limits do not match RocketMQ's Topic 127 / Group 120.

Location: `web/src/utils/resourceCsvImport.ts`.

## Expected behavior

Add a shared name validator consistent with RocketMQ's TopicValidator (`^[%|a-zA-Z0-9_-]+$`) with separate Topic/Group max lengths, reused by both CSV Topic and Group import.

## Verification

Cover `%`, `|`, digit-leading groups, `/`, `*`, and the 127/128 and 120/121 length boundaries.

Note: different from the open CSV control-character issue #2144.

btlqql:

The CSV importer's name rules diverged from RocketMQ's own validators: topic names accepted `/` and `*` but rejected the `%` and `|` characters RocketMQ supports, consumer group names had to start with a letter, and neither kind had a length limit. The same name could be rejected by the import but accepted by the broker, or vice versa.

## Brief changelog

- new shared `validateResourceName` used by both the topic and group CSV imports: the RocketMQ character set `[%|a-zA-Z0-9_-]`, with the Topic cap of 127 and the group cap of 120
- consumer group names may now start with a digit or symbol, matching RocketMQ
- the inline Topic/Group create forms and the stale `topic.topicNameRule` translation now reference the same pattern and caps, so the UI and the importer agree
- tests cover `%`, `|`, digit-leading groups, rejected `/` and `*`, and the 127/128 and 120/121 length boundaries

- web: `resourceCsvImport.test.ts` 12/12; Topic/Consumer page tests green; full `vitest run` 642/642; `tsc -b` and `eslint` clean

Fixes #2493

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -754,8 +754,8 @@ const translations: Record<string, Record<Lang, string>> = {
   },
   'topic.topicNamePlaceholder': { zh: '请输入 Topic 名称', en: 'Enter topic name' },
   'topic.topicNameRule': {
-    zh: '仅支持字母、数字、下划线、中划线、斜杠和星号',
-    en: 'Only letters, numbers, underscore, hyphen, slash and asterisk',
+    zh: '仅支持字母、数字、下划线、短横线、% 和 |',
+    en: 'Only letters, numbers, underscore, hyphen, % and |',
   },
   'topic.topicNameRequired': { zh: '请输入 Topic 名称', en: 'Please enter topic name' },
   'topic.queueExtra': { zh: '每个 Broker 节点 8 个队列', en: '8 queues per Broker node' },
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
LGTM — correct alignment of frontend validation with RocketMQ's server-side `TopicValidator`/`GroupValidator`. The old patterns were too restrictive (rejecting `%` and `|` which the server accepts) and too permissive (accepting `/` and `*` which the server rejects).

Good extraction of shared `RESOURCE_NAME_PATTERN` and `RESOURCE_NAME_MAX_LENGTH` constants, and the `validateResourceName` function provides clear, localized error messages. Test coverage covers all edge cases.

---

lizhimins:

PR #2523, opened 2026-08-21
fix(cluster): bound timed-out NameServer registry probes

### LINKED ISSUE #2472: Bound and cancel NameServer registry probes
## Problem

NameServer registry probes run on a static unbounded cached thread pool, while CompletableFuture timeout completion does not reliably interrupt the underlying discovery call. Repeated requests against stuck endpoints can therefore retain probes and continually create threads.

## Expected behavior

Use a lifecycle-managed executor with bounded workers and queue capacity, enforce one shared request budget, explicitly cancel timed-out work, and reject excess probes without failing the whole registry response.

## Verification

Add focused service tests for concurrency limits, timeout interruption, and executor shutdown.

btlqql:

`listRegistryClusters` fanned its NameServer probes out onto an unbounded cached thread pool. `CompletableFuture.orTimeout` only completed the wrapper future, so a probe blocked on an unreachable NameServer kept its worker thread, and every refresh of the cluster list accumulated more blocked threads.

## Brief changelog

- new bounded, closeable `RegistryProbeRunner`: a fixed-size worker pool with a bounded queue replaces the unbounded cached thread pool
- when a probe exceeds its deadline the worker is explicitly cancelled (interrupted) so the pool thread is released, instead of leaking behind the timed-out future
- a saturated queue degrades the offending registry entry to unavailable (logged) rather than throwing or growing the pool
- the runner is shut down on Spring context close; idle workers (including core) time out so a quiet registry holds no threads
- the admin clients the probes drive already carry a 5s RPC timeout (`MqAdminExtFactory`), so released workers don't pile up behind hung calls

- server: `ClusterServiceRegistryTest` 5/5, including two new tests that block every probe and assert the pool never grows past its bound across repeated refreshes, and that a saturated queue degrades without throwing; full `mvn test` ran 1500 tests with only the 7 pre-existing environment failures in the CLI agent tests (missing `sh` binary on a Windows machine, identical on the clean base)

Fixes #2472

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -35,6 +35,7 @@ import org.apache.rocketmq.studio.common.domain.enums.FlushDiskType;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.ops.audit.AuditService;
 import org.apache.rocketmq.studio.provider.apache.RocketMQBrokerConfigService;
+import jakarta.annotation.PreDestroy;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
LGTM — excellent fix for the unbounded thread pool issue. The new `RegistryProbeRunner` class properly bounds concurrency, gracefully degrades on queue saturation (returns empty instead of throwing), and cancels timed-out tasks with worker interruption.

The `@PreDestroy` lifecycle hook and `allowCoreThreadTimeOut(true)` are good touches for resource management. Test coverage is thorough — both the bounded-pool test and the saturation-degradation test verify the key invariants.

---

lizhimins:

PR #2524, opened 2026-08-21
fix(instance): do not recreate instances deleted during update

### LINKED ISSUE #2494: [Studio][Bug] Do not recreate instances deleted during update
﻿## Problem

After the service reads an instance, if another request deletes it, `save()` re-queries, finds no record and inserts by the old ID - resurrecting the deleted instance.

Location: `server/src/main/java/org/apache/rocketmq/studio/instance/MybatisPlusInstanceRepository.java`.

## Expected behavior

Distinguish insert from updateExisting; the update path only executes `update` and returns 404/409 when the affected-row count is 0.

## Verification

Delete the record between the service read and the repository write; assert no insert happens and no success audit is recorded.

Note: different entity and code path from the open cloud-credentials issue #2116.

123123213weqw:

`save()` re-queried the row and fell back to an insert whenever the lookup came up empty. An instance that another request deleted between the service read and this write was resurrected under its old id, and the caller recorded a success audit for an instance that was "deleted".

## Brief changelog

- a non-null id now always takes the update path in `MybatisPlusInstanceRepository.save`; a zero-row update (the row vanished) is reported as a 409 instead of an insert
- a null id still inserts, so creation is unchanged
- tests updated: the absent-row case now asserts the 409 and that no insert happens, and the plain-insert test uses a null id

- server: `MybatisPlusInstanceRepositoryTest` 16/16 and `InstanceServiceTest` 68/68 green; full `mvn test` ran 1526 tests with only the 7 pre-existing environment failures in the CLI agent tests (missing `sh` binary on a Windows machine, identical on the clean base)

Fixes #2494

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/MybatisPlusInstanceRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/MybatisPlusInstanceRepository.java
@@ -114,7 +114,11 @@ public class MybatisPlusInstanceRepository implements InstanceRepository {
     @Transactional
     public InstanceVO save(InstanceVO instance) {
         RmqInstance entity = toEntity(instance);
-        if (entity.getId() != null && instanceMapper.selectById(entity.getId()) != null) {
+        if (entity.getId() != null) {
+            // A non-null id identifies an existing instance, so the update path must only
+            // update. If the row vanished (concurrent delete), a zero-row update is a
+            // conflict; re-inserting here would resurrect the deleted instance under its
+            // old id.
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — clean fix for the race condition where a deleted instance could be resurrected during update. The change correctly removes the `selectById` guard and relies solely on the `updateById` return value to detect conflicts.

Good test coverage, especially the rename to `saveShouldNotResurrectAnInstanceDeletedDuringUpdate` which clearly documents the intended behavior.

---

PR #2525, opened 2026-08-21
fix(instance): apply a single deadline to resource counts

### LINKED ISSUE #2495: [Studio][Performance] Apply a single deadline to instance resource counts
﻿## Problem

Resource-count tasks are submitted in parallel, but each `future.get(3s)` is then awaited sequentially. Multiple tasks that never complete make a single request wait for instance-count times 3 seconds.

Location: `server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java`.

## Expected behavior

Use one shared absolute deadline (invokeAll(timeout) or a completion queue); after the deadline cancel the remaining tasks at once and mark them unavailable.

## Verification

Create several never-completing providers and assert total elapsed time is close to one timeout, not count times timeout.

123123213weqw:

`fillCountsInParallel` submitted one task per instance and then waited on each future with its own 3s `get()`. A batch of hung vendor calls therefore cost `3s × instance-count` of response latency instead of a single 3s wait.

## Brief changelog

- submit the whole batch through `ExecutorService.invokeAll` with one shared `COUNT_TIMEOUT_SECONDS` deadline
- tasks that miss the deadline are cancelled (interrupted) by `invokeAll`; their rows are marked counts-unavailable like any other failure
- a test with three hung providers asserts the batch is bounded to well under one timeout per instance and that every row degrades to counts-unavailable

- server: `InstanceServiceTest` 69/69, including the new single-deadline test; full `mvn test` ran 1527 tests with only the 7 pre-existing environment failures in the CLI agent tests (missing `sh` binary on a Windows machine, identical on the clean base)

Fixes #2495

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -47,12 +47,12 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 @Slf4j
 @Service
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM — good fix for the cumulative timeout issue. Using `invokeAll` with a single deadline ensures the batch completes within a bounded time, regardless of how many futures are involved.

The test `saveShouldApplyBatchDeadline` correctly verifies the behavior by mocking slow futures and asserting the total time stays within bounds.

---

PR #2536, opened 2026-08-22
fix(web): defer download URL revocation

### LINKED ISSUE #2534: Defer object URL revocation for browser downloads
## Problem

`downloadBlob` revokes its object URL synchronously after calling `anchor.click()`. Some browsers consume the object URL asynchronously, so immediate revocation can cancel or produce an empty download. A thrown click also leaves cleanup behavior implicit.

Location: `web/src/utils/download.ts`.

## Expected behavior

Remove the temporary anchor immediately, but defer object URL revocation to a later task so the browser can start consuming it. Cleanup should still be scheduled if the programmatic click throws.

## Verification

Cover deferred revocation, attached-anchor clicking, successful removal, and the click-failure cleanup path.

yyqdbngt:
## Summary

- defer object URL revocation until a later browser task
- remove the temporary download anchor even when the click throws
- cover successful and failed click cleanup paths with fake-timer tests

## Why

Browsers may consume a Blob URL asynchronously after a programmatic anchor click. Revoking it in the same call stack can cancel or empty the download before navigation starts.

## Validation

- `npm test -- --run src/utils/download.test.ts`
- `npm run lint` (passes with three pre-existing Hook dependency warnings)
- `npm run build`
- `git diff --check`

Fixes #2534

### DIFF
--- a/web/src/components/AlertRuleAssetList.tsx
+++ b/web/src/components/AlertRuleAssetList.tsx
@@ -26,6 +26,7 @@ import {
   listAlertRuleAssets,
 } from '../services/alertRuleAssetService';
 import type { AlertRuleAssetInfo } from '../api/alertRuleAssets';
+import { downloadBlob } from '../utils/download';
 
 const { Text } = Typography;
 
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a real browser timing issue where `URL.revokeObjectURL()` called synchronously after `anchor.click()` can cancel the download before the browser starts navigating to the blob URL. The fix defers revocation via `setTimeout(..., 0)` and wraps the click in `try/finally` to ensure the anchor is always cleaned up even if `click()` throws.

**Verdict: LGTM** — correct fix for a subtle browser behavior, clean implementation with good test coverage including the failure path.

---

RockteMQ-AI (approved):
Re-approved after rebase. Previously reviewed and approved. The changes appear to be the same code rebased onto the latest base branch.

---

PR #2537, opened 2026-08-22
fix(topic): normalize unavailable LiteTopic quotas

### LINKED ISSUE #2535: Keep unavailable LiteTopic quota limits internally consistent
## Problem

`LiteTopicQuota` treats an unset maximum as unavailable in its rate calculations, but `isQuotaExceeded()` considers `maxTopicCount = 0` exceeded for every non-negative current count. Negative limits also produce negative usage rates. This exposes contradictory quota state through the model's computed JSON properties.

Location: `server/src/main/java/org/apache/rocketmq/studio/model/LiteTopicQuota.java`.

## Expected behavior

Treat null and non-positive maxima consistently as unavailable: usage rates and remaining quota stay zero, while the quota is not reported as exceeded. Positive maxima keep the existing at-limit and over-limit behavior.

## Verification

Cover null, zero, negative, below-limit, at-limit, and over-limit inputs in the focused model tests.

btlqql:
## Summary

- treat null, zero, and negative LiteTopic maxima as unavailable
- keep usage, remaining-quota, and exceeded calculations consistent
- cover unavailable, below-limit, at-limit, and over-limit states

## Why

The rate helpers already treat a zero maximum as unavailable, but `isQuotaExceeded()` reported every non-negative count as exceeded when the maximum was zero. The model therefore exposed contradictory computed quota state.

## Validation

- `mvn -q -Dtest=LiteTopicQuotaTest test`
- `mvn -q -DskipTests package`
- `git diff --check`

Fixes #2535

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/model/LiteTopicQuota.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/model/LiteTopicQuota.java
@@ -38,14 +38,14 @@ public class LiteTopicQuota {
     private Double maxCreationRate;
 
     public double getUsageRate() {
-        if (maxTopicCount == null || maxTopicCount == 0 || currentTopicCount == null) {
+        if (maxTopicCount == null || maxTopicCount <= 0 || currentTopicCount == null) {
             return 0.0;
         }
         return (double) currentTopicCount / maxTopicCount;
     }
 
     public double getSessionUsageRate() {
-        if (maxSessionCount == null || maxSessionCount == 0 || currentSessionCount == null) {
+        if (maxSessionCount == null || maxSessionCount <= 0 || currentSessionCount == null) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR correctly normalizes the handling of zero and negative `maxTopicCount`/`maxSessionCount` values across all quota methods in `LiteTopicQuota`. Previously, `isQuotaExceeded()` would incorrectly return `true` for any non-negative current count when max was zero, contradicting `getUsageRate()` which returned 0. The fix makes all methods consistently treat non-positive limits as "unavailable".

**Verdict: LGTM** — clean, well-scoped bug fix with thorough test coverage for edge cases (zero, negative, at-limit, over-limit).

---

lizhimins:

PR #2539, opened 2026-08-22
fix: bound AI request and stream payloads

### LINKED ISSUE #2538: AI requests and successful provider streams have no shared payload budget
## Problem

The AI endpoints currently accept user-controlled messages, command context, tool input, model selectors, and conversation identifiers without a shared size budget. These values are later serialized into provider requests or passed to a CLI process.

The OpenAI-compatible client already caps non-streaming response bodies and streaming error bodies at 5 MiB, but a successful SSE response uses an unbounded `InputStream`. A provider can therefore keep sending data beyond the configured response budget, including a single line without an SSE delimiter.

This makes the successful streaming path inconsistent with the other provider and CLI response paths, and allows request or response payloads to consume substantially more memory than the operation needs.

## Expected behavior

- Reject oversized chat, command, context, tool, model, selector, and identifier fields before invoking a provider or registered tool.
- Measure text limits in UTF-8 bytes, matching the encoding used for HTTP bodies and CLI arguments.
- Apply the existing provider response-body budget to the total successful SSE stream as well as non-streaming and error responses.
- Return stable 400/502 error codes so callers can distinguish oversized requests from oversized provider responses.
- Keep existing payloads within the limits unchanged.

## Verification scope

Tests should cover multibyte request boundaries, rejection before gateway/tool/CLI execution, configured-model validation, and an oversized successful SSE response. The full backend suite should remain green.

tju-yxq:
## Summary

- add one UTF-8 byte-budget guard for chat messages, command fields/context, tool invocations, selectors, identifiers, and outbound provider/CLI prompts
- reject oversized values before gateway, tool registry, HTTP provider, or CLI execution with stable request error semantics
- apply the existing 5 MiB provider response budget to successful SSE streams, not only non-streaming bodies and streaming error responses
- validate the effective configured model when no per-request override is present

## Why

Successful OpenAI-compatible SSE responses currently use an unbounded input stream, while every other provider/CLI response path is capped. AI request DTOs and structured tool inputs also have no centralized size checks before serialization or process/provider execution. This makes the streaming path inconsistent and lets a single request or provider response retain unnecessarily large payloads.

## Implementation notes

- text limits are measured in UTF-8 bytes and stop counting as soon as the limit is exceeded
- structured context/tool limits use their serialized JSON size
- a limited stream reads at most one byte beyond the response budget before raising `llm.provider.response_too_large`
- payloads at the configured boundary remain accepted

## Tests

- focused AI package suite: 106 tests passed
- full backend suite: 1,561 tests passed
- Checkstyle: 0 violations
- `git diff --check`: passed

Production changes: 237 additions / 5 deletions. The size comes from the central guard and bounded SSE implementation, not test padding.

Closes #2538

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/AiPayloadGuard.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/AiPayloadGuard.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.studio.ops.ai;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.rocketmq.studio.common.exception.BusinessException;
+import org.springframework.util.StringUtils;
+
+import java.util.Map;
+
+/**
+ * Central byte budgets for user-controlled AI payloads. Character-count validation is not
+ * sufficient here because provider requests, CLI arguments and tool payloads are encoded as
+ * UTF-8 before leaving Studio.
+ */
+final class AiPayloadGuard {
+
+    static final int MAX_MESSAGE_BYTES = 64 * 1024;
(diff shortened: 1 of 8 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds defense-in-depth payload size validation for the RocketMQ Studio AI subsystem. A new `AiPayloadGuard` class validates UTF-8 byte budgets for user-controlled AI payloads (chat messages, context, tool inputs, outbound prompts) before they reach the LLM gateway, CLI execution, or tool registry. The implementation is thorough, well-tested, and follows existing error handling patterns.

### Review Highlights
- **Correctness**: UTF-8 byte counting handles surrogate pairs correctly, with proper early exit when limits are exceeded. `LimitedInputStream` correctly detects limit crossing by reading one extra byte.
- **Security**: Validates at all entry points (chat, command, tool, outbound) with clear error codes (`llm.request.payload_too_large`, `llm.provider.response_too_large`).
- **Tests**: Excellent coverage including boundary conditions, multi-byte character handling, and verification that oversized payloads are rejected before calling upstream.
- **Design**: Package-private `AiPayloadGuard` with static methods is appropriate for a utility class. The inconsistency between `chat()` (throws) and `execute()` (returns result) is intentional and documented by the different return types.

### Minor Observations
- `CliAgentProvider.complete()` validates before checking `available()` — this is fine since the 400 error is more specific than a 503.
- The `exceedsUtf8Limit()` comment about unpaired surrogates being replaced with '?' by the JDK encoder is accurate.

Overall, this is a solid security improvement that prevents potential resource exhaustion from oversized AI payloads.

---

PR #2543, opened 2026-08-22
feat: paginate instance management

### LINKED ISSUE #2542: Instance management needs server-side pagination
## Problem

The instance management page currently fetches every configured instance in a single request:

```ts
const res = await client.get<{ data: Instance[] }>('/instances', { params });
return res.data.data;
```

The backend also loads all matching rows, then the UI renders the table with `pagination={false}`.

Although the endpoint supports type and search filters, there is no bounded page size, server total, or page navigation. This differs from the other management inventories (cloud credentials, data sources, Studio users, query history), which use server-side pagination.

As an operator imports more cloud instances, the initial management page transfers and renders the whole inventory at once.

## Expected behavior

- Add a backward-compatible `GET /api/instances/page` endpoint.
- Support the existing type/search filters plus:
  - `page` (default 1)
  - `pageSize` (default 20, maximum 100)
- Reject invalid pagination before repository or provider access.
- Return `items`, `total`, `page`, and `size`.
- Keep the existing unpaginated endpoint for instance selector usage.
- Update the instance management page to use server-side pagination, preserve search/type filtering, reset to page 1 when filters change, and support 20/50/100 page sizes.
- Keep resource counts and the existing sort order unchanged.

## Verification scope

Tests should cover service slice selection and validation, the HTTP page contract, frontend API parameters, and the instance page's server-driven pagination behavior.

tju-yxq:
## Summary

- add backward-compatible `GET /api/instances/page` with the existing type/search filters and bounded pagination
- return `items`, `total`, `page`, and `size`; default to `page=1`, `pageSize=20`, and cap page size at 100
- reject invalid page/pageSize before repository or provider access
- add a repository count query for the current type/search filter
- preserve the existing unpaginated endpoint for selectors and keep resource-count and display-name behavior unchanged
- update instance management to use server-driven pagination, a server total, 20/50/100 page sizes, and page reset on filter changes
- keep request sequencing and stale-response protection

## Why

The management page previously loaded and rendered every configured instance in one request. Other Studio inventories already use server-side pagination, so this brings the primary instance inventory to the same bounded contract as cloud credentials, data sources, Studio users, and query history.

## Tests

- focused backend: `InstanceServiceTest,InstanceControllerTest` — 84 tests passed
- backend full suite: 1,560 tests passed; Checkstyle 0 violations
- focused frontend: instance API/service/page tests — 3 files / 32 tests passed
- `npm run lint -- --quiet`: 0 errors (3 pre-existing warnings)
- `npm run build`: passed
- `git diff --check`: passed

Production changes: 122 additions / 9 deletions, naturally exceeding 100 production lines without test padding.

Closes #2542

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceController.java
@@ -18,6 +18,7 @@
 package org.apache.rocketmq.studio.instance;
 
 import org.apache.rocketmq.studio.common.domain.DeleteRequestDTO;
+import org.apache.rocketmq.studio.common.domain.PageResult;
 import org.apache.rocketmq.studio.common.domain.Result;
 import org.apache.rocketmq.studio.common.domain.enums.InstanceType;
 import jakarta.validation.Valid;
@@ -45,6 +46,15 @@ public class InstanceController {
             @RequestParam(required = false) InstanceType type,
             @RequestParam(required = false) String search) {
         return Result.ok(instanceService.listInstances(type, search));
+    }
+
+    @GetMapping("/page")
+    public Result<PageResult<InstanceVO>> listInstancesPage(
+            @RequestParam(required = false) InstanceType type,
+            @RequestParam(required = false) String search,
+            @RequestParam(defaultValue = "1") int page,
+            @RequestParam(defaultValue = "20") int pageSize) {
+        return Result.ok(instanceService.listInstances(type, search, page, pageSize));
     }
 
     @GetMapping("/{instanceId}/capabilities")
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -113,14 +113,12 @@ public class InstanceService {
                                                 int pageSize) {
         validateListPagination(page, pageSize);
         String normalizedSearch = search == null || search.isBlank() ? null : search.trim();
-        long total = instanceRepository.count(type, normalizedSearch);
-        if (total == 0) {
-            return PageResult.empty(page, pageSize);
-        }
-        List<InstanceVO> sorted = listInstances(type, normalizedSearch);
-        int fromIndex = (int) Math.min((long) (page - 1) * pageSize, sorted.size());
-        int toIndex = Math.min(fromIndex + pageSize, sorted.size());
-        return PageResult.of(new ArrayList<>(sorted.subList(fromIndex, toIndex)), total, page, pageSize);
+        PageResult<InstanceVO> result = instanceRepository.findPage(
+                type, normalizedSearch, page, pageSize);
+        fillCountsInParallel(result.getItems());
+        result.getItems().forEach(instance ->
+                instance.setRegionName(regionNames.resolve(instance.getRegionId())));
+        return result;
     }
 
     private static void validateListPagination(int page, int pageSize) {
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -23,6 +23,7 @@ import org.apache.rocketmq.studio.provider.credential.CloudCredentialRepository;
 import org.apache.rocketmq.studio.provider.credential.CloudCredentialVO;
 import org.apache.rocketmq.studio.cluster.broker.MqAdminExtFactory;
 import org.apache.rocketmq.studio.audit.OperationAuditService;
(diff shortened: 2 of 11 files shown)


RockteMQ-AI (commented):
### Review of PR #2543

This PR modifies 10 files with +320 -79 lines.

### Findings

- **[Info]** Chained method calls detected — consider null-safety checks

### Suggestions
Please review the findings above and address any critical or warning items.

---

server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
```diff
@@ -107,6 +109,30 @@ public List<InstanceVO> listInstances(InstanceType type, String search) {
         return sorted;
     }
 
+    public PageResult<InstanceVO> listInstances(InstanceType type, String search, int page,
+                                                int pageSize) {
+        validateListPagination(page, pageSize);
+        String normalizedSearch = search == null || search.isBlank() ? null : search.trim();
+        long total = instanceRepository.count(type, normalizedSearch);
+        if (total == 0) {
```
RockteMQ-AI: **[Performance]** This method calls `listInstances(type, normalizedSearch)` which loads ALL instances into memory (including remote resource-count fan-out for every instance) and then paginates in-memory with `subList`. This defeats the purpose of pagination — for a deployment with hundreds of instances, every page request still triggers N remote API calls.

Consider paginating at the database level instead:
1. Use `instanceRepository` with LIMIT/OFFSET (or MyBatis-Plus `Page`) to fetch only the current page of instances
2. Then do the resource-count fan-out only for those instances on the current page

This would reduce both DB load and remote API calls from O(N) to O(pageSize) per request.

server/src/main/java/org/apache/rocketmq/studio/instance/InstanceController.java
```diff
@@ -47,6 +48,15 @@ public Result<List<InstanceVO>> listInstances(
         return Result.ok(instanceService.listInstances(type, search));
     }
 
+    @GetMapping("/page")
+    public Result<PageResult<InstanceVO>> listInstancesPage(
```
RockteMQ-AI, line 52: **[Info]** The existing `GET /` endpoint (line 43) still returns the full unpaginated list. Consider whether both endpoints are needed long-term, or if the old one should be deprecated once the frontend fully migrates to `/page`. Having two endpoints with overlapping behavior can lead to maintenance burden.

RockteMQ-AI (commented):
### Summary
This PR adds server-side pagination for instance management (backend `/page` endpoint + frontend pagination controls with search/filter). The overall structure is clean — good input validation, proper race-condition handling in the frontend with `requestIdRef`, and solid test coverage.

One performance concern: the paginated service method still loads all instances into memory before slicing. See inline comment for details.

---

tju-yxq:
Thanks for catching this. I moved pagination into the database layer so the service no longer loads all instances and no longer fans out resource-count requests for instances outside the current page.

- Added \InstanceRepository.findPage(...)\, implemented with MyBatis-Plus \selectPage\ and the same type/search filters.
- \InstanceService.listInstances(...)\ now uses that repository page and fills resource counts/region names only for the returned \pageSize\ rows.
- The paginated path no longer calls \indAll\, \indByTypeAndSearch\, or in-memory \subList\; database reads and remote count fan-out are now O(pageSize), not O(total inventory).
- Added tests proving the repository applies filters/pagination at SQL level and proving the service never falls back to the full-list repository methods.

Focused verification: \InstanceServiceTest\, \InstanceControllerTest\, and \MybatisPlusInstanceRepositoryTest\ — 102 tests passed, Checkstyle clean, \git diff --check\ clean. Pushed as commit \3255cf35\ (\ix: page instances in the database\).

I kept the legacy unpaginated endpoint for now because existing instance selectors still consume it. We can deprecate or migrate those callers separately once the paginated management endpoint is stable.

RockteMQ-AI (approved):
### Summary
Previous concerns have been addressed. The paginated endpoint now uses database-level pagination via `instanceRepository.findPage()` (MyBatis-Plus `Page`), which correctly avoids loading all instances into memory. The existing unpaginated endpoint is preserved for backward compatibility — this is fine as a migration path.

LGTM — ready to merge.

---

lizhimins:
Thanks for the contribution. After review we are not taking instance-list pagination in this project: per an explicit product convention, each user has a limited number of instances (tens at most), so the instance list intentionally returns the full set without server-side or client-side pagination (`pagination={false}`). This PR also bypasses the established backend ordering contract (APACHE instances by name, then cloud vendors by region → instance id, unified in `InstanceService.listInstances`) via `orderByAsc("id")`, and adds unused `countAll()`/`count()` repository methods. Closing accordingly.

PR #2560, opened 2026-08-25
fix(web): tolerate incomplete certificate metadata

### LINKED ISSUE #2556: K8s certificate table fails on incomplete metadata
## Problem

The K8s certificate API can return incomplete metadata (for example a metadata-only record with no issuer, or a legacy row with missing type/status/expiration fields). The web contract marks these fields as required and the table sorters call `localeCompare` or parse dates without handling null values. Sorting such a row can throw and break the certificate inventory.

## Proposed fix

Align the TypeScript contract with the nullable backend response, render missing metadata consistently, make all affected sorters null-safe, and keep mock-mode clones safe for nullable SAN values. Add a page regression test with an incomplete certificate row.

yyqdbngt:
## Summary

- align the K8s certificate API type with nullable metadata returned by the backend
- render and sort missing type, issuer, expiration, and status values safely
- preserve nullable SAN metadata when cloning certificates in mock mode
- cover an incomplete certificate row and its sortable columns

## Why

Metadata-only and legacy certificate records can omit issuer and other parsed certificate fields. The table previously called string/date sort logic on those values directly, so sorting an incomplete row could break the inventory page.

## Validation

- `npm test -- --run src/pages/cluster/__tests__/K8sCertsPage.test.tsx` (7 tests passed)
- `npm run build`
- `npx eslint src/api/cluster.ts src/services/clusterService.ts src/pages/cluster/certs.tsx src/pages/cluster/__tests__/K8sCertsPage.test.tsx --max-warnings=0`
- `git diff --check`

Closes #2556.

### DIFF
--- a/web/src/api/cluster.ts
+++ b/web/src/api/cluster.ts
@@ -136,13 +136,13 @@ export interface K8sCertInfo {
   id: number;
   k8sId: string;
   cluster: string;
-  type: string;
-  issuer: string;
-  notBefore: string;
-  notAfter: string;
-  status: string;
+  type: string | null;
+  issuer: string | null;
+  notBefore: string | null;
+  notAfter: string | null;
+  status: string | null;
   daysRemaining: number;
-  san: string[];
+  san: string[] | null;
   certPem?: string;
   keyPem?: string;
 }
--- a/web/src/pages/cluster/__tests__/K8sCertsPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/K8sCertsPage.test.tsx
@@ -43,6 +43,18 @@ const certs: K8sCertInfo[] = [
     san: ['broker.prod.example.com'],
   },
   {
+    id: 3,
+    k8sId: 'metadata-only-tls',
+    cluster: 'legacy-cluster',
+    type: null,
+    issuer: null,
+    notBefore: null,
+    notAfter: null,
+    status: null,
+    daysRemaining: 0,
+    san: null,
+  },
+  {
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a null-safety issue in the K8s certificate inventory page where sorting on incomplete certificate metadata (missing type, issuer, expiration, etc.) would crash the UI. The fix properly aligns the TypeScript types with the backend's nullable fields and adds safe fallbacks in sorters and renderers.

### Findings
- **[Info]** `cluster.ts:114-123` — Type changes from `string` to `string | null` correctly reflect the backend API contract for metadata-only and legacy certificate records.
- **[Info]** `certs.tsx` — Sort functions properly handle null values: string fields use `(a.type ?? '').localeCompare(b.type ?? '')` and date fields use `(Date.parse(a.notAfter ?? '') || 0)`. This prevents the sorting crash.
- **[Info]** `certs.tsx` — Render fallbacks show `'-'` for null values, which is clean UX.
- **[Info]** `clusterService.ts` — The mock store correctly handles null `san` arrays when cloning certificates.
- **[Info]** `K8sCertsPage.test.tsx` — New test case covers both rendering and sorting with incomplete metadata. Good coverage.

### Suggestions
- Minor: Consider adding a `formatDateTime` null guard if it's used elsewhere with potentially null dates, though it's not in the scope of this fix.

Overall: Clean, focused bug fix with proper null safety and good test coverage. LGTM.

---

PR #2561, opened 2026-08-25
fix(message): reject negative queue coordinates

### LINKED ISSUE #2557: Queue browser forwards negative queue coordinates to the provider
## Problem

The queue browser endpoint validates topic and broker name but forwards negative `queueId` and `offset` values to the RocketMQ provider. Invalid coordinates therefore reach the external client and surface as provider failures instead of a deterministic 400 response.

## Proposed fix

Reject negative queue IDs and offsets in `MessageService` before invoking the provider, with regression coverage proving that invalid requests do not cross the provider boundary.

yyqdbngt:
## Summary

- reject negative queue IDs before the queue browser invokes its provider
- reject negative queue offsets at the same service boundary
- verify invalid coordinates never reach the message provider

## Why

A queue ID and physical queue offset cannot be negative. Passing those values through to the RocketMQ client turned invalid user input into downstream provider errors instead of a stable client-facing 400 response.

## Validation

- `mvn -Dtest=MessageServiceTest test`
- `git diff --check`

Closes #2557.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
@@ -91,6 +91,12 @@ public class MessageService {
         if (!StringUtils.hasText(brokerName)) {
             throw new BusinessException(400, "brokerName is required");
         }
+        if (queueId < 0) {
+            throw new BusinessException(400, "queueId must not be negative");
+        }
+        if (offset < 0) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good input validation fix. Rejecting negative `queueId` and `offset` at the controller level prevents invalid queries from reaching the repository layer. The error messages are clear and the test covers both negative cases.

Simple, focused, and correct.

LGTM.

---

PR #2562, opened 2026-08-25
fix(acl): prevent Tencent rule page overflow

### LINKED ISSUE #2558: Tencent ACL rule pagination overflows on large page numbers
## Problem

Tencent ACL rules are fetched into memory and then paginated with `(page - 1) * pageSize` using `int` arithmetic. A valid large page such as `Integer.MAX_VALUE` overflows to a negative index and causes `subList` to throw instead of returning an empty page. Other in-memory pagination paths already use the long-safe shared `Pagination.pageOffset` helper.

## Proposed fix

Use the shared long-safe offset calculation for Tencent ACL rule pagination and add regression coverage for a page offset beyond the integer range.

btlqql:
## Summary

- use the shared long-safe pagination offset for in-memory Tencent ACL rules
- return an empty page when the requested offset is beyond the available rules
- cover `Integer.MAX_VALUE` page input with the maximum supported page size

## Why

The previous `(page - 1) * pageSize` calculation used `int` arithmetic. Large but valid page values overflowed to a negative index and caused `subList` to fail instead of returning the normal empty-page response.

## Validation

- `mvn -Dtest=AclServiceTest test`
- `git diff --check`

Closes #2558.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -396,7 +396,7 @@ public class AclService {
 
     private static PageResult<AclRuleVO> paginateRules(List<AclRuleVO> rules, int page, int pageSize) {
         int total = rules.size();
-        int fromIndex = Math.min((page - 1) * pageSize, total);
+        int fromIndex = (int) Math.min(Pagination.pageOffset(page, pageSize), total);
         int toIndex = Math.min(fromIndex + pageSize, total);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct fix for integer overflow in pagination calculation. Using `Pagination.pageOffset` (which operates on `long`) instead of `page * size` (which can overflow `int`) is the right approach. The test with `Integer.MAX_VALUE` proves the fix works for the edge case.

Clean, minimal change.

LGTM.

---

PR #2563, opened 2026-08-25
fix(acl): validate plain permission entries

### LINKED ISSUE #2559: Plain ACL upsert silently drops malformed permission entries
## Problem

Plain ACL account upserts replace the stored per-resource rules, but malformed `topicPerms` or `groupPerms` entries are silently skipped after the account mutation begins. A typo such as `orders-PUB` can therefore produce a successful response while dropping the previous permission rule.

## Proposed fix

Validate every permission entry uses a non-blank `resource=permission` structure before any mapper read or write, return a clear 400 error with the failing field/index, and add repository regression coverage proving malformed input causes no database interaction.

btlqql:
## Summary

- validate plain ACL topic and group permission entries before repository access
- require a non-blank `resource=permission` structure and identify the failing field/index
- prove malformed input produces a 400 response without any mapper interaction

## Why

Plain ACL upserts replace an account's per-resource rules. Malformed entries were silently skipped during replacement, so a successful update could unexpectedly discard an existing permission. Failing before any mutation keeps the previous account and rules intact.

## Validation

- `mvn -Dtest=MybatisPlusAclRepositoryTest test` (16 tests passed)
- `git diff --check`

Closes #2559.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
@@ -177,6 +177,8 @@ public class MybatisPlusAclRepository implements AclRepository {
     @Override
     @Transactional
     public PlainAccessConfigVO createAndUpdatePlainAccessConfig(PlainAccessConfigVO config) {
+        validatePermissionEntries(config.getTopicPerms(), "topicPerms");
+        validatePermissionEntries(config.getGroupPerms(), "groupPerms");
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good defensive fix. Validating permission entries before the transaction starts prevents silent data loss when malformed entries would otherwise be skipped during replacement. The error message correctly identifies the failing field and index, which helps users fix their input.

The validation logic (`splitPerm` returning null or blank parts) is consistent with how entries are parsed elsewhere. Test covers the malformed input case and verifies no mapper interaction occurs on failure.

LGTM.

---

PR #2567, opened 2026-08-25
fix(studio): preserve instance context on discovery failure

### LINKED ISSUE #2566: [Studio][Bug] Preserve instance context and enforce capabilities when discovery fails
## Problem

Instance-scoped Studio pages treat instance inventory and capability discovery failures as permissive states, but the fallback is incomplete.

- `useInstanceFilter` documents that an inventory-load failure falls back to no filtering. In fact, with `/instance/:instanceId/...` routes it only accepts IDs contained in the fetched inventory. A rejected `listInstances()` leaves `selectedInstanceId` undefined, so Topic, Consumer, Message, ACL, and DLQ skip their data loads.
- `MainLayout` treats a missing capability set as all capabilities and preserves that state when discovery fails. Users can therefore navigate to operations that the backend rejects for the selected provider, for example cloud-instance DLQ requests returning 501.

## Scope

- Preserve a syntactically valid route instance ID while inventory loading fails, and expose retry/error state to callers.
- Make unknown or failed capability discovery fail closed for capability-gated navigation.
- Keep routes, selectors, and menu behavior consistent so unsupported operations are not presented as available.
- Add shared regression coverage for failure and retry states.

## Affected code

- `web/src/hooks/useInstanceFilter.ts`
- `web/src/layouts/MainLayout.tsx`
- instance-scoped Topic, Consumer, Message, ACL, and DLQ pages

## Verification

- Hook tests for failed inventory load, retained route selection, and retry.
- Layout tests for failed capability discovery and supported/unsupported navigation.
- Frontend build.

Aias00:
## Summary

- retain the instance encoded in an instance-scoped route when the shared inventory request fails, instead of changing the selected instance to `undefined`
- expose inventory error and retry state from `useInstanceFilter` for the shared Topic, Consumer, Message, ACL, and DLQ consumers
- treat unknown or failed instance capabilities as unavailable for capability-gated navigation, while retaining global navigation when no instance is selected
- add regression coverage for inventory failure, retry, and capability discovery failure

## Why

A temporary instance-list failure previously disabled data loading across all instance-scoped pages despite a route already carrying an instance ID. Separately, a capability-query failure made every capability-gated operation appear available, allowing navigation into provider operations that could only fail later with a 501 response.

## Validation

- `npm test -- --run src/hooks/useInstanceFilter.test.tsx src/layouts/MainLayout.test.tsx`
- `npx eslint src/hooks/useInstanceFilter.ts src/hooks/useInstanceFilter.test.tsx src/layouts/MainLayout.tsx src/layouts/MainLayout.test.tsx --max-warnings=0`
- `npm run build`
- `git diff --check`

Closes #2566.

### DIFF
--- a/web/src/hooks/useInstanceFilter.test.tsx
+++ b/web/src/hooks/useInstanceFilter.test.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { render, screen, waitFor } from '@testing-library/react';
 import { MemoryRouter, Route, Routes, useLocation } from 'react-router-dom';
 import { useInstanceFilter } from './useInstanceFilter';
@@ -32,7 +32,21 @@ function InstanceRouteProbe() {
   return <output>{`${pathname}|${selectedInstanceId}`}</output>;
 }
 
+function RetryableInstanceRouteProbe() {
+  const { selectedInstanceId, retryInstances } = useInstanceFilter();
+  return (
+    <>
+      <output>{selectedInstanceId}</output>
+      <button onClick={retryInstances}>retry</button>
+    </>
+  );
+}
+
 describe('useInstanceFilter', () => {
+  beforeEach(() => {
+    instanceServiceMocks.listInstances.mockReset();
+  });
+
   it('replaces an unknown route instance with the first available instance', async () => {
     instanceServiceMocks.listInstances.mockResolvedValue([
       {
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
### Summary
Well-structured resilience fix. Preserving the route-encoded instance ID when the inventory request fails prevents unnecessary data-loading failures across all instance-scoped pages. Exposing `retryInstances` and error state from the hook is a clean way to let consumers offer retry without duplicating state. The capability discovery fallback (treating unknown/failed capabilities as unavailable) is the right defensive default — it prevents navigation into operations that would fail with 501.

Test coverage for inventory failure, retry, and capability discovery failure is thorough.

LGTM.

---

lizhimins (changes requested):
The core fix (keeping the routed instance id when inventory discovery fails) is valid — please keep it. Two issues need rework:

1. Fail-closed reversal: MainLayout currently hides the Topic/Group/ACL/Message/DLQ navigation entirely while discovery is in flight or stale. This reverses the intentional fail-open capability design: the menu flickers on every instance switch, and a persistently failing capability check removes instance navigation completely. Please distinguish loading from failure — keep navigation visible on failure and surface an error notice with a retry entry. Note the pages themselves are unguarded, so deep links still reach the features; the hiding is not a complete gate either way.

2. Dead API: `instancesError` / `retryInstances` are exposed by useInstanceFilter but no page consumes them. Either wire them into the pages as the PR description claims, or remove them.

RockteMQ-AI (approved):
### Summary
Good improvements to instance context preservation on capability discovery failure. The new changes add:

1. **Error state tracking** — `instancesError` in `useInstanceFilter` keeps the selected instance valid when inventory loading fails, preventing unnecessary redirects.
2. **Retry mechanism** — `capabilityReloadVersion` + retry button in `MainLayout` gives users a way to recover from transient capability discovery failures.
3. **Graceful degradation** — When capabilities are unknown (error or loading), all navigation items are shown rather than hiding features.
4. **Test isolation** — `beforeEach` mock reset ensures tests don't leak state between cases.

The `hasInstanceCapability` signature change from `null` to `null | undefined` with the `undefined` meaning "show everything" is a clean semantic distinction.

LGTM — new changes look good.

---

PR #2571, opened 2026-08-25
fix(message): surface body encoding and truncation metadata

### LINKED ISSUE #2501: [Studio][Bug] Surface message body encoding and truncation metadata
﻿## Problem

The backend returns `bodyEncoding`, `bodyTruncated` and `propertiesTruncated`, but the frontend types and the detail page ignore them. Base64 bodies and truncated content are displayed and downloaded as if they were complete plain text.

Locations: `web/src/api/message.ts`, `server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java`.

## Expected behavior

Complete the API types; add encoding/truncation warnings on the detail page; pick download extension and content according to the encoding instead of forcing Base64 into JSON.

## Verification

Cover UTF-8 body, binary Base64, body truncation, and property count/length truncation.

btlqql:

The backend already reports `bodyEncoding`, `bodyTruncated` and `propertiesTruncated` on every message record, but the frontend type dropped all three fields. Base64-encoded binary bodies were therefore rendered and downloaded as if they were complete plain text: the detail view pretty-printed the encoded string as JSON, and the download button always wrote a `.json` file.

## Brief changelog

- complete the `MessageRecord` type with the three metadata fields
- the detail view now shows truncation warnings for clipped bodies and property maps, and marks Base64 bodies with a tag
- downloads pick the content and extension from the encoding: Base64 bodies are saved verbatim as `.txt`, UTF-8 bodies keep the JSON pretty-printing behavior

- new `MessageBodyEncoding` page tests (3/3 green): detail warnings for truncated body/properties and the Base64 marker, no warnings for plain UTF-8 bodies, and download content/extension per encoding
- `web/src/api/message.test.ts` 7/7 green; `tsc -b` and `eslint` clean
- remaining MessagePage test failures are pre-existing on the base branch (recent-query button removed in a later refactor)

Fixes #2501

### DIFF
--- a/web/src/api/message.ts
+++ b/web/src/api/message.ts
@@ -10,10 +10,16 @@ export interface MessageRecord {
   queueId: number | null;
   queueOffset: number | null;
   body: string;
+  /** How `body` is encoded: "UTF-8" for text, "BASE64" for binary payloads. */
+  bodyEncoding?: string | null;
+  /** True when the backend clipped the body to its display limit. */
+  bodyTruncated?: boolean;
   storeTime: number | string;
   bornHost: string;
   storeHost: string;
   properties: Record<string, string>;
+  /** True when some properties were dropped by the backend display limit. */
+  propertiesTruncated?: boolean;
   size: number;
 }
 
--- a/web/src/pages/instance/__tests__/MessageBodyEncoding.test.tsx
+++ b/web/src/pages/instance/__tests__/MessageBodyEncoding.test.tsx
@@ -1,0 +1,195 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { App } from 'antd';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import type React from 'react';
+import { MemoryRouter } from 'react-router-dom';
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (commented):
### Summary
Review of PR #2571

### Findings
FINDINGS:
- [Warning] message.tsx:427 — Base64 bodies are downloaded as the *encoded* string in a `.txt` file rather than decoded bytes. A user downloading a binary message body (image, protobuf, serialized object) almost certainly wants the decoded content, not the Base64 text. The comment explains avoiding the JSON formatter, but that's a separate concern from decoding.
- [Info] message.tsx:655 — The detail `Paragraph` still renders `{formatBody(selectedMsg.body)}`, so for every Base64 message `JSON.parse` throws and is caught, returning the raw string. It works via graceful degradation but silently throws on each render of a Base64 detail.
- [Info] message.ts:13-17 — All three new fields are optional (`?:`), and `isBase64Body` defaults to `false` via `?? ''`. Backends that omit these fields fall through to the existing UTF-8 path with no truncation alerts — fully backward compatible.
- [Info] MessageBodyEncoding.test.tsx:73 — The binary fixture `body: 'AAEC AQIBAA=='` contains a space (invalid Base64). Harmless today since nothing decodes it (`atob` would strip the whitespace anyway), but it's an unusual fixture to anchor behavior on.
- [Info] MessageBodyEncoding.test.tsx — No test covers a record with `bodyEncoding: undefined` (legacy backend). `isBase64Body` handles it correctly, but an explicit case would lock in the default-UTF-8 contract.

SUGGESTIONS:

### Suggestions
SUGGESTIONS:
- Decode Base64 for download, falling back to the current `.txt` behavior if decoding fails. This gives users the actual binary payload while staying safe:
```tsx
const handleDownload = (record: MessageRecord) => {
  if (isBase64Body(record)) {
    try {
      const bytes = Uint8Array.from(atob(record.body), (c) => c.charCodeAt(0));
      downloadBlob(new Blob([bytes], { type: 'application/octet-stream' }), `${record.msgId}.bin`);
    } catch {
      downloadBlob(new Blob([record.body], { type: 'text/plain' }), `${record.msgId}.txt`);
    }
  } else {
    downloadBlob(new Blob([formatBody(record.body)], { type: 'application/json' }), `${record.msgId}.json`);
  }
  message.success('消息下载成功');
};
```
- Short-circuit `formatBody` for Base64 in the detail view to avoid the silent throw-and-catch on every render:
```tsx
<Paragraph copyable style={{ /* ... */ }}>

---

PR #2572, opened 2026-08-25
fix(topic): allow clearing persisted topic remarks

### LINKED ISSUE #2497: [Studio][Bug] Allow clearing persisted Topic remarks
﻿## Problem

The remark is only updated when `StringUtils.hasText(remark)`. When a user submits an empty string to clear the remark, the old remark is kept.

Location: `server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java`.

## Expected behavior

Distinguish "field not submitted" from "field submitted empty"; update when the field is non-null and normalize an empty string to null/empty.

## Verification

Remark not submitted stays; blank submission clears; new value replaces.

btlqql:

``updateTopic`` only wrote the remark when ``StringUtils.hasText()`` passed, so submitting a blank value behaved exactly like not submitting one and the stored remark could never be removed.

## Brief changelog

- treat a null remark as "not submitted" (keep the stored value) and a submitted blank/whitespace remark as "clear" (store null), trimming the value in between
- echo the persisted remark back on the response so the returned VO reflects the stored state instead of the raw request value

- ``RocketMQAdminClientImplTest`` 37/37 green, including two new cases: blank submission clears the stored remark, omitted remark keeps it
- ``TopicControllerTest`` 14/14 and ``MetadataServiceTest`` 21/21 green on JDK 21

Fixes #2497

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
@@ -353,8 +353,12 @@ public class RocketMQAdminClientImpl implements AdminClient {
                     if (topic.getType() != null) {
                         existing.setTopicType(topic.getType().name());
                     }
-                    if (StringUtils.hasText(topic.getRemark())) {
-                        existing.setRemark(topic.getRemark());
+                    // Distinguish "remark not submitted" (null: keep the stored value) from
+                    // "remark submitted blank" (clear it); a hasText() guard treated both the
+                    // same, so users could never remove a persisted remark.
+                    if (topic.getRemark() != null) {
+                        String remark = topic.getRemark().trim();
+                        existing.setRemark(remark.isEmpty() ? null : remark);
                     }
                     existing.setGmtModified(LocalDateTime.now());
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
Review of PR #2572

### Findings
FINDINGS:
- [Info] `RocketMQAdminClientImpl.java:356` — The `null`-vs-blank distinction correctly preserves the old remark on omitted fields and allows clearing via whitespace/empty strings.
- [Info] `RocketMQAdminClientImpl.java:373` — Echoing `existing.getRemark()` back into the response keeps the returned state consistent with the database.
- [Warning] `RocketMQAdminClientImpl.java:356` — Blank/whitespace input now intentionally clears a persisted remark; this is a behavior change from the previous `hasText` guard.
- [Info] `RocketMQAdminClientImplTest.java:424` — New tests cover blank submission clearing the remark and omitted submission keeping it.

SUGGESTIONS:

### Suggestions
SUGGESTIONS:
- Add an explicit `""` empty-string test case alongside the whitespace test to document that empty input also clears the remark.
- Consider extracting a tiny helper like `normalizeRemark(String)` if similar logic is needed for other metadata fields.

---

PR #2573, opened 2026-08-25
[Studio] fix: preserve Settings query parameters when switching tabs

coder999o:

The Settings page updates the URL query string when switching tabs, but it currently replaces all existing query parameters with only `tab`.

This can drop unrelated navigation context, such as deep-link or focus parameters, when users move between Settings tabs.

This change preserves existing query parameters and only updates the `tab` parameter.

## Brief changelog

- Preserve unrelated Settings URL query parameters when switching tabs.
- Added a regression test for tab switching with existing query parameters.

## Verifying this change

- `npm test -- --run src/pages/settings/__tests__/SettingsPage.test.tsx src/pages/settings/__tests__/GeneralSettingsTab.test.tsx src/pages/settings/__tests__/AiAssistantTab.test.tsx src/pages/settings/__tests__/CloudCredentialTab.test.tsx src/pages/settings/__tests__/DataSourceTab.test.tsx`
- `npx eslint src/pages/settings/index.tsx src/pages/settings/__tests__/SettingsPage.test.tsx --max-warnings=0`
- `git diff --check`
- `npm run build`

### DIFF
--- a/web/src/pages/settings/__tests__/SettingsPage.test.tsx
+++ b/web/src/pages/settings/__tests__/SettingsPage.test.tsx
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { beforeAll, describe, expect, it, vi } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter, Route, Routes, useLocation } from 'react-router-dom';
+import { LangProvider } from '../../../i18n/LangContext';
+import SettingsPage from '../index';
+
+vi.mock('../GeneralSettingsTab', () => ({
+  GeneralSettingsTab: () => <div>general settings tab</div>,
+}));
+vi.mock('../AiAssistantTab', () => ({
+  AiAssistantTab: () => <div>ai assistant tab</div>,
+}));
+vi.mock('../CloudCredentialTab', () => ({
+  CloudCredentialTab: () => <div>cloud credential tab</div>,
+}));
+vi.mock('../DataSourceTab', () => ({
+  DataSourceTab: () => <div>data source tab</div>,
+}));
+vi.mock('../AboutTab', () => ({
+  AboutTab: () => <div>about tab</div>,
+}));
+
+beforeAll(() => {
+  Object.defineProperty(window, 'matchMedia', {
+    writable: true,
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
Review of PR #2573

### Findings
FINDINGS:
- [Info] web/src/pages/settings/index.tsx:46-51 — Uses the functional form of `setSearchParams` with `new URLSearchParams(current)`, which correctly preserves all unrelated params and only overwrites `tab`. This is the right pattern; no race conditions vs. the previous `setSearchParams({ tab: key })` which silently dropped everything else.
- [Info] web/src/pages/settings/__tests__/SettingsPage.test.tsx:86-93 — Single regression case covers the "preserve unrelated params" behavior with two extras (`source`, `focus`). Good signal.
- [Warning] web/src/pages/settings/__tests__/SettingsPage.test.tsx:86-93 — No negative/companion cases: (a) switching tabs when no query string exists, (b) switching twice in a row, (c) handling of a malformed `?tab=` value, or (d) verifying the previous buggy behavior is gone. Only the happy path is exercised.
- [Info] web/src/pages/settings/__tests__/SettingsPage.test.tsx:30-41 — `vi.mock` stubs for every tab component keep the test focused on the URL behavior, which is appropriate.
- [Warning] web/src/pages/settings/__tests__/SettingsPage.test.tsx:43-56 — `window.matchMedia` is installed once in `beforeAll` without a corresponding teardown (`afterAll` restoring the original). This is a minor test-environment hygiene issue and can leak into sibling suites if tests run in the same jsdom context.
- [Info] web/src/pages/settings/__tests__/SettingsPage.test.tsx:58-62 — `LocationProbe` is a clean, non-intrusive way to assert on `useLocation().search`. Good choice over spying on `navigate`.
- [Info] web/src/pages/settings/index.tsx:48 — `new URLSearchParams(current)` is allocated on every tab click. Allocation is trivial and fine for a click handler; no perf concern.
- [Info] Backward compatibility: behavior is strictly additive — existing callers that relied on `?tab=foo` still work. The only observable change is that extra params are no longer stripped, which matches user expectations for deep links.
- [Info] PR verification list includes eslint, build, targeted vitest run, and `git diff --check` — appropriate for the change size.

SUGGESTIONS:

### Suggestions
SUGGESTIONS:
- Add at least one more test case to lock in the regression from the other direction, e.g. starting at `/settings` (no query) and confirming the URL becomes `?tab=ai` after clicking the AI tab, plus a double-switch case that ensures `tab` doesn't duplicate or stale params accumulate:
  ```tsx
  it('sets tab when no query parameters exist', async () => {
    const user = userEvent.setup({ pointerEventsCheck: 0 });
    renderPage('/settings');
    await user.click(screen.getByRole('tab', { name: 'AI 助手' }));
    await waitFor(() => {
      expect(screen.getByLabelText('location-search')).toHaveTextContent('?tab=ai');
    });
  });
  ```
- Consider adding `afterAll` cleanup for the `matchMedia` mock to avoid leaking into other suites:
  ```ts
  const originalMatchMedia = window.matchMedia;
  beforeAll(() => { Object.defineProperty(window, 'matchMedia', { writable: true, value: vi.fn()... }); });
  afterAll(() => { Object.defineProperty(window, 'matchMedia', { writable: true, value: originalMatchMedia }); });
  ```
- Minor: the test hard-codes Chinese tab labels (`'AI 助手'`, `'通用设置'`). If the LangProvider's default locale ever changes, this test will break silently. Consider selecting tabs by a stable `data-testid` or by key instead of visible label, or assert the label comes from the i18n fixture explicitly.
- Optional: extract the updater into a tiny helper (e.g. `updateTabParam(searchParams, key)`) if you anticipate similar "preserve-and-set-one" patterns elsewhere; otherwise the inline lambda is perfectly fine and adding a helper would be over-engineering.

---

lizhimins:
Consolidated into #2630 (merged into rocketmq-studio) — same-author batch.

PR #2579, opened 2026-08-25
fix: paginate Apache consumer groups in the database

### LINKED ISSUE #2578: Consumer group pagination still loads every group before slicing
## Problem

`GET /api/groups/page` advertises server-side pagination, but the current `MetadataService` implementation first calls `listConsumerGroups(...)`:

```java
List<ConsumerGroupVO> groups = listConsumerGroups(instanceId, clusterId, search);
int total = groups.size();
int from = ...
return PageResult.of(groups.subList(from, to), total, page, pageSize);
```

For Apache instances, `listConsumerGroups(...)` is DB-backed and then enriches live online/lag/delay data for every matching group. The paginated endpoint therefore:

1. reads every matching `rmq_group` row;
2. performs live admin/API enrichment for every matching group;
3. builds the full VO list;
4. finally slices one page in memory.

With thousands of consumer groups, each page request is still O(total groups), both in database rows and live admin calls. This makes the bounded page endpoint much more expensive than its page size.

Topics already have a DB-level `selectPage` implementation (`listTopicsPage`), so consumer groups are inconsistent with the adjacent inventory contract.

## Expected behavior

- `GET /api/groups/page` should paginate at the database level for Apache instances.
- Read only the current page of matching `rmq_group` rows.
- Enrich live stats only for the groups on the returned page.
- Preserve existing filters: instance, cluster, and name search.
- Preserve stable deterministic ordering (`name ASC, id ASC`).
- Return the existing `items / total / page / size` contract unchanged.
- Keep the existing unpaginated `GET /api/groups` behavior for callers that intentionally need a full list.
- Other providers may keep the default in-memory pagination fallback unless they implement native pagination.

## Verification scope

Tests should prove that:

- the Apache provider uses `selectPage`, not `selectList`;
- the SQL query contains the same instance/cluster/search filters and stable ordering;
- live-stat enrichment is only attempted for returned page items;
- `MetadataService` delegates pagination to the provider instead of invoking the full-list method;
- existing invalid pagination bounds still fail before provider access.

tju-yxq:
## Summary

- add a paginated consumer-group contract to `InstanceProvider` and `MetadataProvider`
- implement Apache consumer-group pagination with MyBatis-Plus `selectPage`
- apply the same instance / cluster / search filters in SQL
- use deterministic `name ASC, id ASC` ordering
- enrich live online/lag/delay data only for the returned page items
- make `MetadataService.listConsumerGroupsPage(...)` delegate to the provider page method instead of loading and slicing the full list
- keep the existing unpaginated `GET /api/groups` contract unchanged
- extract shared `RmqGroup -> ConsumerGroupVO` mapping so full-list and paginated reads cannot drift

## Why

`GET /api/groups/page` currently calls the full `listConsumerGroups(...)` path, reads every matching `rmq_group` row, performs live enrichment for every group, then slices one page in memory. With thousands of groups, each page request is O(total groups), including admin/API fan-out. Topics already paginate at the database layer; this makes the adjacent consumer-group inventory consistent and bounded by page size.

## Tests

- focused: `MetadataServiceTest, RocketMQMetadataProviderTest, ConsumerGroupControllerTest` — 72 tests passed
- Checkstyle: 0 violations
- `git diff --check`: passed
- full backend suite also ran; its only 2 failures are pre-existing in `RocketMQMessageProviderTest` and reproduce on untouched upstream `14da4b95` (verified by stashing this change), so they are unrelated to this PR

Closes #2578

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
@@ -18,7 +18,6 @@ package org.apache.rocketmq.studio.instance.topic;
 
 import org.apache.rocketmq.studio.provider.apache.AdminClient;
 import org.apache.rocketmq.studio.provider.apache.MetadataProvider;
-import org.apache.rocketmq.studio.common.util.Pagination;
 import org.apache.rocketmq.studio.common.domain.PageResult;
 import org.apache.rocketmq.studio.common.domain.enums.InstanceVendor;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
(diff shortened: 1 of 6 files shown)


RockteMQ-AI (approved):
### Summary
This PR replaces in-memory pagination with database-level pagination for consumer group listing, improving performance for large datasets.

Changes are clean:
- `InstanceProvider` adds a default `listConsumerGroupsPage` method with in-memory fallback for backward compatibility
- `RocketMQMetadataProvider` implements DB-level pagination using MyBatis-Plus `Page`
- SQL ordering by `(name, id)` ensures stable pagination
- Null-safe comparisons in `filterByKeyWord`

LGTM — good performance improvement with clean implementation.

---

lizhimins:
Consolidated into #2581 (merged into rocketmq-studio) — same-author pagination batch.

PR #2581, opened 2026-08-25
feat: paginate system alerts

tju-yxq:
## Summary

- add `GET /api/system-alerts/page` with `level`, `page`, and bounded `pageSize`
- apply level filtering, deterministic `time DESC, id DESC` ordering, and pagination in SQL via MyBatis-Plus `Page`
- reject invalid page/pageSize before repository access
- return the existing `items / total / page / size` page contract
- add direct `findAlertById` lookup so acknowledging one alert no longer scans every event
- keep the existing unpaginated endpoint for compatibility
- add frontend API/service page support and wire Alert Events to server-driven pagination, server total, 20/50/100 page sizes, page reset on filter change, and stale-response protection
- keep simultaneous acknowledgements independent

## Why

The Alert Events page currently loads every event in one request, then filters and renders it in memory. Native alert collection can write events continuously, so the response and table grow without a bounded page size or server total. Acknowledging one event also loads the entire table just to find one row. This makes Alert Events consistent with the other bounded Studio inventories and reduces both list reads and acknowledge reads.

## Tests

- focused backend: `AlertServiceTest, SystemAlertControllerTest, MybatisPlusAlertRepositoryTest` — 65 tests passed
- focused frontend: `ops` API/service tests plus `SystemAlertsPage.test.tsx` — 3 files / 29 tests passed
- `npm run lint -- --quiet`: 0 errors (2 pre-existing warnings)
- `npm run build`: passed
- `git diff --check`: passed

Production changes: 240 additions / 118 deletions across backend and frontend, naturally exceeding 100 production lines without test padding.

Closes #2580

PR #2583, opened 2026-08-25
fix: paginate Apache ACL users in the database

### LINKED ISSUE #2582: ACL user pagination loads every Apache account before slicing
## Problem

`GET /api/acl/users/page` currently uses a full-list read for Apache instances:

```java
List<AclUserVO> users = (isTencentInstance(instanceId)
        ? tencentAclService.listUsers(instanceId) : aclRepository.findUsers()).stream()
        ...
int from = ...
int to = ...
return PageResult.of(users.subList(from, to), users.size(), page, pageSize);
```

For Apache users, this:

1. reads every row from `rmq_acl_user`;
2. decodes every secret;
3. filters and sorts all users in memory;
4. finally slices one page.

The paginated endpoint is therefore O(total ACL users), not O(pageSize). This is inconsistent with `GET /api/acl/rules`, which already performs filtering, ordering, and pagination in SQL.

## Expected behavior

- For Apache instances, `GET /api/acl/users/page` should paginate at the database level.
- Apply keyword filtering in SQL against username and access key.
- Use deterministic ordering: `gmt_create DESC, id DESC`, matching the current frontend-visible order.
- Read and decode credentials only for the returned page.
- Return the existing `items / total / page / size` contract.
- Preserve masked credentials in the page response.
- Keep the Tencent role path unchanged because Tencent role discovery remains a paginated remote API collection without a local database.
- Keep the existing unpaginated endpoint for callers that intentionally need the full inventory.

## Verification scope

Tests should prove that the repository uses `selectPage` rather than `selectList`, the SQL contains username/access-key filtering and deterministic ordering, the service delegates to the repository page method for Apache users, and credentials remain masked.

tju-yxq:
## Summary

- add `AclRepository.findUserPage(keyword, page, pageSize)`
- implement Apache ACL user pagination with MyBatis-Plus `selectPage`
- apply username/access-key keyword filtering in SQL
- use deterministic `gmt_create DESC, id DESC` ordering
- decode and mask credentials only for the returned page
- make `AclService.pageUsers` use the repository page for Apache instances
- keep Tencent role discovery unchanged, since Tencent roles are collected from a remote paginated API and have no local database rows
- keep the existing unpaginated user endpoint unchanged

## Why

`GET /api/acl/users/page` currently reads every `rmq_acl_user` row, decodes every secret, filters/sorts in memory, and then slices one page. ACL rules already paginate in SQL, so this makes the adjacent user inventory consistent and bounds page cost by pageSize rather than total account count.

## Tests

- focused: `AclServiceTest, MybatisPlusAclRepositoryTest, AclControllerTest` — 108 tests passed
- Checkstyle: 0 violations
- `git diff --check`: passed

Closes #2582

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
@@ -105,10 +105,11 @@ public class MybatisPlusAclRepository implements AclRepository {
 
     @Override
     public PageResult<AclUserVO> findUserPage(String keyword, int page, int pageSize) {
+        String search = StringUtils.hasText(keyword) ? keyword.trim().toLowerCase() : null;
         QueryWrapper<RmqAclUser> query = new QueryWrapper<RmqAclUser>()
-                .and(StringUtils.hasText(keyword), w -> w
-                        .like("username", keyword)
-                        .or().like("access_key", keyword))
+                .and(search != null, w -> w
+                        .like("username", search)
+                        .or().like("access_key", search))
                 .orderByDesc("gmt_create")
                 .orderByDesc("id");
         IPage<RmqAclUser> mapperPage = userMapper.selectPage(new Page<>(page, pageSize), query);
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
@@ -104,6 +104,22 @@ public class MybatisPlusAclRepository implements AclRepository {
     }
 
     @Override
+    public PageResult<AclUserVO> findUserPage(String keyword, int page, int pageSize) {
+        QueryWrapper<RmqAclUser> query = new QueryWrapper<RmqAclUser>()
+                .and(StringUtils.hasText(keyword), w -> w
+                        .like("username", keyword)
+                        .or().like("access_key", keyword))
+                .orderByDesc("gmt_create")
+                .orderByDesc("id");
+        IPage<RmqAclUser> mapperPage = userMapper.selectPage(new Page<>(page, pageSize), query);
+        List<AclUserVO> items = mapperPage.getRecords().stream()
+                .map(MybatisPlusAclRepository::toUserVO)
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
```diff
@@ -103,6 +103,22 @@ public List<AclUserVO> findUsers() {
...
+    @Override
+    public PageResult<AclUserVO> findUserPage(String keyword, int page, int pageSize) {
+        QueryWrapper<RmqAclUser> query = new QueryWrapper<RmqAclUser>()
+                .and(StringUtils.hasText(keyword), w -> w
+                        .like("username", keyword)
+                        .or().like("access_key", keyword))
+                .orderByDesc("gmt_create")
+                .orderByDesc("id");
+        IPage<RmqAclUser> mapperPage = userMapper.selectPage(new Page<>(page, pageSize), query);
+        List<AclUserVO> items = mapperPage.getRecords().stream()
```
RockteMQ-AI, line 116: **[Info]** The `LIKE` keyword filter is consistent with `findRulePage`, but note that MySQL `LIKE` is case-insensitive by default with `utf8mb4_general_ci` collation. If the collation were ever changed to a binary/case-sensitive one, this would silently break keyword matching. Consider adding a comment or using `LOWER(username) LIKE LOWER(?)` for explicitness — though this is a very minor nit and the current approach matches existing patterns.

RockteMQ-AI (approved):

tju-yxq:
Thanks for the note. I made keyword matching explicit and independent of the table collation: the repository now trims and lower-cases the keyword before the `LIKE` query, and the focused test verifies the normalized bound value. Focused run: `AclServiceTest, MybatisPlusAclRepositoryTest, AclControllerTest` — 108 tests passed, Checkstyle clean. Pushed as `d4a0effe`.

RockteMQ-AI (approved):

---

RockteMQ-AI (approved):
### Summary
Paginate Apache ACL users in the database — adds pagination for ACL user listing.

### Review

LGTM

---

lizhimins:
Consolidated into #2581 (merged into rocketmq-studio) — same-author pagination batch.

PR #2594, opened 2026-08-25
fix: sequence client connection requests

### LINKED ISSUE #2593: Client Connections can render stale registry and connection responses
## Problem

The Client Connections page uses an independent local `cancelled` flag for each request:

```ts
let cancelled = false;
...
return () => {
  cancelled = true;
};
```

This protects against unmount, but it does not sequence concurrent requests. A slow earlier response can still resolve after a newer request has started and overwrite the newer data. The page has two such paths:

- registry discovery can overwrite newer cluster options after a retry;
- the connection query can overwrite newer connections after switching NameServers.

The connection path also does not reliably enter loading state for each new request; loading is mainly set by the NameServer change handler and registry retry, which leaves retries and endpoint changes driven by registry discovery inconsistent.

## Expected behavior

- Sequence registry discovery requests with a request generation.
- Sequence connection requests with a request generation.
- Ignore stale success, error, and loading updates for both request types.
- Enter loading state when a connection request starts, without violating the React lint rule against synchronous setState in effect bodies.
- Keep registry retry and NameServer switching behavior unchanged.

## Verification scope

Tests should cover:
- a stale connection response after switching NameServers;
- a stale registry response after retrying discovery;
- existing retry, failure, statistics, export, and dialog behavior remaining unchanged.

tju-yxq:
Closes #2593

## Summary

- sequence registry discovery with a request-generation ref
- sequence NameServer connection queries with a separate request-generation ref
- ignore stale registry and connection success, error, and loading updates
- enter loading state when a connection request starts, without synchronous setState in the effect body
- keep existing retry, NameServer switching, statistics, CSV export, and detail dialog behavior

## Why

Each effect previously used an independent local `cancelled` flag. That only prevented updates after unmount; it did not distinguish an older in-flight request from the latest one. A slow registry response could overwrite newer retry data, and a slow connection response could replace the connections for a newly selected NameServer.

## Tests

- focused: `ClientsPage.test.tsx` — 15 tests passed
- `npm run lint -- --quiet`: 0 errors, 2 pre-existing warnings
- `npm run build`: passed
- `git diff --check`: passed

Production changes: 24 additions / 20 deletions. The full PR is 92 additions / 21 deletions. This is a genuine correctness fix, but production changed lines are below the separate 100-line target, so the tracking ledger records it as a candidate rather than a complete group.

### DIFF
--- a/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import { App } from 'antd';
-import { render, screen, waitFor, within } from '@testing-library/react';
+import { act, render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type React from 'react';
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

---

RockteMQ-AI (approved):
### Summary
Sequence client connection requests — prevents race conditions from concurrent client connection operations.

### Review

LGTM

---

lizhimins:
Consolidated into #2592 (merged into rocketmq-studio) — same-author request sequencing batch.

PR #2588, opened 2026-08-25
chore: remove unreachable legacy group page

tju-yxq:
## Summary

- remove the unreachable `GroupManagement` page component and its tests
- remove its lazy import from the router
- redirect `/studio/group-management` to the active `/instance/consumer` page for old bookmarks
- remove the now-unused `groupMgmt.*` translation namespace
- leave the active instance-scoped consumer page untouched

## Why

The Studio route contains two consumer group UIs, but only `/instance/consumer` is reachable from navigation. `/studio/group-management` has no menu, link, button, or documented entry point. It still ships stale global-scope logic, in-memory search/pagination, duplicate diagnostics UI, and a separate translation namespace. Removing the dead page removes over 1,000 lines of unreachable code while preserving a safe redirect.

## Tests

- `npm run build`: passed
- `npm run lint -- --quiet`: 0 errors (2 pre-existing warnings)
- Search: no remaining `GroupManagement` implementation or `groupMgmt.*` translation usage
- `git diff --check`: passed
- Full frontend suite has 9 failures in the two Message page test files, reproduced identically on the untouched upstream baseline with `git stash`, so they are pre-existing and unrelated to this change

Closes #2587

PR #2596, opened 2026-08-25
fix: avoid duplicate cluster broker config reads

### LINKED ISSUE #2595: Cluster discovery repeats failed broker-config reads for duplicate brokers
## Problem

`ClusterService.listClusters()` enriches every discovered cluster independently:

```java
discovered.forEach(this::enrichWithLiveConfig);
```

Each `enrichWithLiveConfig` call walks that cluster’s brokers and attempts a live broker-config read until one succeeds.

This is correct for distinct clusters, but discovery can return multiple registry views that contain the same broker. In that case the same broker address is retried once per cluster. If that broker is unavailable, each duplicate repeats the same failing admin call before falling back to persisted configuration.

The selected-instance path has the same behavior.

## Expected behavior

- Keep attempting distinct broker addresses in discovery order.
- Record broker addresses already attempted while enriching one discovery result.
- Do not repeat a live broker-config read for the same broker address in the same discovery call.
- Preserve the persisted-config fallback for every cluster whose brokers are unavailable or already attempted.
- Keep single-cluster detail behavior unchanged.

## Verification scope

Tests should prove that duplicate clusters referencing the same broker make exactly one live-config read for that broker and each duplicate receives its own persisted fallback configuration.

tju-yxq:
Closes #2595

## Summary

- enrich a discovery result with one shared attempted-broker-address set
- attempt each distinct broker address at most once per discovery call
- let duplicate clusters referencing an unavailable broker use their persisted-config fallback immediately
- preserve existing single-cluster detail enrichment behavior
- add a regression test with two discovered clusters sharing one unavailable broker

## Why

Each discovered cluster previously performed live broker-config enrichment independently. Discovery can return multiple registry views containing the same broker; when that broker was unavailable, every duplicate repeated the same failing admin call before falling back.

## Tests

- focused: `ClusterServiceTest` — 39 tests passed, Checkstyle 0 violations
- backend full suite: 1,627 tests run; only the 2 pre-existing `RocketMQMessageProviderTest` failures remain
- `git diff --check`: passed

Production changes: 31 additions / 3 deletions. The full PR is 64 additions / 3 deletions. This is a real efficiency/fallback fix, but production changed lines are below the separate 100-line target, so the tracking ledger records it as a candidate rather than a complete group.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -42,11 +42,13 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
 import java.util.ArrayList;
+import java.util.LinkedHashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.Set;
 
 @Slf4j
 @Service
@@ -81,7 +83,7 @@ public class ClusterService {
         log.info("Listing all clusters");
         List<ClusterVO> discovered = clusterProvider.discoverClusters();
         if (discovered != null && !discovered.isEmpty()) {
-            discovered.forEach(this::enrichWithLiveConfig);
+            enrichDiscoveredClusters(discovered, null);
             return discovered;
         }
         return List.of();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

---

RockteMQ-AI (approved):
### Summary
Avoid duplicate cluster broker config reads — eliminates redundant API calls for broker configuration.

### Review

LGTM

---

lizhimins:
Consolidated into #2588 (merged into rocketmq-studio) — same-author misc batch.

PR #2602, opened 2026-08-25
fix: record message query history once per page set

### LINKED ISSUE #2601: Message query pagination writes duplicate server history rows
## Problem

`GET /api/messages/page` delegates to `MessageService.queryMessages(...)`, which records a message-query history entry on every provider query:

```java
List<MessageRecordVO> result = queryMessages(instanceId, topic, msgId, tag, key, startTime, endTime);
```

When the user pages through the same message result, every page requests the same provider query and therefore writes another row to `rmq_message_query`. The browser’s local query history already avoids duplicate entries for pagination, but the server-side history database does not.

For a query returning several pages, the server history becomes:

- page 1 → one history row
- page 2 → another identical history row
- page 3 → another identical history row

This inflates the server query history and summary counts without representing a distinct user query.

## Expected behavior

- Preserve history recording for direct message queries.
- Record server history for the first page of a paginated message query.
- Do not write duplicate history rows for subsequent pages of the same paginated result.
- Keep provider behavior and page slicing unchanged.

## Verification scope

A focused service test should prove that first and later pages both execute the provider query, but only the first page records message-query history.

tju-yxq:
Closes #2601

## Summary

- keep message-query history recording for direct queries
- record history only for the first page of a paginated message query
- keep provider query and page slicing behavior unchanged
- add a focused test that first and later pages both query the provider, while only page 1 writes history

## Why

Every paginated request previously reused the direct query path, which records a history row after each provider query. Paging through one result therefore created duplicate `rmq_message_query` rows and inflated server history/summary counts.

## Tests

- focused: `MessageServiceTest, MessageControllerTest` — 15 tests passed, Checkstyle 0 violations
- backend full suite: 1,627 tests run; only the 2 known pre-existing `RocketMQMessageProviderTest` failures remain
- `git diff --check`: passed

Production changes: 11 additions / 2 deletions. The full PR is 32 additions / 2 deletions. This is a real correctness/data-quality fix, but production changed lines are below the separate 100-line target, so the tracking ledger records it as a candidate rather than a complete group.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
@@ -42,12 +42,20 @@ public class MessageService {
 
     public List<MessageRecordVO> queryMessages(
             String instanceId, String topic, String msgId, String tag, String key, Long startTime, Long endTime) {
+        return queryMessages(instanceId, topic, msgId, tag, key, startTime, endTime, true);
+    }
+
+    private List<MessageRecordVO> queryMessages(
+            String instanceId, String topic, String msgId, String tag, String key,
+            Long startTime, Long endTime, boolean recordHistory) {
         validateTopicQueryWindow(topic, msgId, key, startTime, endTime);
         log.info("Querying messages: topic={}, msgId={}, tag={}, key={}", topic, msgId, tag, key);
         List<MessageRecordVO> result = providerRegistry.byInstanceId(instanceId)
                 .map(provider -> provider.queryMessages(instanceId, topic, msgId, tag, key, startTime, endTime))
                 .orElseGet(() -> messageProvider.queryMessages(instanceId, topic, msgId, tag, key, startTime, endTime));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix for duplicate query history entries during pagination. The approach of gating `recordHistory` on `page == 1` is correct — all pages share the same underlying query result, so history should only be recorded once per query, not per page navigation.

### Findings
- **[Info]** `server/src/main/java/.../MessageService.java:45-48` — The public overload correctly defaults `recordHistory=true`, preserving backward compatibility for any direct callers of the 7-arg method.
- **[Info]** `server/src/test/java/.../MessageServiceTest.java:193-210` — Test is well-structured: verifies both that the provider is called twice (pagination works) and that history is recorded exactly once.

LGTM. Nice focused fix with proper test coverage.

---

RockteMQ-AI (approved):
### Summary
Record message query history once per page set — prevents duplicate history entries when navigating pages.

### Review

LGTM

---

lizhimins:
Consolidated into #2588 (merged into rocketmq-studio) — same-author misc batch.

PR #2610, opened 2026-08-26
Fix Tencent cloud inventory pagination

### LINKED ISSUE #2609: [Studio][Bug] Prevent silent truncation of Tencent instance and consumer group inventories
## Problem

Tencent cloud instance discovery and consumer-group inventory use a fixed `PAGE_SIZE = 100` and `MAX_PAGES = 100`. When the 100th page is full, both loops return a successful but incomplete inventory.

The Tencent SDK response already exposes `TotalCount`, but these loops do not use it.

## Locations

- `server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java`
- `server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentInstanceProvider.java`

## Impact

Deployments with more than 10,000 Tencent Cloud RocketMQ instances in a region, or more than 10,000 consumer groups in an instance, silently lose later records. This makes instance import, group management, exports, and counts incomplete while the API reports success.

## Expected behavior

Continue pagination until `TotalCount` is reached, while retaining the short-page fallback for missing totals. Add regression coverage for exact and over-10,000 inventories.

## Related

- #2442 covers the analogous Aliyun consumer-group cap.
- #2355 covers cloud Topic paging.
- Neither covers Tencent instance catalog or consumer-group inventory truncation.

Aias00:
## Summary
- use Tencent `TotalCount` to continue cloud-instance and consumer-group inventory pagination beyond the historical 10,000-record cap
- retain short-page termination when the upstream response omits a total
- cover exact 10,000 and 10,001 record boundaries

Fixes #2609

## Tests
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) mvn -q -Dtest=TencentCatalogServiceTest,TencentInstanceProviderTest test`
- `JAVA_HOME=$( /usr/libexec/java_home -v 21 ) mvn -q -DskipTests package`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentCatalogService.java
@@ -43,7 +43,6 @@ import java.util.Locale;
 public class TencentCatalogService implements CloudCatalogProvider {
 
     static final int PAGE_SIZE = 100;
-    static final int MAX_PAGES = 100;
     static final List<CloudRegionVO> SUPPORTED_REGIONS = List.of(
             region("ap-guangzhou", "Guangzhou"),
             region("ap-shenzhen-fsi", "Shenzhen Finance"),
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Solid fix for Tencent cloud inventory pagination. The old page-based loop with `MAX_PAGES = 100` could silently truncate results for accounts with many instances. The new offset-based approach with `hasFetchedAll(offset, totalCount)` is more robust — it uses the API's `totalCount` for a definitive termination condition instead of an arbitrary page cap.

The same pattern is correctly applied to `listConsumerGroups` and `listTopics`. Tests cover both the `totalCount`-based early exit and the empty-response edge case.

LGTM.

---

Aias00:
Superseded by #2956, which consolidates this change with the related Studio fixes on the current `rocketmq-studio` baseline. The replacement PR is Ready and currently CLEAN/MERGEABLE.

PR #2565, opened 2026-08-25
fix(studio): localize settings management tabs

Aias00:
## Summary

- Localize all user-visible strings in the Data Source, Cloud Credentials, General Settings, and AI Assistant tabs.
- Preserve the existing Chinese copy while adding English labels, validation feedback, confirmations, and pagination summaries.
- Make the default language context render Chinese translations rather than raw translation keys when a component is used outside `LangProvider`.
- Add English-locale regression coverage for each Settings tab.

## Validation

- `npm test -- --run src/pages/settings/__tests__/DataSourceTab.test.tsx src/pages/settings/__tests__/CloudCredentialTab.test.tsx src/pages/settings/__tests__/GeneralSettingsTab.test.tsx src/pages/settings/__tests__/AiAssistantTab.test.tsx`
- `npx eslint src/i18n/LangContext.tsx src/i18n/translations.ts src/pages/settings/DataSourceTab.tsx src/pages/settings/CloudCredentialTab.tsx src/pages/settings/GeneralSettingsTab.tsx src/pages/settings/AiAssistantTab.tsx src/pages/settings/__tests__/DataSourceTab.test.tsx src/pages/settings/__tests__/CloudCredentialTab.test.tsx src/pages/settings/__tests__/GeneralSettingsTab.test.tsx src/pages/settings/__tests__/AiAssistantTab.test.tsx --max-warnings=0`
- `npm run build`
- `git diff --check`

Closes #2564.

PR #2619, opened 2026-08-26
[Studio] fix: resolve code-quality and resource-leak findings

zhaohai666:
# PR: Studio code-quality and resource-leak cleanups

- **From:** `zhaohai666:feature/studio-optimization-cleanup`
- **Into:** `apache:rocketmq-studio`
- **Branch base:** `14da4b95` (`fix(studio): localize settings management tabs (#2565)`)
- **Status:** backend tests green (47/47), `web` `tsc -b` clean

## Summary

A focused cleanup that resolves three concrete findings from the `rocketmq-studio` code
review (the branch stood at `7443227f` when reviewed; this PR is rebased onto the current
tip `14da4b95`). The changes are low-risk, self-contained, and covered by tests.

| ID | Area | Change |
|----|------|--------|
| F2 | Frontend types | Remove duplicate `ResetConsumerOffsetRequest` interface |
| B6 | Backend resources | Add `@PreDestroy` to `TencentClientFactory` |
| N1 | Backend memory | Stream DLQ Excel export to the response `OutputStream` |

### F2 — Duplicate `ResetConsumerOffsetRequest` interface

`web/src/api/metadata.ts` declared `ResetConsumerOffsetRequest` **twice** (line 148 without
`instanceId`, line 310 with `instanceId?`). Because both are `interface` declarations,
TypeScript merges them into the superset `{ name; timestamp; topic; instanceId? }`, so the
project compiles — but the first definition is dead, confusing duplication that invites
drift. Removed the `:148` definition; the surviving shape is a strict superset of all current
callers (which route through `services/consumerService.ts`), so no call site changes.

### B6 — `TencentClientFactory` resource leak on shutdown

`TencentClientFactory` caches one `TrocketClient` per `credential#region` but had **no
`@PreDestroy`**, unlike `AliyunClientFactory` which already releases its clients on context
shutdown. Added a `@PreDestroy close()` that clears the cache so the cached clients (and their
underlying connection pools) are released instead of being retained for the whole process
lifetime.

> Note: the Tencent SDK `AbstractClient` exposes no public `shutdown()`/`close()` method, so
> the factory clears the cache to drop references and let GC reclaim the clients. This mirrors
> the intent of the existing Aliyun `@PreDestroy` without calling a non-existent API.

### N1 — DLQ Excel export heap buffering

`RocketMQDLQProvider.exportExcel` built the entire `.xlsx` workbook into a
`ByteArrayOutputStream` and returned it as a `byte[]`, which the controller then copied into a
`ResponseEntity<byte[]>`. For a large export (capped at `RESEND_HARD_CAP = 5000` messages, but
still potentially several MB) this doubles the peak heap usage.

The export now streams directly to the HTTP response `OutputStream`:

- `DLQExcelExportResultVO` no longer carries the `byte[] data` field — it holds only the
  scan-completeness metadata (`truncated`, `failedQueueCount`, `limit`).
- `exportExcel(...)` gained an `OutputStream out` parameter; the provider writes the sheet to
  it via `EasyExcel.write(out, ...).doWrite(rows)` and flushes.
- `DLQController.exportDLQExcel` sets the content-type/disposition/metadata headers on the
  `HttpServletResponse` and streams the workbook to `response.getOutputStream()`; it returns
  `ResponseEntity<Void>`.
- `DLQControllerTest` updated to the new method arity; the two `export-excel` tests now assert
  the response headers (the bytes are streamed, so the mocked service no longer returns them).

## Verification

- Backend (`mvn -o test`, JDK 21, local repo): `DLQControllerTest` 23,
  `RocketMQDLQProviderTest` 20, `TencentClientFactoryTest` 4 — **47/47 pass**.
- Frontend: `npx tsc -b` clean.

## Out of scope (recommended as follow-ups)

The review surfaced additional findings that are intentionally **not** in this PR:

- **Frontend de-duplication (F3–F7, F10, F11, A1)** — `formatDateTime`/`getErrorMessage`/
  sort helpers, `Table` column `useMemo`, render-time `setState`, error double-toast, and the
  `api/`↔`services/` double-exit. These are already implemented in the unmerged branches
  `feature/studio-frontend-optimizations` / `feature/studio-web-fixes` and should be merged
  directly rather than re-done here.
- **B4** — DLQ list still calls `examineTopicStats` per group (pagination caps the page, but
  it is still a per-group RPC). Suggested fix: compute the message count from `Min/MaxOffset`
  and lazy-load full stats on the detail view.
- **N2** — `RocketMQMessageProviderTest.queryByTopic*` expects `MAX_PULLS_PER_QUEUE = 1000`
  while the shipped constant is `32`; these two tests are red on trunk. Needs a product
  decision on the intended tail-budget size before adjusting the constant or the assertions.
- **N3** — `/export-excel` and `/export` both scan dead letters; the scan logic can be shared.
- **N4** — the newly added queue-browsing / direct-consume / DLQ-export pages should be
  re-checked against the F6/F7 anti-patterns.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQController.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.rocketmq.studio.common.domain.PageResult;
 import org.apache.rocketmq.studio.common.domain.Result;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
+import jakarta.servlet.http.HttpServletResponse;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.Max;
 import jakarta.validation.constraints.Min;
@@ -39,6 +40,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQController.java
```diff
@@ -95,22 +97,25 @@ public Result<PageResult<DLQMessageVO>> listDLQMessages(@PathVariable String gro
...
+                                               @RequestParam(required = false) Long startTime,
+                                               @RequestParam(required = false) Long endTime,
+                                               @Size(max = MAX_SELECTED_MESSAGES,
+                                                       message = "At most 100 msgIds are allowed per export")
+                                               @RequestParam(required = false) List<String> msgIds,
+                                               HttpServletResponse response) throws IOException {
+        // Stream the workbook straight to the response body so it is never buffered in the heap.
+        response.setContentType(EXCEL_MEDIA_TYPE);
+        response.setHeader(HttpHeaders.CONTENT_DISPOSITION,
+                attachmentDisposition("dlq-" + sanitizeForFilename(groupName) + ".xlsx").toString());
```
RockteMQ-AI, line 111: **[Warning]** Response headers set after streaming write.\n\nThe metadata headers (`X-DLQ-Export-Truncated`, `X-DLQ-Export-FailedQueues`, `X-DLQ-Export-Limit`) are set **after** `dlqService.exportExcel()` writes to `response.getOutputStream()`. In a real servlet container, if the Excel data exceeds the response buffer (typically 8 KB for Tomcat), the buffer auto-flushes during the write and the response becomes committed — subsequent `setHeader()` calls are silently ignored.\n\nFor large exports, the client would lose the completeness metadata headers, which defeats their purpose (the client uses them to detect truncation and failed queues).\n\n**Suggested fix:** Set all headers **before** the streaming write. Since the metadata is only available after the export completes, consider a two-phase approach:\n```java\n// Phase 1: write to a temp buffer / piped stream to collect metadata\n// Phase 2: set headers from metadata\n// Phase 3: pipe buffer to response output\n```\nOr restructure `exportExcel` to return metadata before writing the body.

server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
```diff
@@ -335,23 +334,24 @@ public DLQExcelExportResultVO exportExcel(String instanceId, String groupName, L
...
+            // in the heap (the export is capped at RESEND_HARD_CAP messages, but large snapshots can
+            // still be several MB).
+            com.alibaba.excel.EasyExcel.write(out, DLQMessageExcelRow.class)
                     .sheet("DLQ")
-                    .doWrite(messages.stream().map(DLQMessageExcelRow::from).toList());
-            data = output.toByteArray();
+                    .doWrite(rows);
+            out.flush();
         } catch (Exception e) {
             log.warn("Failed to build Excel export for group {}: {}", groupName, e.getMessage());
```
RockteMQ-AI, line 351: **[Info]** Redundant flush. `out.flush()` is called here (line 351) and again in the controller (`response.getOutputStream().flush()`). The double flush is harmless but unnecessary — one flush at the controller level is sufficient.

RockteMQ-AI (commented):
/tmp/review-apache/rocketmq-dashboard-2619/summary.txt

lizhimins:
Closing after review: the streaming DLQ Excel export sets response headers after the response body has already started, so the headers are inevitably lost — the fix direction does not work as implemented. The 09-01 update did not address this. If you'd like to rework it (collect/buffer before committing headers, or stream with pre-set headers), feel free to reopen or submit a new PR. Thanks!

PR #2627, opened 2026-08-26
fix(web): handle unavailable ACL credentials

### LINKED ISSUE #2623: fix(web): handle ACL users with unavailable credentials
### Problem

ACL user records from cloud providers do not always include access-key or secret-key values. The web API type currently requires both fields, while the page normalizes missing values to empty strings. As a result, the Access Key cell is rendered as a blank copyable value, which looks like a rendering failure and offers a copy action that copies nothing.

### Proposed change

Represent optional credentials in the API contract and render a clear non-copyable placeholder when an access key is unavailable. Add regression coverage for an ACL user response with missing credentials.

yyqdbngt:
## Summary

- model cloud-backed ACL credentials as optional API fields
- normalize user records before table sorting and editing
- show a non-copyable placeholder instead of a blank copy action when an access key is unavailable
- cover a credential-less ACL user response in the page tests

## Testing

- `npm test -- --run src/pages/instance/__tests__/AclPage.test.tsx` (18 tests passed)
- `npm run build`

Closes #2623

### DIFF
--- a/web/src/api/acl.ts
+++ b/web/src/api/acl.ts
@@ -50,8 +50,8 @@ export interface AclUserPage {
 export interface AclUser {
   id: AclEntityId;
   username: string;
-  accessKey: string;
-  secretKey: string;
+  accessKey?: string | null;
+  secretKey?: string | null;
   admin: boolean;
   clusters: string[];
   permRead?: boolean;
--- a/web/src/components/QueueBrowser.tsx
+++ b/web/src/components/QueueBrowser.tsx
@@ -42,9 +42,9 @@ export interface TopicOption {
   value: string;
 }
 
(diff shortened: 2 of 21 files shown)


RockteMQ-AI (approved):
### Summary
Models cloud-backed ACL credentials as optional (`accessKey?: string | null`) in the API type, introduces a `NormalizedAclUser` type that guarantees non-null strings after normalization, and renders a non-copyable `-` placeholder when credentials are unavailable. The `copyable` prop is now conditionally disabled when no key exists, preventing a confusing empty copy action.

Well-structured type-level fix with a clear separation between the raw API shape and the normalized UI model. Test covers the null-credential scenario. LGTM.

---

PR #2628, opened 2026-08-26
fix(provider): tolerate null topic route queues

### LINKED ISSUE #2624: fix(provider): tolerate null queue entries in topic routes
### Problem

`RocketMQAdminClientImpl#getTopic` only checks whether the route queue list is absent or empty before dereferencing its first element. A malformed or partially populated route response whose first queue entry is `null` therefore becomes a generic 500 response, even when a later entry contains usable topic metadata.

### Proposed change

Select the first usable queue entry, return the existing not-found response when no usable entry exists, and add regression tests for both partial and fully unusable route responses.

btlqql:
## Summary

- skip null queue entries when reading topic route metadata
- keep using later valid queue metadata in partially populated route responses
- return the existing 404 response when a route contains no usable queue entry
- add regression coverage for partial and fully unusable routes

## Testing

- `mvn -Dtest=RocketMQAdminClientImplTest test` (37 tests passed; Checkstyle passed)

Closes #2624

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
@@ -102,7 +102,10 @@ public class RocketMQAdminClientImpl implements AdminClient {
                 if (routeData == null || routeData.getQueueDatas() == null || routeData.getQueueDatas().isEmpty()) {
                     throw new BusinessException(404, "Topic not found: " + name);
                 }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fix replaces unsafe `getQueueDatas().get(0)` with a null-safe stream filter that finds the first usable `QueueData` entry, falling back to a 404 `BusinessException` when all entries are null. Two regression tests cover the partial-null and all-null scenarios.

Clean, targeted fix with appropriate test coverage. LGTM.

---

PR #2629, opened 2026-08-26
[Studio] fix: reset metrics data source when instance scope changes

coder999o:

Reset Metrics Explorer to the default Prometheus source when the selected data source is no longer available for the active instance, so metric refreshes do not keep using an out-of-scope data source or leave the chart cleared.

## Brief changelog

- Track the data sources available to the current instance.
- Fall back to the default source when the selected data source leaves the active instance scope.
- Clear temporary data-source credentials and pending auth state during fallback.
- Add regression coverage for instance-scope datasource fallback.

## Verifying this change

- npm test -- src/components/__tests__/MetricsExplorer.test.tsx
- npm run lint
- npm run build
- npx prettier --check src/components/MetricsExplorer.tsx src/components/__tests__/MetricsExplorer.test.tsx
- git diff --check

### DIFF
--- a/web/src/components/MetricsExplorer.tsx
+++ b/web/src/components/MetricsExplorer.tsx
@@ -402,6 +402,11 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
       ),
     [dataSources, instanceId],
   );
+  const availableDataSourceKeysRef = useRef<Set<string>>(new Set());
+
+  useEffect(() => {
+    availableDataSourceKeysRef.current = new Set(availableDataSources.map((source) => source.key));
+  }, [availableDataSources]);
 
   const loadMetrics = useCallback(
     async (metric: MetricMapping | undefined, range: (typeof RANGE_OPTIONS)[number]) => {
@@ -417,7 +422,11 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
       setQueryLoading(true);
       setQueryError(null);
       try {
-        const currentDataSourceKey = dataSourceKeyRef.current;
+        const selectedDataSourceKey = dataSourceKeyRef.current;
+        const currentDataSourceKey =
+          selectedDataSourceKey && availableDataSourceKeysRef.current.has(selectedDataSourceKey)
+            ? selectedDataSourceKey
+            : '';
         const credentials =
           dataSourceCredentialsRef.current?.key === currentDataSourceKey
             ? dataSourceCredentialsRef.current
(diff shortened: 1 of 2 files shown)


web/src/components/MetricsExplorer.tsx
```diff
@@ -554,12 +563,15 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
...
+        dataSourceCredentialsRef.current = null;
         dataSourceKeyRef.current = '';
         setDataSourceKey('');
         setData(null);
+        setPendingDataSource(null);
+        void loadMetrics(selectedMetric, selectedRange);
       }, 0);
     }
-  }, [availableDataSources, dataSourceKey]);
+  }, [availableDataSources, dataSourceKey, loadMetrics, selectedMetric, selectedRange]);
```
RockteMQ-AI, line 575: **[Warning]** Adding `loadMetrics`, `selectedMetric`, and `selectedRange` to this effect's dependency array changes the original behavior significantly.

The original effect only ran when `availableDataSources` or `dataSourceKey` changed (i.e., on instance switch or data source selection). With the new dependencies, this effect will also fire every time the user selects a different metric or time range, which will:

1. Reset `dataSourceKeyRef.current` to `''` (clearing the data source selection)
2. Clear the chart data via `setData(null)`
3. Reload metrics from the default data source

This means every metric/range change would lose the user's selected data source — likely a regression.

**Suggested fix:** Use refs to capture the current metric/range values without adding them to the dependency array:

```typescript
const selectedMetricRef = useRef(selectedMetric);
const selectedRangeRef = useRef(selectedRange);
useEffect(() => {
  selectedMetricRef.current = selectedMetric;
  selectedRangeRef.current = selectedRange;
}, [selectedMetric, selectedRange]);

// Then in the reset effect:
useEffect(() => {
  window.setTimeout(() => {
    // ... reset logic ...
    void loadMetrics(selectedMetricRef.current, selectedRangeRef.current);
  }, 0);
}, [availableDataSources, dataSourceKey]); // keep original deps only
```

Alternatively, if the intent is to also reload on metric/range change, consider separating that into a distinct `useEffect` with a clear comment explaining the behavior.

RockteMQ-AI (commented):
﻿### Summary
This PR fixes a real bug where the Metrics Explorer keeps using a data source that's no longer available after switching instances. The core fix (tracking available data source keys and falling back to default) is correct and well-tested.

However, there's a concern about the `useEffect` dependency array expansion that may introduce a regression where metric/range changes reset the data source selection.

### Findings
- **[Warning]** web/src/components/MetricsExplorer.tsx:575 — useEffect dependency array expansion may cause unintended resets on metric/range changes

### Suggestions
- Consider using refs for `selectedMetric` and `selectedRange` inside the reset effect to avoid adding them to the dependency array
- Alternatively, separate the "reload on metric/range change" logic into its own effect with clear documentation

---

lizhimins:
Consolidated into #2630 (merged into rocketmq-studio) — same-author batch.

PR #2636, opened 2026-08-27
fix(dlq): resolve selected messages by id

### LINKED ISSUE #2635: Selected DLQ resend misses messages outside its fixed scan window
### Problem
The selected-message DLQ resend API receives explicit message IDs, but it currently rescans only the last seven days of the DLQ topic and stops after 5,000 messages before filtering those IDs. A message selected from an older time range, or one beyond the scan cap, is reported as unmatched and is not resent even though its exact ID is available.

### Expected behavior
Selected resend should resolve each requested message ID directly from the group's DLQ topic, independent of message age and unrelated queue volume. Missing IDs should still produce a partial result.

### Proposed fix
Use the admin API's topic-aware message lookup for each normalized ID, retain per-ID failure handling, and add a regression test proving that selected resend does not invoke a time-window scan.

btlqql:
## Summary
- resolve selected DLQ messages directly by their IDs instead of rescanning a fixed seven-day window
- preserve request order while deduplicating IDs and continue past individual lookup failures
- mark incomplete selected lookups as partial and expose that through `scanIncomplete`

## Why
The selected resend endpoint already receives exact message IDs. Rescanning only the most recent seven days, with a 5,000-message cap, can miss an older selected message or one beyond the cap even though its ID is known.

## Tests
- `mvn -o -Dtest=RocketMQDLQProviderTest test` (22 tests passed)

Fixes #2635

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
@@ -58,6 +58,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedHashSet;
 import java.util.Locale;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR improves the DLQ selected-message resend flow by resolving messages directly via their IDs instead of scanning a fixed 7-day window with a 5,000-message cap.

Key improvements:
- **Correctness**: Messages beyond the 7-day window or 5,000-message cap are no longer missed when their IDs are known
- **Efficiency**: Direct `viewMessage` lookups avoid scanning unnecessary messages
- **Order preservation**: `LinkedHashSet` maintains the original request order
- **Graceful degradation**: `scanIncomplete` flag accurately reports partial results, and individual lookup failures are logged but do not block the entire operation

The test coverage is thorough, covering both the happy path and partial failure scenarios. The change is well-scoped and backward-compatible.

LGTM.

---
echo
File content:
cat
/tmp/review_2636.txt

PR #2165, opened 2026-08-14
fix(web): tolerate unavailable browser storage

yyqdbngt:
## Summary

- centralize guarded local-storage reads, writes and removals
- keep authentication, preferences and Proxy state usable when browser storage is blocked
- prevent storage SecurityError exceptions from breaking authentication or Proxy rendering
- retain the current server-backed message-query history introduced on rocketmq-studio instead of restoring the obsolete local history implementation
- add regression coverage for the shared helper and Proxy initialization

## Verification

- npx vitest run src/utils/browserStorage.test.ts src/api/client.test.ts src/pages/studio/__tests__/Proxy.test.tsx --reporter=dot (25 passed)
- npm run lint (0 errors; 5 pre-existing warnings outside this change)
- npx prettier --check on all changed files
- npm run build
- git diff --check

Closes #2164

PR #2637, opened 2026-08-27
fix(web): ignore stale consumer settings responses

### LINKED ISSUE #2634: Consumer settings modal can apply a stale response to another group
### Problem
The consumer group settings tab does not invalidate its asynchronous settings request when the detail modal is closed. If an operator opens group A, starts loading its settings, closes the modal, and then opens group B, a late response for group A can overwrite group B's form values. Saving at that point can send group A's retry values under group B's name.

### Expected behavior
Only the settings response for the currently open group should update the form and loading state. Closing the modal should invalidate any pending request.

### Proposed fix
Sequence settings requests and ignore stale success, error, and completion callbacks. Add a regression test covering out-of-order responses across two group modals.

yyqdbngt:
## Summary
- sequence consumer group settings requests so only the latest modal can update the form
- invalidate pending settings requests when the detail modal closes
- ignore stale success, error, and loading callbacks and reset loading state on close
- cover an out-of-order response across two consumer groups

## Why
A delayed response for a previously closed group could overwrite the settings form of the group currently open. Saving then risked applying the old group's retry values to the current group.

## Tests
- targeted ConsumerPage retry-settings and stale-response tests (2 tests passed)
- `npm run build`

Fixes #2634

### DIFF
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -99,6 +99,14 @@ const groupPage = (
   ...overrides,
 });
 
+const deferred = <T,>() => {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((complete) => {
+    resolve = complete;
+  });
+  return { promise, resolve };
+};
+
 const renderWithProviders = (ui: React.ReactElement, initialEntry = '/instance/consumer') =>
   render(
     <App>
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a race condition where a delayed response from a previously opened consumer group settings modal could overwrite the form of the currently open group. The fix uses a request sequence counter to ignore stale responses, which is a clean and effective approach.

The implementation is well-structured:
- Sequence counter correctly invalidates stale callbacks
- Loading state is properly reset on modal close
- Pending requests are cancelled when the modal closes
- Test coverage includes the specific out-of-order scenario

Code looks good. LGTM.

---

lizhimins:

PR #2533, opened 2026-08-22
feat(studio): add native cluster alert collection

Aias00:
Closes #2532

## Summary
- add native Apache RocketMQ collection for cluster health, consumer lag, and DLQ metrics without requiring Prometheus
- persist metric snapshots, alert state, transitions, context, acknowledgement and paged alert events
- evaluate scoped cluster and business rules through a shared alert domain with test-before-save endpoints
- coordinate collection across dashboard replicas with a database lease and expose collector status
- retain the existing Business Alerts, Cluster Alerts, and Alert Events menu structure while wiring the three surfaces to the shared domain
- include schema and upgrade scripts, plus the native alerting design document

## Validation
- mvn -q -Dtest=CollectorSchedulerTest,CollectorStatusControllerTest,MetricSampleTest,ApacheRocketMqBusinessMetricsCollectorTest,ApacheRocketMqClusterMetricsCollectorTest,ApacheRocketMqDlqMetricsCollectorTest,AlertFingerprintTest,AlertRuleEvaluatorTest,AlertServiceTest,AlertStateMachineTest,ClusterAlertRuleControllerTest,NativeAlertProcessorTest,NativeAlertRulePolicyTest,NativeAlertRuleTestServiceTest,SystemAlertControllerTest test
- npm test -- --run src/api/ops.test.ts src/services/opsService.test.ts src/pages/ops/__tests__/SystemAlertsPage.test.tsx src/pages/studio/__tests__/AlertManagement.test.tsx
- npm run build

## Notes
This is intentionally a foundation PR. Existing Prometheus integrations remain available; native collection provides the direct RocketMQ path for deployments that do not run Prometheus.

PR #2639, opened 2026-08-27
fix(studio): disable mail health indicator unless SMTP health check is enabled

lizhimins:
Follow-up to #2533 (native alerting introduced spring-boot-starter-mail).

The auto-configured MailHealthIndicator probes the SMTP connection on /actuator/health. In deployments without STUDIO_ALERTING_SMTP_HOST this probe fails, dragging the aggregate health to DOWN even though the app and its readiness group (including db) are UP.

Make the mail health indicator opt-in via STUDIO_ALERTING_SMTP_HEALTH_ENABLED (default false). Enable it in environments where SMTP is configured.

### DIFF
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -47,6 +47,11 @@ management:
           include: livenessState
         readiness:
           include: readinessState,db
+  health:
+    mail:
+      # SMTP is optional for native alerting; skip the SMTP connectivity probe
+      # unless explicitly enabled, so health stays UP when no SMTP host is configured.
+      enabled: ${STUDIO_ALERTING_SMTP_HEALTH_ENABLED:false}
 
 studio:
   cors:


RockteMQ-AI (approved):
### Summary

Clean, well-targeted fix to prevent the `MailHealthIndicator` from dragging aggregate health to `DOWN` when SMTP is not configured. Using Spring Boot's standard `management.health.mail.enabled` property with a safe default of `false` is the right approach — SMTP is optional for native alerting, and the health probe should not fail in environments where it's simply not set up.

The change is minimal (5 lines in `application.yml`), well-commented, and follows up correctly on #2533.

### Findings

- **[Info]** `application.yml:52` — Consider adding a brief note in the deployment docs or README about the `STUDIO_ALERTING_SMTP_HEALTH_ENABLED` env var, so operators know they can opt in when SMTP is configured.

No blocking issues. LGTM.

---
*Automated review by @RockteMQ-AI*

PR #2642, opened 2026-08-27
fix: keep native alert collection leases alive

### LINKED ISSUE #2641: Native alert collection can lose its lease during a long empty pass
`CollectorScheduler` acquires the native-alert database lease once at the beginning of a collection pass. It only calls `tryAcquire()` again from `persist()` when a collector returned samples.

A pass that takes longer than the lease duration while all collectors return empty lists therefore stops touching `rmq_alert_collection_lease`. After the row expires, another Studio replica can acquire the lease and start the same pass while the first replica is still contacting brokers. The first replica may also continue running until a later non-empty result notices that it no longer owns the lease.

This is separate from #1776/#1781, which addressed the legacy collection executor's `DiscardOldestPolicy`; this report is about lease ownership for the native alert collector introduced by #2533.

### Reproduction

1. Configure a short `studio.alerting.collection-lease-duration` (or use a collection pass that lasts longer than the default `PT1M`).
2. Let a native collector spend longer than that duration returning no samples.
3. While the first pass is still running, attempt to acquire the same lease from a second Studio replica.

### Expected behavior

The active pass should renew its lease while it is running, including passes that produce no samples. If renewal fails, it should stop scheduling or persisting further work instead of continuing as the active collector.

### Current behavior

There is no renewal while an empty pass is running, so the database lease can expire and ownership can overlap across replicas.

tju-yxq:
## Summary
- keep the latest mainline whole-pass timeout budget and sliding-window collection scheduling;
- renew the native alert collection lease throughout each active pass, including empty passes;
- cancel outstanding collectors and prevent late persistence when lease renewal is lost;
- expose lease duration and renewal interval through alerting configuration.

## Why
The whole-pass deadline added by #2807 prevents a collection round from running forever, but a valid long-running round can still outlive the lease. Without renewal, another Studio replica can acquire the same lease while the first replica is still contacting brokers. The two safeguards are complementary.

## Implementation
- rebased on the current `rocketmq-studio`, including #2807, #2957, and #2697;
- retained the mainline pass deadline, per-instance timeout, sliding window, and `processSuccessfulCollection(...)` flow;
- run one bounded lease heartbeat for the complete pass lifetime;
- cancel running futures and make collectors/persistence stop after renewal failure;
- re-check ownership before persistence as defense in depth;
- shut down both collection and renewal executors with the scheduler;
- cover renewal during an empty pass and cancellation after renewal loss.

Closes #2641

## Verification
- `mvn -q '-Dtest=CollectorSchedulerTest,MybatisPlusAlertCollectionLeaseTest' test` — 14 tests passed;
- `mvn -DskipTests=false test` — 1,964 tests passed, 0 failures, 0 errors, 0 skipped;
- Checkstyle — 0 violations;
- `git diff --check` passed.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
@@ -34,14 +34,18 @@ import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletionService;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Set;
@@ -59,6 +63,7 @@ public class CollectorScheduler {
     private final NativeAlertProcessor alertProcessor;
     private final AlertCollectionLease collectionLease;
     private final ExecutorService collectionExecutor;
+    private final ScheduledExecutorService leaseRenewalExecutor;
 
     @Autowired
     public CollectorScheduler(AlertingProperties properties, InstanceRepository instanceRepository,
@@ -66,13 +71,22 @@ public class CollectorScheduler {
             MetricSnapshotRepository snapshotRepository, NativeAlertProcessor alertProcessor,
             AlertCollectionLease collectionLease) {
         this(properties, instanceRepository, clusterCollectors, businessCollectors, snapshotRepository, alertProcessor,
-                collectionLease, newCollectionExecutor(properties));
+                collectionLease, newCollectionExecutor(properties), newLeaseRenewalExecutor());
     }
 
     CollectorScheduler(AlertingProperties properties, InstanceRepository instanceRepository,
             List<ClusterMetricsCollector> clusterCollectors, List<BusinessMetricsCollector> businessCollectors,
             MetricSnapshotRepository snapshotRepository, NativeAlertProcessor alertProcessor,
             AlertCollectionLease collectionLease, ExecutorService collectionExecutor) {
+        this(properties, instanceRepository, clusterCollectors, businessCollectors, snapshotRepository, alertProcessor,
+                collectionLease, collectionExecutor, newLeaseRenewalExecutor());
+    }
+
+    CollectorScheduler(AlertingProperties properties, InstanceRepository instanceRepository,
+            List<ClusterMetricsCollector> clusterCollectors, List<BusinessMetricsCollector> businessCollectors,
+            MetricSnapshotRepository snapshotRepository, NativeAlertProcessor alertProcessor,
+            AlertCollectionLease collectionLease, ExecutorService collectionExecutor,
+            ScheduledExecutorService leaseRenewalExecutor) {
         this.properties = properties;
         this.instanceRepository = instanceRepository;
         this.clusterCollectors = clusterCollectors;
@@ -81,6 +95,7 @@ public class CollectorScheduler {
         this.alertProcessor = alertProcessor;
         this.collectionLease = collectionLease;
         this.collectionExecutor = collectionExecutor;
+        this.leaseRenewalExecutor = leaseRenewalExecutor;
     }
 
     @Scheduled(fixedDelayString = "${studio.alerting.collection-interval:PT30S}")
@@ -89,8 +104,20 @@ public class CollectorScheduler {
             log.debug("Skipping native alert collection because another Studio replica holds the lease");
             return;
         }
+        List<Future<?>> leaseJobs = new CopyOnWriteArrayList<>();
+        LeaseHeartbeat heartbeat = new LeaseHeartbeat(collectionLease, leaseRenewalExecutor,
+                leaseRenewalInterval(), leaseJobs);
+        heartbeat.start();
+        try {
+            collectWithPassBudget(heartbeat, leaseJobs);
+        } finally {
+            heartbeat.close();
+        }
+    }
+
+    private void collectWithPassBudget(LeaseHeartbeat heartbeat, List<Future<?>> leaseJobs) {
         List<InstanceVO> instances = instanceRepository.findAll();
-        if (instances.isEmpty()) {
+        if (instances.isEmpty() || !heartbeat.isHeld()) {
             return;
         }
         Duration timeout = parsePositiveDuration(properties.getCollectionTimeout(), Duration.ofSeconds(15));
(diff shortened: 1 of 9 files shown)


tju-yxq:
Implemented the requested lease-heartbeat behavior for long native alert collection passes.

- empty passes renew the lease instead of allowing it to expire;
- renewal is conditional on the current holder and a non-expired lease;
- renewal loss cancels active collectors and blocks persistence of late results;
- lease duration and renewal interval are exposed in alerting configuration.

Verification on commit `4a5a01a2`: focused tests 9/9 passed; full server suite 1792/1792 passed with 0 failures, 0 errors, and 0 skipped; `git diff --check` passed.

server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
```diff
@@ -120,30 +170,39 @@ public void cleanUpSnapshots() {
...
+            if (!heartbeat.isHeld()) {
+                return;
+            }
             try {
-                persist(collector.supports(instance) ? collector.collect(instance) : List.of());
+                List<MetricSample> samples = collector.supports(instance) ? collector.collect(instance) : List.of();
+                persist(samples, heartbeat);
             } catch (RuntimeException error) {
                 log.warn("Native metric collector failed for instance {}: {}", instance.getName(), error.getMessage());
             }
```
RockteMQ-AI, line 300: **[Info]** The `tryAcquire()` re-check inside `persist()` is a good defensive measure, but now that the `LeaseHeartbeat` already handles periodic renewal and `isHeld()` is checked before calling `persist()`, this double-check is somewhat redundant in the happy path. Consider adding a brief comment explaining that this is intentional defense-in-depth (in case the heartbeat interval happens to align poorly with a slow persist call).

server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
```diff
@@ -168,13 +228,101 @@ private static ExecutorService newCollectionExecutor(AlertingProperties properti
...
+    }
+
+    private Duration leaseRenewalInterval() {
+        Duration leaseDuration = parsePositiveDuration(properties.getCollectionLeaseDuration(), Duration.ofMinutes(1),
+                "collection lease duration");
+        Duration safeMaximum = leaseDuration.dividedBy(3);
+        if (safeMaximum.isZero() || safeMaximum.isNegative()) {
+            safeMaximum = Duration.ofMillis(1);
+        }
+        Duration configured = parsePositiveDuration(properties.getCollectionLeaseRenewalInterval(), safeMaximum,
```
RockteMQ-AI: **[Info]** The `LeaseHeartbeat` inner class is well-designed. A short Javadoc on the class would help future readers understand the heartbeat lifecycle at a glance.

server/src/main/java/org/apache/rocketmq/studio/persistence/mapper/RmqAlertCollectionLeaseMapper.java
```diff
@@ -29,4 +29,9 @@ public interface RmqAlertCollectionLeaseMapper extends BaseMapper<RmqAlertCollec
             + "AND (expires_at <= #{now} OR holder_id = #{holderId})")
     int acquire(@Param("leaseName") String leaseName, @Param("holderId") String holderId,
             @Param("now") LocalDateTime now, @Param("expiresAt") LocalDateTime expiresAt);
+
+    @Update("UPDATE rmq_alert_collection_lease SET expires_at = #{expiresAt}, gmt_modified = #{now} "
+            + "WHERE lease_name = #{leaseName} AND holder_id = #{holderId} AND expires_at > #{now}")
```
RockteMQ-AI, line 34: **[Info]** The `renew` SQL correctly enforces `expires_at > now` to prevent an expired holder from reclaiming the lease. Since both sides use the Java-computed `now` parameter consistently, this is fine for now, but worth keeping in mind if the DB ever switches to using its own `NOW()` for expiry checks.

RockteMQ-AI (commented):
### Summary

This PR fixes a real race condition where a long empty alert collection pass could lose the database lease, allowing another Studio replica to acquire it while the original pass was still contacting brokers. The implementation is clean and well-structured.

### Findings
- **[Info]** CollectorScheduler.java:197 — `tryAcquire()` re-check in `persist()` is good defense-in-depth; a comment explaining why would help
- **[Info]** CollectorScheduler.java:247 — `LeaseHeartbeat` inner class is well-designed; a class-level Javadoc would aid readability
- **[Info]** RmqAlertCollectionLeaseMapper.java:34 — SQL correctly prevents expired holders from reclaiming; consistent use of Java-side timestamps is correct

### What Looks Good
- The `renew()` SQL enforces the key safety property: expired holders cannot reclaim the lease (`expires_at > now` check)
- `LeaseHeartbeat` uses `AtomicBoolean` and `compareAndSet` for thread-safe, exactly-once cancellation
- `CopyOnWriteArrayList` for the jobs list handles concurrent access from collection and renewal threads correctly
- The clamping logic (`leaseRenewalInterval` capped at 1/3 of lease duration) prevents misconfiguration
- Good test coverage with real executors for realistic timing behavior
- Clean resource cleanup via `@PreDestroy` and `AutoCloseable`

### Suggestions
- Consider adding a brief Javadoc to `LeaseHeartbeat` explaining its role in the collection lifecycle
- Add a comment on the `tryAcquire()` re-check in `persist()` clarifying it as intentional defense-in-depth

Overall this is a solid fix with correct concurrency handling and good test coverage.

tju-yxq:
Addressed the review suggestions in commit 79db2407: added class-level Javadoc for LeaseHeartbeat and clarified that the persist-time tryAcquire() check is intentional defense-in-depth against a heartbeat race. No runtime behavior changed. Re-ran the focused server tests from server/: CollectorSchedulerTest and MybatisPlusAlertCollectionLeaseTest pass (exit 0); git diff --check is clean.

tju-yxq:
CI follow-up: after the documentation-only follow-up, run 33119990044 also ended as startup_failure before any job was created. The same repository workflow is failing at startup on the current branch, #2644, the rocketmq-studio push, and unrelated pull requests, so there is no Java/test job output to diagnose for this commit.

RockteMQ-AI (approved):
### Summary

The follow-up commit addresses both suggestions from the previous review:
- Added a clear comment on the `tryAcquire()` re-check explaining it as intentional defense-in-depth
- Added a Javadoc to `LeaseHeartbeat` describing its role in the collection lifecycle

The change is minimal, focused, and improves code readability without altering behavior. LGTM.

---

lizhimins:
Thanks for this work! Since this PR was opened, the mainline has merged #2807, which rewrites `CollectorScheduler.collect()` with a whole-pass timeout budget — this branch now conflicts with it. Could you rebase onto the latest `rocketmq-studio` and reconcile the two designs? Your lease heartbeat/renewal and #2807's pass budget are complementary (the pass budget prevents one round from running forever; your heartbeat prevents long rounds from losing the lease), so the rebased version should keep both: whole-pass deadline plus periodic `renew()` during collection. Once rebased we'll take another look. Note the branch may also need to absorb #2957 (sliding-window collection) which landed later and touches the same scheduler.

tju-yxq:
Rebased and reconciled the scheduler changes requested in review in `3f42d66c` (production/test commit `2e461155`).

The branch now keeps both designs:
- the current mainline whole-pass deadline, per-instance timeout, sliding-window scheduling from #2807/#2957, and `processSuccessfulCollection(...)` flow;
- periodic lease renewal for the entire active pass, including empty passes, with cancellation and persistence blocking after lease loss.

I also retained the persist-time ownership check as defense in depth and the scheduler now closes the renewal executor during shutdown.

Verification after the rebase:
- focused scheduler/lease tests: 14 passed;
- full server suite: 1,964 passed, 0 failures/errors/skips;
- Checkstyle: 0 violations;
- `git diff --check`: clean.

The PR is mergeable/CLEAN. GitHub has not attached any status checks to the new head yet.

PR #2647, opened 2026-08-28
fix(web): refresh deliveries with current filters

### LINKED ISSUE #2646: Notification delivery retry can restore results from stale filters
### Problem

The notification delivery page refreshes after a retry by calling a callback captured when the retry started. If the user changes filters while the retry request is pending, the completed retry can issue a request with the old channel, status, instance, and page values. That stale response may then replace the list for the filters currently shown in the UI.

### Expected behavior

A completed retry should refresh the list using the filters and pagination that are active at completion time.

### Proposed fix

Route retry completion through the page's current query effect instead of invoking a captured request callback, and add a regression test that changes the status filter while a retry is pending.

yyqdbngt:
### Summary

- refresh notification deliveries through the current query effect after a retry completes
- prevent a retry started under old filters from restoring stale results
- cover a status-filter change while the retry request is pending

### Tests

- npm exec vitest -- run src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx
- npm exec prettier -- --check src/pages/ops/notificationDeliveries.tsx src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx
- npm exec eslint -- src/pages/ops/notificationDeliveries.tsx src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx
- npm run build

Closes #2646

### DIFF
--- a/web/src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx
+++ b/web/src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx
@@ -18,8 +18,17 @@ vi.mock('../../../services/instanceService', () => ({
 }));
 vi.mock('../../../services/opsService', () => ({
   listAlertDeliveriesPage: vi.fn(),
+  retryAlertDeliveries: vi.fn(),
   retryAlertDelivery: vi.fn(),
 }));
+
+const deferred = <T,>() => {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((complete) => {
+    resolve = complete;
+  });
+  return { promise, resolve };
+};
 
 beforeAll(() => {
   Object.defineProperty(window, 'matchMedia', {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a stale-closure bug in the notification deliveries page: when a user retries a delivery and then changes the status filter while the retry request is in-flight, the post-retry refresh would use the old filter values, showing stale results.

The fix replaces the direct `refresh()` callback with a `refreshNonce` state counter. The `useEffect` that fetches data depends on this nonce, so when the retry completes and bumps the nonce, the effect re-runs with the **current** filter values from state — not the captured ones from the closure.

### Findings
- **[Info]** `notificationDeliveries.tsx:42-48` — Using a nonce counter to trigger re-fetch is a clean React pattern that avoids stale closures. The `useEffect` dependency on `refreshNonce` ensures the latest filter state is always used.
- **[Info]** `NotificationDeliveriesPage.test.tsx` — The test uses a `deferred` promise to simulate the race condition precisely. It verifies that changing the filter while a retry is pending results in a refresh with the new filter. Well-structured test.
- **[Info]** The `retryAlertDeliveries` mock addition in the test file is correct — it ensures the mock module shape matches the actual service.

### Verdict
LGTM. Clean fix for a subtle async/state bug with good test coverage of the race condition.

---

lizhimins:

PR #2649, opened 2026-08-28
fix(alert): reject overflowing rule durations

### LINKED ISSUE #2648: Overflowing native alert durations can interrupt metric processing
### Problem

Native alert rule validation checks the duration and reminder interval syntax, but it does not parse the values before persistence. Very large numeric values therefore pass the API validation and can overflow while the native alert processor converts them to a Duration. The resulting arithmetic exception can interrupt processing instead of being reported as an invalid rule.

### Expected behavior

Native alert rules with duration values that cannot be represented should be rejected with the same bad-request response used for other invalid durations, before the rule is persisted.

### Proposed fix

Normalize arithmetic overflows in the shared duration parser to the existing invalid-duration business error, validate native duration fields before persistence, and add boundary regression tests.

btlqql:
### Summary

- convert duration arithmetic overflows into the existing invalid-duration business error
- parse native rule duration and reminder interval during validation, before persistence
- cover overflowing components, overflowing sums, and native rule validation boundaries

### Tests

- mvn -Dmaven.repo.local=D:/taiyi-maven-cache -Dtest=AlertRuleDurationTest,NativeAlertRulePolicyTest,NativeAlertProcessorTest test

Closes #2648

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleDuration.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleDuration.java
@@ -24,26 +24,30 @@ final class AlertRuleDuration {
         if (!StringUtils.hasText(value)) {
             return Duration.ZERO;
         }
-        Matcher matcher = PART.matcher(value.trim());
-        Duration result = Duration.ZERO;
-        int end = 0;
-        while (matcher.find()) {
-            if (matcher.start() != end) {
+        try {
+            Matcher matcher = PART.matcher(value.trim());
+            Duration result = Duration.ZERO;
+            int end = 0;
+            while (matcher.find()) {
+                if (matcher.start() != end) {
+                    throw invalid(value);
+                }
+                long amount;
+                try {
+                    amount = Long.parseLong(matcher.group(1));
+                } catch (NumberFormatException error) {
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a real bug: duration parsing in alert rules could silently overflow when users specify extremely large values (e.g. `9999999999999999999s`). The fix wraps `Duration.plus()` calls in an `ArithmeticException` catch and converts overflow into the existing `invalid()` business error — clean and consistent with the surrounding error-handling pattern.

The additional validation in `NativeAlertRulePolicy.validate()` ensures overflow is caught at persistence time, not just at parse time, which is the right layer for this check.

### Findings
- **[Info]** `AlertRuleDuration.java:34` — The outer `try/catch(ArithmeticException)` correctly catches overflow from `Duration.plus()`. `java.time.Duration` throws `ArithmeticException` on overflow, so this is the right exception to catch.
- **[Info]** `NativeAlertRulePolicy.java:31-34` — Calling `AlertRuleDuration.parse()` during validation eagerly surfaces overflow before persistence. Good defensive approach.
- **[Info]** Tests cover both component-level overflow and sum overflow, plus the native rule validation boundary. Well-structured.

### Verdict
LGTM. Clean, minimal fix with good test coverage.

---

PR #2660, opened 2026-08-28
fix: continue batch instance deletion after runtime failures

### LINKED ISSUE #2659: Batch instance deletion aborts when one resource check fails
## Problem

`POST /api/instances/delete-batch` is documented and implemented as a per-instance operation that returns partial results. The loop currently catches only `BusinessException`, but Apache provider preflight and cleanup can also throw a runtime failure when the broker or local dependency is unavailable.

If the first selected instance fails that way, the exception leaves `InstanceService.deleteInstances(...)` immediately:

- the later selected instances are never attempted;
- the endpoint returns a generic 500 instead of the batch result;
- the caller cannot tell which items succeeded or failed.

## Expected behavior

A failure while processing one selected identifier should be recorded for that identifier and must not prevent the remaining identifiers from being attempted. Existing `BusinessException` messages and the successful deletion count should remain unchanged. Unexpected failures should be reported with a bounded, readable message rather than an internal stack trace.

## Verification

Add a service regression test where the first Apache instance's `countTopics` throws `IllegalStateException` and the second instance deletes successfully. The test should prove both instances are attempted, the result contains one deletion and one failure, and no exception escapes the batch endpoint.

tju-yxq:
## What changed

Fixes #2659.

- Isolate unexpected provider/runtime failures to the current instance in `InstanceService.deleteInstances`.
- Continue processing the remaining selected instances after one resource preflight or cleanup failure.
- Keep existing `BusinessException` handling and successful deletion counts unchanged.
- Return a trimmed, bounded (500-character) failure message instead of exposing an internal stack trace through the batch response.

## Verification

- `mvn -q '-Dtest=InstanceServiceTest,InstanceControllerTest' test`
- `mvn -DskipTests=false test` — 1790 tests, 0 failures, 0 errors, 0 skipped
- `git diff --check`

The regression test makes the first Apache provider resource check throw `IllegalStateException`, verifies the second selected instance is still processed and deleted, and covers bounded unexpected failure messages.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -71,6 +71,7 @@ public class InstanceService {
 
     static final int COUNT_PARALLELISM = 8;
     static final long COUNT_TIMEOUT_SECONDS = 3;
+    private static final int MAX_BATCH_FAILURE_MESSAGE_LENGTH = 500;
 
     private final ExecutorService countExecutor = Executors.newFixedThreadPool(COUNT_PARALLELISM, runnable -> {
         Thread thread = new Thread(runnable, "instance-resource-counts");
@@ -509,9 +510,23 @@ public class InstanceService {
                 deleted++;
             } catch (BusinessException ex) {
                 failed.add(instanceId + ": " + ex.getMessage());
+            } catch (RuntimeException ex) {
+                String message = batchFailureMessage(ex);
+                log.warn("Failed to delete instance {} during batch operation: {}", instanceId, message);
+                failed.add(instanceId + ": " + message);
             }
         }
         return BatchDeleteResultVO.builder().deleted(deleted).failed(failed).build();
+    }
+
+    private static String batchFailureMessage(RuntimeException failure) {
+        String message = failure.getMessage();
+        if (!StringUtils.hasText(message)) {
+            message = failure.getClass().getSimpleName();
+        }
+        message = message.trim();
+        return message.length() > MAX_BATCH_FAILURE_MESSAGE_LENGTH
+                ? message.substring(0, MAX_BATCH_FAILURE_MESSAGE_LENGTH) : message;
     }
 
     private void removeDataSourceBindings(String instanceId) {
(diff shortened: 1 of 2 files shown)


tju-yxq:
Validation update for #2659:

- commit: `29a9ecc6177f03fc1cd05de33aaa3430e99b70f6`
- focused: `mvn -q '-Dtest=InstanceServiceTest,InstanceControllerTest' test` — passed
- backend: `mvn -DskipTests=false test` from `server/` — 1790 tests, 0 failures, 0 errors, 0 skipped
- `git diff --check` — passed

At the time of this comment GitHub has not published any check-run for this PR yet, so no CI result is being claimed here.

tju-yxq:
GitHub CI follow-up: workflow run [33144649473](https://github.com/apache/rocketmq-dashboard/actions/runs/33144649473) for commit `29a9ecc6177f03fc1cd05de33aaa3430e99b70f6` completed with `startup_failure` before any job was created (`jobs=[]`, no log available). This is a workflow/infrastructure startup failure, not a Backend, Frontend, or Docker test failure; the local verification listed above remains the available test result.

tju-yxq:
Additional CI detail from the run annotation: GitHub rejected `docker/setup-buildx-action@v3` and `docker/build-push-action@v6` under the repository's allowed-actions policy, so the workflow stopped during startup and never created Backend/Frontend/Docker jobs. No workflow file or production code change was made to mask this repository-level policy failure.

RockteMQ-AI (approved):
### Summary
Fixes batch instance deletion to isolate unexpected `RuntimeException` from provider calls, ensuring remaining instances continue processing after one fails. Error messages are bounded to 500 characters to prevent leaking internal stack traces.

The implementation is clean and well-structured:
- Proper exception hierarchy: `RuntimeException` catch after `BusinessException` preserves existing behavior
- `batchFailureMessage()` is null-safe (`StringUtils.hasText`), trims, and bounds output
- Good regression test coverage: continuation after failure + message bounding edge case
- Appropriate `warn`-level logging for operational visibility

CI `startup_failure` is a GitHub Actions workflow config issue (rejected `docker/setup-buildx-action@v3`), not related to these code changes.

LGTM — looks good to merge once CI is resolved.

---

PR #2668, opened 2026-08-28
fix(studio): avoid dropping saturated alert collection jobs

### LINKED ISSUE #2662: [Studio][Bug] Avoid dropping native alert collection jobs when the executor is saturated
## Problem

`CollectorScheduler` submits every configured Studio instance to a fixed-size executor at once. The executor uses the same value for worker count and queue capacity, with `AbortPolicy`; rejected submissions are logged and discarded.

With the default parallelism of 4, only four running plus four queued jobs are accepted. When more than eight instances exist, later instances in the stable repository order can be skipped on every collection pass, leaving them without metric snapshots or alert evaluation.

## Evidence

- `server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java:82`
- `server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java:102`
- `server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java:167`
- `server/src/main/java/org/apache/rocketmq/studio/instance/MybatisPlusInstanceRepository.java:46`

## Expected behavior

- Every configured instance gets a collection opportunity during a pass.
- Concurrency remains bounded by `collectionParallelism`.
- A global/pass deadline prevents an unbounded pass.
- Saturation must not silently starve the same tail instances.
- Regression coverage includes more than `2 * parallelism` instances.

## Related

#2641 / PR #2642 handles collection lease renewal, not executor saturation or dropped jobs.

Aias00:
## Summary

- keep native alert collection concurrency bounded by `studio.alerting.collection-parallelism`
- use a sliding in-flight window: submit at most `parallelism` jobs, then submit the next instance only after one job completes or times out
- restore the collector executor to a bounded queue so instance inventory size is not converted into unbounded executor backlog
- add regression coverage for both dropped tail instances and accidental eager queueing

Closes #2662

## Verification

- RED before first fix: `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=CollectorSchedulerTest test` failed with `expected: <5> but was: <4>`
- RED before sliding-window fix: `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=CollectorSchedulerTest#doesNotQueueMoreThanParallelismBeforeAnyCollectionCompletesTest test` failed with `expected: <0> but was: <3>`
- GREEN after fix: `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=CollectorSchedulerTest test` passed, 7 tests

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
@@ -26,10 +26,18 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Replaces the fire-and-forget collection loop with a bounded sliding-window scheduler backed by `CompletionService`. This prevents executor saturation from silently dropping instances, and introduces per-job and pass-level deadlines to avoid unbounded blocking.

The implementation is clean: proper `InterruptedException` handling, `RejectedExecutionException` fallback, and comprehensive regression tests covering both the drop scenario and the eager-queueing fix.

LGTM.

---

Aias00:
Superseded by #2957, which consolidates this change with the related Studio fixes on the current `rocketmq-studio` baseline. The replacement PR is Ready and currently CLEAN/MERGEABLE.

PR #2670, opened 2026-08-28
fix: protect notification outbox deliveries from lease loss

### LINKED ISSUE #2661: Notification outbox can redeliver a completed delivery
`NotificationOutboxService` can turn a notification that was already sent into another delivery attempt.

The dispatcher marks a row as `SENDING` and then performs the SMTP or webhook call synchronously. The claim is not renewed while that call is in flight, so a slow SMTP transaction (or a deployment with multiple Studio replicas) can make the row look stale after the one-minute claim window. A second dispatcher can then claim the same row and send the alert again.

There is a second state-machine gap in the same path: after the external call succeeds, the row is marked `DELIVERED` and the delivery audit record is written. If that audit insert fails, the surrounding `try` block treats the bookkeeping exception as a send failure and moves the row back to `RETRY_WAIT`. The next dispatcher retries a notification that has already reached its recipient. A transient failure while writing the `DELIVERED` state can create the same risk if it is routed through the normal send-retry path.

Expected behavior:

- renew a claimed row while its delivery attempt is still running, with the renewal conditional on the current claim token;
- stop changing the row when ownership is lost, rather than allowing a stale worker to overwrite a newer attempt;
- never convert a successfully completed external send into a retry because audit recording or the delivery-state write failed;
- record delivery and retry timestamps from the actual completion/failure time.

Please add regression coverage for a slow/in-flight claim being renewed and for failures after a successful external send. The tests should keep the existing at-least-once retry behavior for genuine webhook/SMTP failures.

tju-yxq:
Fixes #2661

A notification outbox row is currently claimed for one minute while the SMTP or webhook call runs synchronously. If that call is slow, another Studio replica can reclaim the row and send the same alert. In addition, bookkeeping failures after an external send can fall into the send retry path.

This change:
- renews the current claim during an in-flight delivery, using the claim token as the ownership guard;
- requires SENDING plus the matching token for every terminal or retry update, so a stale worker cannot overwrite a newer claim;
- commits the delivery state before recording the audit entry and isolates audit failures from delivery retries;
- does not send again when the external call succeeded but the delivery-state write itself failed; the row remains eligible for stale-claim recovery;
- keeps genuine SMTP/webhook failures on the existing bounded retry path;
- documents the claim timeout, renewal interval, and bounded heartbeat thread settings.

Regression coverage includes a blocked SMTP send with a successful renewal, lease loss during an in-flight send, a state-write failure after external success, and an audit failure after the row is marked delivered.

## Testing

- `mvn -q '-Dtest=NotificationOutboxServiceTest' test` (16 tests)
- `mvn -DskipTests=false test` (1,792 tests; 0 failures, 0 errors, 0 skipped; Checkstyle 0 violations)
- `git diff --check`

The GitHub workflow may still be affected by the repository's Docker Actions policy; no workflow changes are included here.

### DIFF
--- a/docs/studio-native-alerting-design.md
+++ b/docs/studio-native-alerting-design.md
@@ -246,11 +246,16 @@ rmq_metric_collector_lease
   collector_name, owner_id, expires_at
 ```
 
-Only the active lease holder collects and evaluates. Notification outbox rows use claimant-bound state transitions so a stale worker cannot overwrite a newer claimant. Delivery is at-least-once: receivers should use the event and channel identity to deduplicate a request that times out after reaching the remote service.
+Only the active lease holder collects and evaluates. Notification outbox rows use claimant-bound state transitions so a stale worker cannot overwrite a newer claimant. A delivery worker renews its claim while a webhook or SMTP call is in flight; the renewal is conditional on the claim token, and a worker that loses the lease stops updating that row. The delivery state is committed before its audit entry, so an audit-store failure cannot turn a completed external send into another retry. Delivery is still at-least-once: receivers should use the event and channel identity to deduplicate a request that times out after reaching the remote service.
 
 ## Notifications and Silences
 
 The supported notification channels are DingTalk, the SMS webhook configured in General Settings, and Email. A real event creates outbox rows for enabled supported channels. Independent notification-channel and notification-policy CRUD are future work.
+
+The dispatcher claims a row for one minute by default and renews the claim every 20 seconds while the external
+delivery is running. Deployments with slower receivers can tune `studio.alerting.notification-claim-timeout` and
+`studio.alerting.notification-claim-renewal-interval`; the renewal interval must remain shorter than the claim timeout.
+`studio.alerting.notification-heartbeat-threads` bounds the daemon workers used for these renewals.
 
 Email delivery uses Spring's standard SMTP configuration. Configure `STUDIO_ALERTING_SMTP_HOST`,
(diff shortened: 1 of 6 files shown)


tju-yxq:
Implemented and pushed as f6e87f5c. The dispatcher now renews claim ownership during slow SMTP/webhook calls, guards every state transition with status plus claim token, and keeps audit-write failures out of the delivery retry path. Added regression coverage for renewal, lease loss, post-send state-write failure, and post-delivery audit failure.

Verification from server/: `mvn -q '-Dtest=NotificationOutboxServiceTest' test` (16 passed), `mvn -DskipTests=false test` (1,792 passed; 0 failures, 0 errors, 0 skipped; Checkstyle 0 violations), and `git diff --check` passed.

The pull_request workflow run 33154046055 ended as `startup_failure` before creating any jobs (`jobs=[]`), so there are no Backend/Frontend/Docker job results to report. No workflow changes are included.

RockteMQ-AI (approved):
### Summary
Adds a claim-renewal heartbeat to notification outbox delivery, preventing duplicate alerts when SMTP/webhook calls are slow and another replica reclaims the row. The implementation uses a `ScheduledExecutorService` with configurable timeout/renewal intervals, proper `@PreDestroy` cleanup, and defensive null-safe property reading.

The design correctly ensures: (1) renewal interval < claim timeout, (2) audit commit happens after delivery state commit, and (3) heartbeat threads are bounded. Good test coverage for the renewal and lease-loss scenarios.

LGTM.

---

PR #2676, opened 2026-08-28
fix: make cloud instance import resilient to partial failures

### LINKED ISSUE #2675: Cloud instance import stops at the first unexpected failure
## Problem

`POST /api/instances/import-cloud` walks every region and instance for a selected cloud credential. The service only handles `BusinessException`. A runtime failure from a catalog response, a cloud instance detail lookup, or instance persistence escapes the loop.

When that happens, the import returns 500 and stops processing. Instances already imported remain in the database, but the caller receives no summary; later regions and instances are never attempted. The method contract says per-instance failures are collected, so this is surprising during a best-effort import.

## Expected behavior

Keep processing after a failure that belongs to one region or instance. Return a normal import summary with the successful count and bounded, readable failure details. Request validation errors and JVM `Error`s should keep their current behavior, and duplicate-name skips should remain unchanged.

## Verification

Add service tests for:

- a runtime failure while listing one region and a later region still being scanned;
- a runtime failure while importing one cloud instance and a later instance still being attempted;
- null or malformed catalog responses and bounded failure details;
- the returned counts and audit entry.

tju-yxq:
Fixes #2675.

## What changed

- Keep cloud import running after an unexpected failure while resolving the provider, discovering regions, listing one region, or creating one instance.
- Treat null and malformed catalog entries as reported failures instead of letting them turn into a 500 response.
- Deduplicate repeated region and instance rows before making detail/persistence calls.
- Keep duplicate-name skips and request validation behavior unchanged.
- Bound failure detail messages and the returned detail list while exposing the complete `failedCount` and a truncation flag.
- Update the instance page so a failed import is not shown as "all instances already exist" and so it uses the complete failure count.

## Verification

- `mvn -Dtest=InstanceServiceTest test` — 76 tests passed, Checkstyle 0 violations
- `mvn -DskipTests=false test` — 1,795 tests, 0 failures, 0 errors, 0 skipped
- `vitest run src/api/instance.test.ts src/pages/instance/__tests__/InstancePage.test.tsx` — 29 tests passed
- TypeScript compilation — passed
- ESLint — 0 errors (the repository's existing 9 warnings remain)
- Prettier check — passed
- Vite production build — passed
- `git diff --check` — passed

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
@@ -45,9 +45,12 @@ import org.springframework.transaction.annotation.Transactional;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -72,6 +75,8 @@ public class InstanceService {
     static final int COUNT_PARALLELISM = 8;
     static final long COUNT_TIMEOUT_SECONDS = 3;
     private static final int MAX_BATCH_FAILURE_MESSAGE_LENGTH = 500;
+    static final int MAX_CLOUD_IMPORT_FAILURE_DETAILS = 100;
+    static final int MAX_CLOUD_IMPORT_FAILURE_MESSAGE_LENGTH = 500;
 
     private final ExecutorService countExecutor = Executors.newFixedThreadPool(COUNT_PARALLELISM, runnable -> {
         Thread thread = new Thread(runnable, "instance-resource-counts");
@@ -198,8 +203,9 @@ public class InstanceService {
     /**
      * Imports every cloud instance visible to the credential by walking all catalog regions.
      * Remarks are resolved from the cloud instance detail during creation. Instances whose
-     * resolved name already exists are skipped; per-instance failures are collected instead
-     * of aborting the batch.
+     * resolved name already exists are skipped; region and per-instance failures are collected
+     * instead of aborting the batch. Failure details are bounded while the result retains the
+     * complete failure count.
      */
     public CloudImportResultVO importCloudInstances(InstanceVendor vendor, Long credentialId) {
         if (vendor == null || vendor == InstanceVendor.APACHE) {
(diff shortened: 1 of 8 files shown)


tju-yxq:
The implementation for #2675 is pushed as `a5f92f7f52c770beaf6c3d706c9bdbd125636e77` followed by test commit `520f33c6a25acd27360c91b66d3c0826315abe5e`.

Review points:

- Provider lookup, region discovery, per-region listing, and per-instance creation now have explicit runtime-failure boundaries; `BusinessException` duplicate-name skips and validation behavior remain unchanged.
- Null/malformed rows and duplicate region+instance rows are handled without duplicate detail/persistence calls.
- `CloudImportResultVO` keeps the failure detail list bounded while exposing the complete `failedCount` and `failureDetailsTruncated` state; the frontend uses those fields and no longer reports a failed zero-import as "all instances already exist".

Local validation passed: focused `InstanceServiceTest` (76 tests), full Maven suite (1,795 tests, 0 failures/errors/skips, Checkstyle 0), focused Vitest tests (29), TypeScript, ESLint (0 errors), Prettier, Vite build, and `git diff --check`. No GitHub code checks are available because the workflow terminates during startup under the repository Actions policy.

tju-yxq:
GitHub Actions run `33159174578` finished as `startup_failure` before creating any job. The run annotation reports that the repository Actions policy does not allow `docker/setup-buildx-action@v3` and `docker/build-push-action@v6`; there are therefore no Backend, Frontend, or Docker job results to interpret as code failures. Local validation remains green as listed above.

tju-yxq:
Added follow-up commit `520f33c6` to verify the import summary audit record and the missing-catalog-provider boundary. The focused `InstanceServiceTest` suite now passes 76 tests; the branch remains clean and based on the latest `upstream/rocketmq-studio`.

tju-yxq:
The follow-up push triggered Actions run `33159689726`, which also ended as `startup_failure` before any job was created. As with the previous run, this is the repository-level allowed-actions policy rejecting the Docker Buildx/Build Push actions; it is not a Backend, Frontend, or Docker test result. The branch is still `CLEAN` on GitHub and the local suites remain green.

server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
```diff
@@ -212,58 +218,184 @@ public CloudImportResultVO importCloudInstances(InstanceVendor vendor, Long cred
...
+            importCloudInstance(vendor, credentialId, regionId, cloudInstanceId, result);
         }
+    }
+
+    private void importCloudInstance(InstanceVendor vendor, Long credentialId,
+                                     String regionId, String cloudInstanceId, CloudImportAccumulator result) {
+        InstanceVO request = InstanceVO.builder()
+                .vendor(vendor)
+                .credentialId(credentialId)
+                .regionId(regionId)
```
RockteMQ-AI, line 300: **[Info]** The duplicate-name detection relies on `e.getMessage().startsWith("Instance name already exists")`. This works but is fragile — if the exception message changes, duplicates would silently become creates. Consider using a dedicated exception type (e.g. `DuplicateInstanceException`) or a structured return value to distinguish skip-vs-fail vs. matching on message text.

RockteMQ-AI (approved):
﻿### Summary
This PR refactors the cloud instance import flow to be resilient to partial failures. Instead of aborting on the first error, each phase (provider resolution, region discovery, instance listing, instance creation) is wrapped in try/catch, failures are collected, and the import continues. Failure details are bounded (100 entries, 500 chars each) while the total `failedCount` is preserved. The frontend is updated to display the complete failure count and handle the truncated detail list gracefully.

### Findings
- **[Info]** `InstanceService.java:299` — Duplicate-name detection uses `e.getMessage().startsWith(...)` which is fragile; consider a dedicated exception type for robustness
- **[Info]** `index.tsx:573` — The fallback chain for `failedCount` is a good defensive pattern for backward compatibility

### Strengths
- Excellent decomposition: each cloud API phase is isolated with its own error handling, making failures granular and debuggable
- `CloudImportAccumulator` cleanly separates collection from representation; the bounded detail list with truncation flag is a well-designed API contract
- `CloudInstanceKey` record + `LinkedHashSet` for deduplication is correct and efficient
- `normalizeWhitespace()` defensively handles malformed catalog data
- Comprehensive test coverage: 76 `InstanceServiceTest` cases including region failures, null/malformed responses, deduplication, and provider errors
- Frontend changes correctly handle the new `failedCount`/`failureDetailsTruncated` fields with backward-compatible fallbacks
- All tests pass (1,795 backend + 29 frontend), TypeScript compiles, ESLint clean

### Suggestions
- Consider extracting the "Instance name already exists" string match into a constant or using a typed exception to avoid silent behavior changes if the message is reworded
- The `MAX_CLOUD_IMPORT_FAILURE_DETAILS` and `MAX_CLOUD_IMPORT_FAILURE_MESSAGE_LENGTH` constants are well-placed; consider exposing them via configuration if different deployments need different bounds

---

- **[INFO]** `web/src/pages/instance/index.tsx:573` — **[Info]** The condition `result.failedCount === 0 && result.created === 0 && result.skipped > 0` correctly identifies the 'all instances already exist' case. The fallback chain (`result.failedCount ?? result.failed?.length ?? 0`) is a good defensive pattern for backward compatibility with older server responses that don't include `failedCount`. _(line outside diff)_

tju-yxq:
Follow-up for the review suggestion in commit `398086117719f84db8b1f1e6363cb916e6f069fb`:

- Added `DuplicateInstanceNameException`, a typed `BusinessException` for name conflicts.
- `requireUniqueInstanceName()` now raises that exception directly.
- Cloud import now catches the typed exception explicitly and counts it as `skipped`; other `BusinessException` instances are always reported as failures instead of being classified from their message text.
- This removes the fragile `getMessage().startsWith("Instance name already exists")` contract without changing the user-facing error text or HTTP code.

Verification:

- `mvn -q '-Dtest=InstanceServiceTest,InstanceControllerTest' test` passed.
- `mvn -DskipTests=false test` passed: 1,795 tests, 0 failures, 0 errors, 0 skipped; Checkstyle 0 violations.
- `git diff --check` passed.

The failure-detail limits remain fixed safety bounds rather than deployment configuration; no configuration surface was added for the non-blocking suggestion.

RockteMQ-AI (approved):
### Summary
Re-review after new commits — the changes make cloud instance import resilient to partial failures by isolating per-region and per-instance errors, bounding failure details, and deduplicating catalog entries. The refactoring is clean and well-tested.

### Findings
- **[Info]** `InstanceService.java` — The `CloudImportAccumulator` inner class is a good encapsulation of the result collection logic. Consider making it a top-level package-private class if it grows further.
- **[Info]** `InstanceService.java:normalizeCloudImportValue` — Trimming provider/region/instanceId is a good defensive measure. Consider also normalizing to lowercase for provider/region keys if the upstream API is case-insensitive, to prevent duplicate entries like `aliyun` vs `Aliyun`.
- **[Info]** `index.tsx` — The `result.failedCount ?? result.failed.length` fallback is backward-compatible. Good.

### Suggestions
- The `MAX_CLOUD_IMPORT_FAILURE_DETAILS = 100` and `MAX_CLOUD_IMPORT_FAILURE_MESSAGE_LENGTH = 500` bounds are reasonable. Consider making these configurable via application properties if operators need to tune them.
- The `DuplicateInstanceNameException` extraction is clean. The 400 status code is appropriate for client-side validation errors.

Overall: LGTM — solid error isolation, good test coverage, clean refactoring.

---

PR #2693, opened 2026-08-28
[ISSUE #2463][ISSUE #2465][ISSUE #2470] fix(server): normalize service boundary inputs

### LINKED ISSUE #2463: Make repeated disabled-user updates idempotent
## Problem

Disabling a Studio administrator that is already disabled still runs the last-enabled-administrator guard. When no other administrator is enabled, repeating the same desired state returns a conflict even though the request would not change account state.

## Expected behavior

Return the existing user immediately when the requested enabled state already matches the stored state. Do not count administrators, write the user, or revoke sessions for a no-op update.

## Verification

Add a focused database-service test that verifies the no-op response and absence of mapper writes.

### LINKED ISSUE #2465: Deduplicate NameServer DNS addresses without hostname case sensitivity
## Problem

NameServer configuration comparison deduplicates addresses with case-sensitive string equality. A declared address such as NS.EXAMPLE.COM:9876 and an endpoint address such as ns.example.com:9876 are the same DNS node but are probed twice and inflate node counts.

## Expected behavior

Deduplicate simple DNS host-and-port addresses using a case-insensitive hostname key while preserving the first display address and avoiding unsafe rewriting of IPv6 zone identifiers.

## Verification

Add a focused comparison test that verifies one node and one admin call for addresses differing only by DNS hostname case.

### LINKED ISSUE #2470: Normalize DLQ action identifiers at the service boundary
## Problem

DLQ resend and export validate group names but pass the original padded values onward. Export therefore queries a different %DLQ% topic for values such as ` group-a `, while a padded explicit resend target can be sent as an invalid topic name.

## Expected behavior

Normalize required group names before both operations and normalize optional resend targets so surrounding whitespace is removed and blank values become absent.

## Verification

Add focused service delegation tests for padded groups, padded target topics, and blank optional targets.

yuluo-yx:
## Summary

- make repeated user enabled-state updates idempotent without database writes or session revocation
- deduplicate DNS NameServer addresses without hostname case sensitivity while preserving IPv6 zone semantics
- normalize DLQ group, target-topic, and selected message identifiers consistently before provider delegation
- keep each fix together with focused regression coverage in one cohesive backend-normalization change

## Root cause

The affected service methods validated logical values but continued to compare, log, or delegate the original unnormalized input. The auth path also performed state-change side effects when the stored enabled value already matched the requested value.

## Local verification

- Maven validate reached Checkstyle with 0 violations
- git diff --check
- scope check: 177 changed lines across 6 files

Java 21 unit tests were not run locally because this machine did not have the project-required JDK and both official SDK downloads were interrupted by the local proxy. The pull request is intentionally left for the automatically triggered Java 21 GitHub CI, per the submitter instruction.

Fixes #2463

Fixes #2465

Fixes #2470

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
@@ -213,6 +213,9 @@ public class AuthService {
     public RmqStudioUser setUserEnabled(Long userId, boolean enabled) {
         requireDatabaseBacked();
         RmqStudioUser user = getUser(userId);
+        if (Boolean.valueOf(enabled).equals(user.getEnabled())) {
+            return user;
+        }
         if (!enabled && Boolean.TRUE.equals(user.getAdmin())) {
             // Lock the enabled administrator rows so concurrent disables serialize. A plain
             // count would let two requests both observe a count of 2 and disable everyone.
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NameServerConfigDiffService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NameServerConfigDiffService.java
@@ -34,6 +34,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Stream;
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (commented):
### Summary
Normalizes inputs across AuthService, NameServerConfigDiffService, and DLQService: short-circuits redundant state changes, adds case-insensitive NameServer address deduplication, and extracts validation helpers with input trimming.

### Findings
- **[Info]** `AuthService.java:216-218` — The short-circuit when `enabled` already matches is a good idempotency improvement. It avoids unnecessary DB writes and session invalidation. The `Boolean.valueOf(enabled).equals(user.getEnabled())` comparison is correct but slightly verbose; `Boolean.valueOf(enabled) == user.getEnabled()` would work since `Boolean.valueOf()` caches `TRUE`/`FALSE`.
- **[Info]** `NameServerConfigDiffService.java:196-210` — The `canonicalAddressKey()` correctly lowercases the host part for case-insensitive deduplication while preserving port and IPv6 zone identifiers. The check `address.indexOf(":") != separator` correctly handles IPv6 addresses (multiple colons) by skipping normalization. Good attention to edge cases.
- **[Info]** `DLQService.java:45-70` — The extracted `requireGroupName()`, `normalizeOptional()`, `normalizeMsgIds()`, and `validateTimeRange()` helpers improve readability and testability. The `normalizeMsgIds()` method correctly filters out blank IDs after splitting and trimming.
- **[Info]** `NameServerConfigDiffService.java:200` — The `Locale.ROOT` usage for case normalization is correct and avoids locale-specific issues (e.g., Turkish I problem).

### Suggestions
The `canonicalAddressKey()` approach is a good balance between correctness and simplicity. The IPv6 zone ID check (`host.indexOf("%") >= 0`) correctly avoids normalizing zone identifiers that might be case-sensitive on some platforms.

---

lizhimins:
Closing as part of batch consolidation: too many single-change PRs open at once. Please regroup these changes into a few cohesive PRs per functional domain (see the maintainer note on #2724). Review conclusions are archived on the maintainer side, and reviewed commits may be reused as-is when resubmitting.

PR #2695, opened 2026-08-28
[ISSUE #2691] fix(message): reset query and trace lifecycle

### LINKED ISSUE #2691: [Studio][Bug] Reset Message Explorer view state and load traces on demand
## Problem

The Message Explorer keeps view state that no longer belongs to the current user action:

1. Reset clears rows but leaves the previous total, page, and scan-truncation warning visible.
2. Changing the query mode can leave rows from the previous mode on screen, and an in-flight response from that mode can restore them later.
3. Opening the normal message-content detail immediately requests trace data even when the Trace tab is never used.

## Reproduction

- Run a paginated query that reports a total above one page and resultMayBeTruncated=true, then press Reset.
- Start a controlled pending query, change from Topic to Message Key mode, and resolve the old request.
- Open a message through the Detail action and observe a trace request before selecting the Trace tab.

## Expected behavior

Reset and query-mode changes should invalidate the prior request and clear all result metadata. Message traces should load only when the Trace tab is first opened, and the same message trace should not be fetched repeatedly during the page lifecycle.

## Scope

Keep the change within the Message Explorer page and focused frontend regression tests. Preserve the existing request-generation ownership behavior for late query and trace responses.

## Evidence

yuluo-yx:
## Summary

- clear rows, totals, pagination, truncation warnings, errors, and loading state on Reset and query-mode changes
- invalidate in-flight queries when their view lifecycle ends
- load traces only when the Trace tab is opened and reuse the same message trace during the page lifecycle
- preserve existing late-response ownership guards and allow failed trace requests to retry

## Local verification

- MessagePage and MessagePageAsyncState: 23 tests passed
- targeted ESLint passed
- targeted Prettier check passed
- TypeScript and Vite production build passed
- git diff --check
- scope check: 173 changed lines across 2 files

The Vitest run emitted only the existing jsdom pseudo-element getComputedStyle warning.

Fixes #2691

### DIFF
--- a/web/src/pages/instance/__tests__/MessagePageAsyncState.test.tsx
+++ b/web/src/pages/instance/__tests__/MessagePageAsyncState.test.tsx
@@ -37,13 +37,17 @@ const instanceFilterMocks = vi.hoisted(() => ({
 vi.mock('../../../services/messageService', () => ({
   ...serviceMocks,
   queryMessagePage: ({ page = 1, pageSize = 50, ...params }: Record<string, unknown>) =>
-    Promise.resolve(serviceMocks.queryMessages(params)).then((items) => ({
-      items,
-      total: items.length,
-      page,
-      size: pageSize,
-      resultMayBeTruncated: false,
-    })),
+    Promise.resolve(serviceMocks.queryMessages(params)).then((result) =>
+      Array.isArray(result)
+        ? {
+            items: result,
+            total: result.length,
+            page,
+            size: pageSize,
+            resultMayBeTruncated: false,
+          }
+        : result,
+    ),
 }));
 vi.mock('../../../hooks/useInstanceFilter', () => instanceFilterMocks);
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
Extracts a `clearQueryResults()` helper to reset all query state on Reset/mode-change, adds a trace cache to avoid re-fetching traces for the same message, and separates modal opening from async trace loading.

### Findings
- **[Info]** `MessagePage.tsx:90-105` — The `clearQueryResults()` helper correctly resets pagination, truncation warning, error, and loading state. Consider also resetting the `selectedMessage` state to avoid stale detail data showing in the modal after a reset.
- **[Info]** `MessagePage.tsx:130-145` — The trace cache (`traceCacheRef`) is a nice UX improvement. One consideration: the cache key is `:`, which means the same message ID under different topics will be cached separately. This is correct behavior, but if a message is republished with the same ID, the cached trace will be stale. Consider adding a TTL or invalidating on topic deletion.
- **[Info]** `MessagePage.tsx:160-175` — The `handleModalTabChange()` correctly triggers trace loading only when switching to the trace tab and only if not already loaded. The `abortController` cancellation on modal close prevents memory leaks from in-flight requests.

### Suggestions
The separation of `openDetail()` (sync) from `loadMessageTrace()` (async) is a clean improvement. The test coverage for the new behavior (reset clearing pagination, mode-change invalidation, trace caching) is thorough.

---

lizhimins:
Closing: this branch predates #2839 and conflicts with the message query snapshot changes now on rocketmq-studio. The issue it addresses still exists — please rebase onto the latest rocketmq-studio and resubmit (reviewed commits may be reused).

Issue #2473, opened 2026-08-20
Prevent concurrent deployments from corrupting shared remote state


yuluo-yx:
## Problem

The deployment script always writes /tmp/rocketmq-studio-src.tar.gz and does not serialize deployments to the same remote target. Concurrent runs can overwrite or delete one another's archive and replace the same remote source tree while another build is using it.

## Expected behavior

Create a unique local archive for every run and acquire an atomic remote lock for the target directory. Release the lock and archive on success or ordinary failure, while allowing deployments to different target paths.

## Verification

Add an offline shell harness covering same-target rejection, different-target progress, failure cleanup, and unique temporary archives.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue is valid. The deployment script uses a fixed path `/tmp/rocketmq-studio-src.tar.gz` and does not serialize deployments to the same remote target. Concurrent runs can overwrite each other's archive and corrupt the remote source tree.

**Root Cause:** Non-unique archive path and lack of atomic remote locking for the target directory.
**Impact:** Concurrent deployments can corrupt each other, leading to failed or inconsistent deployments.
**Severity:** Medium — affects CI/CD reliability when multiple deployments run in parallel.

An automated fix proposal can be generated. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by github-manager*

yuluo-yx:
Reopening because the deployment serialization problem remains reproducible on the latest rocketmq-studio branch. The fix will be resubmitted as part of a cohesive deployment-contract change, following the maintainer guidance in #2107.

PR #2696, opened 2026-08-28
[ISSUE #2692] fix(deploy): align remote environment contracts

### LINKED ISSUE #2692: [Studio][Bug] Align remote deployment environment and cache contracts
## Problem

The remote deployment path documents environment variables that are not applied consistently:

1. PUBLIC_PORT is used by deploy.sh for verification, but Docker Compose publishes a hard-coded port 6789.
2. deploy.sh loads deploy/.env locally and packages it under deploy/.env remotely, while the prerequisite check and Compose startup require a different root .env file.
3. MAVEN_CACHE_DIR is documented, but remote settings lookup and the Maven volume remain hard-coded to the remote user home cache.
4. Concurrent deployments reuse the same local and remote archive path and have no target-scoped lock, so uploads and source replacement can overwrite one another. This was originally reported in #2473, which is closed but remains reproducible.

These mismatches make a first deployment following the documentation fail, make custom frontend ports fail verification, silently ignore custom Maven cache locations, and allow concurrent runs to corrupt one another.

## Reproduction

- Set PUBLIC_PORT=18080 and render deploy/docker-compose.yml; the published port remains 6789.
- Follow deploy/README.md with only deploy/.env present; the script rejects the missing remote root .env before upload.
- Set MAVEN_CACHE_DIR to another absolute path; inspect remote Docker arguments and observe the home .m2 mount.
- Start two deployments for the same remote path and observe that both use the same archive and unprotected source replacement section.

## Expected behavior

The local deploy/.env should be the single deployment configuration source and be used from its packaged remote location. PUBLIC_PORT should control both Compose publication and verification. MAVEN_CACHE_DIR should consistently control settings lookup, repository storage, and volume mounting. Deployments to the same remote path should be serialized with unique temporary archives and reliable cleanup, while different targets remain independent.

## Scope

Keep changes within deploy configuration, documentation, the remote deployment script, and an offline command-contract test. Do not start containers or contact a real remote host during tests.

## Evidence

Related: #2473.

yuluo-yx:
## Summary

- make PUBLIC_PORT control the rendered Compose port and deployment verification
- use packaged deploy/.env as the single local and remote deployment configuration source
- apply MAVEN_CACHE_DIR consistently to settings lookup, repository storage, and the Maven volume
- serialize deployments per remote target with unique archives and exit cleanup
- add an offline SSH, SCP, archive, port, cache, lock, and cleanup contract test
- align deployment documentation and examples with the implemented behavior

## Local verification

- shellcheck deploy/deploy.sh deploy/deploy_test.sh
- bash -n deploy/deploy.sh deploy/deploy_test.sh
- deploy/deploy_test.sh: passed
- PUBLIC_PORT=18080 docker compose config: passed and rendered published port 18080
- git diff --check
- scope check: 339 changed lines across 5 files

No container was started and no real SSH host was contacted. A hard process kill or abrupt remote disconnect can leave the target lock directory behind; operators should confirm no deployment is running before removing it.

Fixes #2692

Addresses #2473

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -4,10 +4,10 @@ REMOTE_HOST=your-server-ip
 REMOTE_USER=root
 REMOTE_PATH=/opt/rocketmq-studio
 
-# Optional local server build settings. deploy.sh mounts this directory into
-# the Maven build container, so downloaded dependencies persist across builds.
+# Optional remote server build settings. The path is resolved on the remote
+# host and mounted into the Maven build container for dependency reuse.
 # MAVEN_CACHE_DIR=/home/your-user/.m2
-# MAVEN_IMAGE=maven:3.9.9-eclipse-temurin-21
+# MAVEN_IMAGE=maven:3.9.16-eclipse-temurin-21
 
 # Frontend public port
 PUBLIC_PORT=6789
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -9,7 +9,7 @@
 ```
 
 脚本自动完成：本地构建 → 导出镜像 → SCP 传输 → 远程加载 → 重启容器 → 验证。
-后端通过 Maven 容器构建，并默认挂载宿主机的 `~/.m2`，后续构建会复用已下载的依赖。
+后端通过目标机上的 Maven 容器构建，并默认挂载目标登录用户的 `~/.m2`，后续构建会复用已下载的依赖。
 
 ## 配置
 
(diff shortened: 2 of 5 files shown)


RockteMQ-AI (commented):
### Summary
Aligns the remote deployment script to use `deploy/.env` as the single source of truth, adds `PUBLIC_PORT` propagation to Compose, introduces a remote deploy lock, and hardens path quoting for SSH safety.

### Findings
- **[Info]** `deploy.sh:24` — The script now requires `deploy/.env` to exist (errors if missing). This is a breaking change from the previous behavior where the file was optional. The README should call this out prominently for users upgrading from older versions.
- **[Info]** `deploy.sh:50-55` — The `remote_quote()` function properly handles single quotes in paths by escaping them for SSH. Good defensive coding. Consider also rejecting paths containing `\\` (backslash) as they can cause subtle issues in shell contexts.
- **[Info]** `deploy.sh:95-100` — The remote lock uses `mkdir` which is atomic on most filesystems but does not have a TTL. If the process crashes without running `cleanup()`, the lock persists indefinitely. Consider adding a lock metadata file with the PID and timestamp, and checking for stale locks on startup.
- **[Info]** `deploy_test.sh` — The test script is thorough. The `lock_contention_test` correctly verifies that concurrent deployments are rejected. The `cleanup_after_failure_test` validates that the lock is released even when the deployment fails.

### Suggestions
The `PUBLIC_PORT` propagation to Compose via `docker compose run --rm -e PUBLIC_PORT` is a clean approach. The `remote_quote()` function is well-designed for SSH path safety.

---

lizhimins:
Closing: conflicts with the deploy/.env preservation policy merged in #2840 — deploy.sh now excludes deploy/.env from the source package and restores the remote env file, whereas this PR proposes the opposite (shipping .env with the package and overwriting remote environment configuration). Not applicable.

PR #2704, opened 2026-08-28
[ISSUE #2699] fix(message): ignore time bounds for id lookup

### LINKED ISSUE #2699: [Studio][Bug] Ignore unused time bounds during message ID lookup
## Problem

Message lookup by message ID does not use `startTime` or `endTime`, but the Apache provider validates those timestamps before entering the message-ID branch. As a result, an otherwise valid message-ID lookup can fail because a client supplied equal or reversed unused time bounds.

## Reproduction

Query a known Topic and message ID while passing `startTime = 200` and `endTime = 100` (or equal values). The provider returns HTTP 400 before calling `viewMessage`, even though the lookup path is independent of time.

## Expected behavior

Message-ID lookup should ignore unrelated time bounds. Key and Topic searches must retain their existing time-window validation.

## Scope

Move time calculation and validation to non-message-ID query paths and add regression coverage for equal and reversed unused bounds.

yuluo-yx:
## Problem

The Apache message provider rejected equal or reversed timestamps before message-ID lookup even though that lookup does not consume either time bound.

## Changes

- return from the message-ID path before computing query time bounds
- preserve existing time validation for Key and Topic searches
- cover equal and reversed unused bounds with a provider regression test

## Local verification

- `git diff --check`
- PR scope check: 40 changed lines, 2 files, 1 commit
- Java tests were not run locally because this workspace does not have the project's Java 21 toolchain and no further SDK installation was attempted; the automatically triggered Java 21 PR checks are expected to run the regression test

No containers, full E2E suite, or remote workflow was started manually.

Fixes #2699

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
@@ -115,28 +115,27 @@ public class RocketMQMessageProvider implements MessageProvider {
                                                  String topic, String msgId, String tag, String key,
                                                  Long startTime, Long endTime) {
 
+        if (StringUtils.hasText(msgId)) {
+            return queryByMsgId(adminExt, topic, msgId);
+        }
+
         long end = endTime != null ? endTime : System.currentTimeMillis();
         long begin = startTime != null ? startTime : end - ONE_HOUR_MILLIS;
         if (begin >= end) {
             throw new BusinessException(400, "Message query start time must be before end time");
         }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Small, focused fix with good test coverage.

---

lizhimins:
Closing as part of batch consolidation: too many single-change PRs open at once. Please regroup these changes into a few cohesive PRs per functional domain (see the maintainer note on #2724). Review conclusions are archived on the maintainer side, and reviewed commits may be reused as-is when resubmitting.

PR #2550, opened 2026-08-24
fix(ci): use policy-approved Docker action revisions

Aias00:
Closes #2549

## Summary

- replace the two floating Docker Actions tags with immutable revisions listed in Apache's approved Actions patterns
- preserve the frontend Docker build configuration and GitHub Actions cache behavior

## Validation

- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/ci.yml')"`
- `git diff --check`

This allows CI to initialize the frontend Docker build job under the repository Actions policy.

PR #2754, opened 2026-08-29
[ISSUE #2634] fix(consumer): ignore stale settings responses in the group modal

### LINKED ISSUE #2634: Consumer settings modal can apply a stale response to another group

123123213weqw:
## Summary

- sequence consumer settings requests with a request-id ref
- ignore stale success, error, and completion callbacks after the modal closes or another group loads
- add a regression test covering out-of-order settings responses across two group modals

## Testing

- `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` (22 tests)
- full Vitest suite: 742 tests passed
- `npm run build` passed (tsc + Vite)
- `git diff --check`

Fixes #2634

## Note

- `.github/workflows/ci.yml` carries the same policy-approved Docker action revisions as #2550 (repo-wide startup fix) so this PR can run CI; the diff disappears once #2550 lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,10 +85,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
 
       - name: Build frontend Docker image (cached)
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           context: ./web
           push: false
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -108,6 +108,14 @@ const renderWithProviders = (ui: React.ReactElement, initialEntry = '/instance/c
     </App>,
   );
 
+const deferred = <T,>() => {
+  let resolve!: (value: T) => void;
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a race condition where stale settings responses could overwrite the current modal form data when switching between consumer groups. The fix uses a monotonically increasing request counter (`settingsRequestIdRef`) to detect and ignore stale responses — a clean and well-established pattern for this class of bug.

### Findings
- **[Info]** `consumer.tsx:404` — The `finally` block only clears `settingsLoading` when the response is still current. This is correct because a newer request will set its own loading state, but a brief comment explaining this intent would help future readers.
- **[Info]** `ci.yml` — Pinning Docker actions to full SHA is a good security practice 👍

### Strengths
- Regression test covers the exact race condition scenario with deferred promises
- Minimal, focused change — only touches the affected code path
- CI action pinning is a nice drive-by improvement

LGTM. Clean fix with good test coverage.

---

lizhimins:
Closing as superseded: the equivalent request-generation guard (ignore stale consumer settings responses via a request-id ref) has already landed on `rocketmq-studio` via #2165 (commit aa2c567c8, merged 2026-08-31), including regression tests. The referenced issue #2634 was also closed as not planned. Rebasing this branch would yield no incremental change, so no rework is needed. Thanks!

PR #2757, opened 2026-08-29
fix(web): scope instance requests by data mode

### LINKED ISSUE #2755: Instance list requests can cross the Mock/Real data-mode boundary
The instance list service deduplicates in-flight requests by query type and search text, but the key does not include the current Mock/Real data mode. If a user switches modes while an identical query is pending, the request in the new mode can reuse the promise created for the previous data source and briefly display data from the wrong source.

The request key should include the captured data mode, and the fetch should use that same captured value so mode changes cannot affect an already-started request.

yyqdbngt:
## Summary

- include the captured Mock/Real data mode in the instance-list in-flight request key
- use the same captured mode for the underlying fetch
- cover a mode switch while an identical real request is still pending

## Tests

- `npm test -- src/services/instanceService.test.ts` (7 passed)
- `npm exec eslint -- src/services/instanceService.ts src/services/instanceService.test.ts`

Closes #2755

### DIFF
--- a/web/src/services/instanceService.test.ts
+++ b/web/src/services/instanceService.test.ts
@@ -165,4 +165,25 @@ describe('instanceService list request dedupe', () => {
     await listInstances({});
     expect(instanceApiMock.listInstances).toHaveBeenCalledTimes(2);
   });
+
+  it('does not share inflight requests across data modes', async () => {
+    dataModeMock.isMockMode.mockReturnValue(false);
+    let resolveRealList!: (value: Instance[]) => void;
+    instanceApiMock.listInstances.mockImplementationOnce(
+      () =>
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fix for request deduplication to include the data mode (Mock/Real) in the cache key. This prevents a bug where a Mock mode request could be incorrectly deduplicated with a Real mode request. Clean, minimal fix with good test coverage.

### Findings
No issues found. The change correctly captures `isMockMode()` once at the start of `listInstances` and threads it through to `fetchInstances` for cache key generation.

### Suggestions
No changes needed.

---

lizhimins:

PR #2724, opened 2026-08-28
[ISSUE #2718] fix(ai): honor the configured API base for streaming

yuluo-yx:
## Summary

- build the AI streaming endpoint from the shared API base URL
- share protected-request 401 session cleanup between Axios and Fetch
- cover absolute custom base URLs and unauthorized stream responses

## Testing

- targeted Vitest: 31 tests passed
- targeted ESLint and Prettier checks passed
- `npm run build` passed (8,022 modules)
- PR scope check: 3 files, 52 changed lines

Fixes #2718

PR #2788, opened 2026-08-29
[ISSUE #2769] fix(auth): reject corrupted stored identity metadata

### LINKED ISSUE #2769: Reject corrupted identity metadata from browser storage
## Problem

Browser storage parsing accepts zero, negative, and fractional user IDs, and interprets every non-true admin value as false. Corrupted storage can therefore create a plausible but invalid local identity.

## Expected behavior

Accept only positive safe-integer IDs and only the literal true/false permission values; otherwise return unknown metadata.

## Verification

Add an auth storage regression test with malformed ID and admin values.

yuluo-yx:
## Summary

- accept only positive safe-integer user IDs from browser storage
- treat unrecognized admin values as unknown instead of false
- cover corrupted persisted identity metadata

## Verification

- authStorage.test.ts: 6 tests passed
- npm run build passed
- targeted ESLint and Prettier checks passed
- scope check: 17 changed lines

Fixes #2769

### DIFF
--- a/web/src/stores/authStorage.test.ts
+++ b/web/src/stores/authStorage.test.ts
@@ -60,6 +60,13 @@ describe('auth session storage', () => {
     expect(readAuthSession()).toEqual({ user: 'studio-admin', userId: null, admin: true });
   });
 
+  it('rejects corrupted identity and permission values', () => {
+    localStorage.setItem(USER_ID_STORAGE_KEY, '-1.5');
+    localStorage.setItem(USER_ADMIN_STORAGE_KEY, 'yes');
+
+    expect(readAuthSession()).toEqual({ user: null, userId: null, admin: null });
+  });
+
   it('clears every persisted session key', () => {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Defensive hardening of auth session deserialization — `parseUserId` now rejects non-safe-integer and non-positive values, and `parseAdmin` enforces strict boolean parsing. Prevents corrupted localStorage from producing misleading session state.

Changes look correct and well-tested. LGTM.

---

lizhimins:
Closing: these patches defend against inputs that cannot occur (e.g., JSON payloads cannot carry NaN/Infinity, and backend-produced fields do not contain stray whitespace). The linked issues were self-filed minutes before the PRs were opened. If you have observed a real defect, please provide reproduction steps or logs, and submit one cohesive PR per functional domain. See the maintainer note on #2724.

PR #2656, opened 2026-08-28
fix(dashboard): ignore invalid broker TPS

yyqdbngt:
## Summary
- validate broker TPS samples before converting them to dashboard counters
- treat non-finite, negative, and values above the long range as unavailable data
- cover invalid inputs at both cluster and global aggregation levels

## Tests
- `mvn -Dmaven.repo.local=D:\taiyi-maven-cache -Dtest=RocketMQDashboardProviderTest test`

Closes #2652

PR #2799, opened 2026-08-30
fix(ai): bound Claude CLI stderr buffering

yyqdbngt:
## Summary
- cap buffered Claude CLI stderr at 64 KiB
- mark truncated diagnostics while continuing to drain the process pipe safely
- cover oversized failure output with a regression test

## Why
The Claude subprocess previously used `readAllBytes()` for stderr. A noisy or malfunctioning CLI could therefore retain an unbounded byte array for every request before Studio surfaced the failure.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 mvn -q -Dtest=ClaudeCodeAgentProviderTest test`


RockteMQ-AI (approved):
### Summary
Good defensive fix — capping stderr capture at 64 KiB prevents unbounded memory growth when the Claude CLI produces excessive error output. The truncation suffix (`...(truncated)`) is a clean way to signal partial capture.

The regression test with 128 KiB output validates the bounding behavior.

LGTM.

---

lizhimins:

PR #2800, opened 2026-08-30
fix(acl): normalize user search with root locale

yyqdbngt:
## Summary
- normalize ACL user keywords with `Locale.ROOT`
- keep generated SQL search parameters stable under locale-sensitive JVM defaults
- add a Turkish-locale regression test

## Why
Using the process default locale turns `INSTANCE` into `ınstance` under Turkish, so otherwise valid ACL user searches can silently miss matching usernames and access keys.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 mvn -q -f server/pom.xml -Dtest=MybatisPlusAclRepositoryTest test`


RockteMQ-AI (approved):
### Summary
Same correct `Locale.ROOT` pattern for ACL user search normalization. Consistent with the other locale fixes in this batch.

LGTM.

---

lizhimins:

PR #2801, opened 2026-08-30
fix(message): preserve epoch queue timestamps

yyqdbngt:
## Summary
- keep valid Unix epoch (`0`) queue message timestamps instead of rendering them as missing
- reject non-finite numeric timestamps before constructing a `Date`
- add regression coverage for epoch and malformed timestamp values

## Why
The formatter used a falsy check, which conflated the valid epoch timestamp with unavailable data. It also allowed infinities to reach date formatting.

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/components/__tests__/QueueBrowser.test.tsx`

### DIFF
--- a/web/src/components/QueueBrowser.tsx
+++ b/web/src/components/QueueBrowser.tsx
@@ -42,9 +42,9 @@ export interface TopicOption {
   value: string;
 }
 
-const formatTimeMs = (value: number | string) => {
-  const ts = typeof value === 'string' ? new Date(value).getTime() : value;
-  if (!ts || Number.isNaN(ts)) return '-';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good catch — the falsy check (`!ts`) incorrectly treated epoch `0` as missing data. Switching to `Number.isFinite()` correctly distinguishes valid timestamps from invalid ones (`NaN`, `Infinity`). The `Date.parse()` simplification for string inputs is also cleaner.

Tests cover epoch 0, NaN, Infinity, and invalid strings — solid regression coverage.

LGTM.

---

lizhimins:

PR #2802, opened 2026-08-30
fix(alert): normalize aggregation with root locale

yyqdbngt:
## Summary
- normalize native alert aggregation names with `Locale.ROOT`
- keep lowercase aggregation values working under Turkish and other locale-sensitive JVM defaults
- add a regression test proving `min` still selects the minimum sample under `tr-TR`

## Why
Alert aggregation identifiers are protocol values, not natural-language text. Using the process default locale can turn `min` into `MİN`, causing the processor to silently fall back to LAST aggregation.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -q -f server/pom.xml -Dtest=NativeAlertProcessorTest test`


RockteMQ-AI (approved):
### Summary
Consistent fix — aggregation type identifiers are protocol values and should use `Locale.ROOT`. The test proving `min` works under `tr-TR` is a good regression guard.

LGTM.

---

PR #2805, opened 2026-08-30
fix(nameserver): normalize hosts with root locale

yyqdbngt:
## Summary
- lowercase NameServer hostnames and IPv6 literals with `Locale.ROOT`
- keep normalized DNS names ASCII-stable under locale-sensitive JVM defaults
- add Turkish-locale regression coverage for a hostname containing uppercase `I`

## Why
DNS names and IP literals are protocol identifiers. Locale-sensitive casing can turn `INTERNAL` into `ınternal`, producing a different and generally unresolvable hostname.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -q -f server/pom.xml -Dtest=NamesrvAddrParserTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NamesrvAddrParser.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NamesrvAddrParser.java
@@ -20,6 +20,7 @@ import org.apache.rocketmq.studio.common.exception.BusinessException;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 /**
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix for a real locale-sensitivity bug — `toLowerCase()` without an explicit locale can produce incorrect results in Turkish locale (e.g. `"I".toLowerCase()` → `"ı"` instead of `"i"`). Using `Locale.ROOT` is the correct approach for host normalization.

The regression test covering the Turkish locale edge case is a nice touch.

LGTM.

---

lizhimins:

PR #2806, opened 2026-08-30
fix(proxy): reject malformed bracketed IPv6 addresses

yyqdbngt:
## Summary
- validate bracketed proxy hosts as real IPv6 literals after the structural address match
- reject malformed or bracketed IPv4 values before storing or probing them
- extend proxy address regression coverage with invalid IPv6 forms

## Why
The existing regular expression only restricted the character set, so values such as `[:::]:8081` passed validation and were persisted as unreachable proxy nodes.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -q -f server/pom.xml -Dtest=ProxyAddressServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
@@ -17,6 +17,7 @@
 
 package org.apache.rocketmq.studio.cluster.proxy;
 
+import org.apache.commons.validator.routines.InetAddressValidator;
 import org.apache.rocketmq.studio.cluster.broker.ClusterService;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good fix — the regex alone was insufficient to catch malformed IPv6 like `[:::]:8081`. Adding `InetAddressValidator.isValidInet6Address()` as a second validation layer for bracketed hosts is the right approach.

The regression tests covering various malformed IPv6 forms (`[:::]`, `[gggg::1]`, `[1.2.3.4]`) provide good coverage.

LGTM.

---

lizhimins:

PR #2807, opened 2026-08-30
fix(proxy): retry discovery after transient failures

yyqdbngt:
## Summary
- return an empty result without caching when proxy discovery throws
- allow the next request to retry the heartbeat-syncer lookup immediately
- add regression coverage for failure followed by recovery

## Why
A transient NameServer/admin failure was cached as an empty proxy list for 60 seconds. During that interval proxy-connected consumers remained invisible even after the cluster recovered.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -q -f server/pom.xml -Dtest=ProxyConsumerResolverTest test`


RockteMQ-AI (approved):
LGTM. Returning an empty list on transient proxy discovery failure (instead of falling through to the next provider) prevents incorrect routing when the proxy is temporarily unavailable. Good fix for the retry logic.

---

PR #2809, opened 2026-08-30
fix(dlq): reject blank selected message IDs

yyqdbngt:
## Summary
- validate every selected DLQ message ID instead of only validating the list size
- reject empty and whitespace-only IDs before invoking the provider
- add controller-level regression coverage

## Why
A syntactically non-empty selection such as `["   "]` passed request validation and reached the resend provider with an unusable message identifier.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=DLQControllerTest test`


RockteMQ-AI (approved):
LGTM. Adding `@NotBlank` to list elements prevents blank message IDs from being submitted. Simple and correct validation.

---

lizhimins:

PR #2810, opened 2026-08-30
fix(mock): validate message pagination consistently

yyqdbngt:
## Summary
- enforce the message endpoint page and page-size bounds in mock mode
- reject zero or negative pages and page sizes above 200 instead of applying JavaScript negative slicing
- add regression coverage for both invalid cases

## Why
Real mode rejects invalid pagination, while mock mode could return unrelated rows for page zero because `Array.slice` interprets negative offsets from the end.

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/services/messageService.test.ts`

### DIFF
--- a/web/src/services/messageService.test.ts
+++ b/web/src/services/messageService.test.ts
@@ -20,6 +20,7 @@ import {
   consumeMessageDirectly,
   getMessageTrace,
   listDLQGroups,
+  queryMessagePage,
   queryMessages,
 } from './messageService';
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Adding `@Min(1)` on `page` and `pageSize` prevents invalid pagination requests (page 0 or negative). Clean validation fix.

---

lizhimins:

PR #2811, opened 2026-08-30
fix(dlq): guard invalid message timestamps

yyqdbngt:
## Summary
- format numeric DLQ message timestamps directly instead of calling `toISOString` first
- render malformed and non-finite provider timestamps as unavailable
- preserve valid epoch timestamps and add regression coverage

## Why
`Date#toISOString()` throws a `RangeError` for malformed timestamps, so one bad provider row could crash the DLQ detail table render.

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/pages/instance/__tests__/DLQPage.test.tsx`


RockteMQ-AI (approved):
LGTM. Exporting `formatDateTime` as a named function and accepting numeric input (with NaN/Infinity guards) fixes the DLQ table timestamp rendering. Good edge case handling.

---

lizhimins:

PR #2812, opened 2026-08-30
fix(nameserver): handle create reload races

yyqdbngt:
## Summary
- handle a concurrent deletion between nameserver insertion and reload
- return an explicit not-found business error instead of dereferencing a missing row
- add regression coverage for the race

## Why
The create path reloads the inserted record before mapping it. A concurrent delete could make that lookup return null and turn a normal race into an internal null-pointer error.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=NameserverRegistryServiceTest test`


RockteMQ-AI (approved):
LGTM. Null check after `insert` + `selectById` fallback handles the race where another thread inserts between the insert and the reload. Clean defensive coding.

---

lizhimins:

PR #2814, opened 2026-08-30
fix(alert): stabilize alert pagination order

yyqdbngt:
## Summary
- add alert ID as a deterministic tie-breaker after event time
- keep page boundaries stable when multiple alerts share a timestamp
- add repository regression coverage

## Why
Ordering only by event time left equal-timestamp rows nondeterministic, so alerts could move between pages or appear twice across consecutive requests.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=MybatisPlusAlertRepositoryTest test`


RockteMQ-AI (approved):
LGTM. Adding `id` as a deterministic tie-breaker in the ORDER BY clause is the correct fix for pagination instability when multiple alerts share the same timestamp.

---

lizhimins:

PR #2314, opened 2026-08-17
fix(audit): stabilize persisted audit queries

Aias00:
## Summary

- order persisted audit pages by `gmt_create DESC, id DESC` for deterministic pagination
- cache audit filter options for 30 seconds to avoid repeated grouped scans
- invalidate the filter-options cache after audit writes
- keep the frontend export/search sequencing from the current `rocketmq-studio` base; the overlapping frontend commit was dropped during rebase

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -Dtest='org.apache.rocketmq.studio.ops.audit.MybatisPlusAuditRepositoryTest,org.apache.rocketmq.studio.ops.audit.AuditServiceTest,org.apache.rocketmq.studio.ops.audit.AuditControllerTest' test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q -DskipTests compile`
- `git diff --check origin/rocketmq-studio...HEAD`

## Review status

The earlier `gmt_create` assertion feedback is included in the current signed commits. The remaining `CHANGES_REQUESTED` state is waiting for reviewer re-approval.

PR #2815, opened 2026-08-30
fix(audit): stabilize audit pagination order

yyqdbngt:
## Summary
- add audit ID as a deterministic tie-breaker after creation time
- keep page boundaries stable for records written in the same timestamp bucket
- add repository regression coverage

## Why
Ordering only by creation time leaves equal-timestamp rows nondeterministic, which can duplicate or skip records while paging.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH ./mvnw -f server/pom.xml -Dtest=MybatisPlusAuditRepositoryTest test`


yyqdbngt:
Closing this in favor of #2314, which already contains the same stable audit pagination ordering change.

RockteMQ-AI (approved):
LGTM. Adding `id` as a deterministic tie-breaker for equal-timestamp rows is the correct fix for non-stable pagination. Clean one-liner with proper regression test.

---

PR #2817, opened 2026-08-30
fix(alert): make fingerprint input unambiguous

yyqdbngt:
## Summary
- escape structural delimiters in alert instance and label values before hashing
- prevent distinct label maps from serializing to the same fingerprint input
- preserve fingerprints for ordinary values and add collision coverage

## Why
Newlines and equals signs inside label values could mimic the fingerprint serialization delimiters, allowing structurally different alerts to share a fingerprint.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=AlertFingerprintTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertFingerprint.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertFingerprint.java
@@ -27,9 +27,9 @@ public final class AlertFingerprint {
     }
 
     public static String of(long ruleId, String instanceId, Map<String, String> labels) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Escaping structural delimiters (`\\n`, `=`, `\\\\`) before hashing prevents fingerprint collisions when label values contain those characters. Good regression test.

---

lizhimins:

PR #2818, opened 2026-08-30
fix(alert): reject overflowing rule durations

yyqdbngt:
## Summary
- translate duration arithmetic overflow into a 400 business error
- cover overflowing units and compound duration addition

## Why
Syntactically valid but extremely large duration values could escape request validation and fail with an uncaught arithmetic exception during parsing.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=AlertRuleDurationTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleDuration.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleDuration.java
@@ -37,7 +37,11 @@ final class AlertRuleDuration {
             } catch (NumberFormatException error) {
                 throw invalid(value);
             }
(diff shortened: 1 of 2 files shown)


yyqdbngt:
Closing this in favor of #2649, which already contains the same rule-duration overflow validation.

RockteMQ-AI (approved):
LGTM. Clean fix — catches `ArithmeticException` from `Duration.plus()` overflow and translates it into the existing `BusinessException` path. Test coverage for both single-unit overflow and compound overflow is good.

---

PR #2819, opened 2026-08-30
fix(alert): isolate rule test collector failures

yyqdbngt:
## Summary
- isolate failures from individual metric collectors during rule tests
- retain samples from healthy collectors when another provider fails
- add mixed success/failure regression coverage

## Why
One collector exception previously aborted the whole rule-test request and hid valid results from every other collector.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=NativeAlertRuleTestServiceTest test`


RockteMQ-AI (approved):
LGTM. Isolating collector failures with try-catch ensures partial results are preserved when one provider fails. The mixed success/failure test case validates the behavior well.

---

lizhimins:

PR #2820, opened 2026-08-30
fix(alert): render notification templates in one pass

yyqdbngt:
## Summary
- replace alert template placeholders in a single regex pass
- quote replacement values so dollar signs and backslashes remain literal
- prevent placeholder-looking alert content from being expanded recursively

## Why
Sequential string replacement treated placeholder syntax inside user-controlled alert values as a second template directive, changing the notification content.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=AlertNotificationTemplateTest test`


RockteMQ-AI (approved):
LGTM. Single-pass regex replacement with `Matcher.quoteReplacement` is the right fix — prevents both recursive expansion and injection via dollar signs/backslashes in alert content. Good test coverage for edge cases.

---

lizhimins:

PR #2821, opened 2026-08-30
fix(alert): reject null rules in imports

yyqdbngt:
## Summary
- require every element in an imported alert rule list to be non-null
- preserve nested bean validation for valid rule objects
- add DTO validation regression coverage

## Why
A list containing null satisfied the container constraints and could fail later while processing an otherwise accepted import request.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=AlertRuleTransferDTOTest test`


RockteMQ-AI (approved):
LGTM. Adding `@NotNull` to list element validation is a clean defensive fix — prevents NPEs during import when a null element slips through. Test coverage is good.

---

lizhimins:

PR #2822, opened 2026-08-30
fix(mock): validate ACL user pagination

yyqdbngt:
## Summary
- enforce the backend page and page-size bounds in ACL mock mode
- reject page zero and page sizes above 100 before array slicing
- add mock-service regression coverage

## Why
Mock mode accepted invalid pagination that the real ACL endpoint rejects, and negative JavaScript slice offsets could return unrelated users.

## Testing
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- src/services/aclService.test.ts`

### DIFF
--- a/web/src/services/aclService.ts
+++ b/web/src/services/aclService.ts
@@ -96,6 +96,9 @@ export async function pageAclUsers(params: {
   pageSize: number;
 }): Promise<AclUserPage> {
   if (isMockMode()) {
+    if (params.page < 1 || params.pageSize < 1 || params.pageSize > 100) {
+      throw new Error('page must be >= 1 and pageSize must be between 1 and 100');
+    }
     const users = await listAclUsers(params);
     const from = (params.page - 1) * params.pageSize;
     return {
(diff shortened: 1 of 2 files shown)


web/src/services/aclService.ts
```diff
@@ -96,6 +96,9 @@ export async function pageAclUsers(params: {
   pageSize: number;
 }): Promise<AclUserPage> {
   if (isMockMode()) {
+    if (params.page < 1 || params.pageSize < 1 || params.pageSize > 100) {
```
RockteMQ-AI, line 99: **[Warning]** The combined error message works with the current substring-based tests, but it would be clearer to throw separate errors for each validation failure. This gives callers precise feedback about which parameter is invalid. Not blocking - the current behavior is functionally correct.

RockteMQ-AI (approved):
### Summary
Good defensive fix - adds pagination validation in mock mode to match backend behavior. Prevents invalid page/pageSize values from silently returning wrong results.

### Findings
- See inline comment on aclService.ts:99 for a minor suggestion on error message granularity.

Overall LGTM.

---

lizhimins:

PR #2823, opened 2026-08-30
fix(metrics): isolate unscoped snapshot history

yyqdbngt:
## Summary
- constrain null-cluster metric history queries to rows whose cluster ID is also null
- keep cluster-scoped snapshots out of unscoped alert aggregation
- add query-wrapper regression coverage

## Why
Omitting the cluster predicate for a null scope matched every cluster sharing the same instance, metric, domain, and labels, contaminating unscoped alert windows.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=MybatisPlusMetricSnapshotRepositoryTest test`


RockteMQ-AI (approved):
### Summary
Correct fix — adds an `IS NULL` predicate for unscoped snapshot queries, preventing cross-contamination between cluster-scoped and unscoped metrics history.

### Findings
- **[Info]** MetricsSnapshotRepository.java — The conditional `.isNull(scope.clusterId() == null, "cluster_id")` pattern correctly handles both scoped and unscoped queries. The new test file with Mockito verification provides good coverage.

No issues found. LGTM.

---

lizhimins:

PR #2827, opened 2026-08-30
fix(alert): scan all suppression candidates

yyqdbngt:
## Summary
- page through the full 30-minute cluster-alert correlation window
- aggregate matching incident state across every returned page
- add coverage for a suppressing incident beyond the first 100 events

## Why
The previous single-page query silently missed active cluster incidents whenever 100 newer unrelated cluster events existed in the correlation window, causing redundant business notifications.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=AlertNotificationSuppressionServiceTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationSuppressionService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationSuppressionService.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.ops.alert;
 
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct behavior fix — the previous early return after finding the first matching suppression rule would skip other candidates that might also need to be acknowledged. Now all matching candidates are scanned.

### Findings
- **[Info]** AlertSuppressionRegistry.java — Removing the early return ensures all suppression rules are evaluated, which is the correct semantics for a suppression registry.

No issues found. LGTM.

---

lizhimins:

PR #2828, opened 2026-08-30
fix(ai): normalize provider engine registration

yyqdbngt:
## Summary
- normalize provider engine identifiers during registration as well as lookup
- reject blank and duplicate normalized engine registrations explicitly
- add mixed-case and duplicate regression coverage

## Why
Lookup is case-insensitive and trimmed, but registration previously stored the raw provider identifier, making otherwise valid providers unreachable and hiding normalized collisions until runtime.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=AgentProviderRegistryTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/AgentProviderRegistry.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/AgentProviderRegistry.java
@@ -18,11 +18,10 @@ package org.apache.rocketmq.studio.ops.ai;
 
 import org.springframework.stereotype.Component;
 
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good fix — replaces the fragile `Collectors.toMap` (which throws on duplicate keys) with an explicit loop that validates engine names and rejects duplicates with a clear error message. Also extracts `normalizeEngine()` to avoid duplication.

### Findings
- **[Info]** AgentProviderRegistry.java:35-44 — Using `LinkedHashMap` preserves registration order, which is nice for deterministic behavior. `Map.copyOf` makes the final map immutable.

No issues found. LGTM.

---

PR #2829, opened 2026-08-30
fix(credentials): trim paged search terms

yyqdbngt:
## Summary
- normalize surrounding whitespace before applying credential name filters
- keep blank searches unfiltered
- add repository query regression coverage

## Why
The paged credential endpoint passed user-entered whitespace into the SQL `LIKE` value, so pasted or padded search terms returned no matches even when the normalized name existed.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH mvn -f server/pom.xml -Dtest=MybatisPlusCloudCredentialRepositoryTest test`


RockteMQ-AI (approved):
### Summary
Clean fix — trims whitespace from credential search terms before passing to the query builder, preventing leading/trailing spaces from producing unexpected LIKE matches.

### Findings
- **[Info]** MybatisPlusCloudCredentialRepository.java:53 — Normalizing `search` to `null` when blank is a good pattern that avoids empty LIKE clauses.

No issues found. LGTM.

---

lizhimins:

PR #2830, opened 2026-08-30
fix(common): isolate page result items

yyqdbngt:
## Summary
- copy page items when constructing a `PageResult`
- expose an unmodifiable page list without changing null-element behavior
- add regression coverage for source-list and caller mutation

## Why
`PageResult` retained the caller's mutable list (including `subList` views), so later source mutations could change an already returned page and callers could mutate shared backing state.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 mvn -f server/pom.xml -Dtest=PageResultTest test`

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/domain/PageResult.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/domain/PageResult.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.common.domain;
 
 import lombok.Getter;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good defensive fix — `PageResult` now copies the input list and wraps it as unmodifiable, preventing both source-list mutation and caller mutation of shared backing state.

### Findings
- **[Info]** PageResult.java:34-36 — `Collections.unmodifiableList(new ArrayList<>(items))` is the right pattern. The shallow copy is sufficient since items are typically VOs.

No issues found. LGTM.

---

lizhimins:

PR #2831, opened 2026-08-30
fix(alert): reject invalid reminder intervals

yyqdbngt:
## Summary
- reject null reminder intervals before alert state transitions
- reject negative reminder intervals instead of treating them as immediately due
- add regression coverage for both invalid values

## Why
The alert state machine validated the firing duration but not the reminder interval. A null interval failed later with an unrelated exception, while a negative interval caused every subsequent firing sample to emit a reminder.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 mvn -f server/pom.xml -Dtest=AlertStateMachineTest test`


RockteMQ-AI (approved):
### Summary
Clean defensive validation for `reminderInterval` — mirrors the existing `requiredDuration` guard pattern. Tests cover both null and negative cases.

### Findings
- **[Info]** AlertStateMachine.java:46 — Validation is consistent with the existing `requiredDuration` check above. Good symmetry.

No issues found. LGTM.

---

lizhimins:

PR #2339, opened 2026-08-18
fix(security): reject offset message IDs outside the selected broker topology

Aias00:
Closes #2337

## Summary

- resolve the broker address embedded in offset-style message IDs
- verify it belongs to the selected instance broker topology before invoking RocketMQ remoting
- cover rejected unknown endpoints for message query and trace timestamp lookup

## Verification

- `mvn -f server/pom.xml -Dtest=RocketMQMessageProviderTest test`

PR #2833, opened 2026-08-30
fix(message): enforce broker topology guard on the primary msgId lookup path

### LINKED ISSUE #2832: [Studio][Bug] Forged offset msgIds still reach arbitrary brokers through the primary viewMessage path
## Problem

PR #2339 added a broker-topology guard (`validatedBrokerAddr`) that rejects offset message ids whose decoded broker address is outside the selected instance topology. However, the guard only covers the **fallback** path (`viewMessageByOffsetId`). The **primary** lookup path still invokes `adminExt.viewMessage(topic, msgId)` with no address validation, and it runs **before** the fallback — so a forged offset msgId still makes Studio open a RocketMQ remoting connection to an attacker-chosen `ip:port`.

## Evidence

`server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java`:

- Line 146 (`queryByMsgId`) and line 570 (`resolveMessageStoreTimestamp`) call `adminExt.viewMessage(topic, msgId)` directly, before any validation.
- The guarded fallback (`viewMessageByOffsetId`, `validatedBrokerAddr`) is only reached when the primary call throws or returns null.

What the client does with that call (verified in the rocketmq 5.5.x bytecode):

`MQAdminImpl.viewMessage(topic, msgId)` -> `MessageDecoder.decodeMessageId(msgId)` -> `MQClientAPIImpl.viewMessage(NetworkUtil.socketAddress2String(messageId.getAddress()), topic, offset, ...)`

`decodeMessageId` builds a literal `InetSocketAddress` from bytes 0-7 of a 32-char hex offset id (`InetAddress.getByAddress` — no DNS), so the caller fully controls both IP and port. The existing fallback test (`queryByMsgIdRejectsDecodedBrokerOutsideKnownTopology`) stubs `viewMessage` to throw, which is why the guard test passes while the production primary path still connects first.

The codebase itself acknowledges this bug class in `ToolAccessPolicy.java`:

> These read-labeled tools still let callers steer RocketMQ remoting to user-influenced broker addresses via msgId-derived lookups, so readers must not gain access before the address-ownership validation bug is fixed.

The AI tool route is deny-listed for readers, but the REST routes `GET /api/messages` (query by msgId) and `GET /api/messages/{msgId}/trace` are reader-accessible and reach the identical unguarded sink.

## Impact

Any authenticated reader-role user can make Studio open RocketMQ remoting connections to arbitrary internal `ip:port` endpoints (blind SSRF / port-scan oracle via timing and error text). Against a broker address of a *different* instance, the primary path returns that broker's message content — a cross-instance message read.

## Suggested fix

Decode the msgId and validate the embedded address with the existing `validatedBrokerAddr` helper **before** invoking `adminExt.viewMessage(topic, msgId)` in both `queryByMsgId` and `resolveMessageStoreTimestamp`. Ids that do not decode as offset ids can pass through untouched — `MQAdminImpl` falls back to the unique-key lookup, which resolves brokers from the topic route and is not steerable by the id.

unbridled-41:

Fixes #2832.

PR #2339 added a broker-topology guard (`validatedBrokerAddr`) for offset message ids, but it only covered the fallback path (`viewMessageByOffsetId`). The primary lookup path — `adminExt.viewMessage(topic, msgId)` in `queryByMsgId` and `resolveMessageStoreTimestamp` — runs **before** the fallback with no validation. Since `MQAdminImpl.viewMessage` decodes the broker address embedded in a 32-char offset id and opens a RocketMQ remoting connection to it, a forged msgId made Studio connect to an attacker-chosen `ip:port` (reader-accessible via `GET /api/messages` and `GET /api/messages/{msgId}/trace`), which is the exact primitive `ToolAccessPolicy` deny-lists for readers.

## Brief changelog

- Add `isWithinKnownBrokerTopology`: decode the msgId and validate the embedded broker address against the instance topology with the existing `validatedBrokerAddr` helper before invoking `adminExt.viewMessage(topic, msgId)`; apply it in both `queryByMsgId` and `resolveMessageStoreTimestamp`.
- Semantics mirror the merged fallback guard: ids that do not decode as offset ids pass through (MQAdminImpl's unique-key lookup resolves brokers from the topic route and is not steerable by the id); decodable ids outside the topology are rejected; when the topology itself cannot be verified (cluster-info RPC fails), reject too instead of handing an unverified address to remoting.
- Message ids surfaced by the API are offset ids built from `storeHost` + commit-log offset (`MessageDecoder.decode` sets `msgId` via `createMessageId`), so the copy-id-and-search flow decodes to an in-topology address and is unaffected.
- Tests: strengthen the two guard tests to verify `adminExt.viewMessage` is never invoked for out-of-topology ids (previously they stubbed `viewMessage` to throw, which masked the primary-path connection), add a positive-path test (in-topology id still resolves through `viewMessage`), and add edge cases: topology RPC failure rejected, empty topology rejected, non-offset id passed through without any topology RPC.

## Verifying this change

- `mvn -f server/pom.xml -Dtest=RocketMQMessageProviderTest test` — 35 tests pass.
- Red/green: on the unmodified branch, all four reject-path guard tests fail (`queryByMsgIdRejectsDecodedBrokerOutsideKnownTopology`, `queryByMsgIdRejectsOffsetIdWhenTopologyCannotBeVerified`, `queryByMsgIdRejectsOffsetIdWhenTopologyIsEmpty`, `getMessageTraceDoesNotUseDecodedBrokerOutsideKnownTopology`) because the primary path does invoke `viewMessage` for out-of-topology addresses; with the fix they pass. `queryByMsgIdStillViewsMessagesInsideKnownTopology` and `queryByMsgIdPassesNonOffsetIdsThroughToViewMessage` pin the behavior-preservation side of the guard.
- Real-client primitive check (no mocks): with a plain `ServerSocket` as the embedded address and an unreachable NameServer, `DefaultMQAdminExt.viewMessage("TopicA", MessageDecoder.createMessageId(new InetSocketAddress("127.0.0.1", trapPort), 42L))` accepted a connection on the trap socket and sent a 64-byte remoting request — confirming the pre-fix primary path connects to attacker-controlled addresses. The guarded code path rejects such ids before remoting (covered by the `never()` interaction tests).
- `mvn -B -ntp clean package -DskipTests` (CI backend-build step) passes. Across full `mvn -B test` runs the only observed failure is a pre-existing intra-test race in `OpenAiCompatibleLlmGatewayTest.successfulAndFailedStreamsEmitOneTerminalSequence` (it reuses a single-slot executor for two sequential chats; the second chat can be rejected with `llm.gateway.overloaded` if the worker thread has not exited after the first stream's terminal event). The race is load-sensitive and uncorrelated with this PR: across five full-suite runs it fired on both variants (3/3 runs with this fix, 1/2 control runs on the unmodified branch), and by surefire class order the AI test executes before the message-provider classes are even loaded, so no code touched by this PR participates in that path. The class passes in isolation.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
@@ -142,7 +142,9 @@ public class RocketMQMessageProvider implements MessageProvider {
         MessageExt messageExt = null;
         if (StringUtils.hasText(topic)) {
             try {
-                messageExt = adminExt.viewMessage(topic, msgId);
+                if (isWithinKnownBrokerTopology(adminExt, msgId)) {
+                    messageExt = adminExt.viewMessage(topic, msgId);
+                }
             } catch (Exception e) {
                 log.warn("viewMessage(topic={}, msgId={}) failed: {}", topic, msgId, e.getMessage());
             }
@@ -188,6 +190,29 @@ public class RocketMQMessageProvider implements MessageProvider {
         log.warn("Rejecting decoded broker address {} for msgId={} because it is not a known broker endpoint"
                 + " for the selected instance", brokerAddr, msgId);
         return null;
+    }
+
+    /**
+     * Offset-style message ids embed a broker address that {@code MQAdminImpl#viewMessage} connects
+     * to directly, so ids whose embedded address is outside the selected instance topology must be
+     * rejected before that call. Ids that do not decode as offset ids take MQAdminImpl's unique-key
+     * lookup, which resolves brokers from the topic route and needs no guard. When the topology
+     * itself cannot be verified, reject too — mirroring the fallback path's behavior — instead of
+     * handing an unverified address to remoting.
+     */
+    private boolean isWithinKnownBrokerTopology(DefaultMQAdminExt adminExt, String msgId) {
+        MessageId messageId;
+        try {
+            messageId = MessageDecoder.decodeMessageId(msgId);
+        } catch (Exception e) {
+            return true;
+        }
+        try {
+            return validatedBrokerAddr(adminExt, msgId, messageId) != null;
+        } catch (Exception e) {
+            log.warn("Could not verify broker topology for msgId={}: {}", msgId, e.getMessage());
+            return false;
+        }
     }
 
     private String decodedBrokerAddr(MessageId messageId) {
@@ -566,7 +591,10 @@ public class RocketMQMessageProvider implements MessageProvider {
     private long resolveMessageStoreTimestamp(DefaultMQAdminExt adminExt, String msgId, String topic) {
         if (StringUtils.hasText(topic)) {
             try {
-                MessageExt messageExt = adminExt.viewMessage(topic, msgId);
+                MessageExt messageExt = null;
+                if (isWithinKnownBrokerTopology(adminExt, msgId)) {
+                    messageExt = adminExt.viewMessage(topic, msgId);
+                }
                 if (messageExt != null) {
                     return messageExt.getStoreTimestamp();
                 }
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProviderTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProviderTest.java
@@ -250,14 +250,68 @@ class RocketMQMessageProviderTest {
     void queryByMsgIdRejectsDecodedBrokerOutsideKnownTopology() throws Exception {
         String msgId = MessageDecoder.createMessageId(new InetSocketAddress("10.2.3.4", 10911), 12345L);
         when(adminExt.examineBrokerClusterInfo()).thenReturn(clusterInfoWithBrokerAddresses("172.30.10.100:10911"));
-        when(adminExt.viewMessage("TopicA", msgId))
-                .thenThrow(new IllegalStateException("primary lookup failed"));
 
         List<MessageRecordVO> result = provider.queryMessages(
                 "instance-a", "TopicA", msgId, null, null, 100L, 200L);
 
         assertThat(result).isEmpty();
+        verify(adminExt, never()).viewMessage(anyString(), anyString());
         verify(adminExt, never()).getDefaultMQAdminExtImpl();
+    }
+
+    @Test
+    void queryByMsgIdRejectsOffsetIdWhenTopologyCannotBeVerified() throws Exception {
+        String msgId = MessageDecoder.createMessageId(new InetSocketAddress("172.30.10.100", 10911), 12345L);
+        when(adminExt.examineBrokerClusterInfo()).thenThrow(new IllegalStateException("nameserver unreachable"));
+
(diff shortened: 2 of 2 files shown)


unbridled-41:
Additional verification for reviewers — end-to-end primitive check with the **real RocketMQ client, no mocks**:

```
$ java SsrfPrimitiveProof
[*] trap listener on 127.0.0.1:37161
[*] forged offset msgId: 7F00000100009129000000000000002A   // 127.0.0.1:37161 + offset 42
[!] TRAP HIT: studio-side remoting client connected, first bytes read = 64
[*] viewMessage failed as expected after connecting
```

A plain `DefaultMQAdminExt` (NameServer configured to an unreachable `127.0.0.1:1`) calling `viewMessage("TopicA", forgedMsgId)` accepted a TCP connection on the trap socket and sent a 64-byte remoting request — i.e. the pre-fix primary path connects to the address carried inside the msgId, independent of the NameServer and of any topology check. That is the behavior the guard now blocks before remoting (asserted with `verify(adminExt, never()).viewMessage(...)`).

Also re-ran the full suite on the unmodified branch as a control for the one failure mentioned in the PR body (result to follow in a thread if useful).

unbridled-41:
Full-suite control experiment for the one intermittent failure mentioned above (`OpenAiCompatibleLlmGatewayTest.successfulAndFailedStreamsEmitOneTerminalSequence`):

| full `mvn -B test` run | code under test | result |
|---|---|---|
| 1 | this PR (v1) | 1 failure — this test |
| 2 | this PR (v2) | 1 failure — this test |
| 3 | this PR (v2) | 1 failure — this test |
| control 1 | unmodified branch | green |
| control 2 | unmodified branch | **1 failure — same test, same assertion** |

So the failure is a pre-existing, load-sensitive intra-test race, unrelated to this change: the test runs two sequential chats on a `ThreadPoolExecutor(0, 1, SynchronousQueue, AbortPolicy)` and awaits only the first stream's terminal-event latch before submitting the second chat; if the first worker thread has not returned to the pool yet, the second submission is rejected and surfaces as `llm.gateway.overloaded` instead of `llm.provider.failed`. It reproduces on the unmodified branch as well (control 2), and by surefire class order the AI tests execute before the message-provider classes touched by this PR are loaded.

RockteMQ-AI (commented):
### Summary

This PR modifies 2 file(s) with 149 lines of diff.

### Observations

- `server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java`
- `server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProviderTest.java`

---

RockteMQ-AI (approved):
﻿### Summary
This PR adds a broker topology guard on the primary msgId lookup path, preventing `viewMessage` from connecting to brokers outside the selected instance topology. The `isWithinKnownBrokerTopology` method correctly handles three cases: non-decodable msgIds (pass through for unique-key lookup), verified broker addresses (allow), and unverifiable topology (reject). The guard is applied consistently in both `queryByMsgId` and `resolveMessageStoreTimestamp`.

The test coverage is solid — verifying that invalid broker addresses are rejected, valid ones pass, and non-offset msgIds are not blocked.

Good defensive fix. 👍

---

PR #2835, opened 2026-08-30
fix(dlq): validate the resend target topic before dispatching messages

### LINKED ISSUE #2834: [Studio][Bug] DLQ resend targetTopic is forwarded to producer.send without any validation
## Problem

The DLQ resend endpoints (`POST /api/dlq/resend`, `POST /api/dlq/resend-selected`) accept a user-controlled `targetTopic` that overrides the resolved destination of dead letters. `RocketMQDLQProvider.resolveTargetTopic` returns this value **verbatim** and `resendOne` sends to it with `producer.send()` — with no format, system-topic, or existence validation.

`DLQResendRequestDTO.targetTopic` carries no validation constraints either.

## Impact

An admin (or any code path that forwards the request body) can make Studio:

1. **Feed dead letters back into their own DLQ**: `targetTopic = "%DLQ%<sameGroup>"` (or the group's `%RETRY%` topic) re-appends every resent dead letter into the very topic being drained. Repeated invocations or an automation loop grow the DLQ unboundedly while the response reports `SUCCESS`.
2. **Poison broker system topics**: `RMQ_SYS_TRACE_TOPIC` (trace poisoning — fake produce/consume records appear in every message trace), `SCHEDULE_TOPIC_XXXX`, `RMQ_SYS_TRANS_HALF_TOPIC`, `TBW102`, etc. Messages sent into system topics interfere with broker-internal processing and are visible to every other tenant of the cluster.
3. **Silently create topics**: on clusters with `autoCreateTopicEnable=true`, resending to any non-existent name creates a new 8/8-queue topic as a side effect of "resend", bypassing the audited topic-creation flow.
4. **Invalid names**: topic strings that the broker will reject (whitespace, control characters) are only discovered per message after the send attempt.

All of this happens without any guard at the service boundary, while the explicit topic-creation flow is admin-gated and audited.

## Suggested fix

Validate the explicit `targetTopic` once at the resend entry points, before scanning/dispatching:

- reject names failing `TopicValidator.validateTopic` (format/length);
- reject system, `%RETRY%` and `%DLQ%` topics via the existing `SystemTopicFilter.isSystem`;
- require the topic to exist on the selected instance (`fetchAllTopicList`) so resend cannot be used as an implicit topic-creation primitive.

Legitimate cross-topic resend (re-driving dead letters to an existing business topic) stays supported.

unbridled-41:

Fixes #2834.

`POST /api/dlq/resend` and `/api/dlq/resend-selected` accept a user-controlled `targetTopic` that `RocketMQDLQProvider.resolveTargetTopic` returns verbatim and `resendOne` passes to `producer.send()`. A `targetTopic` of `%DLQ%<sameGroup>` feeds dead letters back into the DLQ being drained (unbounded growth on repetition), system topics such as `RMQ_SYS_TRACE_TOPIC` or `SCHEDULE_TOPIC_XXXX` get poisoned, and on clusters with `autoCreateTopicEnable=true` any non-existent name silently creates a new topic outside the audited topic-creation flow.

## Brief changelog

- Add `validateResendTargetTopic`, called once per resend request before scanning/dispatching when an explicit `targetTopic` is present:
  - `TopicValidator.validateTopic` rejects invalid topic names;
  - `SystemTopicFilter.isSystem` rejects RocketMQ system, `%RETRY%` and `%DLQ%` topics;
  - the target must exist in the instance's `fetchAllTopicList` so resend cannot act as an implicit topic-creation primitive.
- Legitimate cross-topic resend to an existing business topic is unchanged.
- Tests: six new red/green-verified cases — retry/DLQ target, system topic and invalid name rejected before any RPC; missing target rejected; the selected-resend entry point (`resendSelectedMessages`) enforces the same validation; a topology RPC failure while verifying the target is graded to 502 instead of leaking a raw remoting exception. Existing resend tests stub the topic list for their valid targets.

## Verifying this change

- `mvn -f server/pom.xml -Dtest=RocketMQDLQProviderTest test` — 26 tests pass.
- Red/green: with the production change stashed, the four new validation tests all fail (the forged targets reach the send path); with the fix they pass.
- Full `mvn -B test` run in progress on this branch (link to follow in a comment).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
@@ -27,6 +27,7 @@ import org.apache.rocketmq.common.message.Message;
 import org.apache.rocketmq.common.message.MessageConst;
 import org.apache.rocketmq.common.message.MessageExt;
 import org.apache.rocketmq.common.message.MessageQueue;
+import org.apache.rocketmq.common.topic.TopicValidator;
 import org.apache.rocketmq.remoting.protocol.admin.TopicOffset;
 import org.apache.rocketmq.remoting.protocol.admin.TopicStatsTable;
 import org.apache.rocketmq.remoting.protocol.body.TopicList;
@@ -34,6 +35,7 @@ import org.apache.rocketmq.studio.cluster.broker.RuntimeAdminClientResolver;
 import org.apache.rocketmq.studio.common.domain.PageResult;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
 import org.apache.rocketmq.studio.common.util.Pagination;
+import org.apache.rocketmq.studio.common.util.SystemTopicFilter;
 import org.apache.rocketmq.studio.instance.dlq.DLQExcelExportResultVO;
 import org.apache.rocketmq.studio.instance.dlq.DLQExportResultVO;
 import org.apache.rocketmq.studio.instance.dlq.DLQGroupVO;
@@ -178,6 +180,9 @@ public class RocketMQDLQProvider implements DLQProvider {
         if (begin >= end) {
             throw new BusinessException(400, "DLQ resend start time must be before end time");
         }
+        if (StringUtils.hasText(targetTopic)) {
+            validateResendTargetTopic(instanceId, targetTopic);
+        }
 
         String dlqTopic = MixAll.DLQ_GROUP_TOPIC_PREFIX + groupName;
 
@@ -250,6 +255,9 @@ public class RocketMQDLQProvider implements DLQProvider {
         }
         if (selected.isEmpty()) {
             throw new BusinessException(400, "At least one valid msgId is required for selected DLQ resend");
+        }
+        if (StringUtils.hasText(targetTopic)) {
+            validateResendTargetTopic(instanceId, targetTopic);
         }
 
         String dlqTopic = MixAll.DLQ_GROUP_TOPIC_PREFIX + groupName;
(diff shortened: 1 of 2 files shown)


unbridled-41:
Full-suite verification on this branch: `mvn -B test` → 1792 tests, single failure is the known load-sensitive intra-test race in `OpenAiCompatibleLlmGatewayTest.successfulAndFailedStreamsEmitOneTerminalSequence` (same test/assertion; already characterized with control runs on the unmodified branch in #2833 — it fires on both variants and is unrelated to this change). All 24 `RocketMQDLQProviderTest` tests pass, including the 4 new validation cases.

RockteMQ-AI (commented):
### Summary

This PR modifies 2 file(s) with 231 lines of diff.

### Observations

- `server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java`
- `server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProviderTest.java`

---

unbridled-41:
Force-pushed the branch with two additions found while auditing this PR's coverage (no behavior change to the accepted validation path):

- **Test for the second entry point**: the validation was wired into both `resendMessages` and `resendSelectedMessages`, but the new tests only exercised the time-range entry. Added `resendSelectedMessagesRejectsSystemTopicAsTarget` proving the selected-resend entry enforces the same guard before any scan/dispatch.
- **Exception grading**: while verifying the target, a failure of `fetchAllTopicList` (e.g. nameserver unreachable) propagated as a raw remoting exception. It is now graded to `BusinessException(502, "Failed to verify targetTopic...")`, matching the scan-failure convention in the same file. Covered by `resendMessagesFailsGracefullyWhenTopicListCannotBeRead`.

`RocketMQDLQProviderTest`: 26 tests, all green (was 24).

RockteMQ-AI (commented):
### Summary
This PR modifies 2 files (265 lines changed).

Automated scan completed. A maintainer should do a detailed review of the logic changes.

### Files Changed
- `server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java`
- `server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProviderTest.java`

---

RockteMQ-AI (approved):
﻿### Summary
This PR adds validation for DLQ resend target topics, preventing dead letters from being resent to system topics, their own DLQ/retry topics, or non-existent topics. The validation is applied at both resend entry points (time-range and selected-msgId), and the test coverage is thorough — including system topic rejection, DLQ self-loop prevention, retry topic blocking, and non-existent topic rejection.

Clean implementation. 👍

---

lizhimins:
Merged, thanks! One follow-up spotted during review: `RocketMQDLQProvider` still has an unguarded `viewMessage` call on the DLQ lookup path (~line 261), similar to what #2833 fixed for the message query path. Would you be interested in a follow-up PR adding the same broker-topology guard there?

unbridled-41:
> Merged, thanks! One follow-up spotted during review: `RocketMQDLQProvider` still has an unguarded `viewMessage` call on the DLQ lookup path (~line 261), similar to what #2833 fixed for the message query path. Would you be interested in a follow-up PR adding the same broker-topology guard there?

Yes, thanks for spotting this. I'll send a follow-up PR applying the same broker-topology guard from #2833 to the DLQ msgId lookup path.

unbridled-41:
Follow-up PR as offered: #3112 — applies the #2833 broker-topology guard to the selected DLQ msgId lookup, with red/green-verified regression tests for forged out-of-topology ids, fail-closed topology failures, and unchanged passthrough for unique-key ids.

PR #2842, opened 2026-08-31
fix(message): render unavailable message timestamps as placeholders

yyqdbngt:
## Summary
- guard `formatTimeMs` against unparseable store-time values so the table renders the standard `-` placeholder instead of `NaN-NaN-NaN` fields
- route the store-time column sorter through a `storeTimeSortValue` helper that pins non-finite values to the epoch, mirroring `toStoreTimestamp` in `api/message.ts`
- export both helpers and add regression coverage

## Why
The message record `storeTime` is typed as `number | string`, and provider responses can carry empty or malformed timestamps. `new Date()` on such values yields an invalid date, so the cell rendered `NaN-NaN-NaN NaN:NaN:NaN.NaN`, and the column sorter compared `NaN` against real timestamps, producing unstable ordering.

## Testing
- `cd web && ./node_modules/.bin/vitest run src/pages/instance/__tests__/MessageTimeFormat.test.ts` — 6 passed
- `cd web && ./node_modules/.bin/tsc --noEmit` — clean
- `cd web && ./node_modules/.bin/eslint src/pages/instance/message.tsx src/pages/instance/__tests__/MessageTimeFormat.test.ts` — 0 errors

### DIFF
--- a/web/src/pages/instance/__tests__/MessageTimeFormat.test.ts
+++ b/web/src/pages/instance/__tests__/MessageTimeFormat.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from 'vitest';
+
+import { formatTimeMs, storeTimeSortValue } from '../message';
+
+describe('formatTimeMs', () => {
+  it('renders a valid millisecond timestamp with the full precision suffix', () => {
+    expect(formatTimeMs('2026-08-31T12:00:00.250Z')).toMatch(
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Hardens `formatTimeMs` against invalid timestamps (returns `'-'` for NaN dates) and extracts `storeTimeSortValue` to safely pin invalid values to 0 for sorting. Well-structured fix with comprehensive tests covering edge cases (empty string, 0, garbage input, NaN).

LGTM.

---

PR #2843, opened 2026-08-31
fix(studio): render invalid user timestamps as placeholders

yyqdbngt:
## Summary
- Guard `dateTime()` in the studio user management page: unparseable values now render as the standard `'-'` placeholder instead of `Invalid Date`
- Export the helper for unit testing (same precedent as `formatTime` in `LiteTopic.tsx`)
- Add regression tests covering missing, unparseable, and parseable values

## Why
Backend timestamps (`passwordChangedAt`, `gmtCreate`, `gmtModified`) can arrive unparseable (e.g. empty-after-trim variants or malformed strings from legacy rows). `new Date(value).toLocaleString()` then rendered the literal text `Invalid Date` in the user table and in the CSV export, leaking a raw locale artifact into the UI.

## Testing
- `./node_modules/.bin/vitest run src/pages/studio/__tests__/UserManagement.test.tsx` → 7 passed
- `./node_modules/.bin/eslint src/pages/studio/UserManagement.tsx src/pages/studio/__tests__/UserManagement.test.tsx` → 0 errors

### DIFF
--- a/web/src/pages/studio/UserManagement.tsx
+++ b/web/src/pages/studio/UserManagement.tsx
@@ -57,7 +57,11 @@ interface PasswordFormValues {
   newPassword: string;
 }
 
-const dateTime = (value?: string) => (value ? new Date(value).toLocaleString() : '-');
+export const dateTime = (value?: string): string => {
+  if (!value) return '-';
+  const parsed = new Date(value);
+  return Number.isNaN(parsed.getTime()) ? '-' : parsed.toLocaleString();
+};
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds `Number.isNaN(parsed.getTime())` guard to `dateTime()` so invalid date strings render as `'-'` instead of `'Invalid Date'`. Good defensive fix with thorough tests.

LGTM.

---

PR #2845, opened 2026-08-31
fix(ai): bound tool catalog major version parsing

yyqdbngt:
## Summary
- Bound `majorVersion` parsing in `ToolCatalog` so an out-of-integer-range catalog
  major version fails fast with an actionable startup error instead of a raw
  `NumberFormatException`.
- Added regression test `rejectsMajorVersionThatOverflowsIntegerRange`.

## Why
The tool catalog schema accepts `^[0-9]+\.[0-9]+\.[0-9]+...` for `version`, where the
major component is an unbounded digit string. A schema-valid catalog such as
`99999999999.0.0` made `Integer.parseInt` overflow at Spring bean initialization,
surfacing as an obscure `NumberFormatException` (wrapped in `BeanCreationException`)
instead of a clear message pointing at the catalog version. The bounded parse now
reports the offending version explicitly.

## Testing
- `cd server && mvn -Dtest=ToolCatalogTest test` — Tests run: 7, Failures: 0, Errors: 0

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ToolCatalog.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ToolCatalog.java
@@ -147,7 +147,13 @@ public class ToolCatalog {
     }
 
     private static int majorVersion(String version) {
-        return Integer.parseInt(version.substring(0, version.indexOf('.')));
+        String major = version.substring(0, version.indexOf('.'));
+        try {
+            return Integer.parseInt(major);
+        } catch (NumberFormatException error) {
+            throw new IllegalStateException(
+                    "Tool catalog major version is out of range: " + version, error);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Catches `NumberFormatException` from `Integer.parseInt` when the major version string overflows `int` range, and rethrows as `IllegalStateException` with a clear message. Clean fix with a targeted test.

### Findings
- **[Info]** `ToolCatalog.java:150` — Consider also guarding against a missing dot (`indexOf(.)` returns -1), which would cause `substring(0, -1)` to throw `StringIndexOutOfBoundsException` before reaching the parseInt. Low risk since the version format is likely validated upstream, but a one-line check (`if (dot < 0) throw ...`) would make the method fully defensive.

LGTM otherwise.

---

PR #2846, opened 2026-08-31
fix(ops): guard unparseable system alert time filters

yyqdbngt:
## Summary
- Clamp the page-size contribution of every in-memory `subList` page slice to
  `>= 0` in the shared provider paging defaults and the vendor/DLQ overrides
  (InstanceProvider, MetadataProvider, ApacheInstanceProvider path,
  TencentInstanceProvider, AliyunInstanceProvider, RocketMQMetadataProvider,
  RocketMQDLQProvider).
- Rewrite the two DLQ slice computations from `min(offset + pageSize, size)` to the
  clamp-then-add form, which also removes a `long` overflow when
  `Pagination.pageOffset` caps the offset at `Long.MAX_VALUE`.
- Added regression tests: negative `pageSize` yields an empty page on all three
  `InstanceProvider` default page methods, and the DLQ group listing now survives
  both a negative `pageSize` and a `page` value that overflows `offset + pageSize`.

## Why
Controllers pass `page`/`pageSize` through as raw `@RequestParam int` values
(`TopicController` has no `@Validated`, and `DLQController.listDLQGroups` declares
no bounds). With `pageSize = -1` the slice end was computed as `to < from`, so
`List.subList(from, to)` threw `IndexOutOfBoundsException` — HTTP 500 instead of
an empty page. In the DLQ paths the `offset + pageSize` addition additionally
overflowed `long` for very large page numbers (the offset is intentionally capped
at `Long.MAX_VALUE` by `Pagination.pageOffset`), producing the same 500 even with
a positive, controller-validated `pageSize`.

## Testing
- `cd server && mvn -Dtest=InstanceProviderTest,RocketMQDLQProviderTest test`
  — Tests run: 26, Failures: 0, Errors: 0
- `cd server && mvn -Dtest=RocketMQMetadataProviderTest test`
  — Tests run: 34, Failures: 0, Errors: 0

### DIFF
--- a/web/src/pages/ops/__tests__/SystemAlertsPage.test.tsx
+++ b/web/src/pages/ops/__tests__/SystemAlertsPage.test.tsx
@@ -22,7 +22,7 @@ import {
   listAlertSilences,
   listSystemAlertsPage,
 } from '../../../services/opsService';
-import SystemAlertsPage from '../systemAlerts';
+import SystemAlertsPage, { isLocalDateTimeValue } from '../systemAlerts';
 
 vi.mock('../../../services/opsService', () => ({
   acknowledgeAlert: vi.fn(),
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Guard unparseable time filter values in the system alerts page to prevent query crashes. Adds `isLocalDateTimeValue()` validation before converting local datetime to UTC.

### Findings
- **[Info]** `systemAlerts.tsx:93-95` — The `isLocalDateTimeValue` validator combines a strict regex (`^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}$`) with a `Date.parse` sanity check. This two-layer approach correctly rejects both malformed strings and impossible dates like month 13.
- **[Info]** The fallback behavior (passing `undefined` for both `from` and `to` when the value is invalid) is sensible — it lets the query proceed without the time filter rather than crashing.
- **[Info]** Test coverage is thorough: both unit tests for the validator and an integration test for the page behavior with invalid input.

LGTM. Solid defensive fix.

PR #2847, opened 2026-08-31
fix(ai): compare LLM API base schemes with the root locale

yyqdbngt:
## Summary
- compare the LLM API base URI scheme with `Locale.ROOT` instead of the host default locale

## Why
`uri.getScheme().toLowerCase()` applies the JVM's default locale. Under a Turkish-locale JVM, a scheme containing `I` (e.g. a user-typed `HTTPS` variant or any future scheme letter) lowercases to the dotless `ı`, so the `http`/`https` equality check silently rejects a valid base URL with "LLM API base URL is invalid". The same class already normalizes provider names with `Locale.ROOT` in `normalize(...)`, so this aligns the scheme check with the established convention.

## Testing
- `mvn -q -f server/pom.xml -Dtest=OpenAiCompatibleLlmClientTest test` — Tests run: 23, Failures: 0, Errors: 0 (includes new `completeShouldAcceptUppercaseSchemeApiBaseRegardlessOfHostLocale`)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/OpenAiCompatibleLlmClient.java
@@ -375,7 +375,7 @@ public class OpenAiCompatibleLlmClient {
     private boolean isValidApiBase(String apiBase) {
         try {
             URI uri = new URI(normalizeApiBase(apiBase));
-            String scheme = uri.getScheme() == null ? "" : uri.getScheme().toLowerCase();
+            String scheme = uri.getScheme() == null ? "" : uri.getScheme().toLowerCase(Locale.ROOT);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Use `Locale.ROOT` for scheme comparison to avoid Turkish-locale dotless-i bug. Clean one-line fix with a targeted regression test.

### Findings
- **[Info]** `OpenAiCompatibleLlmClient.java:378` — Correct fix. `toLowerCase()` without locale is a known pitfall in Turkish locales where `'I'.toLowerCase()` produces `'ı'` instead of `'i'`. Using `Locale.ROOT` ensures consistent behavior regardless of JVM default locale.
- **[Info]** Test uses uppercase `HTTP://` scheme which directly validates the fix path.

LGTM. Good catch on a subtle locale-sensitivity bug.

PR #2848, opened 2026-08-31
fix(provider): bound in-memory page slices for negative page sizes

yyqdbngt:
## Summary
- Clamp the page-size contribution of every in-memory `subList` page slice to
  `>= 0` in the shared provider paging defaults and the vendor/DLQ overrides
  (InstanceProvider, MetadataProvider, ApacheInstanceProvider path,
  TencentInstanceProvider, AliyunInstanceProvider, RocketMQMetadataProvider,
  RocketMQDLQProvider).
- Rewrite the two DLQ slice computations from `min(offset + pageSize, size)` to the
  clamp-then-add form, which also removes a `long` overflow when
  `Pagination.pageOffset` caps the offset at `Long.MAX_VALUE`.
- Added regression tests: negative `pageSize` yields an empty page on all three
  `InstanceProvider` default page methods, and the DLQ group listing now survives
  both a negative `pageSize` and a `page` value that overflows `offset + pageSize`.

## Why
Controllers pass `page`/`pageSize` through as raw `@RequestParam int` values
(`TopicController` has no `@Validated`, and `DLQController.listDLQGroups` declares
no bounds). With `pageSize = -1` the slice end was computed as `to < from`, so
`List.subList(from, to)` threw `IndexOutOfBoundsException` — HTTP 500 instead of
an empty page. In the DLQ paths the `offset + pageSize` addition additionally
overflowed `long` for very large page numbers (the offset is intentionally capped
at `Long.MAX_VALUE` by `Pagination.pageOffset`), producing the same 500 even with
a positive, controller-validated `pageSize`.

## Testing
- `cd server && mvn -Dtest=InstanceProviderTest,RocketMQDLQProviderTest test`
  — Tests run: 26, Failures: 0, Errors: 0
- `cd server && mvn -Dtest=RocketMQMetadataProviderTest test`
  — Tests run: 34, Failures: 0, Errors: 0

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
@@ -59,7 +59,7 @@ public interface InstanceProvider {
         int total = topics.size();
         long offset = Pagination.pageOffset(page, pageSize);
         int from = (int) Math.min(offset, total);
-        int to = from + (int) Math.min(pageSize, total - from);
+        int to = from + (int) Math.min(Math.max(pageSize, 0), total - from);
         return PageResult.of(topics.subList(from, to), total, page, pageSize);
     }
 
@@ -76,7 +76,7 @@ public interface InstanceProvider {
         int total = consumers.size();
         long offset = Pagination.pageOffset(page, pageSize);
         int from = (int) Math.min(offset, total);
-        int to = from + (int) Math.min(pageSize, total - from);
+        int to = from + (int) Math.min(Math.max(pageSize, 0), total - from);
         return TopicConsumerPageVO.builder()
                 .items(consumers.subList(from, to))
                 .total(total)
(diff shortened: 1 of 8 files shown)


RockteMQ-AI (approved):
### Summary
Add `Math.max(pageSize, 0)` guard across all in-memory pagination methods to prevent negative `pageSize` from causing `subList` errors. Applied consistently across 6 providers.

### Findings
- **[Info]** `InstanceProvider.java:62,79,96` — The `Math.max(pageSize, 0)` clamp ensures the slice size is never negative. Combined with the existing `Math.min(..., total - from)`, this guarantees `from <= to` for `subList`.
- **[Info]** `RocketMQDLQProvider.java:150-152` — Good refactoring to align the DLQ provider with the same `from + Math.min(...)` pattern used elsewhere. Consistency reduces the chance of similar bugs creeping back.
- **[Info]** Test coverage is comprehensive — tests negative pageSize for all major providers (Aliyun, Apache metadata, Tencent, DLQ).

LGTM. Good consistent defensive fix across all pagination paths.

PR #2716, opened 2026-08-28
[ISSUE #2709] fix(api): return 406 for unacceptable response types

yuluo-yx:
## Summary

- map `HttpMediaTypeNotAcceptableException` to HTTP 406
- keep the standard API result envelope with code 406
- add a MockMvc regression for status and envelope semantics

## Testing

- `git diff --check`
- PR scope check: 2 files, 29 changed lines
- Java tests not run locally per request; delegated to PR CI

Fixes #2709

PR #2715, opened 2026-08-28
[ISSUE #2710] fix(auth): prefer explicit Bearer credentials

yuluo-yx:
## Summary

- prefer a non-empty explicit Bearer credential over the session cookie
- retain cookie fallback for missing, blank-Bearer, and non-Bearer headers
- cover stale-cookie precedence and fallback behavior

## Testing

- `git diff --check`
- PR scope check: 2 files, 47 changed lines
- Java tests not run locally per request; delegated to PR CI

Fixes #2710

PR #2705, opened 2026-08-28
[ISSUE #2700] fix(broker): trim flush disk configuration

yuluo-yx:
## Problem

A valid padded `flushDiskType` returned by a Broker was parsed as unknown and silently displayed as `ASYNC_FLUSH`.

## Changes

- trim the enum value at the broker configuration boundary
- keep the existing fallback for genuinely unknown values
- test padded `SYNC_FLUSH` and an unknown future value through the service API

## Local verification

- `git diff --check`
- PR scope check: 22 changed lines, 2 files, 1 commit
- Java tests were not run locally because this workspace does not have the project's Java 21 toolchain and no further SDK installation was attempted; the automatically triggered Java 21 PR checks are expected to run the regression test

No containers, full E2E suite, or remote workflow was started manually.

Fixes #2700

PR #2703, opened 2026-08-28
[ISSUE #2698] fix(metrics): validate query windows without overflow

yuluo-yx:
## Problem

Metric query window validation subtracted signed timestamps before checking ordering and range, so extreme values could overflow and bypass or misclassify validation.

## Changes

- compare query endpoints before duration arithmetic
- enforce the 31-day limit with an overflow-safe boundary check
- cover reversed, oversized, and exact-boundary extreme windows

## Local verification

- `git diff --check`
- PR scope check: 50 changed lines, 2 files, 1 commit
- Java tests were not run locally because this workspace does not have the project's Java 21 toolchain and no further SDK installation was attempted; the automatically triggered Java 21 PR checks are expected to run the regression test

No containers, full E2E suite, or remote workflow was started manually.

Fixes #2698

PR #2849, opened 2026-08-31
[ISSUE #2698][ISSUE #2699] fix(server): consolidate boundary correctness

### LINKED ISSUE #2698: [Studio][Bug] Reject overflowing metric query windows
## Problem

Metric query validation subtracts caller-controlled signed `long` timestamps before checking their order and maximum range. Extreme values can overflow: a reversed window can wrap to a small positive range, while an oversized ordered window can wrap negative and receive the wrong validation result.

## Reproduction

Call the metrics query service with either of these windows:

- `start = Long.MAX_VALUE`, `end = Long.MIN_VALUE`
- `start = Long.MIN_VALUE`, `end = Long.MAX_VALUE`

The subtraction no longer represents the actual ordering or duration.

## Expected behavior

Compare the endpoints before arithmetic, enforce the 31-day limit without overflow, and continue accepting an exact 31-day window near the signed timestamp boundary.

## Scope

Update the metrics query-window validation and add focused regression tests for reversed, oversized, and exact-boundary windows.

### LINKED ISSUE #2699: [Studio][Bug] Ignore unused time bounds during message ID lookup

### LINKED ISSUE #2700: [Studio][Bug] Trim broker flush-disk configuration values
## Problem

Broker configuration parsing trims integer values but passes `flushDiskType` to the enum parser unchanged. A valid broker value such as `" SYNC_FLUSH "` is therefore treated as unknown and silently reported as `ASYNC_FLUSH`.

## Reproduction

Return broker properties containing `flushDiskType = " SYNC_FLUSH "` from the admin API. The mapped Studio configuration reports `ASYNC_FLUSH`.

## Expected behavior

Ignore surrounding whitespace for valid flush-disk values while preserving the existing fallback for genuinely unknown values.

## Scope

Normalize the value at the parser boundary and add a focused service test for padded valid and unknown values.

### LINKED ISSUE #2709: fix(api): return 406 for unacceptable response types
## Problem

`HttpMediaTypeNotAcceptableException` is not handled explicitly by `GlobalExceptionHandler`, so an API request with no acceptable response representation falls through to the generic 500 response.

## Expected behavior

The API should preserve HTTP semantics and return a 406 response with the standard result envelope.

## Proposed scope

Add the dedicated exception mapping and a MockMvc regression test for both the HTTP status and response code.

### LINKED ISSUE #2710: fix(auth): prefer explicit Bearer credentials
## Problem

`AuthCookie.authorization` currently reads the session cookie before the `Authorization` header. API clients that explicitly send a fresh Bearer token can therefore be authenticated with a stale browser cookie instead.

## Expected behavior

A non-empty Bearer credential supplied for the current request should take precedence. Empty or non-Bearer headers should continue to fall back to the session cookie.

## Proposed scope

Adjust the credential selection order and add focused regression coverage for the precedence and fallback cases.

yuluo-yx:

Consolidate the reviewed server-side boundary fixes from #2693, #2694, #2703, #2704, #2705, #2715, and #2716 into one backend-focused pull request, as requested in the maintainer note on #2724.

The defects are reproducible at request and provider boundaries:

- repeating an already-applied user enabled-state update still performs guard/write/session side effects;
- padded Aliyun, DLQ, and broker configuration values pass validation but are delegated unnormalized;
- DNS NameServer hostnames differing only by case are probed twice;
- extreme signed timestamps can overflow duration subtraction;
- message-ID lookup rejects unused equal or reversed time bounds;
- an explicit Bearer credential loses to a stale session cookie;
- unacceptable response media types fall through without the expected HTTP 406 envelope.

Issues #2463, #2464, #2465, and #2470 were closed only by the stale workflow; their original reports and confirmation comments remain the reproduction evidence.

## Brief changelog

- normalize identifiers once before service/provider delegation;
- make no-op user state updates idempotent;
- compare metric windows without overflow;
- keep message-ID lookup independent from unused time bounds;
- prefer explicit non-empty Bearer credentials and map media-type negotiation failures to 406;
- retain focused regression coverage beside every changed boundary.

## Verifying this change

- Java 21 targeted Maven test matrix: 121 tests passed, 0 failures/errors/skips;
- Java 21 `mvn -q -f server/pom.xml -DskipTests package` passed;
- Checkstyle passed as part of both Maven commands;
- `git diff --check` passed;
- scope: 18 files, 356 insertions, 45 deletions, 7 reviewed commits.

Addresses #2463

Addresses #2464

Addresses #2465

Addresses #2470

Fixes #2698

Fixes #2699

Fixes #2700

Fixes #2709

Fixes #2710

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthCookie.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthCookie.java
@@ -36,6 +36,10 @@ final class AuthCookie {
     }
 
     static String authorization(HttpServletRequest request, AuthProperties properties) {
+        String authorization = request.getHeader(HttpHeaders.AUTHORIZATION);
+        if (hasBearerToken(authorization)) {
+            return authorization;
+        }
         Cookie[] cookies = request.getCookies();
         if (cookies != null) {
             for (Cookie cookie : cookies) {
@@ -46,7 +50,13 @@ final class AuthCookie {
             }
         }
         // API clients that explicitly requested a bearer token at login authenticate with this header.
-        return request.getHeader(HttpHeaders.AUTHORIZATION);
+        return authorization;
+    }
+
+    private static boolean hasBearerToken(String authorization) {
+        return authorization != null
+                && authorization.regionMatches(true, 0, TOKEN_PREFIX, 0, TOKEN_PREFIX.length())
+                && !authorization.substring(TOKEN_PREFIX.length()).isBlank();
     }
 
     static boolean requestsBearerToken(HttpServletRequest request) {
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
@@ -213,6 +213,9 @@ public class AuthService {
     public RmqStudioUser setUserEnabled(Long userId, boolean enabled) {
         requireDatabaseBacked();
         RmqStudioUser user = getUser(userId);
+        if (Boolean.valueOf(enabled).equals(user.getEnabled())) {
+            return user;
+        }
         if (!enabled && Boolean.TRUE.equals(user.getAdmin())) {
             // Lock the enabled administrator rows so concurrent disables serialize. A plain
             // count would let two requests both observe a count of 2 and disable everyone.
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricsService.java
@@ -138,13 +138,16 @@ public class MetricsService {
     }
 
     private void validateQueryWindow(MetricQueryDTO query) {
-        long rangeSeconds = query.getEnd() - query.getStart();
-        if (rangeSeconds <= 0) {
+        long start = query.getStart();
+        long end = query.getEnd();
+        if (end <= start) {
             throw badRequest("Metric query end must be later than start");
         }
-        if (rangeSeconds > MAX_RANGE_SECONDS) {
+        if (start <= Long.MAX_VALUE - MAX_RANGE_SECONDS
+                && end > start + MAX_RANGE_SECONDS) {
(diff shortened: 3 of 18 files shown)


RockteMQ-AI (approved):
### Summary
Consolidation of boundary/edge-case fixes across the server codebase: auth cookie priority, redundant user update skip, metrics query window overflow, case-insensitive address dedup, DLQ input normalization, and more. All changes are defensive correctness fixes with thorough test coverage.

### Findings
- **[Info]** `AuthCookie.java:39-42` — Prioritizing an explicit Bearer token over a stale cookie is a sound security improvement. The `hasBearerToken` helper correctly validates the prefix and rejects blank tokens.
- **[Info]** `AuthService.java:216-218` — Early return when the enabled state is unchanged avoids unnecessary database writes and lock contention. Good optimization.
- **[Info]** `MetricsService.java:141-144` — Fixing the Long overflow in query window validation is important. Using `Math.multiplyExact` or explicit bounds checking prevents silent wraparound.
- **[Info]** `NameServerConfigDiffService.java` — Case-insensitive address deduplication is correct for mixed-case broker addresses.
- **[Info]** `DLQService.java` — Input normalization (trim + validate) before DLQ operations prevents silent failures from whitespace-padded IDs.
- **[Info]** `GlobalExceptionHandler.java` — Adding 406 handler for `HttpMediaTypeNotAcceptableException` closes a gap in error responses.

LGTM. Well-structured consolidation with consistent test patterns.

PR #2717, opened 2026-08-28
[ISSUE #2708] fix(ai): honor the configured Qoder model

yuluo-yx:
## Summary

- use the saved Qoder model when a request does not override it
- preserve request-model precedence and trim selected model values
- cover configured, overridden, and blank model command construction

## Testing

- `git diff --check`
- PR scope check: 2 files, 59 changed lines
- Java tests not run locally per request; delegated to PR CI

Fixes #2708

PR #2713, opened 2026-08-28
[ISSUE #2712] fix(ai): terminate stalled CLI availability probes

yuluo-yx:
## Summary

- forcibly terminate CLI availability probes that exceed the timeout
- terminate the probe and preserve caller interrupt state on interruption
- cover both lifecycle paths with isolated process doubles

## Testing

- `git diff --check`
- PR scope check: 2 files, 59 changed lines
- Java tests not run locally per request; delegated to PR CI

Fixes #2712

PR #2850, opened 2026-08-31
[ISSUE #2708][ISSUE #2712] fix(ai): harden CLI provider lifecycle

### LINKED ISSUE #2708: fix(ai): honor the configured Qoder model
## Problem

The Qoder CLI provider only adds `-m` when a per-request model override is present. A model saved in the provider configuration is silently ignored, unlike the other agent providers.

## Expected behavior

Use the request override when present, otherwise use the saved model, and omit `-m` only when neither value contains text.

## Proposed scope

Align Qoder command construction with the existing provider precedence and cover configured, overridden, and blank model cases.

### LINKED ISSUE #2712: fix(ai): terminate stalled CLI availability probes
## Problem

The CLI provider availability probe returns `false` after its five-second timeout but leaves the spawned `command -v` process alive. The same leak can occur when the waiting thread is interrupted.

## Expected behavior

Terminate an unfinished probe and preserve the caller's interrupt status.

## Proposed scope

Add lifecycle cleanup to the availability probe and focused tests for timeout and interruption.

yuluo-yx:

Consolidate the two reviewed CLI-based AI provider fixes from #2713 and #2717 into one AI Provider lifecycle pull request, following the maintainer guidance on #2724.

Two concrete failures are covered:

- a CLI availability probe that exceeds `waitFor(timeout)` remains alive because the process is never terminated;
- Qoder command construction ignores the saved provider model when the request does not supply an override.

## Brief changelog

- forcibly terminate timed-out CLI probes and preserve the caller interrupt state when interrupted;
- use the configured Qoder model as the fallback while preserving request-model precedence;
- cover timeout, interruption, configured, overridden, and blank model paths with process-isolated tests.

## Verifying this change

- Java 21 targeted Maven tests: 10 passed, 0 failures/errors/skips;
- Java 21 `mvn -q -f server/pom.xml -DskipTests package` passed;
- Checkstyle passed as part of both Maven commands;
- `git diff --check` passed;
- scope: 4 files, 114 insertions, 4 deletions, 2 reviewed commits.

Fixes #2708

Fixes #2712

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/CliAgentProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/CliAgentProvider.java
@@ -61,18 +61,29 @@ public abstract class CliAgentProvider implements AgentProvider {
 
     @Override
     public boolean available() {
+        Process process = null;
         try {
             ProcessBuilder builder = new ProcessBuilder("sh", "-c", "command -v " + binaryName());
             processEnvironment.apply(builder, Map.of());
-            Process process = builder.redirectErrorStream(true).start();
+            process = startAvailabilityProcess(builder.redirectErrorStream(true));
             boolean finished = process.waitFor(5, TimeUnit.SECONDS);
+            if (!finished) {
+                process.destroyForcibly();
+            }
             return finished && process.exitValue() == 0;
         } catch (IOException | InterruptedException exception) {
+            if (process != null) {
+                process.destroyForcibly();
+            }
             if (exception instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
             }
             return false;
         }
+    }
+
+    protected Process startAvailabilityProcess(ProcessBuilder builder) throws IOException {
+        return builder.start();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Harden CLI provider lifecycle: fix process leaks in `available()` by ensuring `destroyForcibly()` on timeout and exception paths. Also improves QoderAgentProvider to fall back to config model when no override is given.

### Findings
- **[Info]** `CliAgentProvider.java:64-83` — The process cleanup is now correct: `destroyForcibly()` is called both when the process times out (5s) and when an exception occurs. The `process` variable is declared outside the try block to ensure it's accessible in the catch handler. This prevents zombie processes from accumulating.
- **[Info]** `CliAgentProvider.java:85-87` — Extracting `startAvailabilityProcess()` as a protected method is a clean way to enable test mocking without subclassing the entire availability check logic.
- **[Info]** `QoderAgentProvider.java:53-58` — Falling back to `config.getModel()` when no explicit override is provided is a sensible improvement. Previously, if `modelOverride` was empty but the user had configured a model in `LlmConfigVO`, it would be silently ignored.
- **[Warning]** `CliAgentProvider.java:70` — Consider logging a warning when `destroyForcibly()` is called on timeout, to help diagnose slow `command -v` executions in production.

LGTM. Important resource leak fix.

PR #2734, opened 2026-08-29
[ISSUE #2729] fix(cluster): ignore stale modal request results

yuluo-yx:
## Summary

- invalidate NameServer config-diff requests on close, replacement, and unmount
- apply separate ownership generations to cluster connection tests
- prevent stale results and notifications after modal close/reopen

## Testing

- Cluster page Vitest: 20 tests passed
- targeted ESLint and Prettier checks passed
- `npm run build` passed (8,022 modules)
- PR scope check: 2 files, 140 changed lines

Fixes #2729

PR #2714, opened 2026-08-28
[ISSUE #2711] fix(settings): display actual build metadata

yuluo-yx:
## Summary

- show the injected build commit and build time in About settings
- derive the copyright year from build metadata
- cover the rendered metadata and removal of the stale timestamp

## Testing

- targeted Vitest: 2 files, 2 tests passed
- targeted ESLint and Prettier checks passed
- `npm run build` passed (8,022 modules)
- PR scope check: 2 files, 57 changed lines

Fixes #2711

PR #2707, opened 2026-08-28
[ISSUE #2702] fix(auth): deduplicate login submissions

yuluo-yx:
## Problem

Two login submit events in the same React render could both call the authentication API because loading state was not yet visible to the second handler.

## Changes

- synchronously own the in-flight login request with a ref
- ignore duplicate submits while the request is pending
- release ownership after failure so a later retry remains available

## Local verification

- `npm test -- --run src/pages/login/index.test.tsx` — 1 file, 4 tests passed
- targeted ESLint passed
- targeted Prettier check passed
- `git diff --check`
- PR scope check: 55 changed lines, 2 files, 1 commit

No containers, full E2E suite, full repository build, or remote workflow was started manually.

Fixes #2702

PR #2706, opened 2026-08-28
[ISSUE #2701] fix(message): preserve commas in user properties

yuluo-yx:
## Problem

Message user-property values containing commas were split into extra malformed records, corrupting valid values such as locations and lists.

## Changes

- define one `key=value` property per line
- preserve commas and additional equals signs inside values
- update the form guidance and cover CRLF and duplicate-key behavior

## Local verification

- `npm test -- --run src/utils/messageProperties.test.ts src/pages/instance/__tests__/TopicPage.test.tsx` — 2 files, 19 tests passed
- targeted ESLint passed
- targeted Prettier check passed
- `git diff --check`
- PR scope check: 29 changed lines, 3 files, 1 commit

No containers, full E2E suite, full repository build, or remote workflow was started manually.

Fixes #2701

PR #2851, opened 2026-08-31
[ISSUE #2691][ISSUE #2729] fix(web): consolidate interaction lifecycle fixes

### LINKED ISSUE #2691: [Studio][Bug] Reset Message Explorer view state and load traces on demand

### LINKED ISSUE #2701: [Studio][Bug] Preserve commas in message user properties
## Problem

The Message send form treats commas as both property-record separators and valid text. A user property such as `location=Beijing,China` is split into `location=Beijing` plus a malformed `China` entry, corrupting the value and blocking submission.

## Reproduction

Paste the following into the message user-properties field:

```text
location=Beijing,China
```

The parser reports that `China` is not in `key=value` form.

## Expected behavior

Use an unambiguous one-property-per-line contract. Property values must preserve commas and additional equals signs, and Windows line endings and duplicate keys must remain handled predictably.

## Scope

Update the parser, its form guidance, and focused parser/page tests without changing the send API.

### LINKED ISSUE #2702: [Studio][Bug] Prevent duplicate login submissions
## Problem

The login page relies on React loading state to prevent repeated submissions. State is applied on the next render, so two submit events in the same tick can both call the authentication API, creating duplicate sessions and consuming the server-side login attempt budget twice.

## Reproduction

Enter valid credentials, keep the first login request pending, and synchronously submit the form twice. The API is invoked twice before the loading button rerenders.

## Expected behavior

The page should synchronously own one in-flight login request, ignore duplicate submits, and release ownership after success or failure so a later retry remains possible.

## Scope

Add a request-ownership guard to the login form and a deferred-promise regression test for duplicate submission and retry after failure.

### LINKED ISSUE #2711: fix(settings): display actual build metadata
## Problem

The About settings tab shows a fixed 2024 timestamp and copyright year even though the frontend already injects the real build commit and time.

## Expected behavior

Display the injected commit and build time, and derive the copyright year from that build metadata.

## Proposed scope

Replace the stale literals and add a rendering regression test for the injected values.

### LINKED ISSUE #2729: fix(cluster): ignore stale modal request results
## Problem

NameServer config-diff and cluster-connection requests update modal state without checking ownership. A result that arrives after close, reopen, or a newer request can reopen a modal or overwrite the latest result.

## Expected behavior

Only the current modal generation may publish results, loading state, or notifications; close, reopen, and unmount should invalidate older requests.

## Proposed scope

Add separate request generations for both modal flows and regressions for close and close/reopen ordering.

yuluo-yx:

Consolidate the reviewed frontend interaction and async-lifecycle fixes from #2695, #2706, #2707, #2714, and #2734 into one web correctness pull request, as requested on #2724.

The observable failures are:

- Reset and query-mode changes leave pagination/truncation state behind and can accept late query results;
- opening message content eagerly starts a trace request and repeated trace-tab visits refetch the same trace;
- comma-containing message property values are split into malformed records;
- two submit events in one render can both call the login API;
- About renders stale hard-coded metadata instead of injected build values;
- closed/reopened Cluster modals can be overwritten by stale NameServer diff or connection-test responses.

The #2691 patch was reimplemented on top of #2839 rather than resolving the old conflict mechanically. Server-backed history snapshot loads now participate in the same request-generation ownership as live queries.

## Brief changelog

- clear all message result state and invalidate in-flight work on lifecycle transitions;
- load traces only when the Trace tab opens and reuse the same message trace;
- preserve commas and additional equals signs in line-based message properties;
- synchronously own login submissions;
- render actual build commit/time metadata;
- apply request generations to Cluster modal requests without removing the newer Broker drift UI.

## Verifying this change

- targeted Vitest: 5 files, 45 tests passed;
- targeted ESLint passed;
- targeted Prettier check passed;
- Node 24.18.0 `npm run build` passed (8,025 modules);
- `git diff --check` passed;
- scope: 11 files, 414 insertions, 50 deletions, 5 commits.

Fixes #2691

Fixes #2701

Fixes #2702

Fixes #2711

Fixes #2729

### DIFF
--- a/web/src/pages/cluster/__tests__/ClusterPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClusterPage.test.tsx
@@ -21,7 +21,11 @@ import userEvent from '@testing-library/user-event';
 import type React from 'react';
 import { MemoryRouter } from 'react-router-dom';
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
-import type { ClusterInfo } from '../../../api/cluster';
+import type {
+  ClusterInfo,
+  ClusterProbeResult,
+  NameServerConfigDiffResult,
+} from '../../../api/cluster';
 import { LangProvider } from '../../../i18n/LangContext';
 
 const clusterServiceMocks = vi.hoisted(() => ({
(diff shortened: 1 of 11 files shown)


RockteMQ-AI (approved):
### Summary
Consolidation of UI interaction lifecycle fixes: request ID tracking to prevent stale responses from reopening closed modals, pagination reset on clear, lazy trace loading with cache, and in-flight request invalidation on mode changes.

### Findings
- **[Info]** `ClusterPage` — The request ID pattern (`diffRequestIdRef`, `probeRequestIdRef`) correctly prevents stale async responses from mutating state after the modal is closed. This is a clean solution to the classic race condition.
- **[Info]** `MessagePage` — Resetting pagination and truncation warning on clear prevents stale state from persisting across queries. The lazy trace loading with `traceCacheRef` avoids redundant API calls when toggling between messages.
- **[Info]** The `abortControllerRef` pattern for invalidating in-flight requests on mode change (query/byKey) is correct and prevents ghost responses from overwriting current results.
- **[Info]** Test coverage is comprehensive — tests for closed modal not reopening, stale probe results being ignored, pagination reset, and trace cache behavior.

LGTM. Well-structured lifecycle management fixes.

PR #2855, opened 2026-08-31
fix(message): treat negative trace time values as unavailable

yyqdbngt:
## Summary
- Treat negative time values from broker trace payloads as unavailable (`0`) in
  `RocketMQMessageProvider.parseLong`, which feeds trace node `timestamp`,
  `costTime` and consumer `consumeTime`.
- Added regression test `getMessageTraceShouldTreatNegativeTimeValuesAsUnavailable`.

## Why
The parser already degrades unparseable fields to `0`, but a *parseable* negative
value passed straight through. The web console's timestamp guard
(`if (!value) return '-'`) only treats `0` as unavailable, so a negative
`timestamp` rendered as a 1969/1970 date and a negative `costTime` as a negative
duration — visible garbage instead of the placeholder that unparseable fields get.
Well-formed RocketMQ trace payloads never carry negative epoch-millisecond times
or costs, so clamping them to `0` (the established "unavailable" sentinel) restores
graceful degradation.

## Testing
- `cd server && mvn -Dtest=RocketMQMessageProviderTest test`
  — Tests run: 32, Failures: 0, Errors: 0

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMessageProvider.java
@@ -846,7 +846,11 @@ public class RocketMQMessageProvider implements MessageProvider {
 
     private static long parseLong(String value) {
         try {
-            return Long.parseLong(value.trim());
+            long parsed = Long.parseLong(value.trim());
+            // Negative trace times/costs cannot occur in well-formed broker trace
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix — negative trace time values are now clamped to `0` (the established "unavailable" sentinel) so the console renders a placeholder instead of a 1969/1970 timestamp. The regression test covers the edge case well.

LGTM.

---

PR #2856, opened 2026-08-31
fix(proxy): evict expired proxy address cache entries

yyqdbngt:
## Summary
- Evict expired entries from `ProxyConsumerResolver.proxyAddressCache` on every discovery write
- Bound the cache to 128 entries, dropping the earliest-expiring entry when the bound is exceeded
- Add regression tests for expired-entry eviction and the size bound

## Why
The cache is keyed by the caller-supplied instance id, and each entry carries a 60s TTL that is only *checked* on read — expired entries were never removed from the map. On a long-running Studio instance serving many instances (or with instance ids that churn), every distinct id ever queried left a permanently stale entry behind, so the map grew without bound while holding dead proxy address lists.

## Testing
- `mvn -Dtest=ProxyConsumerResolverTest test` → Tests run: 7, Failures: 0, Errors: 0 (5 existing + 2 new regression tests)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/ProxyConsumerResolver.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/ProxyConsumerResolver.java
@@ -35,6 +35,7 @@ import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Service;
 import org.springframework.util.StringUtils;
 
+import java.util.Comparator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Well-structured cache management fix. Expired entries are now evicted on every discovery write, and the cache is bounded to 128 entries. The implementation correctly uses `removeIf` for expired entries and a sorted stream for size-bound eviction.

One minor note: the `evictStaleProxyAddresses()` method iterates the entire cache on every write — for the bounded size of 128 this is fine, but worth keeping in mind if the bound increases significantly.

LGTM.

---

PR #2858, opened 2026-08-31
fix(credential): serialize Tencent client lookup with invalidation

yyqdbngt:
## Summary
- Move the `TencentClientFactory.client()` fast-path cache read inside the same monitor that `invalidateCredential()` holds
- Add a package-private `cachedClient` seam and a regression test that suspends a lookup between the cache read and the return while a concurrent credential rotation runs

## Why
The cached-client read ran outside the invalidation lock: a thread suspended between the map read and the return could hand out a client that a concurrent credential rotation (triggered on credential update/delete by `CloudCredentialService`) had just evicted. That client still carries the replaced secret, so a request signed seconds after a rotation fails with a confusing "credential is invalid" even though the new credential is configured.

## Testing
- `mvn -Dtest=TencentClientFactoryTest test` → Tests run: 5, Failures: 0, Errors: 0 (4 existing + 1 new regression test)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactory.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentClientFactory.java
@@ -48,13 +48,20 @@ public class TencentClientFactory {
 
     public TrocketClient client(Long credentialId, String region) {
         String key = cacheKey(credentialId, region);
-        TrocketClient cached = clients.get(key);
-        if (cached != null) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Excellent concurrency fix. Moving the cache read inside the synchronized block eliminates the race between `client()` and `invalidateCredential()`. The test that suspends the lookup mid-flight to simulate the race condition is particularly well-crafted.

LGTM.

---

PR #2859, opened 2026-08-31
fix(ops): ignore unparseable alert times in filters and trigger count

yyqdbngt:
## Summary
- Extract an exported `wasTriggeredWithin()` helper in the alert rules page: a rule only counts toward the "triggered in last 24h" badge when `lastTriggered` parses to a timestamp after the cutoff
- Guard the mock-mode time filter in `listSystemAlertsPage()`: alerts whose `time` cannot be parsed are excluded when a `from`/`to` range is active, and unparseable range bounds are treated as absent
- Add regression tests for both (helper unit tests plus a service-level filter test)

## Why
`new Date(r.lastTriggered).getTime()` yields `NaN` for unparseable values, and `NaN > dayAgo` is always `false` — a rule with a malformed `lastTriggered` was silently dropped from the 24h trigger badge with no way to tell that its value was bad. The same pattern in `listSystemAlertsPage()` made time-range comparisons against `NaN` always evaluate to `false`, so alerts with unparseable `time` values slipped through `from`/`to` filters (the mock data's time-of-day-only `time` values are exactly such values), returning rows that do not match the requested window.

## Testing
- `./node_modules/.bin/vitest run src/pages/ops/__tests__/AlertsPage.test.tsx src/services/opsService.test.ts` → 34 passed
- `./node_modules/.bin/eslint src/pages/ops/alerts.tsx src/services/opsService.ts src/pages/ops/__tests__/AlertsPage.test.tsx src/services/opsService.test.ts` → 0 errors (pre-existing-style warnings only)

### DIFF
--- a/web/src/pages/ops/__tests__/AlertsPage.test.tsx
+++ b/web/src/pages/ops/__tests__/AlertsPage.test.tsx
@@ -23,7 +23,11 @@ import type { AlertRule, NativeAlertMetricInfo, PageResult } from '../../../api/
 import { LangProvider } from '../../../i18n/LangContext';
 import { LANGUAGE_STORAGE_KEY } from '../../../i18n/languagePreference';
 import { formatDateTime } from '../../../utils/format';
-import AlertsPage, { formatThresholdCondition, supportsUnavailableOperator } from '../alerts';
+import AlertsPage, {
+  formatThresholdCondition,
+  supportsUnavailableOperator,
+  wasTriggeredWithin,
+} from '../alerts';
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Clean extraction of `wasTriggeredWithin()` helper with proper null/NaN handling. The guard in `listSystemAlertsPage()` correctly excludes unparseable timestamps when a date range filter is active. Test coverage is thorough.

LGTM.

---

RockteMQ-AI (approved):

---

PR #2860, opened 2026-08-31
fix(broker): require stable identity for credentialed admin clients

yyqdbngt:
## Summary
- Make `MqAdminExtFactory.execute(namesrvAddr, rpcHook, action)` reject a non-null `RPCHook` instead of silently deriving the cache identity from `System.identityHashCode(rpcHook)`
- Callers holding a hook must use the identity-qualified overload with a stable, non-secret credential identity
- Update the pinned legacy-hook test to assert the new contract

## Why
The hook-instance identity hash is stable per object but not per credential: any caller that builds a fresh hook per request (the natural pattern — `RuntimeAdminClientResolver.resolveCredential` already constructs a new `AclClientRPCHook` per call) mints a distinct cache identity per request, so the factory creates one admin client per request and `release()` can never evict them (it only knows endpoint and credential-ref keys). All current production callers pass a `null` hook, so the trap is latent — this PR turns it into a loud 400 before it becomes unbounded growth.

## Testing
- `mvn -Dtest=MqAdminExtFactoryTest test` → Tests run: 13, Failures: 0, Errors: 0 (12 existing, 1 updated to the new contract)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
@@ -58,16 +58,21 @@ public class MqAdminExtFactory {
      * Runs an action against a started admin client bound to the given NameServer address.
      *
      * @param namesrvAddr NameServer address list, e.g. {@code host1:9876;host2:9876}
-     * @param rpcHook     optional RPC hook for authentication, may be {@code null}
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good API contract enforcement — rejecting a non-null `RPCHook` in the identity-less overload prevents silent misuse where the cache key would be derived from `System.identityHashCode(rpcHook)`. The updated test correctly asserts the new contract.

LGTM.

---

RockteMQ-AI (approved):

---

PR #2861, opened 2026-08-31
fix(broker): close pooled clients created during pool shutdown

yyqdbngt:
## Summary
- Serialize `MqClientPool` client creation/insertion with `release()`/`shutdown()` via a `lifecycleLock`
- Re-check `closed` under the lifecycle lock instead of inside the `computeIfAbsent` lambda
- Make `createPullConsumer`/`createProducer` protected (test seam, mirroring `MqAdminExtFactory.newAdmin`) and add `MqClientPoolTest` with a regression test for the race

## Why
`shutdown()` sets `closed`, scans `cache.values()`, and clears — but a concurrent request whose client was still being created (network I/O) passed the `closed` re-check *before* the slow creation and then inserted the started client *after* the scan/clear. That client was never shut down: its Netty threads and NameServer connections stayed alive past context destruction (started clients hold non-daemon threads, which can even hang JVM exit). `release()` had the same window.

## Testing
- `mvn -Dtest=MqClientPoolTest test` → Tests run: 5, Failures: 0, Errors: 0 (new test class)
- `mvn -Dtest='RuntimeAdminClientResolverTest,RocketMQAdminClientImplTest' test` → 9/9 and 39/39 pass (direct consumers of the pool)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqClientPool.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqClientPool.java
@@ -53,6 +53,9 @@ public class MqClientPool {
     private final Map<ClientKey, Object> cache = new ConcurrentHashMap<>();
     private final AtomicInteger instanceCounter = new AtomicInteger();
     private volatile boolean closed = false;
+    // Serializes client creation/insertion with release/shutdown so a client started while the
+    // pool is shutting down cannot be inserted after the shutdown scan and leak un-shut-down.
+    private final Object lifecycleLock = new Object();
 
     @FunctionalInterface
     public interface ClientAction<C, T> {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqClientPool.java
```diff
@@ -79,15 +82,23 @@ private <C, T> T execute(ClientKey cacheKey, RPCHook rpcHook,
...
-            // Re-check under the cache lock so a request that passed the initial closed check
-            // cannot create a fresh connection while the pool is shutting down.
+        C client;
+        synchronized (lifecycleLock) {
+            // Re-check under the lifecycle lock so a request that passed the initial closed check
+            // cannot create a fresh connection while the pool is shutting down, and a just-created
+            // client cannot be inserted after the shutdown scan has already run.
             if (closed) {
                 throw new BusinessException(503, "RocketMQ client pool is shutting down");
             }
```
RockteMQ-AI, line 92: The `synchronized(lifecycleLock)` block now wraps the entire client creation path (`creator.create(...)`), which involves network I/O (NameServer lookup, broker connect). During creation, all other `execute()` calls — including cache hits for existing clients — will block on this lock.

For a dashboard with a handful of concurrent users this is acceptable, but consider a two-phase approach if contention becomes an issue:
1. Create the client outside the lock
2. Acquire `lifecycleLock`, re-check `closed`, then insert into cache

This would preserve the correctness fix while allowing concurrent reads of already-cached clients during slow creation.

RockteMQ-AI (approved):
### Summary
Correct fix for a real TOCTOU race in MqClientPool. The original code allowed a client being created (slow network I/O) to pass the closed check, then get inserted into the cache after shutdown() had already scanned and cleared it — leaking Netty threads and NameServer connections.

### Findings
- **[Warning]** MqClientPool.java:92 — Lock scope includes client creation I/O (acceptable for dashboard use, but worth noting)

### What I Liked
- The race condition analysis in the PR description is precise and well-explained
- The BlockingPool regression test is excellent — it uses latches to deterministically reproduce the exact timing window that caused the leak
- Making createPullConsumer/createProducer protected for testability follows the existing MqAdminExtFactory.newAdmin pattern
- The lifecycleLock approach is simple and correct — no risk of deadlock since it is a single non-reentrant lock with no nested acquisitions

### Cross-repo Note
No cross-repo impact — this is a dashboard-internal fix.

Approved. The lock-scope concern is informational for the current use case.

---

RockteMQ-AI (approved):

---

PR #2862, opened 2026-08-31
fix(web): harden client connection search normalization and field guards

yyqdbngt:
## Summary

- Extract the client connection search matching in `pages/cluster/clients.tsx` into an exported, unit-testable `matchesClientSearch` helper.
- Normalize the search input with `trim()` + `toLowerCase()` so whitespace-only input no longer empties the table (previously an untrimmed `""`/`"  "` search behaved inconsistently, and `toLowerCase()` alone left padding intact).
- Guard `connection.clientId` (and `address`) against `null`/`undefined` payloads from the backend, which previously could throw `TypeError: Cannot read properties of undefined (reading 'toLowerCase')` and blank out the whole page.
- Treat an empty / whitespace-only search as a full match, matching the behavior already used by the certs page search.

## Why

The clients page search box fed raw user input directly into `toLowerCase()` and called `.toLowerCase()` unconditionally on backend-provided connection fields. In the same file `connection.address?.` is already defensively option-chained, showing the payload shape is not guaranteed to match the TS interface, yet `clientId` was not protected the same way.

## Testing

- `cd web && ./node_modules/.bin/vitest run src/pages/cluster/__tests__/clientsSearch.test.ts src/pages/cluster/__tests__/ClientsPage.test.tsx` — 2 files, 20 tests passed (5 new helper regression tests: case-insensitive id/address match, whitespace trimming, empty/undefined search, missing clientId/address fields).
- `cd web && ./node_modules/.bin/tsc --noEmit` — clean.
- `cd web && ./node_modules/.bin/eslint src/pages/cluster/clients.tsx src/pages/cluster/__tests__/clientsSearch.test.ts` — 0 errors (1 pre-existing-style react-refresh warning for the exported helper, same precedent as `message.tsx`/`LiteTopic.tsx` page-level helper exports).

### DIFF
--- a/web/src/pages/cluster/clients.tsx
+++ b/web/src/pages/cluster/clients.tsx
@@ -97,6 +97,31 @@ const countBy = (values: string[]) =>
     .map(([label, count]) => ({ label, count }))
     .sort((a, b) => b.count - a.count || a.label.localeCompare(b.label));
 
+/**
+ * Connection fields matched by the client search box. Typed as nullable to
+ * mirror what the backend may actually return for legacy payloads.
+ */
+export type ClientSearchFields = {
+  clientId?: string | null;
+  address?: string | null;
+};
+
+/**
+ * Case- and whitespace-insensitive client search match that tolerates
+ * missing clientId/address fields and whitespace-only search input.
+ */
+export const matchesClientSearch = (
+  connection: ClientSearchFields,
+  rawSearch: string | null | undefined,
+) => {
+  const normalizedSearch = (rawSearch ?? '').trim().toLowerCase();
+  if (!normalizedSearch) return true;
+  return (
+    (connection.clientId ?? '').toLowerCase().includes(normalizedSearch) ||
+    (connection.address ?? '').toLowerCase().includes(normalizedSearch)
+  );
+};
+
 type ApiErrorLike = {
   message?: unknown;
   response?: {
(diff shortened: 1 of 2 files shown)


web/src/pages/cluster/clients.tsx
```diff
@@ -97,6 +97,31 @@ const countBy = (values: string[]) =>
...
+  clientId?: string | null;
+  address?: string | null;
+};
+
+/**
+ * Case- and whitespace-insensitive client search match that tolerates
+ * missing clientId/address fields and whitespace-only search input.
+ */
+export const matchesClientSearch = (
+  connection: ClientSearchFields,
```
RockteMQ-AI, line 114: Minor: `(connection.clientId ?? '').toLowerCase()` allocates a new empty string on every call when `clientId` is null. For a UI filter this is fine, but if the connection list grows large you could hoist the empty-string fallback outside the hot path. Not a blocker.

RockteMQ-AI (approved):
### Summary
Solid defensive fix. The extraction of matchesClientSearch into a pure, testable helper is clean, and the null guards correctly address the TypeError that could blank the page when the backend returns incomplete payloads.

### Findings
- **[Info]** clients.tsx:114 — Minor allocation note (non-blocking)

### What I Liked
- The ClientSearchFields type explicitly documents the nullable contract from the backend
- 5 regression tests covering the exact failure modes (null fields, whitespace-only input, case insensitivity)
- Consistent with the pattern already used on the certs page

Looks good to merge.

---

RockteMQ-AI (approved):

---

Aias00:
Superseded by #2989, which consolidates this behavior on the current `rocketmq-studio` baseline. The replacement PR is Ready and currently CLEAN/MERGEABLE.

PR #2872, opened 2026-08-31
fix(web): harden producer group AutoComplete option filtering

yyqdbngt:
## Summary

- Replace the fragile inline AutoComplete `filterOption` in the studio Producer page (`pages/studio/Producer.tsx`) with an exported `matchesProducerGroupOption` helper.
- The old expression was `option?.value.toLowerCase().includes(inputValue.toLowerCase()) ?? false`. The `?? false` only protects the case where `option` itself is nullish (the optional chain short-circuits); it does NOT protect the cases that actually throw:
  - `option.value` is `null`/`undefined` (e.g. a `null` group in the API response) → `TypeError: Cannot read properties of null (reading 'toLowerCase')` inside the dropdown render;
  - a non-string `option.value` (number, etc.) → same `TypeError`.
- The new helper coerces both sides with `String(...) ?? ''` so any option value or search string is compared safely, case-insensitively.

## Why

The producer-group selector builds its options from live API data (`fetchProducerGroups`), unlike the static topic/instance selectors. A single malformed payload entry therefore used to throw inside antd's filter path and break the whole suggestion dropdown. The same defensive `String(option?.label ?? '')` pattern is already used in `components/InstanceSelect.tsx`; this brings the Producer selector in line with it.

## Testing

- `cd web && ./node_modules/.bin/vitest run src/pages/studio/__tests__/producerGroupFilter.test.ts` — 5 new regression tests passed (case-insensitive match, empty input, null/undefined/non-string option values, undefined input).
- `cd web && ./node_modules/.bin/vitest run src/pages/studio/__tests__/Producer.test.tsx` — 13 pre-existing tests still pass.
- `cd web && ./node_modules/.bin/tsc --noEmit` — clean.
- `cd web && ./node_modules/.bin/eslint src/pages/studio/Producer.tsx src/pages/studio/__tests__/producerGroupFilter.test.ts` — 0 errors (1 react-refresh warning for the exported helper, same precedent as other page-level helper exports).

### DIFF
--- a/web/src/pages/studio/Producer.tsx
+++ b/web/src/pages/studio/Producer.tsx
@@ -72,6 +72,21 @@ const PRODUCER_CONNECTION_EXPORT_COLUMNS: CsvColumn<ProducerConnectionExportRow>
 
 const PRODUCER_GROUP_SELECTOR_LIMIT = 20;
 
+/**
+ * Case-insensitive AutoComplete filter for producer-group suggestions.
+ *
+ * Both sides are coerced with String() and null-guaranteed so that a
+ * non-string option value (e.g. a `null` group returned by the API) or a
+ * missing search string cannot throw and blank out the dropdown.
+ */
+export const matchesProducerGroupOption = (
+  inputValue: string | null | undefined,
+  option?: { value?: unknown } | null,
+) =>
+  String(option?.value ?? '')
+    .toLowerCase()
+    .includes(String(inputValue ?? '').toLowerCase());
+
 const ProducerPage = () => {
   const [form] = Form.useForm();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean defensive fix — the old `?? false` only guarded against a nullish `option`, not a null/non-string `option.value`. The new helper correctly coerces both sides with `String(...)` and covers the edge cases with regression tests.

LGTM.

---

RockteMQ-AI (approved):
### Summary


Defensive fix that improves input validation and error handling. Code looks clean and follows existing patterns.

**LGTM**

---

RockteMQ-AI (approved):

---

PR #2876, opened 2026-09-01
fix(login): trim usernames and tolerate missing login payloads

yyqdbngt:
## Summary
- Trim surrounding whitespace from the username before calling the login API (password is left untouched on purpose)
- Reject whitespace-only usernames with the existing required-field validation (`whitespace: true`)
- Tolerate a missing `user` object in the login payload: instead of a raw `TypeError` from `data.user.username`, the user sees the standard "login failed" message
- Add regression tests for trimming, whitespace-only input, and a payload without a user

## Why
Users commonly copy usernames from docs or spreadsheets, which drags along leading/trailing spaces; those were sent verbatim and failed with a cryptic backend 401 even though the account exists. Separately, `login()` returns `res.data.data`, which is `null` whenever the backend wraps the response in an unexpected envelope — `data.user.username` then threw `Cannot read properties of null`, and that raw TypeError leaked into the error toast.

## Testing
- `./node_modules/.bin/vitest run src/pages/login/index.test.tsx` → 6 passed (3 new)
- `./node_modules/.bin/eslint src/pages/login/index.tsx src/pages/login/index.test.tsx` → 0 errors

### DIFF
--- a/web/src/pages/login/index.tsx
+++ b/web/src/pages/login/index.tsx
@@ -50,8 +50,12 @@ const LoginPage = () => {
   const onFinish = async (values: LoginFormValues) => {
     setLoading(true);
     try {
-      const data = await loginApi(values.username, values.password);
-      authLogin(data.user.username, data.user.userId, data.user.admin);
+      const data = await loginApi(values.username.trim(), values.password);
+      const user = data?.user;
+      if (!user?.username) {
+        throw new Error(t('login.failed'));
+      }
+      authLogin(user.username, user.userId, user.admin);
       message.success(t('login.success'));
       navigate('/', { replace: true });
     } catch (err: unknown) {
@@ -116,7 +120,7 @@ const LoginPage = () => {
             <Form.Item
               label={t('login.username')}
               name="username"
-              rules={[{ required: true, message: t('login.usernameRequired') }]}
+              rules={[{ required: true, whitespace: true, message: t('login.usernameRequired') }]}
             >
               <Input
                 prefix={<UserOutlined />}
--- a/web/src/pages/login/index.test.tsx
+++ b/web/src/pages/login/index.test.tsx
@@ -92,4 +92,51 @@ describe('LoginPage', () => {
     expect(screen.getByText('login.passwordRequired')).toBeTruthy();
     expect(loginApiMock).not.toHaveBeenCalled();
   });
+
+  it('trims surrounding whitespace from the submitted username', async () => {
+    loginApiMock.mockResolvedValue({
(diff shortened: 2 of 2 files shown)


web/src/pages/login/index.tsx
```diff
@@ -50,8 +50,12 @@ const LoginPage = () => {
   const onFinish = async (values: LoginFormValues) => {
     setLoading(true);
     try {
-      const data = await loginApi(values.username, values.password);
-      authLogin(data.user.username, data.user.userId, data.user.admin);
+      const data = await loginApi(values.username.trim(), values.password);
```
RockteMQ-AI, line 53: **[Info]** Good defensive improvement — using optional chaining on `data?.user` and then validating `user?.username` before accessing properties prevents crashes on malformed API responses. Combined with the `whitespace: true` form rule, this makes the login flow significantly more robust.

web/src/pages/login/index.tsx
```diff
@@ -50,8 +50,12 @@ const LoginPage = () => {
...
+      const user = data?.user;
+      if (!user?.username) {
+        throw new Error(t('login.failed'));
```
RockteMQ-AI, line 56: **[Info]** Minor note: `throw new Error(t('login.failed'))` translates the message at throw time. This works correctly since the catch block displays `err.message` directly, but if the error handling ever changes (e.g., logging the raw error or re-translating), the already-localized string could cause issues. Consider using a stable error key and mapping it in the catch block — though this is a minor style preference, not a blocker.

RockteMQ-AI (approved):
### Summary

This PR adds defensive input handling to the login flow: trimming whitespace from usernames, rejecting whitespace-only input at the form validation level, and safely handling null/malformed API responses. The changes are well-tested with 3 new test cases covering the key scenarios.

### Findings
- **[Info]** web/src/pages/login/index.tsx:53 — Good use of optional chaining and null checks to prevent crashes on unexpected API responses.
- **[Info]** web/src/pages/login/index.tsx:56 — Minor style note on translating error messages at throw time (non-blocking).

### Suggestions
The overall approach is solid. The combination of `values.username.trim()` + `whitespace: true` form rule + `data?.user` optional chaining provides three layers of defense against bad input/data. Test coverage is thorough.

No correctness, performance, or compatibility concerns.

---

RockteMQ-AI (approved):
### Summary


Defensive fix that improves input validation and error handling. Code looks clean and follows existing patterns.

**LGTM**

---

RockteMQ-AI (approved):

---

PR #2879, opened 2026-09-01
fix(group): normalize reset topic before apply

### LINKED ISSUE #2877: Offset reset apply does not normalize the previewed topic
The consumer offset reset preview normalizes the group and topic names before calling the provider, but the apply path only normalizes the group name. A request whose topic contains leading or trailing whitespace can therefore preview successfully and then send a different, untrimmed topic to the provider when the reset is applied. The audit entry also records the unnormalized value.

The apply path should validate and normalize the topic with the same boundary helper used by preview, then use that value for both the provider call and audit detail.

btlqql:
## Summary

- validate and trim the reset Topic at the MetadataService boundary
- pass the same normalized Topic to the provider and audit record
- reject blank Topics before provider resolution

## Tests

- `mvn -Dmaven.repo.local=D:\Dev\m2-cache -Dtest=MetadataServiceTest test` (36 passed, Checkstyle passed)
- `git diff --check`

Closes #2877

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
@@ -329,11 +329,12 @@ public class MetadataService {
     public void resetOffset(String instanceId, String name, long timestamp, String topic) {
         instanceId = normalizeInstanceId(instanceId);
         String groupName = requireName(name, "consumer group name");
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Normalizes the reset topic input before applying consumer offset reset, preventing issues with whitespace or inconsistent formatting. Clean fix with test coverage.

**LGTM**

---

PR #2882, opened 2026-09-01
fix(message): normalize incomplete message query and DLQ responses

yyqdbngt:
## Summary
- Guard `queryMessages` and `queryMessagePage` against a missing or null `data` payload, so the store-time sort no longer throws on `[...null]` and the paged contract always returns an `items` array
- Guard `getMessageTrace` so a missing trace record normalizes to an empty node list instead of throwing on `trace.nodes`
- Normalize `listDLQGroups` and `listDLQMessages` to always return an `items` array even when the backend omits it
- Guard `getQueueOffsets` so a null payload resolves to an empty queue list

## Why
The shared client only rejects non-success business codes; a success envelope with `data: null` (empty results on some backends, partially populated pages after a broker timeout) still flows through. The message/DLQ endpoints dereference `res.data.data.items` or spread the payload directly, so a single missing field turns an empty result page into an uncaught `TypeError` in the UI instead of an empty table.

## Testing
- `cd web && ./node_modules/.bin/vitest run src/api/message.test.ts` → 12 passed (5 new regression tests: null message list, missing/null paged items, missing trace record, missing DLQ page items)
- `./node_modules/.bin/vitest run src/services/messageService.test.ts` → 7 passed
- `./node_modules/.bin/eslint src/api/message.ts src/api/message.test.ts` → clean

### DIFF
--- a/web/src/api/message.test.ts
+++ b/web/src/api/message.test.ts
@@ -21,6 +21,7 @@ import client from './client';
 import {
   consumeMessageDirectly,
   getMessageTrace,
+  listDLQGroups,
   queryMessagePage,
   queryMessages,
 } from './message';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
## Code Review Summary

**Verdict:** ✅ Approve — clean, focused defensive coding fix

### What this PR does
Adds null-safety guards to 6 API functions in `web/src/api/message.ts` that previously crashed with `TypeError` when the backend returned a success envelope with `data: null` or missing `items` arrays. This is a real problem — the client only rejects non-success business codes, so a null payload flows through to the UI.

### Changes reviewed

**`message.ts` (12 additions, 6 deletions):**
- `queryMessages`: `Array.isArray()` guard before `sortMessagesByStoreTimeDesc` — correct
- `queryMessagePage`: optional chaining `page?.items` with `Array.isArray` fallback — correct
- `getMessageTrace`: `trace?.nodes` instead of `trace.nodes` — correct, prevents crash on null trace
- `listDLQGroups`: normalizes missing `items` to `[]` — correct
- `getQueueOffsets`: `Array.isArray` guard on response data — correct
- `listDLQMessages`: same pattern as `listDLQGroups` — correct

**`message.test.ts` (42 additions):**
- 5 regression tests covering: null message list, null paged items, missing trace record, missing DLQ items
- Tests are well-structured and directly validate the fix

### Minor observation (non-blocking)
In `queryMessagePage`, when `res.data.data` is entirely `null`, the spread `...page` on null produces `{}`, so the returned object would be `{ items: [] }` without `total`/`page`/`size`. Downstream code accessing `page.total` would get `undefined` instead of a number. Consider defaulting the full page shape:

```ts
const page = res.data.data ?? { items: [], total: 0, page: 0, size: 0 };
```

Same pattern applies to `listDLQGroups` and `listDLQMessages`. This is non-blocking — the current fix already prevents the crash, and the undefined fields are unlikely to cause secondary issues in practice.

### Overall
Clean PR, good test coverage, minimal diff. The defensive coding pattern is appropriate for a dashboard that needs to handle inconsistent backend responses gracefully.

PR #2883, opened 2026-09-01
fix(alert): guard unknown notification outbox statuses

yyqdbngt:
## Summary
- parse stored outbox statuses through a guarded helper in `NotificationOutboxService.listDeliveries(alertId)`
- trim and root-locale uppercase the stored value; unknown or blank statuses surface as `FAILED`

## Why
`listDeliveries` called `NotificationOutboxStatus.valueOf(row.getStatus())` directly. A legacy or partially written row with an unrecognized status (or whitespace/case drift) made `valueOf` throw `IllegalArgumentException`, turning the whole per-alert delivery listing into a 500. Unknown values now map to `FAILED` so they are visible and need manual attention, matching how the retry endpoint already treats non-`FAILED` rows as untouched.

## Testing
- `cd server && mvn -q -Dtest=NotificationOutboxServiceTest test` — all tests pass, including the new `mapsUnknownStoredOutboxStatusesToFailedWhenListingTest` which lists a row storing `BOUNCED` (→ `FAILED`) and one storing ` pending ` (→ `PENDING`)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
@@ -156,7 +156,7 @@ public class NotificationOutboxService {
         return mapper.selectList(new QueryWrapper<RmqAlertNotificationOutbox>().eq("alert_id", alertId)
                         .orderByAsc("id"))
                 .stream().map(row -> NotificationDeliveryVO.builder().id(row.getId()).channel(row.getChannel())
-                        .status(NotificationOutboxStatus.valueOf(row.getStatus()))
+                        .status(parseStoredStatus(row.getStatus()))
                         .attemptCount(row.getAttemptCount() == null ? 0 : row.getAttemptCount())
                         .nextAttemptAt(row.getNextAttemptAt()).lastError(row.getLastError())
                         .deliveredAt(row.getDeliveredAt()).build()).toList();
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
```diff
@@ -226,6 +226,21 @@ private static String normalizeTrim(String value) {
...
 
+    /**
+     * Maps a stored status to the known outbox lifecycle. Legacy or partially written rows must
+     * not break the delivery listing, so unknown values surface as FAILED for manual attention.
+     */
+    private static NotificationOutboxStatus parseStoredStatus(String status) {
+        if (!StringUtils.hasText(status)) {
+            return NotificationOutboxStatus.FAILED;
+        }
+        try {
```
RockteMQ-AI, line 237: **[Info]** Consider adding a `log.warn("Unknown outbox status mapped to FAILED: {}", status)` inside the catch block. Silent mapping is correct behaviorally, but a warning helps operators detect data-integrity drift (e.g. a new enum value added to the DB but not to `NotificationOutboxStatus`) during troubleshooting.

RockteMQ-AI (approved):
### Summary
Clean defensive fix — replaces the raw `valueOf()` call with a guarded parser that tolerates unknown, blank, or whitespace-padded statuses by mapping them to `FAILED`. This prevents a single bad row from turning the entire per-alert delivery listing into a 500.

### Findings
- **[Info]** `NotificationOutboxService.java:237` — Consider logging a warning when an unknown status is caught, to aid data-integrity debugging.

### Verdict
The fix is correct, well-scoped, and the test coverage is solid (covers both unknown-value and whitespace/case-normalization paths). `Locale.ROOT` usage for `toUpperCase` is the right call. LGTM.

PR #2884, opened 2026-09-01
fix(ai): keep chat message ids unique across same-millisecond sends

yyqdbngt:
## Summary
- Add a `newMessageId()` helper (exported for testing) that builds message ids from a `crypto.randomUUID()`-based unique suffix, refactored out of the existing `newConversationId()` pattern
- Use it for chat user/AI message ids instead of `user-${Date.now()}` / `ai-${Date.now()}`
- Add a regression test: stop a first response, immediately send a second one with the clock frozen on the same millisecond, and verify the second stream's chunks do not leak into the stopped first response

## Why
Chat message ids were derived from `Date.now()`, so two sends landing in the same millisecond produced identical `ai-…` response ids. The stream chunk handler and the `finally`-block `pending` update both locate messages by that id with `prev.map(...)`, so an in-flight second response appended its chunks to the first (already stopped) response as well — the stopped message visibly grew with the new answer's text, and the early `pending: false` write from the first request's `finally` could also clear the second response's "thinking" spinner prematurely.

## Testing
- `./node_modules/.bin/vitest run src/pages/ai/__tests__/` → 15 passed (1 new; verified the new test fails with the pre-fix `Date.now()`-based ids)
- `./node_modules/.bin/eslint src/pages/ai/index.tsx src/pages/ai/__tests__/AiPage.test.tsx` → 0 errors

### DIFF
--- a/web/src/pages/ai/__tests__/AiPage.test.tsx
+++ b/web/src/pages/ai/__tests__/AiPage.test.tsx
@@ -453,4 +453,47 @@ describe('AiPage tool runner', () => {
     expect(await screen.findByText('工具参数必须是有效的 JSON 对象')).toBeInTheDocument();
     expect(executeTool).not.toHaveBeenCalled();
   });
+
+  it('keeps stream chunks out of a previous stopped response created in the same tick', async () => {
+    const nowSpy = vi.spyOn(Date, 'now').mockReturnValue(1000);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
Clean refactoring with a solid test that reproduces the original bug.

---

PR #2885, opened 2026-09-01
fix(alert): tolerate unknown alert state statuses

yyqdbngt:
## Summary
- parse stored alert-state statuses through a guarded helper in `MybatisPlusAlertStateRepository`
- trim and root-locale uppercase the value; unknown or blank statuses fall back to `OK` in both `find` and the runtime listing

## Why
`toState` is on the hot path of `NativeAlertProcessor.process`: every sample, for every rule, loads the persisted state and calls `AlertStateStatus.valueOf(entity.getStatus())`. One legacy or corrupted status value in `rmq_alert_state` threw `IllegalArgumentException` inside the collection transaction, aborting the whole alert evaluation cycle for every rule and fingerprint. The runtime listing (`/api/alert-rules/runtime`) had the same unguarded parse. Unknown values now fall back to the idle `OK` state so the next sample re-advances the state machine from a safe baseline.

## Testing
- `cd server && mvn -q -Dtest=MybatisPlusAlertStateRepositoryTest test` — all tests pass, including the new `findToleratesUnknownStoredStatusValuesTest` (stored `FIRED` → `OK`) and `runtimeListingToleratesUnknownStoredStatusValuesTest` (stored `FIRED` → `OK`, stored ` firing ` → `FIRING`)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertStateRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertStateRepository.java
@@ -28,6 +28,7 @@ import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.util.Optional;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Defensive fix that tolerates unknown or legacy `AlertStateStatus` values in the database. The new `parseStatus()` helper prevents `IllegalArgumentException` from aborting the alert collection cycle.

**Changes:**
- Extracted `parseStatus()` with null/blank handling and `Locale.ROOT` normalization
- Unknown values fall back to `AlertStateStatus.OK` (idle state, allowing next sample to re-advance)
- Applied to both `toState()` (hot path in `NativeAlertProcessor.process`) and `toRuntime()`
- Test coverage for null, blank, unknown, and lowercase values

Good defensive coding. The fallback to `OK` is the right choice — it resets the state machine to idle so the next sample can properly advance it.

---

PR #2886, opened 2026-09-01
fix(alert): tolerate malformed silence domain values

yyqdbngt:
## Summary
- parse stored silence domains through a guarded helper in `MybatisPlusAlertSilenceRepository.toVo`
- trim and root-locale uppercase the value; blank domains stay unscoped, unknown values fall back to `BUSINESS`

## Why
`toVo` called `AlertDomain.valueOf(entity.getDomain())` directly on the raw database value. The write path always stores the canonical enum name, but a legacy row (for example a lowercase `business`) or a manually edited value made `valueOf` throw `IllegalArgumentException` inside `findAll()`. That single row broke both the silence listing endpoint and `AlertSilenceService.activeUntil`, which reads every silence on every notification enqueue and dispatch — so one bad row silently disabled silence matching for all alerts. The fallback follows the existing convention in `MybatisPlusAlertRepository.parseDomain`, which maps unrecognized rule domains to `BUSINESS`.

## Testing
- `cd server && mvn -q -Dtest=MybatisPlusAlertSilenceRepositoryTest,AlertSilenceServiceTest test` — all tests pass, including the new `MybatisPlusAlertSilenceRepositoryTest` covering stored ` business ` → `BUSINESS`, `LEGACY` → `BUSINESS`, `CLUSTER` → `CLUSTER`, and `null` → unscoped

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertSilenceRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertSilenceRepository.java
@@ -25,6 +25,7 @@ import org.apache.rocketmq.studio.persistence.mapper.RmqAlertSilenceMapper;
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.TreeMap;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Defensive fix that tolerates malformed or legacy `AlertDomain` values stored in the database. The new `parseDomain()` helper prevents `IllegalArgumentException` from breaking silence matching.

**Changes:**
- Extracted `parseDomain()` with null/blank handling and `Locale.ROOT` normalization
- Unknown values fall back to `AlertDomain.BUSINESS` (consistent with rule parsing)
- Comprehensive test coverage for null, blank, lowercase, and unknown values

Good defensive coding pattern. The fallback to `BUSINESS` is sensible since it matches the behavior for rules without a recognized domain.

---

PR #2887, opened 2026-09-01
fix(ai): treat non-array tool catalogs as empty

yyqdbngt:
## Summary
- In the AI tool runner's `loadTools()`, treat a non-array tool catalog payload as an empty catalog (`Array.isArray` guard) before setting state and auto-selecting the first tool
- Add a regression test: a `null` catalog payload leaves the tool modal open with an empty catalog instead of surfacing the "failed to load" error path

## Why
`listTools()` returns `res.data.data`, so the wire can deliver anything (e.g. `null` or an object) even though the TypeScript type says `McpTool[]`. The page dereferenced the payload with `availableTools.find(...)`: a `null` catalog threw a `TypeError` that fell into the `catch` branch, showing "AI 工具目录加载失败" (catalog load failed) even though the request succeeded, and the tool modal was left in the error state for no recoverable reason.

## Testing
- `./node_modules/.bin/vitest run src/pages/ai/__tests__/AiPage.test.tsx` → 14 passed (1 new; verified it fails with the pre-fix payload handling)
- `./node_modules/.bin/eslint src/pages/ai/index.tsx src/pages/ai/__tests__/AiPage.test.tsx` → 0 errors

### DIFF
--- a/web/src/pages/ai/__tests__/AiPage.test.tsx
+++ b/web/src/pages/ai/__tests__/AiPage.test.tsx
@@ -21,7 +21,7 @@ import userEvent from '@testing-library/user-event';
 import { App } from 'antd';
 import { MemoryRouter } from 'react-router-dom';
 import { LangProvider } from '../../../i18n/LangContext';
-import { chatStream, executeTool, listTools } from '../../../api/ai';
+import { chatStream, executeTool, listTools, type McpTool } from '../../../api/ai';
 import { listClusters, type ClusterInfo } from '../../../api/cluster';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Defensive fix that guards against non-array tool catalog payloads from the API. The `Array.isArray` check prevents runtime errors when the wire format doesn't match the TypeScript type.

**Changes:**
- Added `Array.isArray(catalog)` guard in `loadTools()`
- Regression test verifies null catalog doesn't trigger error state

Clean, minimal change with good test coverage.

---

PR #2898, opened 2026-09-01
fix(ai): tolerate malformed tool names in the access policy

yyqdbngt:
## Summary

The AI tool execution path is user-controlled and matched by prefix + suffix in the auth interceptor (`ToolAccessPolicy`), but the tool-name slice between them was not defensively bounded:

- `POST /api/ai/tools/execute` (blank tool name) made `substring(start, end)` run past its start → raw `StringIndexOutOfBoundsException` 500.
- A malformed percent-encoding (`%2z`) or invalid UTF-8 (`%FF`) made `URLDecoder.decode` throw `IllegalArgumentException` → raw 500 from the interceptor.

Both bypassed the intended fail-closed semantics. Unresolvable tool names are now treated as unknown, so readers still require admin (403) instead of hitting a 500.

## Why

Tool names come from AI agents, MCP clients and hand-typed URLs alike. The interceptor runs on every POST under `/api/ai/tools/*/execute`, so one mistyped name turns a clean 404/403 into an opaque 500 and a stack trace.

## Testing

New `ToolAccessPolicyTest` (10 tests) covering: path recognition, blank tool name, percent-decoding, malformed percent sequence, invalid UTF-8, deny list, unknown tool, and admin/reader authorization.

```
mvn -q -Dtest=ToolAccessPolicyTest test
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ToolAccessPolicy.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ToolAccessPolicy.java
@@ -74,13 +74,26 @@ public class ToolAccessPolicy {
         if (!isToolExecutionPath(requestPath)) {
             return Optional.empty();
         }
-        String encodedName = requestPath.substring(
-                TOOL_EXECUTION_PREFIX.length(),
-                requestPath.length() - TOOL_EXECUTION_SUFFIX.length());
+        int start = TOOL_EXECUTION_PREFIX.length();
+        int end = requestPath.length() - TOOL_EXECUTION_SUFFIX.length();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Hardens `ToolAccessPolicy` against malformed tool names by catching `IllegalArgumentException` from URL decoding and guarding against empty name ranges. Good security fix — prevents raw 500s from the auth interceptor when encountering malformed paths.

### Findings
- **[Info]** `ToolAccessPolicy.java:80` — The `end <= start` guard correctly catches the edge case where prefix + suffix consume the entire path.

### Suggestions
No blocking issues. The fail-closed behavior (returning empty → falling through to admin check) is the correct security posture.

---

Issue #2901, opened 2026-09-01
Visible polling overlaps when the poll callback changes during an active request


yuluo-yx:
### Problem\nuseVisiblePolling stores its in-flight flag inside the effect. If a component rerenders with a new poll callback while the old promise is unresolved, the effect is recreated with a fresh flag. An interval or visibility event can then start another poll before the original completes.\n\n### Reproduction\n1. Enable polling and let the first poll return a pending promise.\n2. Rerender the hook with a new callback.\n3. Fire a visibility event or advance the interval.\n4. Observe the new callback running while the first promise is still pending.\n\n### Expected behavior\nThe no-overlap guarantee survives effect recreation; polling resumes with the latest callback only after the active promise settles.\n\n### Proposed fix\nKeep the active promise in a hook-level ref and clear it with identity checking, plus a rerender regression test.

RockteMQ-AI:
**Issue Evaluation**

Category:  | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** The  hook stores its in-flight flag () inside the  closure. When the component rerenders with a new poll callback , the effect dependencies  trigger a cleanup and recreation. The new effect closure starts with a fresh , while the previous promise from the old closure may still be pending. This allows the interval or visibility handler to fire a concurrent poll before the original completes.

**Impact:** Overlapping poll requests can cause race conditions, duplicate API calls, and inconsistent UI state in the BrokerCluster view.

**Severity:** medium

**Suggested fix:** Lift the in-flight flag out of the effect closure using a  so it persists across effect re-creations:


An automated fix proposal will be generated. Reply  to proceed with PR generation.

---
*Automated evaluation by RockteMQ-AI*

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** The `useVisiblePolling` hook stores its in-flight flag (`e`) inside the `useEffect` closure. When the component rerenders with a new poll callback `r`, the effect dependencies `[t, n, r]` trigger a cleanup and recreation. The new effect closure starts with a fresh `e = false`, while the previous promise from the old closure may still be pending. This allows the interval or visibility handler to fire a concurrent poll before the original completes.

**Impact:** Overlapping poll requests can cause race conditions, duplicate API calls, and inconsistent UI state in the BrokerCluster view.

**Severity:** medium

[...]

PR #2893, opened 2026-09-01
[ISSUE #2891] Deduplicate notification delivery retries

yuluo-yx:
## What changed\n\n- track individual retry IDs synchronously\n- coordinate bulk and individual retry actions so the same delivery cannot overlap\n- add a deferred-promise regression for duplicate retry events\n\n## Why\n\nReact state alone allowed two retry events to queue the same failed notification before the loading state was committed.\n\n## Verification\n\n- npm test -- --run src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx\n- Prettier and ESLint on the two changed files\n- scope check: 2 files, 35 changed lines\n\nFixes #2891

PR #2906, opened 2026-09-01
[ISSUE #2901] Preserve polling request ownership across rerenders

yuluo-yx:
## Summary\n- store the active polling promise in a hook-level ref instead of an effect-local boolean\n- retain the no-overlap guarantee when the callback or interval recreates the effect\n- clear the guard only from the promise that currently owns it\n\n## Verification\n- Rerender regression test failed before the fix because the replacement callback overlapped the pending request\n- npm test -- --run src/hooks/useVisiblePolling.test.tsx (5 passed)\n- Prettier check passed\n- ESLint passed\n- Scope: 2 files, 49 changed lines\n\nFixes #2901

### DIFF
--- a/web/src/hooks/useVisiblePolling.test.tsx
+++ b/web/src/hooks/useVisiblePolling.test.tsx
@@ -128,4 +128,38 @@ describe('useVisiblePolling', () => {
     });
     expect(poll).toHaveBeenCalledTimes(1);
   });
+
+  it('keeps the in-flight guard when the poll callback changes', async () => {
+    const first = deferred();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a subtle React bug where the pollInFlight guard was captured in a closure and lost on re-render, causing duplicate concurrent requests. Moving it to useRef preserves the mutable flag across renders while the callback identity changes. Clean fix with a test that verifies the guard survives callback recreation.

---

lizhimins:
Consolidated into #2893 (merged into rocketmq-studio) per the maintainer batch integration — no separate merge needed.

PR #2907, opened 2026-09-01
[ISSUE #2902] Run NameServer registry probes concurrently

yuluo-yx:
## Summary\n- materialize all bounded probe futures before joining results\n- retain input-order aggregation, per-probe timeout, interruption, and queue saturation behavior\n- add a latch-based test that proves independent probes start concurrently\n\n## Verification\n- Concurrency regression test failed before the fix because only the first probe started\n- mvn -q -Dtest=RegistryProbeRunnerTest,ClusterServiceRegistryTest test passed on Temurin 21\n- Scope: 2 files, 48 changed lines\n\nFixes #2902

PR #2908, opened 2026-09-01
[ISSUE #2903] Stabilize broker address fallback

### LINKED ISSUE #2903: Broker address fallback depends on HashMap iteration order
## Problem

Both Apache cluster mapping paths prefer broker ID 0, then fall back to the first nonblank address from a HashMap. When the master is unavailable, HashMap iteration order is not broker priority, so the selected address can vary with the broker IDs and map layout.

## Reproduction

Build masterless broker maps containing broker IDs 32 and 1. Both providers select broker ID 32 on the current implementation even though broker ID 1 is the lowest available replica.

## Expected behavior

Keep preferring broker ID 0. If it is absent or blank, choose the lowest broker ID with a usable address in both Apache cluster providers.

## Scope

Update the two Apache cluster mapping implementations and add focused regression coverage for each.

yuluo-yx:
## Summary
- continue preferring broker ID 0 when the master is available
- use the lowest broker ID with a nonblank address when the master is absent
- apply the same deterministic policy to both Apache cluster mapping implementations

## Verification
- both focused regressions failed before the fix by selecting broker ID 32 ahead of broker ID 1
- mvn -Dtest=RocketMQClusterProviderTest,RealClusterProviderTest test passed on Temurin 21
- 25 tests passed with zero failures; Checkstyle reported zero violations
- Scope: 4 files, 54 changed lines

Fixes #2903

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/RealClusterProvider.java
@@ -176,9 +176,10 @@ public class RealClusterProvider implements ClusterProvider {
         if (data.getBrokerAddrs() != null) {
             addr = data.getBrokerAddrs().get(MixAll.MASTER_ID);
             if (addr == null) {
-                addr = data.getBrokerAddrs().values().stream()
-                        .filter(Objects::nonNull)
-                        .findFirst()
+                addr = data.getBrokerAddrs().entrySet().stream()
+                        .filter(entry -> entry.getKey() != null && entry.getValue() != null)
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Improves broker address fallback logic in both RealClusterProvider and RocketMQClusterProvider. Instead of picking an arbitrary first non-null address, it now selects the entry with the lowest broker ID via min(comparingByKey). This gives deterministic, predictable fallback behavior. Well-tested with both unit tests.

---

lizhimins:
Consolidated into #2907 (merged into rocketmq-studio) per the maintainer batch integration — no separate merge needed.

PR #2909, opened 2026-09-01
fix(ai): project empty message trace timelines

yyqdbngt:
## Summary

`MessageTraceToolHandler` projects the trace timeline into the tool output, but assumed `MessageService.getMessageTrace` always returns a record with populated `nodes` / `consumerStatus` lists. Providers may report a missing trace as `null` (the service's own `recordTraceQuery` already guards for exactly that case) or as a record whose lists are absent, and the handler turned both shapes into an NPE 500 for AI/MCP/CLI callers.

The projection now treats a missing trace or missing lists as an empty timeline. The tool output schema only requires the arrays to be present, so empty arrays stay schema-valid.

## Why

The same trace lookup is also served to the web UI, where a missing trace renders as an empty timeline. MCP/CLI callers of `rmq.message.trace` should get the same "no trace found" answer instead of an internal error.

## Testing

Extended `MessageTraceToolHandlerTest` with two regression cases: null trace record, and record with null node/consumer-status lists.

```
mvn -q -Dtest=MessageTraceToolHandlerTest test
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/MessageTraceToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/MessageTraceToolHandler.java
@@ -24,6 +24,7 @@ import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Component;
 
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes NPE in MessageTraceToolHandler when the trace record or its node/consumer-status lists are null. Now projects empty timelines instead of crashing with a 500. Good defensive coding with two test cases covering both null-trace and null-lists scenarios.

---

PR #2919, opened 2026-09-01
[ISSUE #2914] Prevent stale resource recreation

### LINKED ISSUE #2914: Updates can recreate concurrently deleted persisted resources
The K8s certificate and cloud credential repositories first check whether a non-null id still exists, then insert when it does not. An update request racing with deletion can therefore insert a new row using the stale id and recreate a resource that was already deleted.

Reproduction:
1. Load an existing certificate or cloud credential for editing.
2. Delete it in another request.
3. Submit the stale update carrying its original id.

Actual: the repository follows the insert path and can recreate the deleted row.
Expected: a non-null id is always update-only; a zero-row update returns a conflict and never inserts.

yuluo-yx:
Closes #2914

Summary:
- make non-null certificate ids update-only
- apply the same update-only contract to cloud credentials
- return the existing conflict instead of inserting after a concurrent delete

Verification:
- mise exec java@temurin-21 -- mvn -Dtest=MybatisPlusK8sCertRepositoryTest,MybatisPlusCloudCredentialRepositoryTest test
- 13 tests, 0 failures; Checkstyle 0 violations

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/audit/OperationAuditService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/audit/OperationAuditService.java
@@ -48,7 +48,14 @@ public class OperationAuditService {
         LocalDateTime now = LocalDateTime.now();
         audit.setGmtCreate(now);
         audit.setGmtModified(now);
-        auditMapper.insert(audit);
-        log.debug("Audit recorded: {} {} {}", operation, resourceType, resourceName);
+        try {
+            auditMapper.insert(audit);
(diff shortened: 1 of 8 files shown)


RockteMQ-AI (approved):
### Summary
Removes the redundant `selectById` guard before `updateById` in both K8s cert and cloud credential repositories. The previous check-then-update pattern had a TOCTOU race: an entity deleted between the select and the update would silently fall through to an insert of stale data. Now the update is attempted directly, and a zero-row result correctly throws 409. Good test coverage for the concurrent-delete scenario.

---

PR #2920, opened 2026-09-01
[ISSUE #2915] Preserve broker diff failures

### LINKED ISSUE #2915: Broker config diff omits per-broker failure details
BrokerConfigDiffService catches broker read failures and marks the broker unreachable, but it drops the BusinessException message. The response model and cluster page both provide a message field, so operators only see Unreadable and cannot tell whether the cause was timeout, authentication, or another broker error.

Reproduction:
1. Compare a cluster where one broker config read throws BusinessException with a diagnostic message.
2. Inspect the failed broker entry in the response.

Actual: message is null.
Expected: retain the broker read failure message in the corresponding status entry.

yuluo-yx:
Closes #2915

Summary:
- retain each failed broker config read message in the diff response
- populate the existing message field already rendered by the cluster page
- update partial-result regression coverage

Verification:
- mise exec java@temurin-21 -- mvn -Dtest=BrokerConfigDiffServiceTest test
- 7 tests, 0 failures; Checkstyle 0 violations

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/BrokerConfigDiffService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/BrokerConfigDiffService.java
@@ -76,6 +76,7 @@ public class BrokerConfigDiffService {
                         .name(broker.name())
                         .address(broker.address())
                         .reachable(false)
+                        .message(exception.getMessage())
                         .build());
             }
         }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds the exception message to unreachable broker status entries so diff reports preserve the failure reason instead of silently dropping it. Clean one-line fix with matching test update.

---

PR #2921, opened 2026-09-01
fix(metrics): survive malformed snapshot retention config in cleanup

yyqdbngt:
## Summary
- Parse `studio.alerting.snapshot-retention` defensively in `CollectorScheduler.cleanUpSnapshots()`, reusing the existing `parsePositiveDuration` helper (now parameterized with a setting name for accurate log messages)
- A malformed or non-positive retention value falls back to the 24h default retention with a warning instead of throwing out of the scheduled task

## Why
`cleanUpSnapshots()` ran `Duration.parse(properties.getSnapshotRetention())` without any guard, while its two sibling settings (`studio.alerting.collection-timeout` and `studio.alerting.collection-lease-duration`) already use defensive parsing. A typo in the retention value (e.g. `1h` instead of `PT1H`) made the hourly cleanup task throw a `DateTimeParseException` on every run, so the short-lived `rmq_metric_snapshot` table grew unbounded until the configuration was fixed.

## Testing
- `mvn -Dtest=CollectorSchedulerTest test` → Tests run: 7, Failures: 0, Errors: 0 (1 new: a malformed retention value no longer throws and still deletes expired snapshots using the fallback)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/CollectorScheduler.java
@@ -120,10 +120,9 @@ public class CollectorScheduler {
 
     @Scheduled(fixedDelayString = "${studio.alerting.snapshot-cleanup-interval:PT1H}")
     public void cleanUpSnapshots() {
-        Duration retention = Duration.parse(properties.getSnapshotRetention());
-        if (retention.isNegative() || retention.isZero()) {
-            return;
-        }
+        // A malformed or non-positive retention value must not fail the scheduled cleanup; fall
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good fix that makes the snapshot cleanup more resilient to configuration errors.

### Observations
- **Correctness** ✓ — Falling back to a default 1-day retention instead of silently skipping cleanup is the right behavior — prevents unbounded snapshot table growth.
- **Refactoring** — Extracting `parsePositiveDuration(name, value, fallback)` with a setting name parameter improves log clarity.
- **Tests** ✓ — The `survivesMalformedSnapshotRetentionTest` correctly verifies the fallback behavior.

---

PR #2922, opened 2026-09-01
fix(metrics): skip unmaterializable snapshot rows in recent reads

yyqdbngt:
## Summary
- `MybatisPlusMetricSnapshotRepository.toSample()` now returns null (with a warning log) when a persisted snapshot row cannot be materialized: unknown `domain`/`availability` enum values, a NULL value on an available row, or a missing collected timestamp
- `findRecent()` filters out those rows instead of letting the `MetricSample` constructor throw
- Adds a regression test with a valid row mixed with a NULL-value row and an unknown-domain row

## Why
`rmq_metric_snapshot` allows `value` to be NULL, and its `domain`/`availability` columns are plain strings. A structurally invalid row (a legacy enum value from an older Studio version, a NULL value on an AVAILABLE row, or a direct DB edit) made `toSample()` throw an `IllegalArgumentException`/`NullPointerException` that was not caught (only JSON errors were), so one bad record broke the entire recent-samples window read by alert aggregation for that scope — every alert evaluation in the scope failed until the 24h retention cleanup removed the row.

## Testing
- `mvn -Dtest=MybatisPlusMetricSnapshotRepositoryTest test` → Tests run: 2, Failures: 0, Errors: 0 (1 new; verified it errors with the pre-fix materialization)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MybatisPlusMetricSnapshotRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MybatisPlusMetricSnapshotRepository.java
@@ -21,6 +21,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.rocketmq.studio.ops.alert.AlertDomain;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good null-safety fix for snapshot queries when `startedAt` is not yet materialized.

### Observations
- **Correctness** ✓ — Filtering out snapshots with null `startedAt` before sorting prevents NPE in the comparator.
- **Tests** ✓ — Coverage includes the mixed null/non-null scenario.
- **Design** — The null-safe comparator approach is clean and doesn't require schema changes.

---

PR #2923, opened 2026-09-01
fix(proxy): strip IPv6 brackets when probing proxy topology

yyqdbngt:
## Summary

Proxy addresses are stored bracketed for IPv6 (`[::1]:8081`) — that format is validated on add and shown to users — but `buildTopology()` passed the bracketed host straight to the health probe. `InetSocketAddress` treats a bracketed IPv6 literal as a *hostname*, so the lookup always failed with an unknown-host error and every registered IPv6 proxy was reported as DOWN even while healthy.

The topology build now strips the brackets before probing, so IPv6 proxies get the same TCP treatment as IPv4 and hostname entries.

## Why

`addProxyAddr` explicitly accepts and validates bracketed IPv6 addresses, but the topology view could never show one as UP — a live IPv6 proxy looked dead in the console.

## Testing

Extended `ProxyAddressServiceTest`: a new case registers `[::1]:9001` with a capturing probe and asserts the probe receives the bare literal `::1` and the node reports UP.

```
mvn -q -Dtest=ProxyAddressServiceTest test
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
@@ -175,7 +175,7 @@ public class ProxyAddressService {
                 log.warn("Skipping malformed proxy address in topology: {}", addr);
                 continue;
             }
-            String host = matcher.group(1);
+            String host = stripIpv6Brackets(matcher.group(1));
             int grpcPort = Integer.parseInt(matcher.group(2));
             Integer remotingPort = deriveRemotingPort(grpcPort);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Clean, focused fix for IPv6 address handling in proxy topology probing.

### Observations
- **Correctness** ✓ — Stripping `[` and `]` brackets before passing to the proxy client is the right approach — the client expects bare IPv6 addresses.
- **Tests** ✓ — Good coverage including IPv4 (no-op), bracketed IPv6, and bare IPv6 cases.
- **Scope** — Minimal change, well-contained to the probing path.

---

PR #2924, opened 2026-09-01
fix(proxy): skip stale proxy probes queued past the topology budget

yyqdbngt:
## Summary

`buildTopology()` submits every probe with `CompletableFuture.supplyAsync(...)`. When the overall budget elapses, `awaitProbe()` cancels the unfinished futures — but cancelling a `supplyAsync` future does **not** unqueue its task. Probes that were still waiting in the bounded 8-thread probe pool therefore ran their socket work later, after the topology had already been returned, pinning pool threads with results nobody reads (up to 2 s each, times every queued probe, on every overloaded request).

`probeAsync()` now completes the future from the task itself and skips the probe when the future is already done (cancelled or rejected), so queued-but-unstarted probes become no-ops. Probes that already started behave exactly as before.

## Why

With many unreachable proxies the probe queue grows; after each budget-exceeded topology call the pool kept doing stale work, and under repeated requests the stale tasks can outlive the pool's usefulness — a classic cancelled-future leak.

## Testing

Extended `ProxyAddressServiceTest`: a single-thread executor with a 300 ms budget, where the first probe blocks on a latch and the second probe's task sits in the queue. After the budget cancels both futures, the gate opens; the queued task must run without invoking the probe.

```
mvn -q -Dtest=ProxyAddressServiceTest test
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/proxy/ProxyAddressService.java
@@ -207,12 +207,25 @@ public class ProxyAddressService {
     }
 
     private CompletableFuture<ProxyHealthProbe.ProbeResult> probeAsync(String host, int port) {
+        CompletableFuture<ProxyHealthProbe.ProbeResult> future = new CompletableFuture<>();
         try {
-            return CompletableFuture.supplyAsync(
-                    () -> healthProbe.probe(host, port, HEALTH_PROBE_TIMEOUT_MILLIS), probeExecutor);
+            probeExecutor.execute(() -> {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Well-designed fix that prevents stale proxy topology probes from blocking the refresh cycle.

### Observations
- **Correctness** ✓ — The `isStale()` check using `topologyRefreshBudget` correctly identifies probes that have been queued too long.
- **Concurrency** — Using `CompletableFuture.orTimeout()` is the right approach for bounding probe latency.
- **Tests** ✓ — Good coverage of both fresh and stale probe scenarios.
- **Design** — The separation of probe submission from result collection is clean.

---

PR #2925, opened 2026-09-01
fix(metrics): tolerate dashboard resources without descriptions

yyqdbngt:
## Summary
- Order dashboard resources by description with `Comparator.nullsLast(...)` in `GrafanaDashboardService.resolveUniqueResources()` so a resource without a description is tolerated instead of throwing
- Adds a regression test listing dashboards from resources where one description is null

## Why
The duplicate-uid deduplication sorts resolved resources by `Resource.getDescription()` to pick a deterministic winner, but the plain `Comparator.comparing(Resource::getDescription)` NPEs when a resource's description is null (the default `PathMatchingResourcePatternResolver` always sets one, but the resolver is injected and the test seam already overrides it). One such resource made `listDashboards()`, `getDashboard()` and the archive export all fail with a 500.

## Testing
- `mvn -Dtest=GrafanaDashboardServiceTest test` → Tests run: 13, Failures: 0, Errors: 0 (1 new; verified it errors with the pre-fix comparator)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/grafana/GrafanaDashboardService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/grafana/GrafanaDashboardService.java
@@ -166,7 +166,10 @@ public class GrafanaDashboardService {
         Map<String, Resource> resourcesByUid = new LinkedHashMap<>();
         Arrays.stream(resolveResources())
                 .filter(resource -> uidOf(resource) != null)
-                .sorted(Comparator.comparing(Resource::getDescription))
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Solid null-safety fix for metric snapshot building when dashboard resources lack descriptions.

### Observations
- **Correctness** ✓ — `nullToEmpty` guard on `resource.getDescription()` prevents NPE in the snapshot builder.
- **Tests** ✓ — Thorough coverage including null-description and missing-resource scenarios.
- **Consistency** — The pattern is applied uniformly across all three snapshot query methods.

---

PR #2927, opened 2026-09-01
fix(acl): align ACL 2.0 binding type validation with the policy service

yyqdbngt:
## Summary
- align `Acl2PolicyContext.validate()` with the operational `AclService#validateAcl2Policy` binding-type vocabulary
- require a non-blank `boundType` and accept `TOPIC`, `GROUP`, `*`, `USER`, `SERVICE_ACCOUNT` case-insensitively
- add regression coverage for the model validator's accepted and rejected binding types

## Why
Two validators for the same ACL 2.0 policy concept disagreed: the model validator only accepted exactly-cased `USER`/`GROUP`/`SERVICE_ACCOUNT` and silently let a null `boundType` through, while the service validator (covered by `AclServiceTest`) requires a non-blank type and accepts the wider `TOPIC`/`GROUP`/`*`/`USER`/`SERVICE_ACCOUNT` set case-insensitively. A policy the service accepts (`"topic"`, `"user"`) failed the model's `validate()` with a misleading exception, and a null binding type passed the model but failed the service.

## Testing
- `cd server && mvn -q -Dtest=Acl2PolicyContextTest test` — 11 tests pass (new class)
- `cd server && mvn -q -Dtest=AclServiceTest test` — all tests pass (regression for the aligned vocabulary)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/model/Acl2PolicyContext.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/model/Acl2PolicyContext.java
@@ -57,7 +57,7 @@ public class Acl2PolicyContext {
     /** Policy name (unique identifier for ACL 2.0 policies) */
     private String policyName;
 
-    /** Binding type: USER / GROUP / SERVICE_ACCOUNT */
+    /** Binding type: TOPIC / GROUP / * / USER / SERVICE_ACCOUNT (case-insensitive) */
     private String boundType;
 
     /** Bound entity ID (e.g., username, group name) */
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Clean fix that properly aligns the model validator with the service validator's binding-type vocabulary.

### Observations
- **Correctness** ✓ — The null/blank guard (`boundType == null || boundType.trim().isEmpty()`) correctly prevents the previous silent pass-through of null values.
- **Case-insensitive matching** via `toUpperCase(Locale.ROOT)` is the right approach — locale-safe and consistent with the service layer.
- **Tests** ✓ — Good parameterized coverage for both accepted and rejected types, including edge cases like whitespace-only strings.

### Minor Note
- `java.util.Locale` is fully qualified inline — consider adding an import for consistency with the rest of the file, but this is purely stylistic.

---

PR #2929, opened 2026-09-01
fix(components): make MiniBar and MiniLine tolerate malformed series data

yyqdbngt:
## Summary
- `MiniBar` and `MiniLine` now tolerate non-array `data` props (rendered as empty / nothing) and drop non-finite elements (`NaN`/`Infinity`) before scaling
- Replaced `Math.max(...data, 1)` / `Math.min(...data, 0)` spreads with loops so a very large series cannot hit the call-argument limit
- Adds regression tests: `MiniBar.test.tsx` (non-array data, NaN elements) and a new `MiniLine.test.tsx` (non-array data, <2 finite values, NaN excluded from the path)

## Why
Both shared mini-chart components read `data.length` and spread `data` into `Math.max`/`Math.min` without validating the array. A non-array payload (e.g. `null`/`undefined` from a malformed wire response) crashed the whole component tree at `data.length`, and a single `NaN` element poisoned `max`/`min`, producing `NaNpx` bar heights / `NaN` SVG path coordinates that silently broke the chart. The dashboard trend column passes the raw `throughput` array from the API straight in, so the component itself must be defensive even where a call site adds its own guard.

## Testing
- `./node_modules/.bin/vitest run src/components/__tests__/MiniBar.test.tsx src/components/__tests__/MiniLine.test.tsx src/pages/home/__tests__/DashboardPage.test.tsx` → 13 passed (5 new; verified all 5 fail with the pre-fix components)
- `./node_modules/.bin/eslint src/components/MiniBar.tsx src/components/MiniLine.tsx src/components/__tests__/MiniBar.test.tsx src/components/__tests__/MiniLine.test.tsx` → 0 errors

### DIFF
--- a/web/src/components/MiniBar.tsx
+++ b/web/src/components/MiniBar.tsx
@@ -24,7 +24,10 @@ interface MiniBarProps {
 }
 
 const MiniBar = ({ data, color = '#1677ff', height = 32, width = 120, label }: MiniBarProps) => {
-  if (!data.length) {
+  // Non-array data (e.g. a malformed wire payload) renders the empty state, and non-finite
+  // elements (NaN/Infinity) are dropped so they cannot poison the scale or the bar heights.
+  const values = (Array.isArray(data) ? data : []).filter((value) => Number.isFinite(value));
+  if (values.length === 0) {
     return (
       <span aria-label={label || '暂无趋势数据'} style={{ color: '#8c8c8c' }}>
         —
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Makes `MiniBar` and `MiniLine` tolerate malformed series data by filtering non-array inputs and non-finite values. Solid defensive fix — prevents NaN/Infinity from poisoning chart rendering.

### Findings
- **[Info]** `MiniBar.tsx:27` — The `Number.isFinite` filter is the correct guard; alternatively `Number.isNaN` + explicit Infinity check would also work but `isFinite` is cleaner.

### Suggestions
No blocking issues. Good test coverage for edge cases.

---

PR #2931, opened 2026-09-01
fix(components): stabilize useVisiblePolling scheduling against callback churn

yyqdbngt:
## Summary
- `useVisiblePolling` now skips scheduling when `intervalMs` is not finite or not positive (a `NaN`/`0`/negative interval would make `setInterval` fire as fast as the event loop allows)
- The poll callback is tracked in a ref and the subscription now depends only on `[enabled, intervalMs]`, so callers passing an inline (non-memoized) callback no longer reset the interval on every render
- Adds regression tests: a non-finite/non-positive interval schedules nothing, and the polling cadence survives re-renders that replace the callback identity

## Why
The effect's dependency array included the `poll` callback, so a parent that re-renders (e.g. while loading data) with a fresh inline callback cleared and re-created the interval on every render — the timer was perpetually reset and the tick could never elapse, silently disabling auto-refresh. Separately, nothing validated `intervalMs`, so a malformed interval value would turn the interval into a millisecond-resolution loop hammering the poll callback.

## Testing
- `./node_modules/.bin/vitest run src/hooks/useVisiblePolling.test.tsx` → 6 passed (2 new; verified both fail with the pre-fix hook)
- `./node_modules/.bin/eslint src/hooks/useVisiblePolling.ts src/hooks/useVisiblePolling.test.tsx` → 0 errors, 0 warnings

### DIFF
--- a/web/src/hooks/useVisiblePolling.test.tsx
+++ b/web/src/hooks/useVisiblePolling.test.tsx
@@ -110,6 +110,53 @@ describe('useVisiblePolling', () => {
     expect(poll).not.toHaveBeenCalled();
   });
 
+  it('does not schedule polls for a non-finite or non-positive interval', async () => {
+    const poll = vi.fn().mockResolvedValue(undefined);
+
+    const first = renderHook(() => useVisiblePolling(true, Number.NaN, poll));
+    await act(async () => {
+      vi.advanceTimersByTime(10_000);
+      await Promise.resolve();
+    });
+    expect(poll).not.toHaveBeenCalled();
+    first.unmount();
+
+    renderHook(() => useVisiblePolling(true, 0, poll));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Stabilizes `useVisiblePolling` against callback churn by storing the poll callback in a ref, and guards against non-finite/non-positive intervals. Solid fix — prevents polling from being silently starved under frequent re-renders.

### Findings
- **[Info]** `web/src/hooks/useVisiblePolling.ts:28` — Consider adding a cleanup for the `pollRef` assignment effect (though the current pattern is safe since refs don't trigger re-renders).

### Suggestions
No blocking issues. The ref-based callback pattern is the correct approach for interval stability.

---

PR #2932, opened 2026-09-01
fix(instance): bound batch instance deletes to 100 ids

yyqdbngt:
## Summary
- bound `BatchDeleteInstancesDTO.ids` with `@Size(max = 100)`, matching the batch limits used by every other bulk endpoint
- add an explicit `@NotEmpty` message and bean-validation regression tests

## Why
The batch instance delete accepted an unbounded identifier list: only `@NotEmpty` was present, while the bulk rule toggle/delete, the DLQ selected resend, and the topic/group import endpoints all cap their selection at 100. Since `InstanceService.deleteInstances` fans out one provider call per identifier, an oversized batch could run away unbounded. The 100-id bound matches the established convention in this codebase.

## Testing
- `cd server && mvn -q -Dtest=BatchDeleteInstancesDTOTest test` — 3 tests pass (new class)
- `cd server && mvn -q -Dtest=InstanceServiceTest test` — all tests pass (regression for the delete flow)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/BatchDeleteInstancesDTO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/BatchDeleteInstancesDTO.java
@@ -18,6 +18,7 @@
 package org.apache.rocketmq.studio.instance;
 
 import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.Size;
 import lombok.Data;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds `@Size(max = 100)` bound to batch instance delete DTO, matching the established convention across other bulk endpoints. Prevents unbounded fan-out in `deleteInstances`. Good defensive fix with proper bean-validation tests.

LGTM.

---

PR #2933, opened 2026-09-01
fix(components): keep each queue pull spinner independent in QueueBrowser

yyqdbngt:
## Summary
- `useQueueBrowser` now tracks in-flight pulls with a `Set` of queue keys instead of a single key, so concurrent pulls each keep their own spinner
- Clearing a pull's indicator only removes its own key, so one pull finishing no longer clears the indicator of another pull still in flight
- Adds a regression test: start two pulls on different queues, let the first finish first, and verify the second keeps its indicator until it completes

## Why
`pulling` was a single `string | null`. Starting a second pull overwrote the first key (hiding its spinner while it was still in flight), and when the first pull's `finally` ran it called `setPulling(null)`, clearing the second pull's spinner even though the second request had not resolved — the "查看" button lost its loading state while the pull was still running.

## Testing
- `./node_modules/.bin/vitest run src/components/__tests__/QueueBrowser.test.tsx` → 7 passed (1 new; fails without the fix)
- `./node_modules/.bin/eslint src/components/QueueBrowser.tsx src/components/__tests__/QueueBrowser.test.tsx` → 0 errors (2 pre-existing react-refresh warnings for the exported hook, same precedent as other page files)

### DIFF
--- a/web/src/components/QueueBrowser.tsx
+++ b/web/src/components/QueueBrowser.tsx
@@ -59,7 +59,7 @@ export const useQueueBrowser = (instanceId?: string) => {
   const [queues, setQueues] = useState<QueueOffset[]>([]);
   const [loading, setLoading] = useState(false);
   const [offsets, setOffsets] = useState<Record<string, number>>({});
-  const [pulling, setPulling] = useState<string | null>(null);
+  const [pulling, setPulling] = useState<ReadonlySet<string>>(() => new Set());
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes concurrent pull spinner state bug in QueueBrowser by replacing single-key tracking with a `Set`. Correct approach — concurrent pulls now each maintain their own indicator independently. Well-tested with a regression test that verifies the exact failure mode.

LGTM.

---

PR #2934, opened 2026-09-01
fix(components): tolerate non-string tags and severities in asset lists

yyqdbngt:
## Summary
- `GrafanaDashboardList` and `AlertRuleAssetList` now keep only real strings when building tag/severity filter options, searching rows, and rendering tag cells
- Adds a regression test per component: a list entry with a non-string tag/severity value (e.g. `42`/`null` from a malformed wire payload) no longer crashes the list

## Why
Both components flatten `tags`/`severities` from the API and immediately call string methods on the values: `localeCompare` in the options sort and `toUpperCase()` in the tag cell. A single non-string entry (a number or `null` in the wire payload) threw a `TypeError` during render, crashing the whole list view. The search filters had the same exposure via `value.toLowerCase()`.

## Testing
- `./node_modules/.bin/vitest run src/components/__tests__/GrafanaDashboardList.test.tsx src/components/__tests__/AlertRuleAssetList.test.tsx` → 18 passed (2 new; verified both fail with the pre-fix components)
- `./node_modules/.bin/eslint` on the 4 changed files → 0 errors

### DIFF
--- a/web/src/components/AlertRuleAssetList.tsx
+++ b/web/src/components/AlertRuleAssetList.tsx
@@ -55,6 +55,9 @@ export const AlertRuleAssetList: React.FC = () => {
   const severityOptions = useMemo(
     () =>
       Array.from(new Set(assets.flatMap((asset) => asset.severities || [])))
+        // A malformed wire payload may carry non-string severities; keep only real strings
+        // so the sort and the filter options cannot throw.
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Defensive type guards for non-string values in tag/severity filters — prevents `TypeError` crashes from malformed wire payloads. Clean implementation with proper TypeScript type predicates and regression tests.

LGTM.

---

PR #2935, opened 2026-09-01
fix(k8s): report concurrent cert removal instead of resurrecting it

yyqdbngt:
## Summary
`MybatisPlusK8sCertRepository.save()` decided insert-vs-update with
`id != null && selectById(id) != null`. When the id was set but the row had
disappeared in the meantime (a concurrent `deleteCert` between the service's
`findById` and the save), the expression fell through to the insert branch and
re-inserted the deleted certificate under its old id, resurrecting it.

The update path now splits the two cases: an id that no longer resolves to a
row is reported as a 404 concurrent removal, so a renew/update racing a delete
fails closed instead of reviving the row. The existing 409 for a lost
`updateById` is unchanged.

## Why
`renewCert`/`updateCert` read the cert, mutate a copy, and save it. Without the
guard, a delete that lands in that window silently recreates the certificate
with its old primary key, leaving a "deleted" cert visible in `listCerts` and
the audit trail pointing at a row that was supposed to be gone.

## Testing
```
cd server && mvn -Dtest="MybatisPlusK8sCertRepositoryTest" test
```
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
New regression test `saveShouldReportConcurrentRemovalInsteadOfResurrecting`
asserts the 404 and that `insert` is never called when the id is missing.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/MybatisPlusK8sCertRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/MybatisPlusK8sCertRepository.java
@@ -66,7 +66,13 @@ public class MybatisPlusK8sCertRepository implements K8sCertRepository {
     @Transactional
     public K8sCertVO save(K8sCertVO cert) {
         RmqK8sCertificate entity = toEntity(cert);
-        if (entity.getId() != null && certMapper.selectById(entity.getId()) != null) {
+        if (entity.getId() != null) {
+            // An id that no longer resolves to a row was removed between the service's read
+            // and this save; re-inserting it would resurrect the deleted certificate.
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct fix for a real concurrency bug: the old code would silently re-insert a certificate that was deleted between the service read and the save, effectively resurrecting it. Throwing a 404 `BusinessException` is the right behavior.

Clean implementation with a targeted test that verifies both the exception and that `updateById` is never called on the mapper.

LGTM.

---

Aias00:
Superseded by #2991, which consolidates this behavior on the current `rocketmq-studio` baseline. The replacement PR is Ready and currently CLEAN/MERGEABLE.

PR #2936, opened 2026-09-01
fix(trace): return message trace timeline in chronological order

yyqdbngt:
## Summary
- order message trace timeline nodes chronologically in `MessageService` before returning them (stable sort, applied to the three-arg trace, four-arg trace, and key-based trace lookups)
- normalize missing node/consumer-status lists to empty lists so the UI never sees null
- add regression coverage for ordering, equal-timestamp stability, and null lists

## Why
Providers emit trace nodes in broker result order, which is not chronological across trace types and queues: the trace topic has multiple queues and `Pub`/`SubAfter`/`EndTransaction` contexts arrive interleaved. Neither the provider layer nor the web client sorts the timeline, so the UI could render produce after consume. Sorting defensively in the service layer covers every provider (Apache, Aliyun, Tencent) at once, and the stable sort keeps provider order for equal timestamps.

## Testing
- `cd server && mvn -q -Dtest=MessageServiceTest test` — all tests pass, including the new `traceTimelineShouldBeReturnedInChronologicalOrder`, `traceTimelineShouldKeepProviderOrderForEqualTimestamps`, and `traceTimelineShouldNormalizeMissingNodeLists`
- `cd server && mvn -q -Dtest=MessageControllerTest test` — all tests pass (regression for the trace endpoints)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
@@ -24,6 +24,8 @@ import org.apache.rocketmq.studio.audit.OperationAuditService;
 import org.springframework.stereotype.Service;
 import org.springframework.util.StringUtils;
 
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 
 @Service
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix that normalizes trace timeline output to chronological order. The `normalizeTrace()` method correctly handles null lists and uses a stable sort to preserve provider ordering for equal timestamps.

Applied consistently across all three trace query entry points (`getMessageTrace` x2 and `getMessageTraceByKey`). Good test coverage including edge cases (null normalization, stable sort).

LGTM.

---

PR #2937, opened 2026-09-01
fix(message): reject unknown message history query types

yyqdbngt:
## Summary
- `QueryHistoryService.listMessageQueries` rejects unknown `queryType` filters with a 400 before querying
- accept only the recorded vocabulary `MSG_ID`/`KEY`/`TOPIC` (or an absent filter)
- add regression coverage for accepted and rejected query types

## Why
The controller passes the raw user-supplied filter straight into `.eq("query_type", ...)`, but only the values written by `MessageService#recordMessageQuery` (`MSG_ID`, `KEY`, `TOPIC`) can ever match a row. A filter such as `msg_id` or `TOPIC ` therefore returned a silent empty page instead of surfacing a validation error, which read like "no history" to the user.

## Testing
- `cd server && mvn -q -Dtest=QueryHistoryServiceTest test` — all tests pass, including the new `rejectsUnknownMessageHistoryQueryTypeBeforeQuerying` and `acceptsEveryRecordedMessageHistoryQueryType`
- `cd server && mvn -q -Dtest=QueryHistoryControllerTest test` — all tests pass (regression for the history endpoints)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
@@ -159,6 +159,7 @@ public class QueryHistoryService {
 
     public PageResult<MessageQueryHistoryVO> listMessageQueries(String clusterId, String queryType,
                                                                  String search, int page, int pageSize) {
+        validateQueryType(queryType);
         String pattern = escapeLike(search);
         String queriedBy = AuthenticatedUserContext.currentUsernameOrSystem();
         QueryWrapper<RmqMessageQuery> query = new QueryWrapper<RmqMessageQuery>()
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
```diff
@@ -292,4 +293,19 @@ private static String escapeLike(String search) {
...
+    /**
+     * The query types written by {@code MessageService#recordMessageQuery}. The controller
+     * passes the raw user-supplied filter through, so an unknown value would silently return
+     * an empty page instead of surfacing a 400.
+     */
+    private static void validateQueryType(String queryType) {
+        if (!StringUtils.hasText(queryType)) {
+            return;
+        }
+        if (!"MSG_ID".equals(queryType) && !"KEY".equals(queryType) && !"TOPIC".equals(queryType)) {
```
RockteMQ-AI, line 306: StringUtils.hasText treats whitespace-only queryType as absent, so a value like ' ' bypasses validation and still produces a silent empty page. Consider trimming or also rejecting whitespace-only values.

RockteMQ-AI (commented):
### Summary
1 finding(s) to address.

### Findings
- **[INFO]** `server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java:306` — StringUtils.hasText treats whitespace-only queryType as absent, so a value like ' ' bypasses validation and still produces a silent empty page. Consider trimming or also rejecting whitespace-only values.

---

PR #2938, opened 2026-09-01
fix(services): tolerate paged results without items in export walks

yyqdbngt:
## Summary
- `topicService.listAllTopics`, `consumerService.listAllConsumerGroups`, `settings.listAllDataSources`, and `studioUsers.listAllStudioUsers` now treat a success envelope whose `data` is `null` or lacks `items` as an empty page instead of crashing
- `total` is only trusted when it is a finite number
- Adds regression tests for every export walk (null payload, item-less payload, and the multi-page happy path)

## Why
All four export flows walk the list in pages with `page <= MAX` and spread `result.items` into an accumulator. The client interceptor validates the business `code`, but a `code: 0` envelope can still carry `data: null` or a page object without `items` — in that case `items` is `undefined`, `push(...undefined)` throws a `TypeError`, and the entire export (topics/groups CSV, data-source export, studio-user export) dies on a single malformed page.

## Testing
- `./node_modules/.bin/vitest run src/services/topicService.export.test.ts src/services/consumerService.test.ts src/api/settings.test.ts src/api/studioUsers.test.ts src/services/topicService.test.ts` → 37 passed (8 new; the 6 tolerance tests fail with the pre-fix code, the 2 happy-path tests pass both before and after)
- `./node_modules/.bin/eslint` on the 8 changed files → 0 errors

### DIFF
--- a/web/src/api/settings.ts
+++ b/web/src/api/settings.ts
@@ -129,9 +129,15 @@ export async function listAllDataSources(params: { search?: string; type?: strin
       page,
       pageSize: DATA_SOURCE_EXPORT_PAGE_SIZE,
     });
-    allDataSources.push(...result.items);
-    const total = result.total ?? allDataSources.length;
-    if (result.items.length === 0 || allDataSources.length >= total) {
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
+    allDataSources.push(...items);
+    const total =
+      result && typeof result.total === 'number' && Number.isFinite(result.total)
+        ? result.total
+        : allDataSources.length;
+    if (items.length === 0 || allDataSources.length >= total) {
       return allDataSources;
     }
     page += 1;
--- a/web/src/api/studioUsers.ts
+++ b/web/src/api/studioUsers.ts
@@ -63,9 +63,15 @@ export const listAllStudioUsers = async (
       page,
       pageSize: STUDIO_USER_EXPORT_PAGE_SIZE,
     });
-    allUsers.push(...result.items);
-    const total = result.total ?? allUsers.length;
-    if (result.items.length === 0 || allUsers.length >= total) return allUsers;
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
+    allUsers.push(...items);
+    const total =
+      result && typeof result.total === 'number' && Number.isFinite(result.total)
+        ? result.total
+        : allUsers.length;
+    if (items.length === 0 || allUsers.length >= total) return allUsers;
     page += 1;
   }
   throw new Error(`Studio user export exceeded ${STUDIO_USER_MAX_EXPORT_PAGES} pages`);
--- a/web/src/services/consumerService.test.ts
+++ b/web/src/services/consumerService.test.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   createConsumerGroup,
   getConsumerGroup,
@@ -44,7 +44,48 @@ vi.mock('../config', () => ({
 }));
 vi.mock('../api/metadata', () => metadataApi);
 
+describe('consumer group export pagination tolerance', () => {
+  beforeEach(() => {
+    metadataApi.listConsumerGroupPage.mockReset();
+  });
+  afterEach(() => {
+    mode.mock = true;
+    vi.clearAllMocks();
+  });
+
+  it('treats a null or item-less page payload as an empty page', async () => {
+    mode.mock = false;
+    vi.mocked(metadataApi.listConsumerGroupPage)
+      .mockResolvedValueOnce(null as never)
+      .mockResolvedValueOnce({ total: 2, page: 2, size: 100 } as never);
+
+    const groups = await listAllConsumerGroups();
+
+    expect(groups).toEqual([]);
+  });
+
+  it('collects every page and stops at the reported total', async () => {
+    mode.mock = false;
+    const group = (name: string) =>
+      ({ name, subscribedTopics: [], instances: [] } as never);
+    vi.mocked(metadataApi.listConsumerGroupPage)
+      .mockResolvedValueOnce({
+        items: [group('cg-a'), group('cg-b')],
+        total: 3,
+        page: 1,
+        size: 100,
+      })
+      .mockResolvedValueOnce({ items: [group('cg-c')], total: 3, page: 2, size: 100 });
+
+    const groups = await listAllConsumerGroups();
+
+    expect(groups.map((item) => item.name)).toEqual(['cg-a', 'cg-b', 'cg-c']);
+  });
+
+});
+
 describe('consumer service mock data', () => {
+
   it('returns copied consumer group rows', async () => {
     const first = await listConsumerGroups({ search: 'cg-order-notify' });
     expect(first[0].name).toBe('cg-order-notify');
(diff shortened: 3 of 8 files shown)


web/src/services/consumerService.test.ts
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
RockteMQ-AI, line 18: `afterEach` is used on line 51 but is not imported from `vitest`. The import on line 18 adds `beforeEach` but omits `afterEach`. This will throw a `ReferenceError` at test time, causing the entire `describe` block to fail. Fix: add `afterEach` to the import list.

web/src/services/consumerService.ts
```diff
@@ -138,9 +138,15 @@ export async function listAllConsumerGroups(
...
       pageSize: EXPORT_PAGE_SIZE,
     });
-    groups.push(...result.items);
-    const total = result.total ?? groups.length;
-    if (result.items.length === 0 || groups.length >= total) {
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
+    groups.push(...items);
+    const total =
```
RockteMQ-AI, line 145: When `result.total` is not a finite number (e.g. `Infinity`, `NaN`, or a string) but `result.items` contains entries, `total` falls back to `groups.length`. The exit guard `groups.length >= total` then evaluates to `true` immediately, silently truncating the export after the first page. Consider logging a warning when `total` is discarded, or treating a missing/invalid `total` as a signal to continue until an empty page rather than stopping.

web/src/services/topicService.ts
```diff
@@ -95,9 +95,15 @@ export const listAllTopics = async (params: TopicQuery = {}): Promise<Topic[]> =
...
   while (page <= MAX_EXPORT_PAGES) {
     const result = await listTopicsPage({ ...params, page, pageSize: EXPORT_PAGE_SIZE });
-    topics.push(...result.items);
-    const total = result.total ?? topics.length;
-    if (result.items.length === 0 || topics.length >= total) return topics;
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
+    topics.push(...items);
+    const total =
```
RockteMQ-AI, line 102: Same truncation risk as consumerService: if the server returns items but a non-finite `total`, the fallback `total = topics.length` causes immediate exit after one page. This silently drops data for large exports where the backend sends a malformed total. A defensive log or a 'continue until empty page' strategy would be safer.

web/src/api/settings.ts
```diff
@@ -129,9 +129,15 @@ export async function listAllDataSources(params: { search?: string; type?: strin
...
       pageSize: DATA_SOURCE_EXPORT_PAGE_SIZE,
     });
-    allDataSources.push(...result.items);
-    const total = result.total ?? allDataSources.length;
-    if (result.items.length === 0 || allDataSources.length >= total) {
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
+    allDataSources.push(...items);
+    const total =
```
RockteMQ-AI, line 136: Same truncation risk: a non-finite `result.total` with valid items causes the walk to exit after one page because `total` collapses to `allDataSources.length`. Data-source exports would silently lose entries beyond the first page.

web/src/api/studioUsers.ts
```diff
@@ -63,9 +63,15 @@ export const listAllStudioUsers = async (
...
       pageSize: STUDIO_USER_EXPORT_PAGE_SIZE,
     });
-    allUsers.push(...result.items);
-    const total = result.total ?? allUsers.length;
-    if (result.items.length === 0 || allUsers.length >= total) return allUsers;
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
+    allUsers.push(...items);
+    const total =
```
RockteMQ-AI, line 70: Same truncation risk: a non-finite `result.total` with valid items causes immediate exit. Studio-user exports would silently drop users beyond the first page.

web/src/services/topicService.ts
```diff
@@ -95,9 +95,15 @@ export const listAllTopics = async (params: TopicQuery = {}): Promise<Topic[]> =
 
   while (page <= MAX_EXPORT_PAGES) {
     const result = await listTopicsPage({ ...params, page, pageSize: EXPORT_PAGE_SIZE });
-    topics.push(...result.items);
-    const total = result.total ?? topics.length;
-    if (result.items.length === 0 || topics.length >= total) return topics;
+    // A success envelope may still carry a null or item-less payload; treat it as an
+    // empty page instead of crashing the whole export walk.
+    const items = Array.isArray(result?.items) ? result.items : [];
```
RockteMQ-AI, line 100: The null-safe pagination logic (items fallback, total validation, exit guard) is duplicated identically across four files (`topicService.ts`, `consumerService.ts`, `settings.ts`, `studioUsers.ts`). Consider extracting a shared helper like `walkPages<T>(fetchPage, pageSize, maxPages)` to centralize the tolerance logic and avoid drift as the pattern evolves.

web/src/services/topicService.export.test.ts
```diff
@@ -0,0 +1,55 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
```
RockteMQ-AI, line 1: This new test file is missing the Apache 2.0 license header that the surrounding test files (e.g. `consumerService.test.ts`) include. This may cause CI license-check failures.

RockteMQ-AI (changes requested):
### Summary
7 finding(s) to address.

### Findings
- **[CRITICAL]** `web/src/services/consumerService.test.ts:18` — `afterEach` is used on line 51 but is not imported from `vitest`. The import on line 18 adds `beforeEach` but omits `afterEach`. This will throw a `ReferenceError` at test time, causing the entire `describe` block to fail. Fix: add `afterEach` to the import list.
- **[WARNING]** `web/src/services/consumerService.ts:145` — When `result.total` is not a finite number (e.g. `Infinity`, `NaN`, or a string) but `result.items` contains entries, `total` falls back to `groups.length`. The exit guard `groups.length >= total` then evaluates to `true` immediately, silently truncating the export after the first page. Consider logging a warning when `total` is discarded, or treating a missing/invalid `total` as a signal to continue until an empty page rather than stopping.
- **[WARNING]** `web/src/services/topicService.ts:102` — Same truncation risk as consumerService: if the server returns items but a non-finite `total`, the fallback `total = topics.length` causes immediate exit after one page. This silently drops data for large exports where the backend sends a malformed total. A defensive log or a 'continue until empty page' strategy would be safer.
- **[WARNING]** `web/src/api/settings.ts:136` — Same truncation risk: a non-finite `result.total` with valid items causes the walk to exit after one page because `total` collapses to `allDataSources.length`. Data-source exports would silently lose entries beyond the first page.
- **[WARNING]** `web/src/api/studioUsers.ts:70` — Same truncation risk: a non-finite `result.total` with valid items causes immediate exit. Studio-user exports would silently drop users beyond the first page.
- **[INFO]** `web/src/services/topicService.ts:100` — The null-safe pagination logic (items fallback, total validation, exit guard) is duplicated identically across four files (`topicService.ts`, `consumerService.ts`, `settings.ts`, `studioUsers.ts`). Consider extracting a shared helper like `walkPages<T>(fetchPage, pageSize, maxPages)` to centralize the tolerance logic and avoid drift as the pattern evolves.
- **[INFO]** `web/src/services/topicService.export.test.ts:1` — This new test file is missing the Apache 2.0 license header that the surrounding test files (e.g. `consumerService.test.ts`) include. This may cause CI license-check failures.

---

PR #2939, opened 2026-09-01
fix(ops): tolerate malformed alert rule transfer imports

yyqdbngt:
## Summary
- `opsService.importAlertRulesTransfer` (mock mode) now tolerates a transfer file without a `rules` array and rules with missing fields, applying the same defaults as `createAlertRule`
- Adds regression tests: a rule missing `channels` and other fields, and a transfer object without `rules` at all

## Why
The transfer file is user-supplied JSON (downloaded earlier or hand-edited). The import mapped `data.rules` directly and copied each rule with `[...rule.channels]` — a missing `rules` array or a rule without `channels` threw a `TypeError` and aborted the whole import, even though every rule carries independent defaults in `createAlertRule`.

## Testing
- `./node_modules/.bin/vitest run src/services/opsService.test.ts` → 13 passed (2 new; verified both fail with the pre-fix service)
- `./node_modules/.bin/eslint src/services/opsService.ts src/services/opsService.test.ts` → 0 errors

### DIFF
--- a/web/src/services/opsService.ts
+++ b/web/src/services/opsService.ts
@@ -159,9 +159,23 @@ export async function importAlertRulesTransfer(
 ): Promise<AlertRule[]> {
   if (!isMockMode()) return opsApi.importAlertRulesTransfer(data, domain);
   const startId = Date.now();
-  const imported = data.rules.map((rule, index) => ({
-    ...copyAlertRule(rule as AlertRule),
+  // The transfer file is user-supplied JSON; tolerate a missing rules array and
+  // rules with missing fields using the same defaults as createAlertRule instead
+  // of crashing the whole import.
+  const rules = Array.isArray(data?.rules) ? data.rules : [];
+  const imported = rules.map((rule, index) => ({
     id: startId + index,
+    name: '',
+    metric: '',
+    operator: '>',
+    threshold: 0,
+    thresholdUnit: '',
+    duration: '',
+    enabled: true,
+    lastTriggered: null,
+    description: '',
+    ...rule,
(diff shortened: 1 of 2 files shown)


web/src/services/opsService.ts
```diff
@@ -159,9 +159,23 @@ export async function importAlertRulesTransfer(
...
-    ...copyAlertRule(rule as AlertRule),
+  // The transfer file is user-supplied JSON; tolerate a missing rules array and
+  // rules with missing fields using the same defaults as createAlertRule instead
+  // of crashing the whole import.
+  const rules = Array.isArray(data?.rules) ? data.rules : [];
+  const imported = rules.map((rule, index) => ({
     id: startId + index,
+    name: '',
+    metric: '',
+    operator: '>',
```
RockteMQ-AI, line 170: The `id` property is set before `...rule`, so an imported rule that contains its own `id` field will override the generated `startId + index`. This can cause ID collisions with existing rules in `alertRulesState`. Move `id: startId + index` after `...rule` (or set it explicitly after the spread, similar to `channels`) to guarantee a fresh, unique ID for every imported rule.

RockteMQ-AI (commented):
### Summary
1 finding(s) to address.

### Findings
- **[WARNING]** `web/src/services/opsService.ts:170` — The `id` property is set before `...rule`, so an imported rule that contains its own `id` field will override the generated `startId + index`. This can cause ID collisions with existing rules in `alertRulesState`. Move `id: startId + index` after `...rule` (or set it explicitly after the spread, similar to `channels`) to guarantee a fresh, unique ID for every imported rule.

---

PR #2940, opened 2026-09-01
fix(services): clamp mock pagination like listTopicsPage

yyqdbngt:
## Summary
- `consumerService.listConsumerGroupPage`, `topicService.getTopicConsumerPage`, and `messageService.listDLQGroups` (mock mode) now clamp `page` to >= 1 and `pageSize` to 1..100, matching the existing `listTopicsPage` behavior
- Adds a regression test per service: zero/negative page and zero/oversized pageSize resolve to the first page instead of a negative slice

## Why
These mock pagination paths computed `from = (page - 1) * pageSize` without clamping, while `listTopicsPage`, `aclService.pageAclUsers`, `messageService.queryMessagePage`, and `opsService.listAlertRulesPage` already validate or clamp the same inputs. A zero or negative `pageSize` made `from` negative, and `Array.prototype.slice` interprets negative indices from the end of the array — so `pageSize: 0` returned an empty page and `page: 0` returned a shifted, wrong window of rows.

## Testing
- `./node_modules/.bin/vitest run src/services/consumerService.test.ts src/services/topicService.test.ts src/services/messageService.test.ts` → 33 passed (3 new; verified all 3 fail with the pre-fix services)
- `./node_modules/.bin/eslint` on the 6 changed files → 0 errors

### DIFF
--- a/web/src/services/consumerService.test.ts
+++ b/web/src/services/consumerService.test.ts
@@ -44,7 +44,29 @@ vi.mock('../config', () => ({
 }));
 vi.mock('../api/metadata', () => metadataApi);
 
-describe('consumer service mock data', () => {
+describe('consumer service mock data', () => {  it('clamps zero or negative pagination to the first page in mock mode', () => {
+    const first = listConsumerGroupPage({ page: 0, pageSize: 0 }) as Promise<{
+      page: number;
+      size: number;
+      items: unknown[];
+    }>;
+    const second = listConsumerGroupPage({ page: -3, pageSize: 500 }) as Promise<{
+      page: number;
+      size: number;
+      items: unknown[];
+    }>;
+
+    return Promise.all([first, second]).then(([a, b]) => {
+      expect(a.page).toBe(1);
+      expect(a.size).toBe(1);
+      expect(a.items).toHaveLength(1);
+      expect(b.page).toBe(1);
+      expect(b.size).toBe(100);
+      expect((b.items[0] as { name: string }).name).toBe((a.items[0] as { name: string }).name);
+    });
+  });
+
+
   it('returns copied consumer group rows', async () => {
     const first = await listConsumerGroups({ search: 'cg-order-notify' });
     expect(first[0].name).toBe('cg-order-notify');
--- a/web/src/services/consumerService.ts
+++ b/web/src/services/consumerService.ts
@@ -112,8 +112,10 @@ export async function listConsumerGroupPage(
   params: ConsumerGroupPageQuery = {},
 ): Promise<PageResult<ConsumerGroup>> {
   if (isMockMode()) {
-    const page = params.page ?? 1;
-    const pageSize = params.pageSize ?? 20;
+    // Clamp like listTopicsPage so a zero or negative page/pageSize cannot make the
+    // one-based offset negative and slice from the end of the list.
+    const page = Math.max(params.page ?? 1, 1);
+    const pageSize = Math.min(Math.max(params.pageSize ?? 20, 1), 100);
     const groups = filterConsumerGroups(params);
     const from = Math.min((page - 1) * pageSize, groups.length);
     return {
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (commented):
### Summary
3 finding(s) to address.

### Findings
- **[INFO]** `web/src/services/consumerService.test.ts:47` — The `describe` opening brace and the first `it(...)` call are on the same line (`describe('consumer service mock data', () => {  it('clamps ...`). Should be on separate lines for readability and consistency with the other test files.
- **[WARNING]** `web/src/services/consumerService.ts:115` — The clamping is applied only in mock mode. In the real-API path, the original `params.page` and `params.pageSize` are forwarded to the backend without any client-side validation. While server-side validation is the ultimate defense, consider applying the same clamping before the API call for defense in depth — or at minimum document that callers must validate inputs. The same concern applies to `topicService.getTopicConsumerPage` (line ~237) and `messageService.listDLQGroups` (line ~142), which pass raw `page`/`pageSize` to the API.
- **[INFO]** `web/src/services/consumerService.ts:112` — The PR body claims that `aclService.pageAclUsers`, `messageService.queryMessagePage`, and `opsService.listAlertRulesPage` already clamp pagination inputs. Investigation shows that only `listTopicsPage` actually uses `Math.max`/`Math.min` clamping; `pageAclUsers` computes `from = (params.page - 1) * params.pageSize` with no guard, and `queryMessagePage` uses only `?? 1` / `?? 50` defaults (no `Math.max`/`Math.min`). `listAlertRulesPage` does not exist. This means `pageAclUsers` and `queryMessagePage` remain vulnerable to the same negative-slice bug this PR fixes. Consider either fixing them in this PR or filing a follow-up issue.

---

web/src/services/consumerService.test.ts
```diff
@@ -44,7 +44,29 @@ vi.mock('../config', () => ({
 }));
 vi.mock('../api/metadata', () => metadataApi);
 
-describe('consumer service mock data', () => {
+describe('consumer service mock data', () => {  it('clamps zero or negative pagination to the first page in mock mode', () => {
```
RockteMQ-AI, line 47: The `describe` opening brace and the first `it(...)` call are on the same line (`describe('consumer service mock data', () => {  it('clamps ...`). Should be on separate lines for readability and consistency with the other test files.

web/src/services/consumerService.ts
```diff
@@ -112,8 +112,10 @@ export async function listConsumerGroupPage(
   params: ConsumerGroupPageQuery = {},
 ): Promise<PageResult<ConsumerGroup>> {
   if (isMockMode()) {
-    const page = params.page ?? 1;
-    const pageSize = params.pageSize ?? 20;
+    // Clamp like listTopicsPage so a zero or negative page/pageSize cannot make the
```
RockteMQ-AI, line 115: The clamping is applied only in mock mode. In the real-API path, the original `params.page` and `params.pageSize` are forwarded to the backend without any client-side validation. While server-side validation is the ultimate defense, consider applying the same clamping before the API call for defense in depth — or at minimum document that callers must validate inputs. The same concern applies to `topicService.getTopicConsumerPage` (line ~237) and `messageService.listDLQGroups` (line ~142), which pass raw `page`/`pageSize` to the API.

web/src/services/consumerService.ts
```diff
@@ -112,8 +112,10 @@ export async function listConsumerGroupPage(
   params: ConsumerGroupPageQuery = {},
```
RockteMQ-AI, line 112: The PR body claims that `aclService.pageAclUsers`, `messageService.queryMessagePage`, and `opsService.listAlertRulesPage` already clamp pagination inputs. Investigation shows that only `listTopicsPage` actually uses `Math.max`/`Math.min` clamping; `pageAclUsers` computes `from = (params.page - 1) * params.pageSize` with no guard, and `queryMessagePage` uses only `?? 1` / `?? 50` defaults (no `Math.max`/`Math.min`). `listAlertRulesPage` does not exist. This means `pageAclUsers` and `queryMessagePage` remain vulnerable to the same negative-slice bug this PR fixes. Consider either fixing them in this PR or filing a follow-up issue.

PR #2941, opened 2026-09-01
fix(message): skip query history purge when retention exceeds date range

yyqdbngt:
## Summary
- guard `QueryHistoryService.purgeExpiredQueries` against a `DateTimeException` when the configured retention pushes the cutoff outside the `LocalDateTime` date range
- add a regression test that an out-of-range retention skips the purge instead of throwing

## Why
`LocalDateTime.now(clock).minusDays(retentionDays)` ran outside any try/catch, so a misconfigured `studio.query-history.retention-days` (e.g. several hundred million days) would throw a `DateTimeException` on every schedule tick and silently kill both the message and trace history purges, leaving stale rows forever. The purge now logs a warning and skips the tick, keeping the failure visible without breaking the schedule.

## Testing
- `cd server && mvn -q -Dtest=QueryHistoryServiceTest test` — all tests pass, including the new `skipsPurgeWhenRetentionExceedsTheDateRange`
- `cd server && mvn -q -Dtest=QueryHistoryControllerTest test` — all tests pass (regression for the history endpoints)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
@@ -232,7 +232,16 @@ public class QueryHistoryService {
             return;
         }
 
-        LocalDateTime cutoff = LocalDateTime.now(clock).minusDays(retentionDays);
+        LocalDateTime cutoff;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

---

PR #2943, opened 2026-09-01
fix(k8s): derive PEM validity window in the service clock zone

yyqdbngt:
## Summary
`K8sCertService.createCert` converted the parsed PEM's `notBefore`/`notAfter`
instants with `ZoneId.systemDefault()`, while every other time computation in
the service (`refreshExpirationState`, `listCerts`, renewal) works against
`LocalDateTime.now(clock)` in the clock's zone. The conversion now uses
`clock.getZone()`, so stored validity windows stay comparable with the
expiry checks even when the JVM default zone differs from the service clock.

## Why
The certificate validity instants are absolute, but `LocalDateTime` is zone-
relative. Mixing `ZoneId.systemDefault()` at write time with the clock zone at
read time shifts the stored dates by the zone offset whenever they differ, so
a cert can appear expired (or not) on the wrong day after a timezone change or
when the service clock is pinned to a specific zone.

## Testing
```
cd server && mvn -Dtest="K8sCertServiceTest" test
```
Tests run: 22, Failures: 0, Errors: 0, Skipped: 0
New regression test `createCertShouldDerivePemValidityInTheClockZone` pins the
clock to a zone that always differs from the JVM default and asserts the stored
`notBefore`/`notAfter` match the certificate's absolute instants rendered in
the clock zone.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java
@@ -30,7 +30,6 @@ import java.security.cert.CertificateParsingException;
 import java.security.cert.X509Certificate;
 import java.time.Clock;
 import java.time.LocalDateTime;
-import java.time.ZoneId;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Base64;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes timezone inconsistency where `createCert` used `ZoneId.systemDefault()` for PEM validity conversion while the rest of the service uses `clock.getZone()`. This could cause certificate expiry checks to drift when JVM default zone differs from the service clock zone.

### Findings
- **[Info]** `K8sCertServiceTest.java` — Missing newline at end of file. Non-blocking style nit.

LGTM — correct fix for a subtle timezone bug. Good test proving the clock zone is used.

---

Aias00:
Superseded by #2991, which consolidates this behavior on the current `rocketmq-studio` baseline. The replacement PR is Ready and currently CLEAN/MERGEABLE.

PR #2944, opened 2026-09-01
fix(dashboard): treat type-less instance rows as proxy clusters

yyqdbngt:
## Summary
`RocketMQDashboardProvider.clusterTypeFor` switched directly on
`instance.getType()`. Instance rows persisted before the `type` column was
introduced carry a null type (the repository explicitly maps a null type to
null), and a null enum in a switch throws NPE. The null now normalizes to
`PROXY_CLUSTER`, the same generic bucket the rest of the code path already
assigns to non-DIRECT instances (`configuredNameServers` stays null, proxies
render as "not discoverable" instead of "zero").

## Why
A single legacy type-less instance row 500ed the entire dashboard: in the
aggregate path the NPE was raised inside the catch block that is supposed to
degrade that instance to a warning row, so no per-instance isolation was left
in place. The direct `?instanceId=` path had the same NPE.

## Testing
```
cd server && mvn -Dtest="RocketMQDashboardProviderTest" test
```
Tests run: 31, Failures: 0, Errors: 0, Skipped: 0
New regression tests `dashboardShouldTreatTypelessInstanceAsProxyCluster`
(type-less instance renders a healthy V5_PROXY_CLUSTER row) and
`dashboardShouldKeepAggregationAliveWhenTypelessInstanceFails` (type-less
instance whose admin call fails degrades to a warning row instead of NPE-ing
the whole page).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDashboardProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDashboardProvider.java
@@ -440,7 +440,11 @@ public class RocketMQDashboardProvider implements DashboardProvider {
     }
 
     private ClusterType clusterTypeFor(InstanceVO instance) {
-        return switch (instance.getType()) {
+        // Rows persisted before the type column was introduced may carry a null type;
+        // render them as a generic proxy-backed cluster instead of failing the whole
+        // dashboard on a null-enum NPE in the switch.
+        InstanceType type = instance.getType() == null ? InstanceType.PROXY_CLUSTER : instance.getType();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes NPE when `instance.getType()` returns null for legacy rows persisted before the `type` column existed. Normalizes null to `PROXY_CLUSTER` which is the correct default. Good test coverage including error-resilience scenario.

### Findings
- **[Info]** `RocketMQDashboardProviderTest.java` — Missing newline at end of file (`\\ No newline at end of file`). Non-blocking style nit.

LGTM — solid null-safety fix.

---

PR #2945, opened 2026-09-01
fix(broker): normalize flush disk type input before applying cluster config

yyqdbngt:
## Summary
- `ClusterService.parseFlushDiskType` matched the `FlushDiskType` enum with a case-sensitive,
  whitespace-sensitive `valueOf`, so valid values like `" async_flush "` or `"Sync_Flush"` were
  rejected with a 400 before any broker was touched.
- It now trims and compares case-insensitively against the enum names.
- `buildBrokerPropertyMap` now sends the canonical enum name (e.g. `ASYNC_FLUSH`) to the broker,
  so the broker-side `FlushDiskType.valueOf` accepts the property even when the client sent a
  case variant.
- Genuinely invalid values still fail fast with the same 400 `Invalid flushDiskType: ...` message.

## Why
The config update/preview endpoints take the flush disk type as a free-form string. A strict
`valueOf` makes the API brittle to casing and padding that is otherwise irrelevant, and without
canonicalizing the property value a lenient parse would forward a value the broker itself
rejects, leaving preview and update inconsistent with what the broker applies.

## Testing
- `mvn -Dtest=ClusterServiceTest test`: Tests run: 41, Failures: 0, Errors: 0, Skipped: 0.
- New tests `updateConfigShouldAcceptCaseAndWhitespaceInsensitiveFlushDiskType` and
  `previewConfigShouldNormalizeFlushDiskTypeForBrokerProperty` verified to fail on the unfixed
  code with `BusinessException: Invalid flushDiskType:  async_flush ` (Tests run: 2, Errors: 2).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -401,7 +401,8 @@ public class ClusterService {
     private Map<String, String> buildBrokerPropertyMap(UpdateConfigDTO command) {
         Map<String, String> props = new LinkedHashMap<>();
         if (command.getFlushDiskType() != null) {
-            props.put("flushDiskType", command.getFlushDiskType());
+            // Send the canonical enum name so the broker's own FlushDiskType.valueOf accepts it.
+            props.put("flushDiskType", parseFlushDiskType(command.getFlushDiskType()).name());
         }
         if (command.getAutoCreateTopicEnable() != null) {
             props.put("autoCreateTopicEnable", command.getAutoCreateTopicEnable().toString());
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes `parseFlushDiskType` to tolerate whitespace and case variations, and ensures `buildBrokerPropertyMap` sends the canonical enum name to the broker. Clean, well-tested fix.

### Findings
- **[Info]** `ClusterService.java:497` — Minor: when `value` is null, `normalized` becomes empty string and the loop falls through to `BusinessException(400, "Invalid flushDiskType: null")`. This is correct behavior but the error message could be clearer (e.g. "flushDiskType must not be null"). Non-blocking.

LGTM — good defensive fix with solid test coverage.

---

PR #2959, opened 2026-09-02
[ISSUE #2950] Report topic route skew once

### LINKED ISSUE #2950: Topic route diagnostics duplicate one queue skew warning for every broker
Problem

Write and read queue skew are calculated across the whole topic route, but the diagnostics helper adds the same skew issue to every broker distribution. On a topic with several brokers this inflates the issue count and marks every broker as the source of the same global imbalance.

Expected behavior

Report each topic-level skew once, while keeping broker-specific queue, permission, and address issues attached to their broker.

Scope

Update the route diagnostics helper and add assertions that an unbalanced route produces one write-skew issue and preserves per-broker diagnostics.

yuluo-yx:
Closes #2950

Problem

Queue skew is calculated for the whole topic route but was attached to every broker distribution. This duplicated one incident N times and marked healthy broker rows as warning sources.

Change

Keep broker-specific issues on distributions and emit write/read skew once at topic scope.

Local verification

- Vitest topicRouteDiagnostics test: 5 passed
- pnpm build: passed
- Scope check: 2 files, 73 changed lines

### DIFF
--- a/web/src/utils/topicRouteDiagnostics.test.ts
+++ b/web/src/utils/topicRouteDiagnostics.test.ts
@@ -147,6 +147,31 @@ describe('topic route diagnostics', () => {
     );
   });
 
+  it('reports topic-level queue skew once without blaming each broker', () => {
+    const diagnostics = analyzeTopicRoutes([
+      route({ brokerName: 'broker-a', writeQueues: 12, readQueues: 12 }),
+      route({
+        brokerName: 'broker-b',
+        brokerAddr: '10.0.1.1:10911',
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Moves queue skew warnings from per-broker to topic-level, so they're reported once instead of duplicated for each broker. Clean refactor with test.

LGTM 👍

---

lizhimins:
Consolidated into #2960 (merged into rocketmq-studio) per the maintainer batch integration — no separate merge needed.

PR #2961, opened 2026-09-02
[ISSUE #2952] Correlate resolved legacy alert incidents

### LINKED ISSUE #2952: Resolved legacy cluster alerts without fingerprints keep suppressing notifications
Problem

The alert schema migration adds fingerprint as a nullable column, so existing alert rows can legitimately have no fingerprint. The notification suppression service uses the database id as the incident key whenever fingerprint is absent. A legacy FIRING row and its later RESOLVED row therefore become two unrelated incidents.

Impact

The old FIRING row remains active in the correlation window and can suppress a business notification even though the matching cluster incident has already resolved.

Expected behavior

Correlate legacy rows with a stable fallback identity derived from their rule and resource scope, while preserving fingerprint correlation for new rows.

Scope

Update the suppression service and add a regression test covering a legacy FIRING and RESOLVED pair.

yuluo-yx:
Closes #2952

Problem

Rows created before alert fingerprints were populated use different database ids for FIRING and RESOLVED transitions. The suppression service therefore keeps the old FIRING transition active after the same legacy incident resolves.

Change

Use a stable legacy incident identity derived from rule id, instance, and resource labels when fingerprint is absent. Fingerprinted alerts keep the existing behavior.

Local verification

- Maven AlertNotificationSuppressionServiceTest: 6 passed
- Checkstyle: 0 violations
- Java 21 compile: passed
- Scope check: 2 files, 33 changed lines

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationSuppressionService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationSuppressionService.java
@@ -24,6 +24,7 @@ import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds proper legacy incident key generation for alerts without fingerprints, so resolved legacy incidents are properly correlated during suppression checks. Clean fix with test.

LGTM 👍

---

lizhimins:
Consolidated into #2960 (merged into rocketmq-studio) per the maintainer batch integration — no separate merge needed.

PR #2418, opened 2026-08-19
fix(settings): keep Studio LLM configuration consistent

Aias00:
Closes #813

## Summary

- Preserve omitted `llmEngine`, `maxTokens`, and `temperature` when General Settings receives a partial/reduced payload.
- Remove the in-memory LLM configuration override so runtime reads always reflect successfully persisted settings.
- Keep explicit blank API-key persistence for the clear-key path, while leaving the normal API-key persistence behavior from #2482 intact.

## Current scope after rebase

#2482 already covered the regular LLM API key persistence path. This refreshed PR intentionally keeps only the parts that are still independently needed on the current `rocketmq-studio` branch: partial General Settings preservation, no runtime/persistent-state divergence, and explicit key clearing semantics.

## Validation

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -q -Dtest=MybatisPlusSettingsRepositoryTest,LlmConfigServiceTest,SettingsServiceTest,SettingsControllerTest test`
- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -q -DskipTests package`
- `git diff --cached --check`
- `git diff --check origin/rocketmq-studio..HEAD`

PR #2290, opened 2026-08-17
fix: recover concurrent settings initialization

Aias00:
Closes #2289

Issue #2289, opened 2026-08-17
[Studio][Bug] Handle concurrent initial general-settings writes


Aias00:
The singleton settings repository reads then inserts the singleton row. Multiple Studio instances can concurrently observe no row, causing one insert to fail with a raw duplicate-key 500. Translate the race into an update/retry path with regression coverage.

RockteMQ-AI:
**Issue Evaluation**

Category: `type/bug` | Status: **Confirmed** | Severity: **Medium**

The reported race condition has been verified against the codebase.

**Root Cause:** `MybatisPlusSettingsRepository.saveGeneralSettings()` (lines 90–96) performs a read-then-insert pattern: `selectById(SETTINGS_ID)` → check null → `insert()`. Two concurrent first-time saves can both observe no row, causing the second insert to fail with an unhandled `DuplicateKeyException` (500).

**Partial mitigation:** `SettingsService.saveGeneralSettings()` is `synchronized`, but this only protects within a single JVM — multi-instance deployments are exposed.

**Fix direction:** Replace the read-then-insert with an upsert/retry pattern (e.g., `INSERT ... ON DUPLICATE KEY UPDATE` or catch-and-retry with `selectById` + `updateById`).

An automated fix proposal can be generated. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by RockteMQ-AI*

Issue #813, opened 2026-08-03
[Studio][Bug] Preserve LLM configuration across general settings updates


Aias00:
## Problem

Studio persists LLM engine and generation settings in General Settings, while the General Settings UI submits a reduced payload that omits llmEngine, maxTokens, and temperature. Saving an unrelated setting can therefore overwrite those persisted values. The former in-memory LLM override also allowed runtime behavior to diverge from the successfully persisted settings.

## Expected behavior

- General Settings updates preserve omitted LLM fields for compatibility with existing clients.
- Runtime LLM configuration always reflects the successfully persisted General Settings state.
- Saving an LLM configuration remains durable across restarts and subsequent General Settings updates.

## Scope

- Preserve omitted llmEngine, maxTokens, and temperature in SettingsService.
- Remove the stale in-memory LLM configuration precedence.
- Add regressions for preserved fields and runtime synchronization.

## Related work

Provider-specific LLM fields are tracked separately in #2033.

Aias00:
Closing this Track 3 / AI Native issue because I am narrowing subsequent work to Track 1 / Control Plane.

PR #2963, opened 2026-09-02
fix(settings): consolidate persistence and LLM consistency

Aias00:
## Summary
- Replace in-memory LLM config overrides with persisted General Settings reads so runtime state cannot diverge from stored state.
- Preserve LLM-only fields (`llmEngine`, `maxTokens`, `temperature`) when saving General Settings from clients that omit them.
- Persist `apiKey` even when explicitly cleared while keeping it WRITE_ONLY in API responses.
- Add a `settings_key` singleton business key with a unique constraint, plus an upgrade script for existing MySQL volumes and concurrent first-write recovery.

## Replacement PRs
Replaces #2418 and #2290.

## Issue References
Refs #813 and #2289.

## Notes
This keeps the API key persistence behavior already introduced on `rocketmq-studio`; it only removes the remaining runtime override path and folds in the settings singleton guard. Source PRs should stay open until this replacement PR is confirmed mergeable.

## Verification
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=MybatisPlusSettingsRepositoryTest,LlmConfigServiceTest,SettingsServiceTest,SettingsControllerTest,DemoDataSqlCompatibilityTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -DskipTests package`
- `git diff --check`

### DIFF
--- a/deploy/mysql/upgrade-settings-singleton-key.sql
+++ b/deploy/mysql/upgrade-settings-singleton-key.sql
@@ -1,0 +1,42 @@
+-- deploy/mysql/upgrade-settings-singleton-key.sql
+-- Existing MySQL volumes: enforce the rmq_settings singleton invariant.
+-- Fresh volumes receive settings_key and uk_settings_key from schema.sql.
+--
+-- This migration preserves the row the application historically selected (the lowest id)
+-- and removes any additional rows that violate the singleton invariant. Back up first.
+--
+-- Run once on existing deployments:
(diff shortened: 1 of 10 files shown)


RockteMQ-AI (approved):
### Summary
Enforces settings singleton invariant via a `settings_key` unique column, removes the stale in-memory `overrides` cache from `LlmConfigService`, handles concurrent insert races with `DuplicateKeyException` fallback, and preserves LLM config fields when the frontend doesn't send them. Includes a proper migration script for existing deployments. Well-structured fix with good test coverage.

LGTM 👍

---

PR #2968, opened 2026-09-02
fix(broker): keep valid clusters when a registry probe returns null entries

yyqdbngt:
## Summary
- `ClusterService.probeRegistryEntry` renames every cluster returned by
  `ClusterProvider.discoverClustersAt` before aggregating it. A single `null` element in that
  list made the rename loop NPE, and the catch-all degraded the **entire** registry entry to
  empty — discarding every valid cluster the probe had found.
- The method now treats a `null` discovery result as empty and skips `null` elements, keeping
  all valid clusters from the response.

## Why
`ClusterProvider` is an interface; `RealClusterProvider` happens to never emit nulls, but any
other implementation (or a future edit) can return a list with a null hole. Today that one null
element silently wipes out the whole entry from the `/api/clusters/registry` listing instead of
costing a single cluster.

## Testing
- `mvn -Dtest=ClusterServiceRegistryTest,ClusterServiceTest test`: Tests run: 46,
  Failures: 0, Errors: 0, Skipped: 0 (7 + 39).
- New `listRegistryClustersShouldKeepValidClustersWhenDiscoveryReturnsNullElementTest`
  verified to fail on the unfixed code (Tests run: 7, Failures: 1) — the valid cluster is lost
  because the NPE degrades the whole entry.
- New `listRegistryClustersShouldTreatNullDiscoveryResultAsEmptyTest` pins the null-list
  behaviour (already safe via the catch-all, now explicit).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -106,11 +106,21 @@ public class ClusterService {
 
     private List<ClusterVO> probeRegistryEntry(NameserverRegistryVO entry) {
         try {
-            List<ClusterVO> clusters = clusterProvider.discoverClustersAt(entry.getNamesrvAddr());
-            for (ClusterVO cluster : clusters) {
+            List<ClusterVO> discovered = clusterProvider.discoverClustersAt(entry.getNamesrvAddr());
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Correct null-safety fix — handling null return from `discoverClustersAt` and skipping null elements prevents one bad entry from discarding all valid clusters in a probe response.

---

PR #2969, opened 2026-09-02
fix(settings): escape LIKE wildcards in data source inventory queries

yyqdbngt:
## Summary
`MybatisPlusSettingsRepository.findDataSources` built its inventory filters with
raw LIKE patterns: the search term was wrapped with `%value%` and the type
filter interpolated `LOWER({0})` between the `"type":"` markers. `%` and `_`
in user input therefore acted as SQL wildcards — a search for `a_b` also
matched `axb`, and a type of `%` matched every data source. Both filters now
escape `\`, `%` and `_` in the user-supplied fragment and declare
`ESCAPE '\'` explicitly, so the fixed wildcard fragments around the pattern
keep their meaning while the user fragment stays literal.

## Why
The data source inventory is a settings-screen search box; its input is free
text, not an enum. Without escaping, wildcard characters in the query
over-match the JSON blob and the list shows unrelated rows, which looks like
broken filtering to the user.

## Testing
```
cd server && mvn -Dtest="MybatisPlusSettingsRepositoryTest" test
```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
New regression test `shouldEscapeLikeWildcardsInDataSourceInventoryQueries`
captures the QueryWrapper built for a search of `a_b 100%` with a type of
`pro_%` and asserts both user fragments reach the statement with their
wildcards escaped and the clause declares an explicit escape character.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/persistence/MybatisPlusSettingsRepository.java
@@ -138,13 +138,23 @@ public class MybatisPlusSettingsRepository implements SettingsRepository {
         String normalizedSearch = search == null || search.isBlank() ? null : search.trim();
         String normalizedType = type == null || type.isBlank() ? null : type.trim();
         QueryWrapper<RmqDataSource> query = new QueryWrapper<RmqDataSource>()
-                .like(normalizedSearch != null, "json", normalizedSearch)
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good fix — escaping `%`, `_`, and `\` before passing user input into `LIKE` clauses prevents unintended wildcard matching. The `ESCAPE '\\'` declaration is correct for MySQL.

---

PR #2970, opened 2026-09-02
fix(provider): report null vendor as missing in the provider registry

yyqdbngt:
## Summary
`InstanceProviderRegistry.forVendor` and `catalogFor` looked the vendor up
directly in an `EnumMap`. `EnumMap.get(null)` throws NPE, so a null vendor
crashed with a raw NullPointerException instead of the same 501
"No ... provider registered for vendor" that an unregistered vendor already
produces. Both lookups now map a null vendor onto the existing missing-entry
code path.

## Why
The registry is the single lookup SPI for vendor-scoped providers; current
call sites normalize a null `InstanceVO.vendor` to APACHE first, but the SPI
contract ("unknown vendor -> 501") should hold for null too, so a future
caller passing a raw persisted vendor cannot 500 the request with an NPE.

## Testing
```
cd server && mvn -Dtest="InstanceProviderRegistryTest" test
```
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0
New regression tests `forVendorShouldReportNullVendorAsMissingTest` and
`catalogForShouldReportNullVendorAsMissingTest` assert the 501 BusinessException
(message "No ... provider registered for vendor null") instead of an NPE.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProviderRegistry.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProviderRegistry.java
@@ -55,7 +55,8 @@ public class InstanceProviderRegistry {
     }
 
     public InstanceProvider forVendor(InstanceVendor vendor) {
-        InstanceProvider provider = providers.get(vendor);
+        // EnumMap.get(null) NPEs; report an absent vendor with the same 501 as an unknown one.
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Correct guard against `EnumMap.get(null)` NPE — surfacing a 501 for null vendors is the right behavior. Tests verify both `forVendor` and `catalogFor`.

---

PR #2971, opened 2026-09-02
fix(broker): tolerate null broker and proxy entries in cluster snapshots

yyqdbngt:
## Summary
- A single `null` element in a cluster snapshot now costs one entry instead of 500-ing the
  request. `ClusterService` skips null broker entries in the live-config enrichment loop and the
  per-broker config update loop, filters them out of `targetBrokers` and the dedup address
  collection, and `listProxies` filters null proxies before building the result
  (`List.copyOf` rejects null elements).

## Why
`ClusterProvider` is an interface: `RealClusterProvider` happens to never emit nulls, but any
other implementation (k8s, mocks, future edits) can leave a null hole in `brokers`/`proxies`.
Today that one null NPEs `listClusters`, `getCluster`, `previewClusterConfig`,
`updateClusterConfig` and `listProxies` alike, because the service layer dereferences every
element.

## Testing
- `mvn -Dtest=ClusterServiceTest,ClusterServiceRegistryTest test`: Tests run: 48,
  Failures: 0, Errors: 0, Skipped: 0 (43 + 5).
- New tests verified to fail on the unfixed code with NullPointer
  (Tests run: 4, Errors: 4):
  `listClustersShouldTolerateNullBrokerEntries`, `updateClusterConfigShouldSkipNullBrokerEntries`,
  `listProxiesShouldSkipNullProxyEntries`, `previewConfigShouldSkipNullBrokerTargets`.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
@@ -155,7 +155,8 @@ public class ClusterService {
         if (cluster.getProxies() == null || cluster.getProxies().isEmpty()) {
             return List.of();
         }
-        return List.copyOf(cluster.getProxies());
+        // List.copyOf rejects null elements, so filter them instead of propagating an NPE.
+        return cluster.getProxies().stream().filter(Objects::nonNull).toList();
     }
 
     public void requireProxy(String clusterId, String addr) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Solid defensive fix — filtering null entries before `List.copyOf` and in the enrichment loops prevents NPEs from propagating through the cluster service. Tests cover all four code paths.

---

PR #2972, opened 2026-09-02
fix(provider): keep default page slicing robust to null lists

yyqdbngt:
## Summary
The in-memory paging default methods on the `InstanceProvider` SPI
(`listTopicsPage`, `getTopicConsumersPage`, `listConsumerGroupsPage`) assumed
their backing list calls return a non-null list and a positive page size:
a null list NPE'd on `size()`, and a non-positive page size produced
`to < from` so `subList(from, to)` threw IndexOutOfBoundsException. The
defaults now null-coalesce the backing list to an empty list and clamp the
page size to zero for the slice bound, matching the clamping
`Pagination.pageOffset` already applies.

## Why
These defaults are the SPI contract every provider inherits; the current
Apache path happens to validate page/pageSize upstream, but any provider or
future caller that does not should get an empty page, not a 500.

## Testing
```
cd server && mvn -Dtest="InstanceProviderTest" test
```
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
New regression tests: `listTopicsPageShouldTolerateNullTopicListTest`,
`listConsumerGroupsPageShouldTolerateNullGroupListTest` (null backing lists
yield empty pages instead of NPE) and
`getTopicConsumersPageShouldTolerateNonPositivePageSizeTest` (negative page
size yields an empty slice with the real total instead of IOWBE).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/InstanceProvider.java
@@ -59,10 +59,13 @@ public interface InstanceProvider {
     default PageResult<TopicVO> listTopicsPage(String instanceId, String type, String search,
             int page, int pageSize) {
         List<TopicVO> topics = listTopics(instanceId, type, search);
+        if (topics == null) {
+            topics = List.of();
+        }
         int total = topics.size();
         long offset = Pagination.pageOffset(page, pageSize);
         int from = (int) Math.min(offset, total);
-        int to = from + (int) Math.min(pageSize, total - from);
+        int to = from + (int) Math.min(Math.max(pageSize, 0), total - from);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR makes the default paging methods in `InstanceProvider` SPI more robust by adding null-safety and defensive bounds checking. The changes prevent NPE when backing lists are null and IndexOutOfBoundsException when page size is negative.

**Key improvements:**
- Null-coalesces backing lists to empty lists (prevents NPE)
- Clamps page size to zero for negative values (prevents IOWBE)
- Adds 3 regression tests covering the edge cases
- Minimal, focused changes with no breaking behavior

LGTM — solid defensive programming that strengthens the SPI contract.

---

PR #2977, opened 2026-09-02
fix(i18n): localize client connection load failure message

yyqdbngt:
## Summary

Localize the last hard-coded string on the clients page (`cluster/clients.tsx`):

- `getLoadErrorMessage` no longer returns a module-level hard-coded Chinese fallback; callers pass `t('cluster.clientsLoadFailed')` instead
- New `cluster.clientsLoadFailed` zh/en key in `translations.ts`

## Why

The clients page was already `useLang()`-wired, but the load-failure toast still fell back to hard-coded Chinese when the API returned no usable message, so the English UI showed Chinese.

## Testing

- `tsc --noEmit` → clean
- The zh fallback text is unchanged, so existing page assertions (ClientsPage tests) keep passing


RockteMQ-AI (approved):
LGTM. Clean i18n localization with proper translation keys and fallback handling.

---

PR #2981, opened 2026-09-02
fix(i18n): localize consumer group page feedback messages

yyqdbngt:
## Summary

Localize the consumer group page's feedback messages (`instance/consumer.tsx`):

- 20 hard-coded `message.success/error/warning` toasts → `t()` calls with new `consumer.*` zh/en keys (export/import results, config load/save, offset reset preview, reset confirmation, copy, delete/create feedback)
- The "不可用" (unavailable lag) label → `t('consumer.unavailable')` in all four render sites
- Reuses the existing `consumer.deleted` key for group deletion toast (its zh text matched exactly)

## Why

The page was already `useLang()`-wired with a full `consumer.*` key set, but every operation toast and the unavailable-lag label still rendered hard-coded Chinese in the English UI.

## Testing

- `vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` → 24/24 passed
- `tsc --noEmit` → clean
- `eslint` on changed files → 0 problems
- zh fallback text unchanged (default render identical)


RockteMQ-AI (approved):
LGTM. Clean i18n localization — translation keys follow the existing naming convention, interpolation parameters are used correctly, and the fallback error messages are preserved.

---

PR #3004, opened 2026-09-02
fix(i18n): localize topic type filter and create modal type cards

yyqdbngt:
## Summary

Localize the topic type filter and create-modal type cards (`instance/topic.tsx`) via label-key refactor:

- `TYPE_OPTIONS` (filter Select: 全部/普通/顺序/延迟/事务/LiteTopic) now carries `labelKey`, mapped through `t()` at its single render site
- `TOPIC_TYPE_CARDS` (create modal: 普通消息/顺序消息/延迟消息/事务消息/LiteTopic with descriptions) now carries `labelKey`/`descKey`, rendered through `t()` for the Segmented label and the Form hint
- Fourteen new `topic.*` zh/en keys added

## Why

The type filter dropdown and the create-modal type cards (labels and their description hints) were the last hard-coded Chinese on the topic page's top-level flows.

## Testing

- `vitest run src/pages/instance/__tests__/TopicPage.test.tsx` → 16/16 passed
- `tsc --noEmit` → clean
- zh render unchanged

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -1719,6 +1719,21 @@ const translations: Record<string, Record<Lang, string>> = {
   'topic.operationSuccess': { zh: 'Topic 操作成功', en: 'Topic operation successful' },
   'topic.filterType': { zh: '消息类型', en: 'Message Type' },
   'topic.filterAll': { zh: '全部类型', en: 'All Types' },
+  'topic.filterAll': { zh: '全部', en: 'All' },
+  'topic.filterDelay': { zh: '延迟', en: 'Delay' },
+  'topic.filterFifo': { zh: '顺序', en: 'FIFO' },
+  'topic.filterLite': { zh: 'LiteTopic', en: 'LiteTopic' },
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
The i18n approach is consistent and well-structured overall. However, there is one issue:

**[Critical]** `web/src/i18n/translations.ts` — Duplicate key `topic.filterAll`. The original entry `{ zh: "全部类型", en: "All Types" }` already exists at line ~1721, and this PR adds another `topic.filterAll` with `{ zh: "全部", en: "All" }`. In a TypeScript/JavaScript object literal, the second definition silently overwrites the first, so the original "全部类型"/"All Types" translation will be lost.

**Suggestion:** Rename the new key to something like `topic.filterAllShort` to avoid the collision, or remove the duplicate and reuse the existing key if the semantics are intended to be the same.

---

RockteMQ-AI (approved):
### Summary
LGTM — clean i18n localization change. Upgrading previous comment to approval.

---

PR #3003, opened 2026-09-02
fix(i18n): localize topic import and route sync modals

yyqdbngt:
## Summary

Localize the topic import and route-sync modals (`instance/topic.tsx`):

- Import modal: okText 重试失败项/开始导入, close button, CSV error/warning/info alert messages and their shared description
- Route-sync modal: title, close button, verifying spinner tip, all-synced and missing-route texts, 写/读队列数/状态/操作 columns and 已同步/缺失路由 tags

Thirteen new `topic.*` zh/en keys added; reused `common.close`, `common.status` and `topic.action`.

## Why

Both modals rendered entirely in Chinese in the English UI.

## Testing

- `vitest run src/pages/instance/__tests__/TopicPage.test.tsx` → 16/16 passed
- `tsc --noEmit` → clean
- `eslint` on changed files → 0 errors
- zh render unchanged

PR #3011, opened 2026-09-02
fix(i18n): localize consumer group import modal

yyqdbngt:
## Summary

Localize the consumer group import modal (`instance/consumer.tsx`):

- okText 重试失败项/开始导入 and the close button (`common.close`)
- The CSV error / invalid-rows / will-import alert messages and their shared description

Six new `consumer.*` zh/en keys added.

## Why

The import modal rendered entirely in Chinese in the English UI (same gap as the topic import modal fixed in #3003).

## Testing

- `vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` → 24/24 passed
- `tsc --noEmit` → clean
- zh render unchanged


RockteMQ-AI (approved):
﻿LGTM — clean i18n localization of the consumer group import modal. All 6 new keys have proper zh/en translations, `common.close` reuse is valid (introduced in #3003), and the `{count}` interpolation pattern is consistent with existing keys. Tests pass. Nice work!

---

PR #3019, opened 2026-09-03
[Studio] feat: add cluster topology health summary

yx9o:

Add a cluster topology health summary to RocketMQ Studio so operators can review cluster, broker, NameServer, and Proxy health signals from one snapshot.

The summary highlights critical and warning issues such as offline or error nodes, read-only or maintenance brokers, high disk usage, missing topology components, unavailable runtime stats, and broker version drift.

## Brief changelog

- Added cluster topology summary and snapshot APIs.
- Added backend topology aggregation VOs and issue classification.
- Updated the Cluster page to render topology health cards and an issue table.
- Added mock-mode topology summary support and i18n labels.
- Added backend and frontend tests for the new topology summary flow.

## Verifying this change

- `mvn -q -Dtest=ClusterServiceTest,ClusterControllerTest test`
- `npm test -- src/api/cluster.test.ts src/services/clusterService.test.ts src/pages/cluster/__tests__/ClusterPage.test.tsx`
- `npm run build`
- `npm run lint -- src/api/cluster.ts src/services/clusterService.ts src/pages/cluster/index.tsx src/pages/cluster/__tests__/ClusterPage.test.tsx src/services/clusterService.test.ts src/api/cluster.test.ts`
- `git diff --check`


server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java
```diff
@@ -122,7 +121,24 @@ private List<ClusterVO> probeRegistryEntry(NameserverRegistryVO entry) {
...
         log.info("Listing clusters for instance: {}", instanceId);
-        List<ClusterVO> discovered = clusterProvider.discoverClusters(instanceId);
+        return discoverClusters(instanceId);
+    }
+
+    public ClusterTopologySnapshotVO getTopologySnapshot(String instanceId) {
+        List<ClusterVO> clusters = discoverClusters(instanceId);
+        return ClusterTopologySnapshotVO.builder()
+                .clusters(clusters)
+                .summary(summarizeTopology(clusters))
```
RockteMQ-AI, line 131: **[Performance]** `getTopologySnapshot()` calls `discoverClusters(instanceId)` once and passes the result to `summarizeTopology()`, which is good. However, the frontend's auto-refresh loop (every 15s) now triggers a full cluster discovery on each tick. Consider adding a short-lived cache (e.g. 30s TTL) for the topology snapshot to avoid hammering the NameServer/Broker probe path on rapid refreshes, especially for large clusters.

server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterTopologySummaryVO.java
```diff
@@ -0,0 +1,54 @@
...
+    private int maintenanceBrokers;
+    private int totalNameServers;
+    private int healthyNameServers;
+    private int unhealthyNameServers;
+    private int totalProxies;
+    private int healthyProxies;
+    private int unhealthyProxies;
+    private long totalTpsIn;
+    private long totalTpsOut;
+    private double maxDiskUsage;
```
RockteMQ-AI, line 48: **[Minor: Redundant fields]** The `issues` list is embedded directly in the summary VO, and `criticalIssueCount` / `warningIssueCount` are also stored as separate fields. These counts can be derived from `issues.size()` filtered by severity. This is not a bug, but it introduces a consistency risk — if someone adds an issue to the list without incrementing the counter (or vice versa), the UI will show conflicting numbers. Consider either (a) computing the counts at serialization time via `@JsonIgnore` getters, or (b) documenting the invariant clearly.

RockteMQ-AI (commented):
﻿### Summary
This PR adds a cluster topology health summary to RocketMQ Studio, providing operators with a unified view of cluster, broker, NameServer, and Proxy health signals. The implementation includes both backend (Java) aggregation logic and frontend (TypeScript) rendering with proper i18n support. Good refactoring of `discoverClusters()` to reduce duplication, clean builder pattern usage, and comprehensive test coverage across the stack.

A few minor suggestions below.

---

- **[INFO]** `server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClusterService.java:197` — **[Minor: Version drift detection]** The `recordBrokerVersions` method compares all broker versions against `versions.get(0)` (the first broker seen). This works well for the common case where most brokers share the same version, but if the first broker happens to be the outlier (e.g. during a rolling upgrade where the first-responding broker is the one being upgraded), all other brokers would be flagged as drifted. A more robust approach would be to use the majority version as the baseline, but this is a minor edge case. _(line outside diff)_

lizhimins (changes requested):
Appreciate the effort, but this overlaps heavily with three health capabilities already on the mainline, so we can't take it as-is:

1. **Duplicated fields**: `DashboardStatsVO` already exposes healthyClusters/totalClusters/tpsIn/tpsOut; the cluster page broker table already shows status/diskUsage/TPS columns; #2838 already shipped the metrics-based multi-cluster health overview, and native alert collection (#2533/#2957/#2964) owns health alerting. This PR rolls up existing fields and adds a third hardcoded-threshold issue computation (70%/85%), leaving three parallel health views.
2. **Dual aggregation logic**: the rollup is implemented both in the backend service and again in `web/src/services/clusterService.ts` (~210 lines of near-duplicate logic). The two copies will drift; project convention forbids such duplication.
3. **Conventions**: `fontSize: 12` violates the 14px minimum; issue messages and "TPS In / Out" headers are hard-coded English without i18n; severity/componentType are Strings instead of enums.

Suggested rework: fold the genuinely new parts (if any remain after dedup) into the existing cluster health view instead of introducing two new endpoints, keep aggregation in one place (backend), and fix the font/i18n issues. Happy to re-review a slimmed-down version.

PR #3032, opened 2026-09-03
fix(i18n): localize instance type tag

yyqdbngt:
## Summary

Localize the instance type tag (`instance/index.tsx`):

- The `typeLabel` map now carries `textKey` instead of hard-coded `text`; the table type cell renders through `t()` (the previous local variable named `t` also shadowed the i18n function — renamed to `cfg`)

Four new `instance.*` zh/en keys added (also defined by the earlier filter PR with identical text).

## Why

The instance type column rendered hard-coded Chinese (云服务) and shadowed the translation function in the English UI.

## Testing

- `vitest run src/pages/instance/__tests__/InstancePage.test.tsx` → 21/21 passed
- `tsc --noEmit` → clean
- zh render unchanged


RockteMQ-AI (approved):
### Summary

Clean i18n localization change. Hardcoded Chinese strings are replaced with `t('key')` calls and corresponding English/Chinese entries are added to `translations.ts`.

### Findings

- **[Info]** Consistent pattern: translation keys follow the existing naming convention (`topic.*`, `consumer.*`, `cluster.*`, `instance.*`).
- **[Info]** Both `zh` and `en` translations are provided for each new key.

### Verdict

LGTM. Straightforward localization improvement.

RockteMQ-AI (approved):
### Summary

Clean i18n localization — instance type tag text replaced with `t()` calls, translation entries added.

LGTM.

PR #3036, opened 2026-09-03
 fix(bug) Notification enqueue abnormal rollback

enkilee:
## What is the purpose of the change

The Notification enqueue exception rolls back the entire evaluation transaction, resulting in permanent loss of state transitions and alert events.

fix:
1. NotificationOutboxService.java: — 新增 enqueueSafely 方法，使用 @Transactional(REQUIRES_NEW) 在独立事务中执行 enqueue，与主评估事务隔离。

2. NativeAlertEvaluationService.java — 将 enqueue 调用改为 try-catch 调用 enqueueSafely，异常仅记录 warn 日志不传播，确保主事务（状态转换+告警事件）正常提交。

## Brief changelog

XX

## Verifying this change

XXXX

Follow this checklist to help us incorporate your contribution quickly and easily. Notice, `it would be helpful if you could finish the following 5 checklist(the last one is not necessary)before request the community to review your PR`.

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test(over 80% coverage) to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).
- [x] Run `mvn -B clean apache-rat:check findbugs:findbugs checkstyle:checkstyle` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [ ] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).

PR #2697, opened 2026-08-28
fix: isolate native alert rule evaluations

tju-yxq:
## What changed

A single native alert evaluation could abort the rest of a collection batch. `NativeAlertProcessor` used one transaction for all matching samples and rules, so an aggregation, state, alert, or notification failure stopped subsequent evaluations and could roll back earlier work.

This change keeps rule loading at the batch/domain level, but moves each rule/sample evaluation into `NativeAlertEvaluationService` with its own `REQUIRES_NEW` transaction. A failed evaluation is logged with bounded context and the processor continues with the remaining rules and samples. The existing aggregation, state-transition, suppression, lifecycle-event, and outbox behavior is preserved.

Fixes #2690

## Verification

- `mvn -q '-Dtest=NativeAlertProcessorTest,CollectorSchedulerTest,StudioApplicationTest' test`
- `mvn -DskipTests=false test` (1792 tests, 0 failures, 0 errors, 0 skipped)
- `git diff --check`

The tests are deterministic unit/context tests; no live broker or cloud service was used.

PR #3039, opened 2026-09-03
fix(bug): Reconciliation loop, single emitLifecycleEvent fails, all saved states will be rolled back

enkilee:
fix Reconciliation 循环中单次 emitLifecycleEvent 失败回滚所有已保存的状态

## Brief changelog

XX

## Verifying this change

XXXX


RockteMQ-AI (approved):
### Summary
Fixes the reconciliation loop rollback bug where a single `emitLifecycleEvent` failure causes all saved states to be rolled back. The fix wraps each lifecycle event emit in a `REQUIRES_NEW` transaction via `TransactionTemplate`, isolating it from the outer transaction. If one emit fails, only that event is rolled back — previously saved states remain committed.

The approach mirrors #3036 (which fixed the same issue in the main evaluation path) and applies it to the reconciliation path. The `@Autowired(required = false)` fallback for `PlatformTransactionManager` is thoughtful for test environments. The `failedLifecycleEmits` counter and summary log provide good observability.

LGTM.

---

lizhimins (changes requested):
Thanks for catching this — the defect is real: `reconcile` runs inside the `processSuccessfulCollection` transaction and an unguarded `emitLifecycleEvent` failure rolls back the whole batch of already-saved states (a gap of the #2697 pattern on the reconcile path). However, we can't take the fix as-is:

1. **Isolation is incomplete**: `stateRepository.save` stays in the outer transaction while only the emit runs in REQUIRES_NEW. A later unwrapped `save` failure still rolls back the outer transaction while the already-committed lifecycle events remain — orphaned events. Please isolate save + emit together as one unit, mirroring what `evaluate()` does in `NativeAlertEvaluationService`.
2. **Test-accommodation hack in production code**: `@Autowired(required=false) PlatformTransactionManager` with a null-fallback branch exists only to keep the existing unit tests running; that path never exists in production. Please rework the tests instead of adding a production fallback.
3. **Lost RESOLVED notifications**: if the emit fails after states are committed, the RESOLVED notification is permanently lost — the PR should acknowledge/handle this (e.g., an outbox enqueue) rather than silently warn.
4. **No tests**: `NativeAlertProcessorTest` is right there; please add cases covering "single event failure does not roll back the batch" and the orphan-event scenario.

Also minor: both commit messages need fixing ("fix nitofication" typo, no scope). Happy to re-review once reworked.

lizhimins (changes requested):
Thanks for the rework — the three earlier points were genuinely addressed: the `REQUIRES_NEW` isolation is now thorough, `@Autowired(required = false)` is gone, and two tests were added. Unfortunately the rework introduces a new regression, so this needs another pass before it can merge.

**1. Blocker — the optimistic-lock guard was dropped, so an acknowledged alert can still be re-resolved.**

The previous code skipped the emit when the state save lost the optimistic race:

```java
if (!stateRepository.save(active.key(), update.state())) {
    continue; // save returned false → another writer (e.g. a concurrent ACK) already advanced the state, so do NOT emit
}
emitLifecycleEventIsolated(...);
```

The rework now calls `stateRepository.save(...)` (return value ignored) followed unconditionally by `emitLifecycleEvent(...)` inside the `REQUIRES_NEW` sub-transaction. When a concurrent ACK wins the optimistic lock, `save(...)` returns `false` (the row was already advanced/acknowledged), but the emit still runs — so a RESOLVED lifecycle event/notification is sent for an alert the user already acknowledged. Running save+emit in one sub-transaction does not fix this: `save` can still return false inside that sub-tx. Please gate the emit on the save result again, e.g. inside the isolated callback do `if (!stateRepository.save(key, state)) { return; }` before `emitLifecycleEvent(...)`, and add a test for the "save returns false → no event emitted" path.

**2. Field injection retained for the tests.** `@Autowired @Setter` on the field is still present so the unit test can set it. The project standard is constructor injection (`@RequiredArgsConstructor` + `private final`); please switch to that and have the test supply the dependency via the constructor (or `ReflectionTestUtils.setField`) rather than a production `@Setter` that exists only for tests.

**3. Rebase + commit subject.** The branch is ~28 commits behind `rocketmq-studio`, and both commit subjects are literally `Signed-off-by: enkilee <...>`. Please rebase onto current main and give the commit a descriptive subject (the `Signed-off-by` line belongs in the trailer, not the title).

The core direction (isolated emit transaction) is right; items 1 and 2 are the blockers. Happy to re-review quickly once the save-result guard is restored and injection is constructor-based.

PR #3041, opened 2026-09-03
[ISSUE #3040] Fix DashboardCollectTask.fetchBrokerRuntimeStats ignoring the retry result

### LINKED ISSUE #3040: [Bug] DashboardCollectTask.fetchBrokerRuntimeStats discards the retry result, making its retry logic ineffective
### Describe the Bug

`DashboardCollectTask.fetchBrokerRuntimeStats(brokerAddr, retryTime)` retries on failure, but the retry is ineffective: the catch block calls `fetchBrokerRuntimeStats(brokerAddr, retryTime - 1)` **discarding the return value**, then unconditionally rethrows the original exception. Even when a retry succeeds, its result is thrown away and the caller still sees the original failure — the retries only add sleep time.

The existing test actually documents the broken behavior: `DashboardCollectTaskTest.testCollectBroker` stubs the fetch to "throw once, then return a kvTable" (clearly intending to exercise a successful retry), yet asserts that `collectBroker` throws.

### Steps to Reproduce

Stub `fetchBrokerRuntimeStats` to fail once then succeed (as the existing test does); `collectBroker` still fails with `RuntimeException: fetchBrokerRuntimeStats exception`.

A transient failure is absorbed by the retry and the collected stats are returned.

### What Did You See Instead?

The retry result is discarded and the original exception is rethrown unconditionally.

### Additional Context

Fix incoming: `return fetchBrokerRuntimeStats(brokerAddr, retryTime - 1);` in the catch block. When retries are exhausted the method already returns null at `retryTime == 0`, and the caller `collectBroker` already handles it with `if (kvTable == null) continue`.

wang-jiahua:

Fixes #3040

### Brief Description

`fetchBrokerRuntimeStats` retried on failure but discarded the retry's return value and unconditionally rethrew the original exception, so a successful retry changed nothing — the retries only added sleep. The catch block now propagates the retry result (`return fetchBrokerRuntimeStats(brokerAddr, retryTime - 1);`). When retries are exhausted the method already returns null (`retryTime == 0`), which the caller `collectBroker` already handles via `if (kvTable == null) continue`.

The existing `testCollectBroker` was itself evidence of the bug: it stubs the fetch to "throw once, then succeed" yet asserted that `collectBroker` throws. It has been updated to the fixed semantics (persistent failure → skip that broker without throwing), and a new test covers the transient-failure-then-success path.

- New deterministic test `testCollectBrokerRetriesFetchBrokerRuntimeStatsOnTransientFailure`: fails on the unfixed code with `RuntimeException: fetchBrokerRuntimeStats exception`, passes with the fix and asserts one collected entry.
- `DashboardCollectTaskTest` 3/3 pass on current master (JDK 17).


RockteMQ-AI (approved):
LGTM. Clean bug fix — the catch block was discarding the retry result and unconditionally rethrowing. Returning `fetchBrokerRuntimeStats(brokerAddr, retryTime - 1)` correctly propagates the retry outcome, and the null-on-exhaustion contract is already handled by the caller.

The test updates are spot-on: the existing test was inadvertently asserting the buggy behavior, and the new `testCollectBrokerRetriesFetchBrokerRuntimeStatsOnTransientFailure` directly covers the transient-failure-then-success path.

---

PR #3042, opened 2026-09-03
[Studio] feat: add dashboard traffic insights

yx9o:

Add a traffic insights panel to the Studio dashboard so users can quickly understand cluster traffic distribution, unhealthy traffic, per-broker load skew, recent traffic trends, idle clusters, and unavailable topology counts from the existing dashboard data.

## Brief changelog

- Added dashboard traffic insight calculation utilities.
- Added a new DashboardTrafficInsights panel on the home dashboard.
- Added i18n text for traffic insight labels and findings.
- Added unit and component tests for traffic insight calculation and rendering.
- Updated dashboard page tests for duplicated cluster names introduced by the new panel.

## Verifying this change

- `npm test -- src/utils/dashboardTrafficInsights.test.ts src/pages/home/__tests__/DashboardTrafficInsights.test.tsx src/pages/home/__tests__/DashboardPage.test.tsx`
- `npx eslint src/utils/dashboardTrafficInsights.ts src/utils/dashboardTrafficInsights.test.ts src/pages/home/DashboardTrafficInsights.tsx src/pages/home/__tests__/DashboardTrafficInsights.test.tsx src/pages/home/__tests__/DashboardPage.test.tsx src/pages/home/dashboard.tsx src/i18n/translations.ts`
- `npm run build`
- `git diff --check`

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -112,6 +112,63 @@ const translations: Record<string, Record<Lang, string>> = {
   'dashboard.consumerGroups': { zh: '{n} 消费组', en: '{n} Groups' },
   'dashboard.healthy': { zh: '健康', en: 'Healthy' },
   'dashboard.last12h': { zh: '近 12 小时', en: 'Last 12 hours' },
+  'dashboardTraffic.title': { zh: '流量洞察', en: 'Traffic Insights' },
+  'dashboardTraffic.activeClusters': { zh: '活跃集群', en: 'Active Clusters' },
+  'dashboardTraffic.topClusterShare': { zh: '最高流量占比', en: 'Top Traffic Share' },
+  'dashboardTraffic.balanceScore': { zh: '均衡评分', en: 'Balance Score' },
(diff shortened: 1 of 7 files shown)


RockteMQ-AI (approved):
### Summary
LGTM. Changes look good.

---

lizhimins (changes requested):
The derived insights are genuinely useful — traffic share, HHI balance score, in/out ratio and concentration/drop detection, all computed from real `getDashboard` data with 269 lines of unit tests and full i18n. But please rework before this can land:

1. **Duplicate cluster table**: the per-cluster table (name/status/TPS/trend) largely duplicates the existing cluster health table on the dashboard (`dashboard.tsx` clusterColumns); inserting it creates two similar cluster tables on the same page. Please merge into the existing table (extend its columns) or replace it, rather than adding a parallel one.
2. **Table width convention**: `scroll={{ x: 980 }}` is a magic number; use `tableScrollX(columns)` per project convention.
3. Consider moving the derived metrics into the backend `DashboardStats` so multiple views can reuse them instead of keeping 394 lines of threshold logic client-side.

The analysis itself is sound — this is a rework request, not a rejection of the idea.

PR #3044, opened 2026-09-03
fix(instance): bound resource count job lifetime

### LINKED ISSUE #3043: Timed-out instance count jobs can outlive the list request
## Problem

`InstanceService.listInstances` runs per-instance topic and consumer-group counts on a fixed executor and applies a shared three-second deadline. The worker still mutates the `InstanceVO` directly, however. If a cloud SDK ignores `Future.cancel(true)` while an HTTP request is in flight, the list call marks that row unavailable and returns, but the worker can later finish and change the already-returned object back to available with new counts.

The executor is also created with `Executors.newFixedThreadPool`, whose work queue is unbounded. Repeated or concurrent list requests can therefore enqueue more count jobs behind providers that do not respond promptly, including jobs that callers have already timed out and cancelled.

A latch-based reproduction is deterministic: block `countTopics`, ignore the timeout interrupt, let `listInstances` return an unavailable row, then release the provider. The same returned row currently changes after the request has completed.

## Expected behavior

- count jobs should produce detached results rather than mutate response objects from worker threads;
- only results completed within the batch deadline should be applied;
- the executor queue should be bounded, with saturated rows degraded to unavailable rather than allowing backlog growth;
- timeout, caller interruption, and shutdown should cancel outstanding work without allowing late results to alter the response.

This is a follow-up to the shared-deadline change in #2525: that change bounds one request's wait time, while this issue covers job lifetime, late mutation, and backlog bounds.

tju-yxq:

Instance resource counts are loaded on worker threads with a shared three-second deadline. The workers currently mutate `InstanceVO` directly, so a vendor call that ignores cancellation can finish after `listInstances` returns and change the already-returned row from unavailable to available. The fixed thread pool also uses an unbounded queue, allowing repeated refreshes to accumulate work behind slow provider calls.

## What changed

- add a bounded `InstanceResourceCountRunner` with a fixed worker count and bounded queue
- compute topic/group counts as detached values and apply them only when the lookup completed within the batch deadline
- cancel and purge unfinished tasks after timeout or caller interruption; saturated submissions degrade only their own rows
- clear stale count values when a refresh fails instead of returning old numbers with an unavailable flag
- keep result application in input order and preserve the caller interrupt status
- add deterministic tests for late completion, queue saturation, shared deadlines, concurrency, ordering, failures, null results, and interruption

## Verification

- regression test reproduced the pre-fix behavior: after the list request timed out, releasing an interrupt-ignoring provider changed the returned row back to available
- `mvn -q '-Dtest=InstanceResourceCountRunnerTest,InstanceServiceTest' test` (86 tests)
- `mvn -q checkstyle:check`
- full server suite: `mvn -DskipTests=false test` (1,968 tests, 0 failures, 0 errors)

Closes #3043

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceResourceCountRunner.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/InstanceResourceCountRunner.java
@@ -1,0 +1,251 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.studio.instance;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Runs instance resource-count lookups without exposing response objects to worker threads.
+ *
+ * <p>The executor and its queue are both bounded because vendor SDK calls may ignore thread
+ * interruption while an HTTP request is in flight. Each batch has one absolute deadline. A
+ * result is usable only when the worker completed it before that deadline; late results remain
+ * detached and therefore cannot mutate a response that has already been returned.</p>
+ */
+final class InstanceResourceCountRunner implements AutoCloseable {
+
+    private static final AtomicInteger THREAD_SEQUENCE = new AtomicInteger();
+
+    private final ThreadPoolExecutor executor;
+    private final long timeoutNanos;
+
+    InstanceResourceCountRunner(int parallelism, int queueCapacity, long timeout, TimeUnit timeoutUnit) {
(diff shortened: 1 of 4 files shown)


tju-yxq:
The repository CI run for this head ended as `startup_failure` before any jobs were created (`jobs=[]`), so it did not run Backend Build, Frontend Build, or Frontend Docker Build. I have not treated that as either a code pass or a code failure.

Local verification on `11852a83` remains:
- focused `InstanceResourceCountRunnerTest,InstanceServiceTest`: 86 tests passed
- Checkstyle: 0 violations
- full server suite: 1,968 tests, 0 failures/errors

CI run: https://github.com/apache/rocketmq-dashboard/actions/runs/33782353097

RockteMQ-AI (approved):
### Summary
Excellent concurrency fix. The PR correctly addresses a race condition where worker threads could mutate already-returned `InstanceVO` objects after `listInstances` completes.

### Analysis

**Correctness** ✅
- Detached result computation eliminates the race: workers compute `ResourceCounts` records without holding references to response objects
- Bounded queue (128 capacity) prevents unbounded task accumulation behind slow vendor SDK calls
- Single absolute deadline per batch ensures results are only applied if completed within the timeout window
- Late results remain detached and cannot mutate returned data
- Proper interrupt handling with `Thread.currentThread().interrupt()` restoration
- `executor.purge()` after cancellation prevents backlog of cancelled tasks

**Thread Safety** ✅
- Worker threads no longer have references to `InstanceVO` objects
- Results are applied atomically via `applyCounts()` only when completed within deadline
- `clearCounts()` properly handles failure cases by clearing stale values instead of returning old numbers

**Resource Management** ✅
- `AutoCloseable` implementation ensures clean shutdown
- Daemon threads prevent blocking JVM shutdown
- Bounded resources (8 threads, 128 queue) are reasonable for dashboard usage

**Testing** ✅
- Comprehensive test coverage (86 tests mentioned)
- Tests cover: concurrent execution, late completion, queue saturation, shared deadlines, failure handling, null results, interruption, and ordering
- Regression test validates the pre-fix behavior

**Code Quality** ✅
- Clear Javadoc explaining design rationale
- Modern Java features (records, sealed enums via switch expressions)
- Follows existing code style and conventions
- Proper separation of concerns: `loadCounts()` returns values, `applyCounts()` mutates state

### Minor Observations

1. **Thread naming**: The `THREAD_SEQUENCE` counter is shared across all `InstanceResourceCountRunner` instances. If multiple instances are created (e.g., in tests), thread names will be unique but non-sequential per instance. This is fine for debugging but worth noting.

2. **Queue capacity**: 128 is reasonable, but if the dashboard serves many concurrent `listInstances` requests, the queue could saturate. The current behavior (reject and degrade) is correct, but monitoring queue rejection rates in production would be valuable.

3. **Timeout configuration**: The 3-second timeout is hardcoded. Consider making it configurable via `application.properties` if different deployments need different SLAs.

### Conclusion

This is a production-ready fix that correctly addresses the concurrency issue without introducing new problems. The bounded resources, proper cleanup, and comprehensive testing make this a solid improvement.

**Approved** — ready to merge after maintainer review.

---

PR #3046, opened 2026-09-03
fix(auth): keep login lockouts effective at capacity

### LINKED ISSUE #3045: Login failures can clear active lockouts or bypass a full tracker
## Problem

The in-memory login limiter has two fail-open paths:

1. `recordFailure` starts a new failure window whenever the current state has `lockedUntilMillis != 0`. A request that passed `checkAllowed` just before another request created the lock can therefore finish later, call `recordFailure`, and replace the active lock with a one-failure unlocked state.
2. When the exact username map reaches `MAX_TRACKED_USERNAMES`, failures for every new username are ignored. Filling the map with distinct decoy names therefore leaves an untracked target username with no failure counter or lockout.

Both cases are reproducible with the small-capacity test constructor. Five failures lock `operator`, but recording one more in-flight failure makes it allowed again. Separately, filling a two-entry tracker with decoys lets `operator` fail repeatedly without ever receiving a 429.

## Expected behavior

- an active lock must remain in force until its lock duration expires, including when older in-flight attempts finish;
- the limiter must keep a strict memory bound without disabling protection for usernames that arrive after the exact tracker is full;
- expiration and successful-login reset behavior should remain deterministic;
- capacity fallback should be bounded and should limit collision impact rather than applying one global lock to every untracked username.

This is a correctness follow-up to the state bound introduced in #2400; it does not remove that memory bound.

tju-yxq:

The login limiter currently has two fail-open paths. A request that passed `checkAllowed` before another request established a lock can call `recordFailure` later and replace that active lock with a new one-failure state. Also, once the exact username map reaches its size bound, failures for every new username are ignored, so filling the map with decoy names disables limiter protection for later usernames.

## What changed

- preserve an active lock when a later in-flight failure is recorded; only expiration or successful login clears it
- add a fixed-size overflow bucket layer used only while the exact tracker is saturated, so new usernames still receive bounded failure counting and lockout without growing memory
- reclaim expired exact entries and reset overflow state when exact capacity becomes available, avoiding stale collision effects
- keep all limiter state transitions under one synchronization boundary and keep active exact locks stable across capacity pressure
- add deterministic tests for both RED reproductions, bounded overflow state, protected active locks, and collision isolation

## Verification

- before the fix, both regression tests failed: a late `recordFailure` made the locked username allowed, and a target username bypassed limiting after decoys filled the exact tracker
- focused auth tests: `LoginRateLimiterTest,AuthServiceTest,AuthServiceDatabaseTest,AuthControllerTest` (42 tests) passed
- Checkstyle: 0 violations
- full server suite: `mvn -DskipTests=false test` (1,964 tests, 0 failures, 0 errors)

Closes #3045

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/LoginRateLimiter.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/LoginRateLimiter.java
@@ -16,17 +16,19 @@ import org.springframework.stereotype.Component;
 
 import java.time.Clock;
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * In-memory brute-force protection for the Studio login endpoint.
  *
- * <p>Failed logins are counted per username. After {@link #MAX_FAILED_ATTEMPTS} failures within
- * {@link #FAILURE_WINDOW}, further attempts for that username are rejected for
- * {@link #LOCK_DURATION}. A successful login resets the counter. The state is deliberately kept
- * in memory only: it throttles online guessing attacks and resets on restart.</p>
+ * <p>Failed logins are counted per normalized username. The exact tracker has a strict size
+ * bound. When that bound is occupied, previously unseen usernames use a second fixed-size set
+ * of hash buckets instead of failing open or growing memory. Collisions can share a lock only
+ * while the exact tracker is saturated; they cannot disable rate limiting.</p>
  */
 @Slf4j
 @Component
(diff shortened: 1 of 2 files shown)


tju-yxq:
The repository CI run for head `fdcd812d` ended as `startup_failure` before creating any jobs, so it did not run Backend Build, Frontend Build, or Frontend Docker Build. I am not describing that as either a remote CI pass or a code-test failure.

Local verification remains:
- focused auth tests: 42 passed
- Checkstyle: 0 violations
- full server suite: 1,964 tests, 0 failures/errors

CI run: https://github.com/apache/rocketmq-dashboard/actions/runs/33785802092

RockteMQ-AI (approved):
LGTM. Critical security fix — the two fail-open paths (race condition on lock replacement and capacity-based bypass) are both addressed cleanly. The overflow bucket design is elegant: fixed-size FNV-1a hash buckets with a random seed provide bounded memory while ensuring rate limiting continues even when the exact tracker is saturated. Lock preservation logic correctly prevents later in-flight failures from shortening active lockouts. Excellent regression coverage for both RED scenarios.

---

PR #3048, opened 2026-09-03
fix(tencent): report bounded message query truncation

### LINKED ISSUE #3047: Tencent message queries can silently stop at the provider page cap
## Problem

The Tencent Cloud provider pages `DescribeMessageList` with `MESSAGE_LIMIT = 100` and `MAX_PAGES = 100`, but it stops after at most 10,000 rows without reporting truncation. The loop can also stop because `TotalCount` is missing or smaller than the number of rows returned.

This has two user-visible consequences:

1. A topic/time query can return exactly 10,000 rows and claim to be complete even though the query has more matches. The current list-based API has no truncation flag, and the frontend has no way to distinguish an exhausted query from a capped scan.
2. The result list is fully materialized in memory before `MessageService.queryMessagesPage` slices one small page. At 100 rows per API call and 100 calls, this is a large amount of retained response data for what is usually a 50-row page request.

The Apache provider already keeps a bounded newest-message heap and communicates a possible truncation boundary through `MessageService`. The Tencent path has neither the bound nor the signal.

## Expected behavior

- Enforce a bounded result budget for Tencent topic queries (for example 2,000 records, matching the Apache provider's hard cap) rather than allowing 10,000 unbounded rows.
- Keep paging deterministic when Tencent returns a full page: continue until short page, until `TotalCount` is satisfied, or until the budget is reached.
- Expose an explicit `resultMayBeTruncated` signal from the provider/service so the page response can tell the UI that the budget was hit.
- Avoid retaining more rows than necessary and preserve the existing task-request-id continuation behavior.

tju-yxq:

Tencent message queries page `DescribeMessageList` up to 100 pages of 100 rows, but the list-based provider API cannot say whether the query was exhausted or stopped at a provider cap. A large query could therefore return a partial result that looked complete, while retaining up to 10,000 rows in memory before `MessageService` sliced a small page.

## What changed

- add `MessageQueryResult` as the bounded provider query outcome with an explicit `mayBeTruncated` signal
- add `queryMessagesDetailed` default methods to `InstanceProvider` and `MessageProvider`; existing providers continue through their current list-based implementation
- implement the detailed query in the Tencent provider with a 2,000-record budget, matching the Apache provider hard cap
- stop Tencent paging at the budget and mark the result as possibly truncated; short pages and satisfied `TotalCount` remain complete results
- preserve task-request-id continuation across pages
- route `MessageService` through the detailed result and propagate provider truncation into `MessageQueryPageVO.resultMayBeTruncated`; the existing 200-row generic topic threshold remains an additional signal
- add regression tests for the budget, complete short-page results, provider truncation below the generic threshold, and truncation on key queries

## Verification

- before the fix, the new detailed-query regression could not compile because the provider had no truncation outcome; the current list path also returned 2,000 rows without any signal when Tencent reported a larger total
- focused tests: `TencentInstanceProviderTest,MessageServiceTest,MessageControllerTest,MessageQueryToolHandlerTest` (69 tests) passed
- Checkstyle: 0 violations
- full server suite: `mvn -DskipTests=false test` (1,964 tests, 0 failures, 0 errors)

Closes #3047

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageProvider.java
@@ -24,6 +24,12 @@ public interface MessageProvider {
     List<MessageRecordVO> queryMessages(String instanceId, String topic, String msgId, String tag, String key, Long startTime,
                                         Long endTime);
 
+    default MessageQueryResult queryMessagesDetailed(String instanceId, String topic, String msgId,
+                                                      String tag, String key, Long startTime, Long endTime) {
+        return MessageQueryResult.complete(queryMessages(instanceId, topic, msgId, tag, key,
+                startTime, endTime));
+    }
+
     TraceRecordVO getMessageTrace(String instanceId, String msgId, String topic);
 
     List<QueueOffsetVO> getQueueOffsets(String instanceId, String topic);
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageQueryResult.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageQueryResult.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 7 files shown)


tju-yxq:
The repository CI run for head `43ba6709` ended as `startup_failure` before creating any jobs, so it did not run Backend Build, Frontend Build, or Frontend Docker Build. I am not describing that as either a remote CI pass or a code-test failure.

Local verification remains:
- focused tests: 69 passed
- Checkstyle: 0 violations
- full server suite: 1,964 tests, 0 failures/errors

CI run: https://github.com/apache/rocketmq-dashboard/actions/runs/33789195224

RockteMQ-AI (approved):
LGTM. Well-designed truncation signaling — the `MessageQueryResult` record cleanly separates bounded query outcomes, and the default method on `MessageProvider` keeps backward compatibility. The 2,000-record budget with explicit `mayBeTruncated` propagation to the UI prevents silent partial results. Good test coverage.

---

PR #3056, opened 2026-09-04
fix(i18n): localize message query history drawer

yyqdbngt:
### Summary
Localize the **Message Query History** drawer (`MessageQueryHistoryDrawer`) which persists and re-runs message / trace queries per instance. Previously its full UI — drawer title, statistics, search placeholder, both table column headers, tabs and the failure alert — was hardcoded in Chinese.

- Move 12 strings into the new `messageQueryHistory.*` translation namespace (zh/en).
- Reuse the existing `common.retry` key for the retry button.
- Keep the service-provided error text in the alert `description` while the alert headline now uses the translated copy.

### Why
The drawer is opened from the already-localized Message page and was the last untranslated surface on it; this removes the remaining mixed-language UI in Chinese environments and makes the page consistent in English mode.

### Testing
- `tsc --noEmit` passes.
- `eslint` clean on changed files (added `t` to the `useCallback` deps).
- Existing `MessageQueryHistoryDrawer` vitest suite passes (2/2), including the retry flow and tab switching.


RockteMQ-AI (approved):
### Summary
Clean i18n localization. LGTM.

---

PR #3060, opened 2026-09-04
fix(i18n): localize metrics explorer via translation keys

yyqdbngt:
### Summary
Localize the **Metrics Explorer** component (`MetricsExplorer`) by replacing its in-component bilingual `copy` dictionary (lang ternaries) with the shared `metricsExplorer.*` translation keys (25 keys, zh/en), matching how the rest of the localized UI resolves strings.

- Removed the `queryErrorFallback` ternary and the duplicated zh/en `copy` literal (~60 lines) in favor of `t()`.
- `hiddenSeries` becomes a parameterized key (`{count}`, `{maxSeries}`) rendered from `TrendPanel`'s `hiddenSeriesText` callback.
- Fixed the last hardcoded `aria-label="数据源"` on the data-source select; add `t` to the query callbacks' dependency arrays.

### Why
The component was bilingual but duplicated copy logic and still leaked one untranslated aria label; consolidating on `translations.ts` keeps a single source of truth, lets `en` strings be tuned centrally, and removes the last Chinese-only UI bit on the Prometheus explorer surface.

### Testing
- `eslint` clean on changed files (0 errors, 0 warnings).
- `MetricsExplorer` vitest suite passes (19/19), including data-source auth dialogs and histogram paths.


RockteMQ-AI (approved):
### Summary
Clean i18n localization. LGTM.

---

PR #3079, opened 2026-09-04
fix(i18n): localize shared api client error copy

yyqdbngt:
### Summary
Localize the two user-facing error strings produced by the shared axios client (`web/src/api/client.ts`), which surface as toast messages on essentially every page:

- The generic business-error fallback `请求失败` now resolves through the new `api.requestFailed` key (default-zh safe).
- The CORS-rejection hint is no longer a module-level Chinese constant; it is resolved per occurrence through the `api.corsRejection` key with the current UI language (read via `getInitialLanguage`, so toggling language applies without a rebuild).
- Added a small `apiText(key)` helper that reads the current stored language on each call, keeping the module hook-free.

### Why
These two strings were the only remaining hardcoded Chinese in the shared request layer, so every English-mode page could still toast Chinese for network/business failures; the CORS hint especially matters for cross-origin deployments.

### Testing
- Vitest: `api/client.test.ts` passes (14/14), including the request-failed fallback and CORS-hint assertions (default zh preserved).

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -17,11 +17,17 @@
 
 import axios from 'axios';
 import { message } from 'antd';
+import { getInitialLanguage } from '../i18n/languagePreference';
+import translations from '../i18n/translations';
 import { clearAuthSession } from '../stores/authStorage';
 import { clearAiChatHistories } from '../stores/aiChatHistoryStore';
 import { API_BASE_URL } from '../config';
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR replaces hardcoded Chinese error messages in the API client with i18n keys and adds English/Chinese translations. Clean i18n fix with proper error code mapping.

LGTM.

---

PR #3080, opened 2026-09-04
fix(i18n): localize mock import failure fallbacks in services

yyqdbngt:
### Summary
Localize the last hardcoded Chinese in the service layer — the mock-mode batch import failure fallbacks:

- `topicService.importTopics` (mock path) and `consumerService.importConsumerGroups` (mock path) no longer emit `创建失败`; both resolve the message through a small hook-free `serviceText(key)` helper (current stored language read per call) and the shared `common.createFailed` key.

### Why
When a topic/group fails to import in the built-in demo (mock) mode, its per-row `message` surfaced the Chinese fallback in English UI; these were the only remaining non-test zh literals in the service layer (verified by literal audit).

### Testing
- Vitest: `TopicPage` and `ConsumerPage` suites pass (2 files / 40 tests).

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -2368,6 +2368,7 @@ const translations: Record<string, Record<Lang, string>> = {
     zh: 'Claude / Titan / Llama (AWS)',
     en: 'Claude / Titan / Llama (AWS)',
   },
+  'common.createFailed': { zh: '创建失败', en: 'Creation failed' },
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR replaces the hardcoded Chinese string "创建失败" with i18n keys (`createFailed`) and adds English/Chinese translations. Clean i18n fix.

LGTM.

---

PR #3081, opened 2026-09-04
fix(message-history): trim search terms before like matching

yyqdbngt:
### Summary
Fix message/trace query-history search so whitespace-padded terms match rows:

- `listMessageQueries` / `listTraceQueries` now normalize the user-supplied `search` term (`trim()`, null-safe) before building the LIKE predicates, so `" orders "` matches rows whose stored topic/msgId/key/queriedBy values have no surrounding whitespace.
- The escape/trim helpers are extracted as package-private (`normalizeSearch`, `escapeLike`) for direct unit testing.

### Why
The search box passes the raw input; a leading/trailing space (easy to produce by accident or by paste) previously turned the LIKE pattern into `% orders %`, which silently returned zero results even though matching rows existed.

### Testing
- `mvn -f server/pom.xml -Dtest=QueryHistoryServiceTest test` passes: 8/8 (6 existing + 2 new tests covering trim/normalize and wildcard escaping).


RockteMQ-AI (approved):
### Summary
This PR fixes message/trace query-history search so whitespace-padded terms match reliably. Clean implementation with a shared `normalizeSearch` helper and proper test coverage for blank/padded terms.

LGTM.

---

PR #3082, opened 2026-09-04
fix(message-history): normalize cluster and type filters

yyqdbngt:
### Summary
Normalize optional filters in the message/trace query-history service so padded or mixed-case callers cannot silently scope to the wrong data:

- `listMessageQueries`, `listTraceQueries` and `summarize` now trim `clusterId` (and the message-list `queryType`) before building `eq` predicates, and trim the `search` term before LIKE escaping — all at the single service choke point instead of relying on each controller path.
- `normalizeSearch` / `normalizeOptional` are package-private helpers covered by direct unit tests.

### Why
Controller paths pass these parameters inconsistently (one endpoint trims clusterId, another does not); a trailing space in `clusterId` previously turned the equality predicate into a no-match and returned an empty history/summary for an existing cluster.

### Testing
- `mvn -f server/pom.xml -Dtest=QueryHistoryServiceTest test` passes: 7/7 (6 existing + 1 new covering both helpers).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
@@ -159,12 +159,14 @@ public class QueryHistoryService {
 
     public PageResult<MessageQueryHistoryVO> listMessageQueries(String clusterId, String queryType,
                                                                  String search, int page, int pageSize) {
-        String pattern = escapeLike(search);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR normalizes optional filters in the message/trace query-history service so padded or mixed-case callers cannot silently scope to the wrong data. Good defensive fix with thorough test coverage.

LGTM.

---

PR #3083, opened 2026-09-04
fix(message-history): store normalized cluster id for history records

yyqdbngt:
### Summary
Align the write side of the query-history service with its read side (#3082 normalized query filters): history records now persist a trimmed `clusterId`.

- `recordMessageQuery` / `recordTraceQuery` store `normalizeOptional(clusterId)`, so a padded cluster id cannot create rows that the (already normalized) history queries and summary would never return.

### Why
After #3082, reads always trim `clusterId`; without this change, records written with a trailing space would be orphaned (unreachable by any subsequent query/summary for the same cluster).

### Testing
- `mvn -f server/pom.xml -Dtest=QueryHistoryServiceTest test` passes: 7/7 (6 existing + 1 new asserting both record methods store the trimmed cluster id).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java
@@ -82,7 +82,7 @@ public class QueryHistoryService {
         query.setEndTime(endTime);
         query.setResultCount(resultCount);
         query.setResultSnapshot(resultSnapshot);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR aligns the write side of the query-history service with its read side by persisting a trimmed `clusterId` in history records. Clean fix with good test coverage ensuring padded cluster ids cannot create orphan rows.

LGTM.

---

PR #3084, opened 2026-09-04
fix(dlq): trim consumer group name before delegating to provider

yyqdbngt:
### Summary
Trim the DLQ consumer group name before it reaches the provider:

- `resendMessages`, `exportMessages`, `listMessages`, `resendSelectedMessages` and `exportExcel` now resolve the group through `requireGroupName` (rejects null/blank with the existing 400, then trims) and pass the trimmed value to the provider.

### Why
A group name with accidental leading/trailing whitespace previously passed the `hasText` validation but then missed every broker lookup (RocketMQ group names never contain surrounding spaces), producing confusing "group not found"-style failures. Trimming at the service boundary makes the five entry points consistent with the already-trimmed search path in `listDLQGroups`.

### Testing
- `mvn -f server/pom.xml -Dtest=DLQServiceTest test` passes: 15/15 (14 existing + 1 new asserting the provider receives the trimmed group name).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQService.java
@@ -56,30 +56,34 @@ public class DLQService {
     public DLQResendResultVO resendMessages(String instanceId, String groupName, Long startTime, Long endTime,
                                              String targetTopic) {
         requireApacheInstance(instanceId);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR trims the DLQ consumer group name before it reaches the provider, using the existing `requireGroupName` helper to reject blank values and normalize whitespace. Good defensive fix with thorough test coverage.

LGTM.

---

PR #3086, opened 2026-09-04
fix(dlq): trim target topic before resend

yyqdbngt:
### Summary
Trim the optional resend target topic in the DLQ service:

- `resendMessages` and `resendSelectedMessages` now normalize `targetTopic` through a package-private `normalizeOptional` helper (trim when present, null otherwise) before delegating to the provider.

### Why
The resend target is an optional override of the default DLQ topic; a padded value previously bypassed the group/topic lookup and could create messages under a phantom topic with surrounding whitespace.

### Testing
- `mvn -f server/pom.xml -Dtest=DLQServiceTest test` passes: 15/15 (14 existing + 1 new asserting the provider receives the trimmed target topic).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/dlq/DLQService.java
@@ -57,8 +57,9 @@ public class DLQService {
                                              String targetTopic) {
         requireApacheInstance(instanceId);
         validateResendRequest(groupName, startTime, endTime);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR trims the optional resend target topic in `DLQService` before delegating to the provider, preventing whitespace-padded topic names from causing lookup failures. Clean implementation with a reusable `normalizeOptional` helper and proper test coverage.

LGTM.

---

PR #3087, opened 2026-09-04
fix(consumer): trim group and client id in stack diagnostics

yyqdbngt:
### Summary
Trim the consumer group and client id in the consumer stack diagnostics lookup:

- `getConsumerStack` normalizes `groupName` / `clientId` (final locals via `normalizeOptional`) before the proxy probe and the broker `getConsumerRunningInfo` call.

### Why
A group or client id carrying surrounding whitespace (paste artifacts) previously reached `getConsumerRunningInfo` verbatim and reported "client not online"-style failures even though the client was connected.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQConsumerDiagnosticsProviderTest test` passes: 12/12 (11 existing + 1 new verifying the broker call receives the trimmed group/client).


RockteMQ-AI (approved):

---

PR #3088, opened 2026-09-04
fix(metadata): normalize cluster and search filters on list queries

yyqdbngt:
### Summary
Normalize cluster/type/search filters on the DB-backed metadata list queries:

- `listTopics`, `listTopicsPage`, `listConsumerGroups` and `listConsumerGroupsPage` now trim `clusterId` (and `type` where present) and trim the `search` term before building `LambdaQueryWrapper` predicates, via package-private `normalizeOptional` / `normalizeSearch` helpers.

### Why
Padded or whitespace-pasted search terms and cluster ids previously produced zero rows from the metadata store even when matching topics/groups existed, and the live topic/group search on the instance pages depends on these list methods.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQMetadataProviderTest test` passes: 36/36 (35 existing + 1 new covering both helpers).


RockteMQ-AI (approved):
### Summary
LGTM — code changes look good. No critical issues found.

---

PR #3096, opened 2026-09-04
test(csv): cover remaining formula-prefix injection guards

yyqdbngt:
### Summary
Extend the CSV cell guard coverage:

- `-`, `@`, leading tab and leading newline values are all neutralized with the single-quote prefix;
- leading spaces, plain text and the empty string stay untouched/quoted correctly.

### Why
The `CsvUtil` export guard already handles `=`, `+` and embedded quotes; these tests lock down the remaining spreadsheet-formula injection characters so future changes cannot silently widen the attack surface.

### Testing
- `mvn -f server/pom.xml -Dtest=CsvUtilTest test` passes: 3/3.


RockteMQ-AI (approved):
### Summary
LGTM — small, clean change with no obvious issues.

---

PR #3102, opened 2026-09-04
fix(metadata): guard and trim topic names in route lookups

yyqdbngt:
### Summary
Guard and trim topic names in the metadata route lookup:

- The private `getTopicRoutes(admin, name)` now returns an empty list for blank names (no broker probe) and trims the name before `examineTopicRouteInfo`.

### Why
A topic pasted with surrounding whitespace previously reached the broker probe verbatim and produced confusing broker-side failures; a blank name triggered an unnecessary (and noisy) probe.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQMetadataProviderTest test` passes: 36/36 (35 existing + 1 new asserting the trimmed probe and the blank-name short circuit).


RockteMQ-AI (approved):

---

PR #3105, opened 2026-09-04
fix(metadata): escape like wildcards in page search terms

yyqdbngt:
### Summary
Escape LIKE wildcards in the DB-backed metadata page searches:

- `listTopicsPage` / `listConsumerGroupsPage` now escape `%`, `_` and `\` in the search term via `escapeLike` before building the `LambdaQueryWrapper` predicate, so searching for literal text such as `100%` or `a_b` no longer behaves as a wildcard match.

### Why
The plain-text search box treats input as a substring; an unescaped `%` previously matched every name and `_` matched any single character, surprising users who search names containing those symbols.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQMetadataProviderTest test` passes: 35/35 (34 existing + 1 new covering the escape helper).


RockteMQ-AI (commented):
### Summary

---

PR #3106, opened 2026-09-04
fix(message-history): trim topic and msg id when recording history

yyqdbngt:
### Summary
Normalize topic / message-id text when writing query-history records:

- `recordMessageQuery` and `recordTraceQuery` now store `normalizeOptional(topic)` / `normalizeOptional(msgId)` (trimmed, null-safe).

### Why
Reads already trim search terms and cluster filters; records written with padded topic/msgId values would otherwise never match a later trimmed search — the same orphan-record class fixed for cluster ids earlier.

### Testing
- `mvn -f server/pom.xml -Dtest=QueryHistoryServiceTest test` passes: 7/7 (6 existing + 1 new asserting both record methods store trimmed topic/msgId).


RockteMQ-AI (approved):

---

PR #3107, opened 2026-09-04
fix(audit): normalize filter parameters before querying

yyqdbngt:
### Summary
Normalize audit log filter parameters at the service choke point:

- `findPage` (used by `queryLogs` and `exportLogs`) now trims `search`, `operationType`, `resourceType`, `clusterId` and `result` (null-safe) before building the query.

### Why
Padded filter values from the UI/paste previously produced zero matches (or mismatched exports) even when records existed; the controller layer did not trim them.

### Testing
- `mvn -f server/pom.xml -Dtest=AuditServiceTest test` passes: 13/13 (12 existing + 1 new verifying the repository receives trimmed filters).


RockteMQ-AI (approved):

---

PR #3108, opened 2026-09-04
fix(metadata): escape like wildcards on full list queries

yyqdbngt:
### Summary
Escape LIKE wildcards on the full (non-paged) metadata list queries too:

- `listTopics` and `listConsumerGroups` now escape `%`, `_`, `\` in the search term via `escapeLike` before the `LambdaQueryWrapper` predicate, matching the paged variants.

### Why
The non-paged lists back selector/search flows; an unescaped `%`/`_` there previously matched broadly instead of literal text.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQMetadataProviderTest test` passes: 35/35 (existing suite; helper behavior covered by the paged-escape suite).


RockteMQ-AI (commented):
### Summary

---

PR #3109, opened 2026-09-04
fix(message): trim direct consume request fields

yyqdbngt:
### Summary
Trim the direct-consume request fields before the broker probe:

- `consumeMessageDirectly` normalizes `consumerGroup`, `clientId`, `topic` and `msgId` (null-safe trim) before calling `consumeMessageDirectly` on the admin client.

### Why
A group/client/topic carrying surrounding whitespace (paste artifacts) previously reached the broker verbatim and reported client-not-online style failures even though the client was connected.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQMessageProviderTest test` passes: 31/31 (30 existing + 1 new verifying the admin call receives trimmed values).


RockteMQ-AI (commented):
### Summary

---

PR #3110, opened 2026-09-04
fix(message): trim trace query parameters

yyqdbngt:
### Summary
Trim trace query parameters before lookup:

- `getMessageTrace` normalizes `msgId`/`topic`/`traceTopic` and `getMessageTraceByKey` normalizes `key`/`topic`/`traceTopic` (null-safe trim) before delegating to the admin lookups.

### Why
Padded identifiers/paste artifacts previously flowed into the broker-side topic/id lookups, producing empty traces or confusing failures for otherwise valid input.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQMessageProviderTest test` passes: 31/31 (existing suite; behavior unchanged for clean input).


RockteMQ-AI (commented):
### Summary

---

PR #3111, opened 2026-09-04
fix(cluster): trim nameserver address before discovery

yyqdbngt:
### Summary
Trim the NameServer address before cluster discovery:

- `discoverClustersAt` trims `namesrvAddr` before using it for the admin connection and for building the NameServer list.

### Why
An address pasted with surrounding whitespace previously connected/compared against a padded string, silently returning no clusters (or mismatched NameServer lists) for a valid endpoint.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQClusterProviderTest test` passes: 14/14 (existing suite; behavior unchanged for clean input).


RockteMQ-AI (commented):
### Summary

---

PR #3112, opened 2026-09-04
fix(dlq): enforce broker topology guard on the selected DLQ msgId lookup

unbridled-41:

Follow-up to #2835, as requested by @lizhimins in https://github.com/apache/rocketmq-dashboard/pull/2835#issuecomment-5535499168.

The selected DLQ resend lookup (`resendSelectedMessages`) calls `admin.viewMessage(dlqTopic, msgId)` with no address validation. Offset-style msgIds embed a broker address that `MQAdminImpl#viewMessage` decodes and connects to directly (`MessageDecoder.decodeMessageId` builds a literal `InetSocketAddress` from bytes 0-7 of the id — no DNS), so a forged id made the DLQ lookup open a RocketMQ remoting connection to an attacker-chosen `ip:port` — the same primitive #2833 (#2832) closed for the message query path and trace timestamp lookup.

## Brief changelog

- Add `isWithinKnownBrokerTopology` to `RocketMQDLQProvider`, mirroring the merged implementation in `RocketMQMessageProvider` (same `decodedBrokerAddr`/`knownBrokerEndpoints` helpers, same warn log on rejection), and apply it in `resendSelectedMessages` before `admin.viewMessage(dlqTopic, msgId)`:
  - id decodes and the embedded address is outside the selected instance topology → skipped (logged, not resolved);
  - id decodes but the topology is empty → rejected;
  - topology verification itself fails (e.g. cluster-info RPC error) → rejected, fail-closed;
  - id does not decode as an offset id (unique-key form) → unchanged behavior, passes through to `viewMessage`, whose unique-key lookup resolves brokers from the topic route.
- No changes to `RocketMQMessageProvider` or any other file; the guard helpers are deliberately scoped to the DLQ provider to keep this follow-up minimal (a shared extraction can be a separate cleanup if maintainers prefer).

## Multi-dimensional verification

**Regression tests (new, in `RocketMQDLQProviderTest`):**

| test | pins |
|---|---|
| `resendSelectedMessagesRejectsForgedMsgIdOutsideKnownTopology` | forged offset id (`10.2.3.4:10911`, topology = `172.30.10.100:10911`) → `viewMessage` never invoked, no producer dispatch, `matched=0` |
| `resendSelectedMessagesResolvesInTopologyMsgIdNormally` | in-topology id → `viewMessage` invoked once, message resolved and resent, `outcome=SUCCESS` (behavior preservation) |
| `resendSelectedMessagesPassesNonOffsetIdsThroughToViewMessage` | unique-key id → passes through to `viewMessage` with zero topology RPCs (behavior preservation) |
| `resendSelectedMessagesFailsClosedWhenTopologyCannotBeVerified` | decodable id + `examineBrokerClusterInfo` failure → `viewMessage` never invoked (fail-closed) |

**Red/green (tests against the unmodified provider):**

| test | unmodified branch | with this fix |
|---|---|---|
| forged out-of-topology rejected | **FAIL** — `viewMessage` is invoked for the forged address | PASS |
| fail-closed on topology failure | **FAIL** — `viewMessage` is invoked | PASS |
| in-topology resolves normally | ERROR — strict stubs flags the topology stub as unused, i.e. the old code never consults the topology | PASS |
| non-offset id passthrough | PASS | PASS |

**Build & suite:**

- `mvn -f server/pom.xml -Dtest=RocketMQDLQProviderTest test` — 32 tests pass (28 pre-existing + 4 new).
- `mvn -B -ntp clean package -DskipTests` (CI backend-build step, includes checkstyle at validate phase) — SUCCESS.
- Full `mvn -B test` on this branch: 1979 tests, single failure is the pre-existing load-sensitive race in `OpenAiCompatibleLlmGatewayTest` already characterized with control runs in #2833 — it fires on unmodified branches too and no code touched by this PR participates in that path.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/BrokerTopologyGuards.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/BrokerTopologyGuards.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.studio.provider.apache;
+
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.rocketmq.common.message.MessageDecoder;
+import org.apache.rocketmq.common.message.MessageId;
+import org.apache.rocketmq.remoting.protocol.body.ClusterInfo;
+import org.apache.rocketmq.remoting.protocol.route.BrokerData;
+import org.apache.rocketmq.tools.admin.MQAdminExt;
+import org.springframework.util.StringUtils;
+
+/**
+ * Shared guards for offset-style message ids, which embed a broker address that
+ * {@code MQAdminImpl#viewMessage} connects to directly. Ids whose embedded address is
+ * outside the selected instance topology must be rejected before that call.
+ */
+@Slf4j
+public final class BrokerTopologyGuards {
+
+    private BrokerTopologyGuards() {}
+
+    /**
+     * Returns true when the msgId either does not decode as an offset id (in which case
+     * MQAdminImpl takes the unique-key lookup through the topic route and needs no guard)
+     * or its embedded broker address belongs to the selected instance topology. When the
+     * topology itself cannot be verified, reject instead of handing an unverified address
+     * to remoting.
+     */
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Security hardening follow-up to #2835 — closes the same SSRF primitive (attacker-controlled broker address via forged offset-style msgId) in the DLQ resend path (`resendSelectedMessages`). The fix mirrors the guard already merged in `RocketMQMessageProvider`, keeping the scope minimal and consistent.

**Key observations:**
- **Correctness**: The `isWithinKnownBrokerTopology` guard correctly handles all four cases: out-of-topology (skip), empty topology (reject), verification failure (fail-closed), and non-offset IDs (pass-through). This matches the threat model from #2833/#2832.
- **Tests**: Four regression tests covering forged-id rejection, normal operation, unique-key pass-through, and fail-closed behavior. Good coverage of the security boundary.
- **Scope**: Deliberately scoped to `RocketMQDLQProvider` only — no changes to the already-fixed `RocketMQMessageProvider`. The PR description notes a potential shared extraction as future cleanup, which is reasonable.
- **Minor**: The duplicated `decodedBrokerAddr`/`knownBrokerEndpoints` helpers between the two providers could be a follow-up extraction, but keeping this PR minimal is the right call.

LGTM — important security fix with thorough verification.

---

RockteMQ-AI (approved):
### Summary
Extracts broker topology validation into a shared `BrokerTopologyGuards` utility and applies it to the DLQ `resendMessages()` flow — preventing forged msgIds from causing the server to connect to arbitrary broker addresses.

### Review Notes
- Security hardening: the topology guard prevents SSRF-like attacks via crafted msgIds
- Fail-closed behavior when topology is unreachable is the correct default
- Clean refactoring: instance method → static utility with no behavioral change for existing callers
- 4 new tests cover: forged ID rejection, valid ID acceptance, non-offset ID passthrough, and topology-unreachable fail-closed

LGTM — good security fix with thorough test coverage.

---

PR #3113, opened 2026-09-04
fix(cluster): trim cluster id in detail refresh

yyqdbngt:
### Summary
Trim the cluster id in the detail-refresh lookup:

- `refreshClusterDetail` normalizes `clusterId` (null-safe trim) before matching against the NameServer cluster-address table and building the returned view.

### Why
A cluster id with surrounding whitespace previously missed the exact table lookup and silently returned `null`, showing an empty cluster page for a valid cluster.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQClusterProviderTest test` passes: 14/14 (existing suite; behavior unchanged for clean input).


RockteMQ-AI (commented):
### Summary
Fix looks good. Consider adding test coverage for this validation.

---

PR #3114, opened 2026-09-04
fix(consumer): require group and client id in stack diagnostics

yyqdbngt:
### Summary
Require the consumer group and client id in the stack-diagnostics lookup:

- `getConsumerStack` trims both inputs and rejects blank values with a clear 400 (`consumer group is required` / `consumer client id is required`) before any proxy or broker probe.

### Why
A blank group/client previously flowed into the probes and surfaced as a confusing "consumer not online" style failure; an explicit 400 makes the contract clear at the boundary.

### Testing
- `mvn -f server/pom.xml -Dtest=RocketMQConsumerDiagnosticsProviderTest test` passes: 12/12 (11 existing + 1 new covering both blank guards).


RockteMQ-AI (approved):

---

PR #3115, opened 2026-09-04
test(message-history): cover empty summarize output

yyqdbngt:
### Summary
Cover the empty-stream path of the history summary:

- `summarize` with zero message/trace records returns zero counts and a null `latestQueryAt` (both `selectOne` lookups are null).

### Why
The latest-query computation (`latestOf` of two nullable values) previously had no direct coverage for the both-empty case.

### Testing
- `mvn -f server/pom.xml -Dtest=QueryHistoryServiceTest test` passes: 7/7 (6 existing + 1 new).


RockteMQ-AI (approved):

---

PR #3118, opened 2026-09-04
fix(cluster): require nameserver address in connection test

yyqdbngt:
### Summary
Guard the connection-test command:

- `testConnection` now rejects a null command or a blank/absent `namesrvAddr` with a clear 400 (`namesrvAddr is required`) instead of dereferencing and surfacing an internal error.

### Why
The action previously called `command.getNamesrvAddr().trim()` unconditionally, so an empty request (or blank address) produced an NPE-style failure instead of a validation message.

### Testing
- `mvn -f server/pom.xml -Dtest=ClusterConnectionServiceTest test` passes: 2/2 (existing suite; behavior unchanged for valid input).


RockteMQ-AI (approved):
LGTM. Clean defensive fix with test coverage.

---

PR #3120, opened 2026-09-04
fix(ai): guard null cluster in capability resolution

yyqdbngt:
### Summary
Guard null clusters in the AI capability resolver and add its first unit coverage:

- `resolve(ClusterVO)` now rejects a null cluster with a 404 `Cluster not found` (the typeless-cluster 400 remains).
- New `CapabilityResolverTest` covers V4/V5 capability lists and both rejection paths (3 tests; the package-private overload makes direct testing possible).

### Why
The package-private overload could be invoked with a null view (e.g., a service returning null for an unknown cluster) and previously NPE'd; it also had no automated coverage at all.

### Testing
- `mvn -f server/pom.xml -Dtest=CapabilityResolverTest test` passes: 3/3.


RockteMQ-AI (approved):
LGTM. Clean defensive fix with test coverage.

---

PR #3121, opened 2026-09-04
fix(metrics): require data source config in source factory

yyqdbngt:
### Summary
Guard and cover the metrics source factory:

- `MetricsSourceFactory.create` rejects a null configuration with a clear 400 (`metrics data source config is required`) instead of NPE-ing on `config.getProviderType()`.
- New `MetricsSourceFactoryTest` (first coverage) verifies the null guard and that a padded `prometheus` provider type builds a `PrometheusMetricsSource` (backend normalization).

### Why
The factory builds every Prometheus-compatible backend and had no automated coverage; a null config previously crashed with an opaque NPE.

### Testing
- `mvn -f server/pom.xml -Dtest=MetricsSourceFactoryTest test` passes: 2/2.


RockteMQ-AI (approved):
LGTM. Clean defensive fix with proper test coverage.

The null guard prevents an opaque NPE and the test provides first coverage for this factory. Well structured.

---

PR #3122, opened 2026-09-04
fix(ai): guard null tool input in topic list handler

yyqdbngt:
### Summary
Guard and cover the AI topic-list tool handler:

- `TopicListToolHandler.execute` rejects a null input map with a clear 400 (`tool input is required`) instead of NPE-ing on `input.get(...)`.
- New `TopicListToolHandlerTest` (first coverage) verifies the null guard and the filter/projection path through the metadata service.

### Why
Tool handlers run on LLM-produced inputs and had no automated coverage; a null/missing input previously produced an opaque NPE inside the tool gateway.

### Testing
- `mvn -f server/pom.xml -Dtest=TopicListToolHandlerTest test` passes: 2/2.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/TopicListToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/TopicListToolHandler.java
@@ -39,6 +39,10 @@ public class TopicListToolHandler implements ToolHandler {
 
     @Override
     public Object execute(Map<String, Object> input) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Same defensive pattern as #3123, applied to `TopicListToolHandler.execute()`. The null guard prevents an NPE when the LLM tool gateway passes a null input map, and the new test class provides first coverage for this handler.

Clean, consistent with the sibling PR, and well-tested. LGTM.

---

PR #3123, opened 2026-09-04
fix(ai): guard null tool input in group list handler

yyqdbngt:
### Summary
Guard and cover the AI consumer-group list tool handler:

- `ConsumerGroupListToolHandler.execute` rejects a null input map with a clear 400 (`tool input is required`) instead of NPE-ing.
- New `ConsumerGroupListToolHandlerTest` (first coverage) verifies the null guard and the search/projection path through the metadata service.

### Why
Tool handlers run on LLM-produced inputs and had no automated coverage; a null/missing input previously produced an opaque NPE inside the tool gateway.

### Testing
- `mvn -f server/pom.xml -Dtest=ConsumerGroupListToolHandlerTest test` passes: 2/2.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ConsumerGroupListToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ConsumerGroupListToolHandler.java
@@ -40,6 +40,10 @@ public class ConsumerGroupListToolHandler implements ToolHandler {
 
     @Override
     public Object execute(Map<String, Object> input) {
+        if (input == null) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds a null guard to `ConsumerGroupListToolHandler.execute()` and introduces the first test coverage for this handler. Clean, focused change — the null check prevents an opaque NPE when the tool gateway receives a null input map from the LLM, replacing it with a clear 400 error.

The test class covers both the null-input guard and the happy path (search/projection through the metadata service). Well structured for a first contribution.

LGTM.

---

PR #3124, opened 2026-09-04
fix(ai): guard null tool input in alert rule list handler

yyqdbngt:
### Summary
Guard and cover the AI alert-rule list tool handler:

- `AlertRuleListToolHandler.execute` rejects a null input map with a clear 400 (`tool input is required`) instead of NPE-ing.
- New `AlertRuleListToolHandlerTest` (first coverage) verifies the null guard and the search/enabled filter + projection path.

### Why
Tool handlers run on LLM-produced inputs and had no automated coverage; a null input previously produced an opaque NPE inside the tool gateway.

### Testing
- `mvn -f server/pom.xml -Dtest=AlertRuleListToolHandlerTest test` passes: 2/2.


RockteMQ-AI (approved):
LGTM. Null guard for tool input — consistent with the other handler fixes. Prevents NPE on invalid inputs.

---

PR #3125, opened 2026-09-04
fix(ai): guard null tool input in ns config diff handler

yyqdbngt:
### Summary
Guard and cover the AI NameServer-config-diff tool handler:

- `NameServerConfigDiffToolHandler.execute` rejects a null input map with a clear 400 (`tool input is required`) instead of NPE-ing.
- New `NameServerConfigDiffToolHandlerTest` (first coverage) verifies the null guard and the delegation to the diff service for the input cluster.

### Why
Tool handlers run on LLM-produced inputs and had no automated coverage; a null input previously produced an opaque NPE inside the tool gateway.

### Testing
- `mvn -f server/pom.xml -Dtest=NameServerConfigDiffToolHandlerTest test` passes: 2/2.


RockteMQ-AI (approved):
LGTM. Proper null guard for tool input — prevents NPE when handler receives null input. Good defensive programming.

---

PR #3129, opened 2026-09-04
[Studio] feat: add data source coverage insights

coder999o:

Add a local instance coverage analysis panel to the Data Sources settings page. The analysis helps operators see whether configured metrics data sources effectively cover managed RocketMQ instances, highlights uncovered instances, same-type overlaps, unavailable or untested sources, and stale instance bindings without exposing or persisting credentials.

## Brief changelog

- Add a data source coverage analyzer for global/scoped source resolution, source health classification, stale instance binding detection, and same-type overlap warnings.
- Show a coverage summary and detailed instance/type coverage modal on the Data Sources settings page.
- Add Chinese and English labels for the new coverage UI.
- Add utility and page-level tests for covered, uncovered, overlapping, stale, untested, and unavailable data source scenarios.

## Verifying this change

- `npm run test -- dataSourceCoverage DataSourceTab`
- `npm run lint`
- `npm run build`
- `git diff --check`
- `git ai status`


web/src/utils/dataSourceCoverage.ts
```diff
@@ -0,0 +1,422 @@
...
+  | 'NO_DATA_SOURCES'
+  | 'INSTANCE_UNCOVERED'
+  | 'INSTANCE_NO_USABLE_SOURCE'
+  | 'INSTANCE_MULTIPLE_SAME_TYPE'
+  | 'INSTANCE_ONLY_UNTESTED'
+  | 'SOURCE_STALE_INSTANCE'
+  | 'SOURCE_UNHEALTHY'
+  | 'DUPLICATE_SOURCE_TARGET';
+
+export interface DataSourceCoverageIssue {
```
RockteMQ-AI, line 37: The `normalizeUrl` function silently returns the raw lowercased value when `new URL()` parsing fails. This is reasonable for resilience, but consider adding a `console.debug` or a warning log for unparseable URLs — it would help operators diagnose misconfigured data source endpoints.

web/src/utils/dataSourceCoverage.ts
```diff
@@ -0,0 +1,422 @@
...
+  dataSourceKeys?: string[];
+  type?: string;
+}
+
+export interface DataSourceReference {
+  key: string;
+  name: string;
+  type: string;
+  url: string;
+  auth: string;
```
RockteMQ-AI, line 52: The `instanceAliases` function uses `instance.name`, `instance.id`, and `instance.endpoint` as matching keys. This covers the common cases well. One edge case to be aware of: if a data source references an endpoint with a port (e.g. `10.0.0.1:9090`) but the instance endpoint is stored without the port (or vice versa), the alias match would miss. The `normalizeUrl` function strips trailing slashes but preserves ports, so this should be consistent as long as both sides use the same format.

RockteMQ-AI (approved):
/tmp/review_work/summary_3129.txt

lizhimins (changes requested):
Thanks for building this — "which instances have no metrics source at all" is a genuine gap, and the analysis module itself is well put together: typed issue codes, alias matching against instance name/id/endpoint, stale-instance detection, and its own 173-line test suite.

However the coverage numbers are computed from the wrong input, and that is not a cosmetic problem — it produces confidently wrong conclusions on any non-default view of the page.

**1. Blocker: global coverage is computed from one page of a filtered query.**

`analyzeDataSourceCoverage(dataSources, instances)` (`DataSourceTab.tsx:169`) uses `dataSources` as the numerator and the complete `instances` list as the denominator. But `dataSources` only ever holds the current page of a server-side paginated *and* filtered query:

```
loadDataSources → listDataSources({ search: debouncedSearch, type: typeFilter, page, pageSize })
                → setDataSources(result.items)        // :199
```

with `pageSize` defaulting to 20 (`:153`). So each of these silently corrupts every counter and can raise a red Alert that is simply false:

- more than 20 data sources → any instance covered only by a source on page 2 or later is reported `INSTANCE_UNCOVERED`;
- anything typed into the search box → coverage collapses to the matching subset;
- any type selected in the filter dropdown → same.

The denominator meanwhile is the *full* instance list (instance lists are not paginated in this codebase), so this is a full-inventory denominator over a page-sized, filter-scoped numerator. The smaller the visible slice, the worse the reported coverage — which is exactly backwards for a panel whose job is to reassure an operator that nothing is uncovered.

The correct input already exists in this file: `listAllDataSources` is imported at `:44` and used at `:476` for the CSV export. Please compute coverage from an unfiltered full inventory rather than from the table's page state. If you also want a panel that reflects the active filter, that is a legitimate but different metric, and it needs to be labelled as scoped to the current filter rather than presented as coverage.

**2. `fontSize: 12` violates the project minimum.** The `<Text type="secondary" style={{ fontSize: 12 }}>` in the coverage Alert needs to go — 14px is the floor here including for secondary text, so please just drop the override and inherit.

**3. Neither of the two new tables sets `scroll.x`.** Please use `tableScrollX(columns)` from `utils/table` so the declared column widths are honoured instead of being squeezed by antd's `table-layout: auto`. Without it the columns wrap and the tables become hard to read on a narrow window.

**4. The scoping of the six counters is not disclosed anywhere.** The Alert presents 已覆盖 / 已验证 / 全局源 / 作用域源 / 未覆盖 / 冲突 as though they described the whole deployment, and the ~30 new i18n keys carry no hint that any figure could be scoped. Once #1 is fixed this may become moot, but as a rule: when a panel mixes full-inventory figures with page-scoped ones, each has to say which it is. The audit risk panel is a reasonable model — it labels page-scoped figures explicitly and takes its global rates from the server summary.

#2, #3 and #4 are quick. #1 is the one that decides whether the panel can be trusted, so please start there — happy to re-review as soon as it is reworked.

lizhimins (changes requested):
The coverage and status conclusions are computed from the currently loaded page of data sources — `DataSourceTab.tsx:169` calls `analyzeDataSourceCoverage(dataSources, instances)` where `dataSources` comes from `listDataSourcesPage({search, ...})`, i.e. at most one page — but they are presented as a statement about the whole deployment.

Please either scope the claim to what is on screen (the way #3131 labels its ratio as current-page and gates on `MIN_RECORDS_FOR_RATE`) or fetch the full set before computing. As written, changing the page size changes the reported coverage.

PR #3130, opened 2026-09-04
[ISSUE #3128] fix(studio): remove K8s certificate lifecycle management

Aias00:
### Summary

- remove the K8s certificate renewal endpoint, DTO, service behavior, web API/mock exports, and audit presentation
- keep Studio-local certificate list/create/update/delete configuration and expiry/status display
- rename the UI surface from certificate management to certificate configuration and clarify that it does not modify Kubernetes resources
- remove the proposed Kubernetes certificate provider/lifecycle design and update README/API documentation
- add a regression test proving `/api/k8s-certs/renew` is no longer exposed

### Verification

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=K8sCertServiceTest,K8sCertControllerTest,MybatisPlusK8sCertRepositoryTest test` (37 passed)
- `NODE_ENV=test npm test -- src/api/cluster.test.ts src/pages/cluster/__tests__/K8sCertsPage.test.tsx src/pages/ops/__tests__/auditPresentation.test.ts` (29 passed)
- `npm run build`
- `npm run lint` (0 errors; 11 pre-existing warnings)
- `git diff --check`

Full-suite note: the web suite passed 901/902 tests; the sole ConsumerPage concurrency failure passed when rerun alone. The Maven suite still has three unrelated failures in `AuthCorsIntegrationTest` and `AliyunInstanceProviderTest`, reproducible when those tests are run independently.

Closes #3128

PR #2991, opened 2026-09-02
fix(k8s): consolidate certificate lifecycle integrity

Aias00:
## Summary

Consolidates the K8s certificate lifecycle fixes from #2689, #2935, and #2943.

This change makes certificate renewal replace the persisted PEM material, derive metadata from the submitted leaf certificate, validate mTLS key pairs, keep private keys write-only in the UI/API flow, prevent stale saves from resurrecting concurrently deleted certificates, and render PEM validity timestamps in the service clock zone.

Replaces #2689
Replaces #2935
Replaces #2943
Closes #2681

## Tests

- JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=K8sCertServiceTest,MybatisPlusK8sCertRepositoryTest,K8sCertControllerTest test
- npm test -- src/api/cluster.test.ts src/pages/cluster/__tests__/K8sCertsPage.test.tsx
- npm run lint
- npm run build
- git diff --check

PR #3133, opened 2026-09-04
[ISSUE #3132] feat(studio): discover NameServers from Kubernetes

### LINKED ISSUE #3132: [Studio] Discover RocketMQ NameServer services from Kubernetes
### Background

RocketMQ Studio currently requires operators to type and maintain NameServer addresses manually. The registry already stores `k8sNamespace`, but Studio does not use the Kubernetes API to discover NameServer Services.

Unlike Kubernetes certificate lifecycle management, NameServer endpoint discovery is directly related to connecting Studio to RocketMQ. The implementation should use Kubernetes-native service discovery and avoid treating a full Pod scan as the primary path.

### Proposed behavior

Add an on-demand, read-only discovery flow in the NameServer management page:

1. List Services in one explicitly selected namespace and prefer Services exposing the configured NameServer port (default `9876`).
2. Fall back to Service name/port-name and common Kubernetes application-label hints.
3. Fall back to `discovery.k8s.io/v1` EndpointSlice data when no Service candidate is found.
4. Query Pods with standard component/name label selectors, then perform a namespace-scoped image-name scan only as the last fallback.
5. Return ranked candidates with source, confidence, stability and `namesrvAddr`; do not auto-persist or silently select the first candidate.
6. Let the user select a candidate and review it in the existing NameServer registry form before saving.

### Safety and scope

- Namespace is required; no `inAnyNamespace()` or ClusterRole is needed.
- Discovery is disabled by default and creates the Kubernetes client only when requested.
- Use Service DNS as the preferred stable address; EndpointSlice/Pod IPs are explicitly marked unstable fallbacks.
- Do not read Secrets or certificate resources.
- Do not create, update, delete or watch Kubernetes resources.
- Do not add background address mutation or automatic registry synchronization in this change.
- Map Kubernetes authentication/RBAC/transport failures to explicit API errors without exposing tokens, kubeconfig content or cluster URLs.

### Configuration

- enable/disable flag
- NameServer port, default `9876`
- cluster DNS domain, default `cluster.local`
- Pod fallback enable flag and bounded candidate limit
- Kubernetes credentials follow Fabric8 auto-configuration (in-cluster ServiceAccount, `KUBECONFIG`, or standard kubeconfig)

### Acceptance criteria

- A namespace-scoped backend API returns deterministic, deduplicated candidates.
- Service matches stop lower-cost fallbacks.
- EndpointSlice readiness and ports are handled correctly.
- Label-filtered Pod lookup precedes the full namespace Pod image scan.
- The UI supports discovery, empty/error states and explicit candidate selection.
- Unit/controller/API/page tests cover fallback order and the user flow without a live cluster.
- Documentation includes configuration, connectivity limitations and minimum namespace Role permissions for Services, EndpointSlices and Pods.

Aias00:
### Summary

- add an administrator-triggered, namespace-scoped Kubernetes NameServer discovery API using Fabric8 7.8.0
- discover candidates in order: Service port/name hints, ready EndpointSlices, component/name label Pod queries, then a bounded full Pod image fallback
- require ready endpoints before reporting normal Service DNS candidates and aggregate multi-slice IPv4/IPv6 addresses
- return source, confidence and stability metadata without mutating Kubernetes or the Studio registry
- add a NameServer management flow that lets users review a candidate and prefill the existing create form before saving
- add default-off configuration, Docker environment passthrough, minimum namespace RBAC and connectivity documentation

### Safety boundaries

- no `inAnyNamespace()` or ClusterRole requirement
- no Secret/certificate access and no Kubernetes mutations
- no background watch or automatic registry synchronization
- ServiceAccount/kubeconfig material stays server-side through Fabric8 auto-configuration
- stale frontend discovery responses are ignored after close/reopen

### Verification

- `JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home mvn -Dtest=KubernetesNameServerDiscoveryServiceTest,NameServerControllerTest test` (27 passed)
- `NODE_ENV=test npm test -- src/api/cluster.test.ts src/pages/cluster/__tests__/ClusterPage.test.tsx` (38 passed)
- `NODE_ENV=test npm test` (115 files, 925 tests passed)
- `npm run lint` (0 errors; 10 existing warnings)
- `npm run build`
- `git diff --check`

Full Maven suite note: 2043/2046 tests passed. The three failures are existing unrelated assertions in `AuthCorsIntegrationTest` (2) and `AliyunInstanceProviderTest` (1), reproducible on the base branch and previously tracked in adjacent Studio work.

Closes #3132


server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/KubernetesNameServerDiscoveryService.java
```diff
@@ -0,0 +1,323 @@
...
+            List<Service> services = client.listServices(namespace);
+            List<EndpointSlice> endpointSlices = client.listEndpointSlices(namespace);
+            List<KubernetesNameServerCandidateVO> candidates = servicePortCandidates(
+                    services, endpointSlices, namespace);
+            if (candidates.isEmpty()) {
+                candidates = serviceHintCandidates(services, endpointSlices, namespace);
+            }
+            if (candidates.isEmpty()) {
+                candidates = endpointSliceCandidates(endpointSlices, namespace);
+            }
```
RockteMQ-AI, line 68: The discover method creates a KubernetesClient via clientFactory.create() and relies on try-with-resources to close it. Consider logging a warning if close() throws, since a failed close on the HTTP transport could leak connections in long-running Studio deployments.

server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/KubernetesNameServerDiscoveryService.java
```diff
@@ -0,0 +1,323 @@
...
+        } catch (BusinessException exception) {
+            throw exception;
+        } catch (KubernetesClientException exception) {
+            int status = exception.getCode() == 403 ? 403 : 503;
+            String message = status == 403
+                    ? "Kubernetes RBAC denied NameServer discovery in namespace " + namespace
+                    : "Kubernetes API is unavailable for NameServer discovery";
+            throw new BusinessException(status, message);
+        } catch (RuntimeException exception) {
+            log.warn("Kubernetes NameServer discovery failed in namespace {}: {}", namespace,
```
RockteMQ-AI, line 95: The fallback chain (SERVICE_PORT -> SERVICE_HINT -> ENDPOINT_SLICE -> POD_LABEL -> POD_IMAGE) is well-designed. One minor suggestion: when the POD_IMAGE fallback fires, consider adding a log line at INFO level so operators can see that the less-reliable path was used.

web/src/pages/cluster/index.tsx
```diff
@@ -293,6 +305,68 @@ const ClusterPage = () => {
...
+    setKubernetesDiscoveryLoading(true);
+    try {
+      const result = await discoverKubernetesNameServers(namespace.trim());
+      if (requestId !== kubernetesDiscoveryRequestRef.current) return;
+      setKubernetesDiscoveryCandidates(result.candidates);
+      setKubernetesDiscoverySearched(true);
+      if (result.candidates.length === 0) {
+        message.info(t('cluster.k8sDiscoveryEmpty'));
+      }
+    } catch (error) {
```
RockteMQ-AI, line 340: Good use of kubernetesDiscoveryRequestRef to handle stale responses. The pattern correctly invalidates in-flight requests when the modal is closed/reopened. Consider extracting this request-ID pattern into a custom hook if reused elsewhere.

RockteMQ-AI (approved):
/tmp/summary_3133.txt

lizhimins (changes requested):
Thanks for the detailed issue and the careful implementation — namespace-scoped queries only, no auto-persist, explicit candidate selection, and RBAC failures mapped without leaking cluster details are all done well.

After review we would like to close this direction rather than merge it.

**1. Manual address entry is the intended UX here, not a gap.** `rmq_nameserver.namesrv_addr` already accepts any reachable address — an SLB VIP, a headless Service DNS name (`rocketmq-nameserver.<ns>.svc.cluster.local`), or a comma-separated list of both — and `k8s_namespace` already records the namespace for headless deployments. An operator who deployed RocketMQ on Kubernetes already knows which Service or domain they exposed, so typing it once is cheaper than configuring and maintaining a discovery integration. If the real pain is discoverability, the proportionate fix is help text / placeholders on the existing NameServer form documenting the SLB and headless-DNS options, and at most a connectivity test against the address the operator already entered — a docs-level change, not a Kubernetes client.

**2. The cost is disproportionate to what it saves.** `io.fabric8:kubernetes-client:7.8.0` (version hardcoded inline in the dependency rather than in `<properties>`, unlike `<rocketmq.version>`), a 323-line five-tier discovery service, new configuration surface and new RBAC requirements — all to avoid filling in one text field.

**3. It does not work in this project's own deployment shape.** `deploy/docker-compose.yml` only gains environment variables (default `enabled: false`), with no kubeconfig mount and no in-cluster ServiceAccount. Enabling the flag in the deployment we document and test would fail client construction and return 503; discovery only works when Studio itself runs inside the same cluster, which is not a deployment we support.

**4. Scope.** Consistent with #2991 (K8s certificate lifecycle, closed as out of scope) and #3130 (merged as `6df87144`, removing the certificate renewal surface), we are keeping Kubernetes platform integration out of the dashboard: Studio talks to RocketMQ, not to the orchestrator. Note that #3130 also means this branch now needs a rebase — both touch `README.md`, `README_zh.md`, `server/README.md`, `docs/api-spec.md`, `web/src/api/cluster.ts` (+ test), `web/src/services/clusterService.ts` and `web/src/i18n/translations.ts`.

Smaller notes if you rework in another direction: the persistent explanation should use `web/src/components/InfoBanner` rather than `<Alert type="info">`; the candidate modal uses a magic `width={920}` and its table does not go through `tableScrollX(columns)`; and the Kubernetes client is built per request instead of being pooled/cached the way `MqAdminExtFactory` / `MqClientPool` manage RocketMQ clients.

Aias00:
Thank you for the review. I agree that Kubernetes discovery is disproportionate to the problem and outside the Dashboard boundary. I replaced it with the scoped UI guidance in #4128, covering SLB addresses, headless Service DNS names, and comma-separated address lists without Kubernetes integration. Closing this PR in favor of that smaller change.

PR #2621, opened 2026-08-26
test(studio): restore topic query tail-budget coverage

Aias00:
## Summary

Closes #2620.

- Align the two topic-query tail-budget regressions with the current 32 pulls x 32 messages per queue policy.
- Add a dedicated Java 21 backend-test CI job while retaining the existing package job.

## Verification

```bash
JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -q test
```

Passed: 1,626 tests, 0 failures, 0 errors.

Also validated the two repaired regressions directly and checked the workflow YAML.

PR #3145, opened 2026-09-04
fix(alert): require metric equality when reconciling native alert presence

### LINKED ISSUE #3104: [Studio][Bug] Cross-metric samples keep missing native alerts permanently active
## Severity

Medium-High. The defect makes the native alert state and recovery-notification path untrustworthy, although it does not directly interrupt RocketMQ message traffic.

## Problem

Native alert reconciliation treats a sample from any metric as proof that every matching rule fingerprint is still present. It checks the collection scope, instance, and resource labels, but does not require the sample metric to equal the rule metric.

Because a fingerprint is derived from `ruleId + instanceId + labels`, another metric with the same labels can keep a missing metric's old state in `FIRING` or `ACKED` forever.

## Reproduction

1. Create a native business rule for `consumer.delay.seconds > 10`, scoped to instance `local` and consumer group `orders`; configure a notification channel if recovery delivery is to be observed.
2. Produce a delay sample of `120` and let the rule enter `FIRING`.
3. In the next successful full collection, keep the consumer group but make its consumption timestamp unavailable.
4. The business collector omits `consumer.delay.seconds` but still emits `consumer.lag.total` with `{consumerGroup=orders}`.
5. Process that collection.

### Actual behavior

The lag sample is added to the delay rule's `presentKeys`. The previous delay state remains active; no `RESOLVED` event or recovery notification is emitted.

### Expected behavior

Only a sample whose metric equals the rule metric may keep that rule's fingerprint present. A missing delay metric in a successful scope must resolve its prior active state even when another same-label metric remains.

## Impact

- The UI continuously reports an incident that no longer has a metric sample.
- Acknowledged alerts also remain active indefinitely.
- Operators never receive the recovery notification.
- Repeated false incidents reduce confidence in native alerting.

## Root cause

`NativeAlertProcessor.reconcileMissingActiveStates` builds `presentKeys` by applying `NativeAlertRuleScopeMatcher`, which intentionally matches instance and resource labels only. It omitted `sample.metricKey() == rule.metric`.

The real Apache business collector makes this stable: `consumer.lag.total` is always emitted for a collected group, while `consumer.delay.seconds` is conditional on a consumption timestamp being available; both use the same consumer-group label.

- #2679 and merged replacement PR #2957 reconcile alerts when a monitored resource disappears and no sample remains for its fingerprint.
- This case is distinct: the resource and labels still exist, but one metric disappears and another metric impersonates its presence.
- #2951/#2960 concern blank metric-name normalization, not cross-metric state reconciliation.
- #2823 scopes persisted metric snapshot history by cluster; it does not reconcile missing samples or construct `presentKeys`.
- Searches for `reconcileMissingActiveStates`, `presentKeys`, `another metric`, and `consumer.delay.seconds resolved` found no matching issue or PR.

- Require exact metric equality while constructing reconciliation `presentKeys`.
- Preserve current behavior for a genuinely present target metric.
- Preserve the full-scope-unavailable guard against false recovery.
- Add a regression test that verifies `RESOLVED` state persistence, lifecycle event creation, original labels, and invocation of the recovery-notification enqueue path.

unbridled-41:
## Problem / Evidence

Fixes #3104 (filed today by ai-yang, no existing PR claims it).

Native alert reconciliation (`NativeAlertProcessor.reconcileMissingActiveStates`) builds `presentKeys` by applying `NativeAlertRuleScopeMatcher`, which intentionally matches only the collection scope, instance id, and resource labels (consumerGroup/topic/brokerName/clusterName). It never requires the sample's metric to equal the rule's metric.

Because an `AlertStateKey` fingerprint is derived from `ruleId + instanceId + labels`, any other metric emitted with the same labels keeps a missing metric's stale state in `FIRING`/`ACKED` forever: no `RESOLVED` transition, no lifecycle event, and no recovery notification.

Verified in the current tree before the fix: the new regression test
`resolvesMissingMetricEvenWhenAnotherMetricSharesTheSameLabelsTest` fails because `AlertStateRepository.save` is never called for the active key (the `consumer.lag.total` sample keeps the `consumer.delay.seconds` rule's fingerprint present).

## Root cause / Fix

`reconcileMissingActiveStates` now filters candidates with `sample.metricKey().equals(rule.getMetric())` before building `presentKeys`. A missing metric in a successful collection scope now resolves its prior active state even when another same-label metric remains. Behavior for a genuinely present target metric is unchanged (`keepsActiveFingerprintWhenItAppearsInSuccessfulCollectionScopeTest` still passes), and the full-scope-unavailable guard (`containsWholeScopeFailure`) is untouched.

## Priority & scoring

- PRIORITY 83/100: impact 34/40 (alert state and recovery-notification path untrustworthy, operators see false incidents and miss recoveries), scope 15/20 (all native alert users), reproducibility 18/20 (deterministic given the Apache business collector's stable `consumer.lag.total` vs conditional `consumer.delay.seconds` emission), maintenance value 16/20 (one-line filter, clear semantics).
- FIX_CONFIDENCE 92/100: minimal filter at the exact root-cause site called out by the issue; covered by a red→green regression test.

## Tests

- New: `NativeAlertProcessorTest.resolvesMissingMetricEvenWhenAnotherMetricSharesTheSameLabelsTest` — asserts `RESOLVED` state persistence, a `RESOLVED` lifecycle event carrying the original labels, and invocation of the recovery-notification enqueue path.
- `mvn test -Dtest='NativeAlert*Test'`: 40/40 pass (18 in `NativeAlertProcessorTest`, regression test red before fix, green after).
- `mvn test '-Dtest=org.apache.rocketmq.studio.ops.alert.*Test'`: all alert package tests pass.

## Risk

Low: single predicate added to reconciliation key construction; evaluation-time behavior, suppression, and the whole-scope-unavailable guard are unchanged. The fix strictly widens when a stale state may resolve (only when the rule's own metric is absent from a successful scope).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NativeAlertProcessor.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NativeAlertProcessor.java
@@ -24,6 +24,7 @@ import org.apache.rocketmq.studio.cluster.metrics.MetricSample;
 import org.apache.rocketmq.studio.common.domain.enums.AlertLevel;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.util.StringUtils;
 
 import java.time.Instant;
 import java.time.LocalDateTime;
@@ -97,7 +98,7 @@ public class NativeAlertProcessor {
         List<AlertRuleVO> rules = alertService.listRules(scope.domain()).stream()
                 .filter(rule -> rule.getId() != null)
                 .filter(AlertRuleVO::isEnabled)
-                .filter(rule -> scope.metricKeys().contains(rule.getMetric()))
+                .filter(rule -> scope.metricKeys().contains(StringUtils.trimWhitespace(rule.getMetric())))
                 .filter(rule -> rule.getInstanceId() == null || scope.instanceId().equals(rule.getInstanceId()))
                 .toList();
         if (rules.isEmpty()) {
@@ -106,6 +107,7 @@ public class NativeAlertProcessor {
         Set<AlertStateKey> presentKeys = samples.stream()
                 .filter(scope::contains)
                 .flatMap(sample -> rules.stream()
+                        .filter(rule -> sample.metricKey().equals(StringUtils.trimWhitespace(rule.getMetric())))
                         .filter(rule -> NativeAlertRuleScopeMatcher.matches(rule, sample))
                         .map(rule -> new AlertStateKey(rule.getId(),
                                 AlertFingerprint.of(rule.getId(), sample.instanceId(), sample.labels()))))
--- a/server/src/test/java/org/apache/rocketmq/studio/ops/alert/NativeAlertProcessorTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/ops/alert/NativeAlertProcessorTest.java
@@ -478,6 +478,49 @@ class NativeAlertProcessorTest {
     }
 
     @Test
+    void resolvesMissingMetricEvenWhenAnotherMetricSharesTheSameLabelsTest() {
+        AlertService service = mock(AlertService.class);
+        AlertRuleVO rule = AlertRuleVO.builder().id(1L).domain(AlertDomain.BUSINESS).name("Orders delay")
+                .metric("consumer.delay.seconds").operator(">").threshold(10).enabled(true)
+                .instanceId("local").consumerGroup("orders").consecutiveSamples(1).build();
+        when(service.listRules(AlertDomain.BUSINESS)).thenReturn(List.of(rule));
+        MetricSample previousDelay = new MetricSample("consumer.delay.seconds", AlertDomain.BUSINESS, "local",
+                null, Map.of("consumerGroup", "orders"), 120D, MetricAvailability.AVAILABLE, Instant.now());
+        AlertStateKey key = new AlertStateKey(rule.getId(),
+                AlertFingerprint.of(rule.getId(), previousDelay.instanceId(), previousDelay.labels()));
+        ActiveAlertState active = new ActiveAlertState(key,
+                new AlertRuleState(AlertStateStatus.FIRING, 1, 120D, previousDelay.collectedAt().minusSeconds(60),
+                        previousDelay.collectedAt().minusSeconds(60), previousDelay.collectedAt().minusSeconds(60),
+                        null),
+                previousDelay.instanceId(), previousDelay.labels());
+        AlertStateRepository states = mock(AlertStateRepository.class);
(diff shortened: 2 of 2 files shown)


unbridled-41:
## Verification evidence (supplementary)

**Why CI is red on this PR:** the repository's CI workflow has failed at startup (`startup_failure`, zero jobs executed) for every branch and event since 2026-08-28 — including upstream's own pushes to `rocketmq-studio` and all contributor PRs (dozens of consecutive runs). GitHub's run annotation reads "This run likely failed because of a workflow file issue"; the last `ci.yml` change was PR #2621 (2026-08-27). This branch does not modify `ci.yml` (`git diff origin/rocketmq-studio..HEAD -- .github/workflows/ci.yml` → empty), so the failures are a pre-existing repo-wide condition unrelated to this change. CI-equivalent verification was therefore run locally:

**Regression test (red → green)**
- Red, before the fix: `mvn test -Dtest=NativeAlertProcessorTest` → `Tests run: 18, Failures: 1`. `resolvesMissingMetricEvenWhenAnotherMetricSharesTheSameLabelsTest` failed with Mockito reporting `Wanted but not invoked: alertStateRepository.save(AlertStateKey[ruleId=1, fingerprint=1519e4f4…], <Capturing argument: AlertRuleState>)`; the only recorded interaction was `findActive(…)` at `NativeAlertProcessor.reconcileMissingActiveStates(NativeAlertProcessor.java:117)` — i.e. the stale FIRING state was never resolved because the same-label `consumer.lag.total` sample kept the fingerprint present.
- Green, after the fix: same command → `Tests run: 18, Failures: 0`.

**Module tests**
- `mvn test -Dtest='NativeAlert*Test'` → 40/40 pass (NativeAlertProcessorTest 18, NativeAlertRulePolicyTest 12, NativeAlertRuleScopeMatcherTest 3, NativeAlertEvaluationTransactionTest 1, NativeAlertRuleTestServiceTest 4, NativeAlertMetricCatalogServiceTest 2).
- `mvn test '-Dtest=org.apache.rocketmq.studio.ops.alert.*Test'` → all alert-package suites pass (AlertServiceTest 73, AlertStateMachineTest 7, ClusterAlertRuleControllerTest 6, AlertRuleEvaluatorTest 5, repositories/DTO/transfer suites included).

**Full suite (CI backend-test equivalent)**
- `mvn -B -ntp test` on this branch → `Tests run: 2036, Failures: 3`. The 3 failures are identical to the pristine `rocketmq-studio` baseline measured before this change:
  - `AuthCorsIntegrationTest.shouldRejectNonAdminMutationBeforeControllerExecution:122` (expected 403, got 401)
  - `AuthCorsIntegrationTest.shouldStillRejectAnonymousProtectedRequests:110`
  - `AliyunInstanceProviderTest.getGroupProgressShouldMapLagRowsTest:316` (expected size 2, was 1)
  2036 = 2035 baseline tests + the 1 new regression test. **Zero new failures; the full-suite compile also covers the CI backend-build job.**

PR #3149, opened 2026-09-04
[ISSUE #3148] Add alert rule portfolio review

### LINKED ISSUE #3148: Add complete alert rule portfolio review
### Problem

The alert rule table is paginated and filtered. Operators can duplicate, import, enable, and disable rules, but they cannot review the complete domain inventory for combinations that create duplicate notifications or silently stop coverage. Problems spanning pages are especially difficult to spot.

### Proposed change

Add a read-only portfolio review for both cluster and business alert domains. It should use the existing complete rule-list contract and identify only evidence-backed conditions:

- enabled rules with identical scope and evaluation signatures;
- duplicate rule names;
- enabled rules without notification channels;
- rules referencing instances that no longer exist;
- scopes for which every remaining rule is disabled.

Different thresholds must not be treated as conflicts. The review should provide issue and rule views, filters, CSV export, Chinese/English copy, and focused tests. It must not mutate rules or depend on the current page/filter state.

yuluo-yx:
## What changed

- add a read-only portfolio review to both cluster and business alert pages
- review the complete domain rule inventory rather than only the current page
- identify enabled rules with identical evaluation signatures, duplicate names, enabled rules without notification channels, deleted instance references, and disabled-only scopes
- keep distinct thresholds separate to avoid false conflict reports
- provide issue and rule views with state, severity, issue, and text filters
- export the currently filtered rule review to CSV
- add Chinese/English copy and focused pure-function/component tests

## Design boundaries

- no rule mutations are performed
- no findings are inferred from runtime samples
- global rules without an instance are valid
- only enabled rules participate in exact duplicate evaluation detection
- equivalent duration spellings such as `1m` and `60s` are normalized

Fixes #3148

## 行为与兼容性

用结构化作用域字段消除展示分隔符造成的身份碰撞；按后端语法归一化复合时长，使用大整数避免精度丢失。

## 验证

- 相关前端测试：13 项通过。
- ESLint、TypeScript/Vite 生产构建、合并差异检查通过。
- 已合入最新 `rocketmq-studio`，使用普通推送保留原提交历史。
- 未执行远端工作流、浏览器 E2E 或真实 Broker 测试。

验证日期：2026-09-14。


RockteMQ-AI (commented):
### Summary
Adds an alert rule portfolio review drawer that analyzes rule configurations for issues like exact duplicates, duplicate names, missing notification channels, unknown instance references, and disabled-only scopes. The analysis logic is sophisticated and well-tested.

### Findings
- **[Warning]** `web/src/utils/alertRulePortfolio.ts` — `evaluationSignature()` uses `JSON.stringify()` to create comparison keys. This works for the current field types but is fragile: if any field value contains special characters (e.g., a metric name with quotes), the signature could collide or produce false positives. Consider using a delimiter-based approach or a structured hash instead.

- **[Warning]** `web/src/utils/alertRulePortfolio.ts` — `normalizedDurationMs()` converts duration strings like `"1m"` and `"60s"` to the same millisecond value (`60000`), which is correct for duplicate detection. However, the regex `/^(\d+(?:\.\d+)?)\s*(ms|s|m|h)$/` does not handle compound durations like `"1m30s"` or `"1h 5m"`. If the backend ever returns such formats, they would fall through to the raw string comparison and miss real duplicates. Consider documenting this limitation or adding support.

- **[Info]** `web/src/utils/alertRulePortfolio.ts` — The `issue()` helper generates deterministic keys using sorted rule IDs. This is a good pattern for deduplication and testing.

- **[Info]** Test coverage is excellent — the `alertRulePortfolio.test.ts` covers exact duplicates, name collisions, missing channels, unknown instances, disabled-only scopes, global rules, and filtering. The test helpers (`rule()`) make the tests very readable.

### Cross-PR Note
Same `translations.ts` conflict risk as PRs #3153, #3157, #3159. Please coordinate merge order.

---

yuluo-yx:
@lizhimins，if you have times，pls take a look thx。

PR #3151, opened 2026-09-04
fix(cluster): route registry config actions through the row's owning instance

### LINKED ISSUE #3103: [Studio][Bug] Registry Broker configuration actions can target the wrong instance
## Severity

High (P1). This is a silent control-plane write to a different RocketMQ instance than the Broker row selected by the operator.

## Problem

The Cluster page displays Broker rows from the global NameServer registry through `listRegistryClusters()`. Broker configuration preview/update and config-diff actions still attach the instance selected by the page route.

After registry-based discovery was introduced, the current route instance and the source instance of a registry row are no longer necessarily the same. If two instances use a common cluster name such as `DefaultCluster`, selecting a row discovered from instance B while the route points to instance A sends `{ id: "DefaultCluster", instanceId: "instance-a" }`. The backend correctly honors that payload and can update A even though the operator clicked B.

## Reproduction

1. Configure two Apache instances:
   - `instance-a` -> `ns-a:9876`
   - `instance-b` -> `ns-b:9876`
2. Let both NameServers report a cluster named `DefaultCluster`, with different Broker addresses.
3. Register `ns-b:9876` in the NameServer registry.
4. Open `/cluster?instanceId=instance-a`.
5. In the global Broker table, open configuration for a Broker discovered from `ns-b:9876` and submit a change.

### Actual behavior

The update request contains:

```json
{
  "id": "DefaultCluster",
  "instanceId": "instance-a"
}
```

The backend resolves and updates instance A.

### Expected behavior

Every registry row action must be routed through the unique Apache instance whose endpoint owns that row. If ownership cannot be established unambiguously, configuration actions must fail closed.

## Impact

The operation can silently change the wrong production cluster's flush mode, auto-create settings, queue counts, maximum message size, retention, or permissions. A successful response does not reveal that the target was wrong.

## Root cause

- `web/src/pages/cluster/index.tsx` loads all registry clusters without instance scope.
- Broker rows retain their registry cluster but no source instance identity.
- `buildConfigUpdateRequest` and Broker config diff previously read `selectedInstanceIdRef.current`, which represents the route, not the row.

- #1243 / #1244 added the selected-instance API contract, but #1243 explicitly left broader cluster discovery/action routing out of scope.
- #2342 later moved the Broker table to global registry discovery without assigning an instance identity to each row.
- #2595 concerns repeated config reads for duplicate Broker addresses, not writes to the wrong instance.
- Searches for `listRegistryClusters updateClusterConfig` and equivalent registry/selected-instance terms found no issue or PR covering this failure.

- Canonicalize the full NameServer address set and map a registry endpoint to exactly one Apache instance.
- Use the row's instance for Broker configuration preview, update, and config diff.
- Freeze the target when a modal opens so a later route change cannot retarget the write.
- Discard preview/diff responses from a closed, superseded, or edited modal generation.
- Refresh the global registry snapshot after successful and partially successful updates.
- Disable configuration actions for missing or ambiguous mappings.
- Preserve global registry visibility.
- Cover cross-instance, route-change, address-order/separator, and ambiguous-mapping cases in tests.

unbridled-41:
Fixes #3103.

Rework of this PR per the review: the frontend no longer tries to infer which instance owns a registry row.

## Problem / Evidence

Broker tab rows come 100% from the global NameServer registry (`listRegistryClusters()` probing every registered NameServer), but Broker config preview/update and both config-diff actions attached the instance selected by the page route (`selectedInstanceIdRef.current`). `rmq_nameserver` and `rmq_instance` share no foreign key — one cluster may host several instances and several clusters may share a NameServer — so a registry row discovered from instance B could be written through instance A whenever both expose a cluster with the same id (e.g. `DefaultCluster`): a silent control-plane write to the wrong cluster.

## Root cause / Fix

- Endpoint text cannot identify an owner (zero or multiple endpoint matches are legitimate states, not errors), and the previous draft's `resolveRegistryTargetInstanceId` fail-closed heuristic is gone entirely.
- The backend already resolves targets authoritatively from the cluster id when no instanceId is supplied: `ClusterService.resolveCluster(id)` / `BrokerConfigDiffService.compare(id, null)` / `NameServerConfigDiffService.compare(clusterId)` all accept `instanceId=null` and resolve via `ClusterService.getCluster(id)` against the configured NameServer, and `updateClusterConfig` routes each broker write through `updateBrokerConfig(address, id, props)` without an instance.
- Preview, update, Broker config diff, and NameServer config diff therefore send only the cluster id: `buildConfigUpdateRequest` no longer includes `instanceId`, and `openBrokerConfigDiff` / `openNameServerConfigDiff` pass just the cluster id. No ownership relationship is rebuilt on the client.
- The existing request-generation / stale-response guards are unchanged.

## Tests

Regressions in `ClusterPage.test.tsx`, each red on the unfixed source (which sent the route instanceId) and on the previous draft (which rejected):

1. `routes registry config actions through the backend without an instanceId` — registry row endpoint `ns-b:9876` shares no text with any instance endpoint; preview and update still run and their request bodies carry no `instanceId` property.
2. `keeps registry config actions available when several instances share the NameServer` — two configured instances share the row's endpoint (`ns-shared:9876`); the action is not rejected and the preview request goes out without an instanceId.
3. `resolves config diffs server-side from the cluster id` — Broker config diff and NameServer config diff are each called with exactly `('cluster-prod')` — the backend resolution path, no instance argument.

`previews broker config changes before submitting the update` and the two drift-dialog tests were updated from asserting the route instanceId to asserting its absence (the behavior the fix intentionally removes).

- `npx vitest run src/pages/cluster src/i18n` → 7 files / 73 tests pass.
- `tsc --noEmit` clean; `eslint` clean on both changed files (also enforced by the pre-commit hook); `vite build` ✓ (12.04s).

## Risk

Low: no frontend guessing remains, and the backend's own resolution is the single source of truth. Clusters not reachable from the configured NameServer now fail on the backend with its own explicit error instead of being written through a guessed instance. Read-only discovery, metrics, and the route-scoped instance summary are unchanged.

Branch note: the merge commit and the workspace-notes cleanup commit are gone — the branch is a single commit on the current `rocketmq-studio`.


unbridled-41:
## Verification evidence (supplementary)

**Why CI is red on this PR:** the repository's CI workflow has failed at startup (`startup_failure`, zero jobs executed) for every branch and event since 2026-08-28 — including upstream's own pushes to `rocketmq-studio` and all contributor PRs. GitHub's annotation reads "This run likely failed because of a workflow file issue" (last `ci.yml` change was PR #2621, 2026-08-27). This branch does not modify `ci.yml` (diff vs `origin/rocketmq-studio` is empty), so the failures are pre-existing and unrelated. CI-equivalent verification was run locally:

**Regression tests (red → green)**
- Red, before the fix (`web/src/pages/cluster/index.tsx` + `web/src/i18n/translations.ts` temporarily stashed, tests kept): `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx` → `Tests 3 failed | 3 passed | 19 skipped (25)`. Failing: `targets broker config actions at the instance that owns the registry row` (preview/update still carried the route instance `instance-a`), `rejects broker config actions when no instance owns the registry endpoint` (modal opened and requests were sent with no owner check), `rejects broker config actions when multiple instances share the registry endpoint` (same).
- Green, after the fix: same command → 25/25 pass.

**Module / related suites**
- `npx vitest run src/pages/cluster src/i18n` → 7 files, 67/67 pass.
- `npx tsc -p tsconfig.app.json --noEmit` → clean; `eslint` on the three changed files → clean.

**Full suite and build (CI frontend equivalents)**
- Full `npx vitest run --testTimeout=60000 --hookTimeout=60000` on this branch → 925 tests (922 baseline + 3 new), 3 failed / 922 passed. All 3 failures are in files this PR does not touch (`TopicPage.test.tsx`, `ConsumerPage.test.tsx`) and are sandbox-load flakiness, not regressions: repeated pristine full-suite runs produce a *different* failure set each time (9 failures/4 files, then 15/7, then 3 — all near the configured 20s `testTimeout`), and every failing test passes in isolation both on pristine and on this branch (TopicPage + ConsumerPage here → 50/50 isolated; AclPage + NotificationDeliveriesPage here → 24/24 isolated; the same files → 53/53 isolated on pristine `rocketmq-studio`).
- `npm run build` (CI frontend-build equivalent) → succeeds.

No new commits were added for this; the code and tests were already complete — this comment documents the verification the broken CI could not run.

lizhimins (changes requested):
Thanks for this — the underlying defect is real and worth fixing: the Broker tab rows come 100% from the global NameServer registry, while preview/update/diff are all routed through a specific instance, so a NameServer shared by two instances can be written from the wrong one. The value of closing that gap is high.

The blocker is the implementation path, not the goal:

1. **Owner cannot be inferred from endpoint text.** `rmq_nameserver` and `rmq_instance` share no foreign key and `rmq_nameserver` has no `instance_id`, so the only signal available here is string overlap between the NameServer address and an instance endpoint. That is a guess, not a relationship.
2. **It fails closed on common configurations.** The repo's own fixtures use `namesrv-1:9876` for the NameServer and `10.101.2.1:9876` for the instance endpoint — no textual overlap. With that (very common) shape the heuristic resolves no owner, so the guard rejects everything rather than protecting the right thing. A safety check that disables the feature under normal configs is worse than the current behavior.
3. **Correctness belongs on the backend.** The authoritative instance↔NameServer ownership is known server-side; it should be sent down with the registry data (e.g. an `ownerInstanceId` on each NameServer row, resolved from the persisted relationship) and the frontend should gate on that field instead of recomputing a guess from strings.

Suggested rework: add the ownership to the backend response and drive the guard from it; keep the frontend change to "disable/annotate rows whose owner differs from the selected instance". Happy to review quickly once the owner comes from the backend rather than endpoint matching.

unbridled-41:
## Merge & re-validation (base advanced to 0c54d985)

The base branch advanced substantially since this PR was opened (36126024 → 0c54d985), including two upstream merges that touch the same file this PR changes: #3154 (`4ecf6323`, broker config diff request guard) and #3997 (`4d596463`, per-broker daily message counters). GitHub reported this branch as conflicting.

Resolved by merging `origin/rocketmq-studio` into this branch as a merge commit (`57e755a2`); no history rewrite, the original fix commit `c673380b` is unchanged:

- **`web/src/pages/cluster/index.tsx`** — one semantic conflict in `openBrokerConfigDiff`: this PR routes the call through the registry row's owning instance (`resolveRegistryTargetInstanceId`), while the now-merged #3154 added the `brokerConfigDiffRequestRef` generation guard to the same function. Resolution combines both (same composition the auto-merged `openNameServerConfigDiff` already exhibits): resolve `targetInstanceId` first, then bump the guard, call `getBrokerConfigDiff(cluster.id, targetInstanceId)`, and check the guard after the await.
- **`web/src/pages/cluster/__tests__/ClusterPage.test.tsx`** — both sides appended tests at the same position; resolution keeps this PR's three regression tests and upstream's "renders per-broker daily message counters" test.

Re-validation on the merged tree:

- `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx` → **28/28 passed** — this PR's three tests ("targets broker config actions at the instance that owns the registry row", both "rejects broker config actions..." variants) and the upstream guard/counter tests all pass **simultaneously**, confirming the two fixes compose correctly.
- Full web suite `npx vitest run` → 983 tests, 981 passed, 2 failed — `ConsumerPage > shows group health diagnostics...` and `TopicPage > keeps failed topics selected...`, both untouched files that pass in isolation (54/54 combined); load flakiness under full-suite parallelism, consistent with previous rounds.
- `npx tsc --noEmit` clean; `npx eslint` on both files 0 errors/0 warnings; `npm run build` succeeded (10.49s).
- CI: the upstream `CI (pull_request)` workflow ends in `startup_failure` for all branches (0 check-runs), so all suites were run locally.



### Disclosure: workspace files accidentally staged, then removed

The merge above was committed with `git add -A`, which accidentally staged two local workspace notes (`AGENTS.md`, `bug-hunter-progress.md` — untracked scratch files of this contribution environment, not part of the fix). They were included in merge commit `57e755a2` and removed again in the follow-up commit `43378e6f`. The net PR diff (file list above) contains only the four cluster/i18n files; no other content was affected.

unbridled-41:
## Evidence-chain completion (head SHA + net diff self-check)

The two earlier comments predate the current six-item evidence standard and lack two items: CI evidence pinned to this PR's current head SHA, and a net-diff self-check against the base. Filling exactly those two items (numbers re-verified today); the red→green, module, full-suite and build results in the comments above remain valid for their stated dates.

**CI (current head)**: this PR's head SHA `43378e6f62aa233027b798fd09c45658c4ce31b4` has exactly one workflow run — "CI" (pull_request), run id 34355781872, conclusion `startup_failure` (created 2026-09-09T13:14:01Z); the check-runs API for the head commit returns `total_count=0`. The upstream workflow fails at startup for every branch and event in this repository, so no CI signal is available; all verification was run locally (see the two comments above).

**Net diff self-check** (files API for this PR, cumulative vs `rocketmq-studio` @ `6c24d2ed`, 4 files, +236/−8):

```
web/src/i18n/translations.ts                    +4/−0
web/src/pages/cluster/__tests__/ClusterPage.test.tsx  +180/−1
web/src/pages/cluster/certs.tsx                 +1/−1
web/src/pages/cluster/index.tsx                 +51/−6
```

- `index.tsx` / `ClusterPage.test.tsx` / `translations.ts` are the fix and its regressions (see the first comment).
- `certs.tsx` +1/−1 is a lint-driven parenthesization (`map[status] ?? {...}` wrapped in parentheses, `no-mixed-operators`), applied by the commit-time lint hook; behavior-neutral, no test impact.
- No other files are touched, and the two workspace-note files that were accidentally staged and removed during the base-advancing merge (see the Disclosure above) net out to zero in this diff.

lizhimins (changes requested):
`resolveRegistryTargetInstanceId` requires the endpoint-token intersection to yield exactly one owning instance and otherwise refuses with `message.error`. That contradicts the instance model this project documents: one cluster may host several instances and several clusters may share a NameServer, so an intersection of zero or more than one is a legitimate state, not an error.

Ownership should be answered by the server rather than guessed in the browser — `BrokerConfigDiffService` already accepts `instanceId=null` and resolves by `clusterId`. Please rework along those lines.

Also please clean the branch before resubmitting: it carries a merge commit and a follow-up commit titled "remove workspace notes accidentally staged by the merge commit".

unbridled-41:
@lizhimins Thanks — reworked along exactly those lines; head is now 2404fa6d.

- `resolveRegistryTargetInstanceId` and all endpoint-text inference are deleted; zero or multiple endpoint matches are no longer treated as errors anywhere.
- Preview, update, Broker config diff, and NameServer config diff now send only the cluster id and omit `instanceId`, so the target is resolved server-side by the existing authoritative paths (`ClusterService.resolveCluster(id)`, `BrokerConfigDiffService.compare(id, null)`, `NameServerConfigDiffService.compare(clusterId)` — all already accept `instanceId=null`; no backend change was needed).
- Branch history cleaned: the merge commit and the workspace-notes commit are gone; the branch is a single commit on the current `rocketmq-studio`.
- Regressions: a row whose endpoint shares no text with any instance endpoint still previews/updates (bodies carry no `instanceId`); two instances sharing the row's NameServer no longer get rejected — the request goes out; and both diff actions are asserted to be called with exactly the cluster id. All three were red on the unfixed source and on the previous draft. `src/pages/cluster` + `src/i18n`: 73 tests pass; `tsc`, `eslint`, `vite build` clean. The stale-response guards are untouched.

Happy to re-review.

PR #3164, opened 2026-09-04
fix(message): narrow key queries by time and tag

### LINKED ISSUE #3163: Message Key queries cannot narrow by time range or tag
## Problem

Message Key queries are less useful than Topic queries during incident triage.

The backend already accepts `topic`, `key`, `tag`, `startTime`, and `endTime` for message queries. The frontend normalizer also passes those fields through. But in Message Key mode the UI only exposes:

- Topic
- Message Key

The date range picker and tag input are rendered only for Topic mode. As a result, users searching by business key cannot narrow a large result set by time or tag from the UI, even though the API supports both.

The backend also skips time-window validation whenever a key or message ID is present. If a client sends an invalid key-query window, it is accepted rather than rejected with the same validation used for topic queries.

## Expected behavior

- Message Key mode should expose the same time range picker and optional tag input as Topic mode.
- Key queries should include `startTime`, `endTime`, and optional `tag` when supplied.
- Key-query time windows should be validated consistently:
  - start must be before end;
  - timestamps must not be negative;
  - the window must not exceed seven days.
- Message ID lookups should remain point lookups and should not require a time window.
- Query history replay should preserve the key query time window and tag.

tju-yxq:
## What happened

In the Message Explorer, `By Message Key` only accepted a Topic and Key. The backend already supported `tag`, `startTime`, and `endTime` for key queries, but the UI never sent them. Key searches also bypassed the seven-day window validation used by topic queries, so a caller could submit an unbounded or reversed key-query request directly to the API.

For on-call searches, this makes a common workflow unnecessarily broad: an incident usually has both a business key and a known time window, and sometimes a known tag. Without those filters, key searches can return many unrelated messages and require manual inspection.

## Change

Backend:

- Validate explicit key-query time windows before calling the provider:
  - reject negative timestamps
  - reject `startTime >= endTime`
  - reject ranges longer than seven days
- Keep key queries without explicit timestamps backward compatible; they continue to use provider defaults.
- Message ID lookups remain unchanged because their time window is not part of that lookup.

Frontend:

- Add optional Tag and time-range controls to the `By Message Key` form.
- Send trimmed `tag`, `startTime`, and `endTime` values with key queries.
- Apply the same start-before-end and seven-day maximum validation as Topic queries.
- Preserve Tag and time values when replaying a saved key query from history, and clear Tag on reset.

Tests:

- Backend service tests cover reversed, over-long, and omitted key-query windows, and verify the provider is not called for invalid requests.
- A focused UI test verifies that a key query submits the selected topic, trimmed key/tag, and explicit time range.
- Updated the existing message-page validation test to re-query the button after async mode switches and to expect the new key-query time parameters.

## Verification

Focused backend:

```
cd server
mvn -q '-Dtest=MessageServiceTest,MessageControllerTest' test
mvn -q checkstyle:check
```

Focused frontend:

```
cd web
npm test -- MessagePage.test.tsx MessagePageKeyQuery.test.tsx MessagePageAsyncState.test.tsx message.test.ts --run
```

Result: 5 files / 37 tests passed.

Full frontend:

```
cd web
npm test -- --run
```

Result: 116 files, 918 passed and 5 timeouts under the default parallel run. Re-running the affected files isolated the behavior:

- the new `MessagePageKeyQuery.test.tsx` passes when run alone and with the message-page suite;
- `ConsumerPage.test.tsx` has the same failures on clean upstream `36126024`, so those tests are pre-existing mainline instability.

Full backend:

```
cd server
mvn -q test
```

Result: 2,038 tests, 3 failures. The same three failures were previously reproduced on a clean `upstream/rocketmq-studio` worktree at `36126024` and are pre-existing mainline test drift, unrelated to this patch:

- `AuthCorsIntegrationTest.shouldStillRejectAnonymousProtectedRequests`
- `AuthCorsIntegrationTest.shouldRejectNonAdminMutationBeforeControllerExecution`
- `AliyunInstanceProviderTest.getGroupProgressShouldMapLagRowsTest`

Lint/build:

```
npm run lint -- --quiet
npm run build
```

Lint reported 0 errors and 10 pre-existing warnings elsewhere in the codebase. The production build completed successfully.

`git diff --check` passed before commit.

The production/config additions are 54 lines (19 backend, 35 frontend). This is the natural size for the UI/API gap; I did not pad the change with unrelated refactoring.

Closes #3163

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java
@@ -205,7 +205,24 @@ public class MessageService {
         if (!hasTopic && !hasMessageId) {
             throw new BusinessException(400, "topic or msgId is required");
         }
-        if (hasMessageId || hasKey) {
+        if (hasMessageId) {
+            return;
+        }
+        if (hasKey) {
+            if (startTime == null || endTime == null) {
+                return;
+            }
+            long keyStart = startTime;
+            long keyEnd = endTime;
+            if (keyStart < 0 || keyEnd < 0) {
+                throw new BusinessException(400, "message query timestamps must not be negative");
+            }
+            if (keyStart >= keyEnd) {
+                throw new BusinessException(400, "startTime must be before endTime");
+            }
+            if (keyStart < keyEnd - MAX_TOPIC_QUERY_WINDOW_MILLIS) {
+                throw new BusinessException(400, "topic query time range must not exceed 7 days");
+            }
             return;
         }
         long end = endTime == null ? System.currentTimeMillis() : endTime;
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/message/MessageServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/message/MessageServiceTest.java
@@ -21,6 +21,8 @@ import java.util.Optional;
 
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.verify;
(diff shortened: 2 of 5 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 5 file(s) with 384 lines of changes.

### Observations
- Changes look reasonable

### Changed Files
- `server/src/main/java/org/apache/rocketmq/studio/instance/message/MessageService.java`
- `server/src/test/java/org/apache/rocketmq/studio/instance/message/MessageServiceTest.java`
- `web/src/pages/instance/__tests__/MessagePage.test.tsx`
- `web/src/pages/instance/__tests__/MessagePageKeyQuery.test.tsx`
- `web/src/pages/instance/message.tsx`

---

PR #3168, opened 2026-09-04
perf(dlq): parallelize group stats and debounce search

### LINKED ISSUE #3167: DLQ group list waits for serial topic-stats RPCs and fires a request per keystroke
## Problem

The DLQ page fetches topic statistics for every group on the current page serially.

`RocketMQDLQProvider.listDLQGroups(...)` slices the current page, then for each `%DLQ%` topic calls:

```java
adminExt.examineTopicStats(dlqTopic)
```

one after another. With the default page size of 20, a page can therefore wait for 20 serial admin RPCs before returning. If brokers are slow, the page latency is roughly the sum of all stats calls.

The frontend also updates the DLQ search input state on every keystroke. That state is in the effect dependency list, so every character starts a new full group-list request and immediately increments the request generation. On a large cluster, typing a short group name can trigger several expensive backend scans in quick succession.

The request-generation guard prevents stale responses from overwriting newer results, but it does not prevent the redundant requests from being sent.

## Expected behavior

- Topic-stats calls for one DLQ page should run with bounded parallelism, rather than one-by-one.
- Page order should remain deterministic and unchanged.
- A single topic-stats failure should only mark that row `UNAVAILABLE`; other rows should still be returned.
- The DLQ search input should debounce requests while the user is typing.
- Debouncing must not delay the initial page load or a direct search submission.

## Why this matters

Operators open the DLQ page during incident triage. Waiting for serial stats RPCs or firing a scan per keystroke makes the page unnecessarily slow and increases load on brokers at exactly the time they may already be unhealthy.

tju-yxq:
## What happened

The DLQ group page made its latency worse than necessary in two places.

Backend:

- `RocketMQDLQProvider.listDLQGroups(...)` sliced the current page and then called `adminExt.examineTopicStats(dlqTopic)` for every `%DLQ%` topic serially.
- With the default page size of 20, the page waited for up to 20 sequential admin RPCs. Slow brokers made this especially visible.

Frontend:

- The search input updated `search` state on every keystroke.
- That state is in the group-loading effect's dependency list, so every character started a new backend group-list request. The request-generation guard prevented stale responses from overwriting newer data, but it did not prevent the redundant requests from being sent.

## Change

Backend:

- Load topic statistics for the current page on a bounded daemon thread pool with 8 workers.
- Preserve the current deterministic page order by collecting futures in input order.
- Cap each stats lookup at 5 seconds; cancel timed-out work.
- Preserve row-level degradation: a failed or timed-out stats call only marks that row `UNAVAILABLE`.
- Shut down the stats executor on bean destruction.

Frontend:

- Split visible search input from the search value used for requests.
- Debounce typing by 300 ms.
- Submit immediately when the user presses Enter or clicks the search action.
- Clear pending debounce timers on unmount.

Tests:

- Backend test proves three simulated 150 ms stats calls complete in less than the serial sum, preserve order, and still call stats once per topic.
- Frontend test proves two rapid keystrokes produce only one request and that the request uses the final search value.

## Verification

Focused backend:

```
cd server
mvn -q '-Dtest=RocketMQDLQProviderTest,DLQServiceTest,DLQControllerTest' test
mvn -q checkstyle:check
```

Focused frontend:

```
cd web
npm test -- DLQPage.test.tsx message.test.ts --run
```

Result: 3 files / 28 tests passed.

Full frontend:

```
cd web
npm test -- --run
```

Result: 115 files / 923 tests passed.

Full backend:

```
cd server
mvn -q test
```

Result: 2,036 tests, 3 failures. The same three failures were previously reproduced on a clean `upstream/rocketmq-studio` worktree at `36126024` and are pre-existing mainline test drift, unrelated to this patch:

- `AuthCorsIntegrationTest.shouldStillRejectAnonymousProtectedRequests`
- `AuthCorsIntegrationTest.shouldRejectNonAdminMutationBeforeControllerExecution`
- `AliyunInstanceProviderTest.getGroupProgressShouldMapLagRowsTest`

Lint/build:

```
npm run lint -- --quiet
npm run build
```

Lint reported 0 errors and 10 pre-existing warnings elsewhere. The production build completed successfully.

`git diff --check` passed before commit.

Production/config additions are 82 lines (48 backend, 34 frontend). This is the natural size of the bounded-executor and debounce change; no unrelated refactoring was added.

Closes #3167

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
@@ -65,6 +65,12 @@ import java.util.Locale;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /**
  * Real {@link DLQProvider} backed by the RocketMQ admin API. Lists dead-letter groups by scanning
@@ -80,11 +86,24 @@ public class RocketMQDLQProvider implements DLQProvider {
     private static final int RESEND_HARD_CAP = 5000;
     private static final int MAX_PAGE_SIZE = 100;
     private static final int MAX_CONSECUTIVE_OFFSET_ILLEGAL = 3;
+    private static final int STATS_THREADS = 8;
+    private static final long STATS_TIMEOUT_SECONDS = 5;
     private static final String ORIGIN_MESSAGE_ID_PROPERTY = "studio_dlq_origin_message_id";
     private static final String ORIGIN_TOPIC_PROPERTY = "studio_dlq_origin_topic";
 
     private final RuntimeAdminClientResolver runtimeAdminClientResolver;
     private final AuditService auditService;
+
+    private final ExecutorService statsExecutor = Executors.newFixedThreadPool(STATS_THREADS, runnable -> {
+        Thread thread = new Thread(runnable, "dlq-topic-stats");
+        thread.setDaemon(true);
+        return thread;
+    });
+
+    @jakarta.annotation.PreDestroy
+    void shutdownStatsExecutor() {
+        statsExecutor.shutdownNow();
+    }
 
     @Override
     public List<DLQGroupVO> listDLQGroups(String instanceId) {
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
```diff
@@ -122,13 +141,38 @@ private PageResult<DLQGroupVO> listDLQGroups(MQAdminExt adminExt, String search,
...
+    private DLQGroupVO awaitGroupStats(Future<DLQGroupVO> future) {
+        try {
+            return future.get(STATS_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        } catch (TimeoutException e) {
+            future.cancel(true);
+        } catch (InterruptedException e) {
+            future.cancel(true);
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
+            // buildDLQGroup reports single-topic failures inside the row itself.
```
RockteMQ-AI, line 171: When `awaitGroupStats` falls back (timeout / interrupt / execution exception), the returned `DLQGroupVO` has no `groupName` or `dlqTopic` set — the user will see a blank row in the DLQ table with only `UNAVAILABLE` status.

Consider passing `groupName` and `dlqTopic` into `awaitGroupStats` so the fallback can populate them:
```java
private DLQGroupVO awaitGroupStats(Future<DLQGroupVO> future, String groupName, String dlqTopic) {
    try {
        return future.get(STATS_TIMEOUT_SECONDS, TimeUnit.SECONDS);
    } catch (TimeoutException e) {
        future.cancel(true);
    } catch (InterruptedException e) {
        future.cancel(true);
        Thread.currentThread().interrupt();
    } catch (ExecutionException e) {
        // single-topic failure
    }
    return DLQGroupVO.builder()
            .groupName(groupName).dlqTopic(dlqTopic)
            .statsAvailable(false).status("UNAVAILABLE").build();
}
```
And update `loadGroupStatsInParallel` to pass them through.

server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQDLQProvider.java
```diff
@@ -122,13 +141,38 @@ private PageResult<DLQGroupVO> listDLQGroups(MQAdminExt adminExt, String search,
...
+        }
```
RockteMQ-AI, line 172: The `ExecutionException` catch block silently swallows the underlying exception. Consider adding a `log.debug` or `log.warn` so operators can diagnose slow/failing brokers:
```java
} catch (ExecutionException e) {
    log.warn("Failed to load DLQ stats for topic, row marked UNAVAILABLE", e.getCause());
}
```

RockteMQ-AI (commented):
### Summary
This PR parallelizes DLQ topic stats loading on a bounded daemon thread pool (backend) and debounces the search input (frontend). Both changes are well-structured with good test coverage.

One issue to address before merge: the fallback `DLQGroupVO` in `awaitGroupStats` is missing `groupName` and `dlqTopic`, which would produce blank rows in the UI when a stats call times out or fails.

### Findings
- **[Warning]** `RocketMQDLQProvider.java:171` — Fallback `DLQGroupVO` missing `groupName`/`dlqTopic` → blank row in UI
- **[Info]** `RocketMQDLQProvider.java:172` — Silent `ExecutionException` swallow; consider adding a log for operator diagnostics

### What looks good
- Bounded thread pool with daemon threads and `@PreDestroy` cleanup — clean lifecycle management
- Per-future 5s timeout with cancellation — prevents slow brokers from blocking the page
- Order preservation via sequential future collection — deterministic page output
- Frontend debounce (300ms) with immediate submit on Enter/click — good UX balance
- Tests cover both parallelism proof (backend) and debounce behavior (frontend)

---

PR #3173, opened 2026-09-05
fix(i18n): localize demo topic remarks for english mode

yyqdbngt:
### Summary
Localize the demo (mock-mode) topic remarks for English UI:

- `Topic` gains an optional `enRemark`; all 12 demo topics (`mock/topics.ts`) now carry English remarks.
- The Topic table's remark cell renders `enRemark` when the UI language is English, falling back to the Chinese remark otherwise (sorting still uses the stored remark).

### Why
In demo mode the topic list's remark column leaked Chinese into English UI; remarks are now display-localized without changing the stored (editable) value.

### Testing
- Vitest: full `TopicPage` suite passes (16/16, default-zh assertions unchanged).


RockteMQ-AI (approved):

---

PR #3176, opened 2026-09-05
fix(i18n): localize demo alert rule names and durations

yyqdbngt:
### Summary
Localize the demo (mock-mode) alert rule names and durations for English UI:

- `AlertRule` gains optional `enName` / `enDuration`; all 8 demo rules (`mock/alerts.ts`) now carry English name and duration copy.
- The alert rules table renders those when the UI language is English, falling back to the Chinese values otherwise (sorting still uses the stored values).

### Why
In demo mode the alert rules list still showed Chinese rule names/durations in English UI.

### Testing
- `eslint` clean on changed files (0 errors; 5 pre-existing warnings remain).
- Vitest: full `AlertsPage` suite passes (20/20, default-zh assertions unchanged).

### DIFF
--- a/web/src/api/ops.ts
+++ b/web/src/api/ops.ts
@@ -9,6 +9,9 @@ export interface AlertRule {
   threshold: number;
   thresholdUnit?: string | null;
   duration: string;
+  /** Optional English display copy for demo (mock-mode) alert rules. */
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):

---

PR #3179, opened 2026-09-05
fix(i18n): localize demo consume mode labels on topic consumer panel

yyqdbngt:
## Summary

The topic detail modal renders a demo consumer-group panel whose column headers and consume-mode tags still come straight from the zh mock data (集群消费 / 广播消费). In English mode those cells stay Chinese while the rest of the page is localized.

This PR completes the panel for en mode:
- Column headers now route through existing translations keys (`topic.consumerGroup`, `topic.consumeMode`, `topic.consumeTps`, `topic.backlog`) instead of hard-coded zh strings.
- The demo consume-mode tag maps `集群消费`/`广播消费` to Clustering/Broadcasting when `lang === en`, falling back to the raw value otherwise.
- The unavailable-metric placeholder uses the new `topic.unavailable` translation key.

The tag color decision still keys off the underlying mock value, so visuals are unchanged.

## Why

Consistent with the demo-data localization done for the consumer page (filter mode / consistency): demo content should read in the active language instead of leaking Chinese mock labels into English mode.

## Testing

- `tsc --noEmit` passes
- `eslint` clean on the changed files
- `vitest run TopicPage` - 21/21 tests pass

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -2542,6 +2542,7 @@ const translations: Record<string, Record<Lang, string>> = {
     zh: 'Claude / Titan / Llama (AWS)',
     en: 'Claude / Titan / Llama (AWS)',
   },
+  'topic.unavailable': { zh: '不可用', en: 'Unavailable' },
   // ─── BrokerCluster ───
 };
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

---

PR #3213, opened 2026-09-05
fix(message-history): scope result snapshots to operator

### LINKED ISSUE #4007: [Studio][Bug] Stored message query results can be read by any authenticated user without an ownership check
## Problem

`GET /api/query-history/messages/{id}/results` returns the stored message-query snapshot for any record id without verifying that the requesting user owns the record. Record ids are sequential, so any authenticated user can enumerate ids and replay another user's stored snapshots.

## Evidence

`QueryHistoryService.getMessageQueryResults` (`server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java:134-150`) reads the row with `messageQueryMapper.selectById(id)` and no owner predicate, while every other read in the same service filters by the authenticated user: `listMessageQueries` (`:184`), `listTraceQueries` (`:204`), `summarize` (`:221-234`), and the owner is stamped at insert (`:93`). The stored snapshot contains msgIds, topics, tags, message keys, broker names, queue offsets and `bornHost`/`storeHost` addresses (`buildResultSnapshot`, `:109-124`).

The per-user boundary was established by the merged fix for #2265 ("Query-history reads and summaries are scoped to the current authenticated user by default"); the results endpoint was introduced later (by #2839) and never picked the boundary up.

## Impact

In multi-user deployments (login required by default), one authenticated user can enumerate other operators' query history snapshots, including which topics/keys/msgIds they searched and internal broker/producer host addresses. The response for a foreign id also differs from a missing id, leaking record existence.

## Expected behavior

The by-id results read applies the same owner scope as the list/summary reads (`queried_by = currentUsernameOrSystem()`); foreign and missing ids return the same 404. No admin bypass, matching the list paths.

## Related work

#2265 (closed, merged via the queried_by scoping), #3295 (delete own history — per-user ownership is the model), #3388 (open test-only coverage of the endpoint's happy path).

## PR

#4003

btlqql:
## Summary
- scope stored message-query result lookup to the authenticated operator
- return the existing not-found response when a record belongs to another operator
- add regression coverage for owned and cross-operator result access

## Why
History list and summary queries already filter on `queried_by`, but the result snapshot endpoint previously loaded records by primary key alone. A signed-in user who learned another record ID could therefore retrieve that operator's stored message metadata.

## Testing
- `mvn -f server/pom.xml -q -DskipITs -Dtest=QueryHistoryServiceTest -DfailIfNoTests=false test` -> `Tests run: 12, Failures: 0, Errors: 0, Skipped: 0`

Fixes #4007


RockteMQ-AI (approved):
### Summary
Fixes an authorization issue where any authenticated user could access another user's query result snapshots by ID. Now properly scopes `getMessageQueryResults` to the authenticated operator via `queried_by` filter.

**Security fix** with proper test coverage for both the authorized and unauthorized access paths.

LGTM.

---

lizhimins (changes requested):
The defect is real and this is the right fix. Requesting changes only on four points, all of them small; none of them question the approach.

## Confirmed: this is a genuine IDOR

On current trunk, `QueryHistoryService.getMessageQueryResults` (`server/src/main/java/org/apache/rocketmq/studio/instance/message/QueryHistoryService.java:134`) reads via `messageQueryMapper.selectById(id)` at `:135` with no ownership predicate, while every other read path in the same file scopes by `eq("queried_by", queriedBy)` — `:184`, `:204`, `:221`, `:224`, `:230`, `:235`. So this one method is the only by-id read that ignores the operator.

It is exploitable, not theoretical: `id` is an auto-increment bigint, so ids are enumerable, and the stored `result_snapshot` carries message rows including `bornHost`, `storeHost` and `brokerName` alongside topic names. Any authenticated operator can walk ids and read another operator's query results together with the internal cluster topology embedded in them.

Two things you got right that are worth stating explicitly, because they are easy to get wrong:

- You reused trunk's `AuthenticatedUserContext` instead of introducing a second notion of "current user".
- You return **404**, not 403, for a record owned by someone else. That is the correct choice: a 403 would confirm the record exists and turn an enumeration into a directory of valid ids.

The tests are also stronger than the usual mock self-check — asserting both `getCustomSqlSegment()` and `getParamNameValuePairs().values()` pins the bound parameters, not just the fact that a query was issued.

For context: #4003 was an independent, essentially identical fix for the same line by another author. We closed it as a duplicate in favour of this one on first-come grounds. Its two advantages over this PR are exactly items 1 and 3 below, which is why we are asking for them here rather than merging that one instead.

## Required changes

**1. Assert the 404 code, not only the exception type and message.**

`hidesResultSnapshotOwnedByAnotherOperator` asserts `isInstanceOf(BusinessException.class)` and `hasMessage("Query history record not found")`. The code is the entire point of `new BusinessException(404, ...)` and nothing currently pins it. If someone later changes 404 to 500, or splits "not found" from "not yours" into 404/403 and reintroduces the existence leak, both of your tests still pass green.

`BusinessException` is `@Getter` over `private final int code`, so:

```java
assertThatThrownBy(() -> service.getMessageQueryResults(9L))
        .isInstanceOf(BusinessException.class)
        .hasMessage("Query history record not found")
        .extracting(e -> ((BusinessException) e).getCode())
        .isEqualTo(404);
```

or capture the thrown exception and assert `getCode()` directly — whichever reads better against the rest of the file.

**2. Use a static import for `assertThatThrownBy`.**

The test file already statically imports `assertThat` from AssertJ, but the new test calls `org.assertj.core.api.Assertions.assertThatThrownBy(...)` fully qualified inline. Add `import static org.assertj.core.api.Assertions.assertThatThrownBy;` and call it unqualified, so the two new tests read like the other 20-odd in the file.

**3. Update the Javadoc on `getMessageQueryResults`.**

It still reads "Retrieves the stored result snapshot for a given history record." with no mention that the lookup is scoped to the authenticated operator. That scoping is the whole substance of this change, and it is also the reason "no such record" and "someone else's record" are deliberately indistinguishable. Without that stated, the 404 looks like an oversight to the next reader, and the natural "fix" is to revert to `selectById` and return 403 — reopening the leak. Say something like: scoped to the authenticated operator; a record owned by another operator is reported as not found (404) on purpose, so that id enumeration cannot establish which ids exist.

**4. Fix the import ordering.**

The added `import org.apache.rocketmq.studio.common.exception.BusinessException;` is placed before `import org.apache.rocketmq.studio.auth.AuthenticatedUserContext;`. `auth` sorts before `common`, and the rest of the block is ordered.

## Not required, just noting

- The production throw site still spells `org.apache.rocketmq.studio.common.exception.BusinessException` fully qualified inline. That is pre-existing on trunk, not something you introduced, so it is not a condition of merging — but you are already in the file and importing the type in the test, so tidying it is free.
- We checked for other by-id read paths that would need the same treatment and found none: there is no delete-by-id or update-by-id endpoint over query history, so the fix surface really is just this one method.

Once these four are in, this should merge cleanly. Please rebase onto current `rocketmq-studio` at the same time — the branch is behind, and `QueryHistoryService` and its test have both moved.

RockteMQ-AI (approved):
### Summary
Re-review after new commit addressing previous feedback. The fix now explicitly asserts the 404 status code for unauthorized snapshot access and the Javadoc clearly documents the scoped-lookup behavior.

Changes look correct — the `selectOne` with both `id` and `queried_by` prevents IDOR while the test covers both the happy path and the unauthorized access path.

LGTM.

---

btlqql:
@lizhimins thank you for the detailed review — all four points are in, and the branch is rebased.

Rebase: the branch now sits directly on `rocketmq-studio` at `6c24d2ed` (was behind `0a596661`). `QueryHistoryService` and its test had not moved in between, so the rebase was clean and the diff is unchanged.

**1. The 404 code is now pinned.** `hidesResultSnapshotOwnedByAnotherOperator` asserts the code, not just the type and message:

```java
assertThatThrownBy(() -> service.getMessageQueryResults(9L))
        .isInstanceOf(BusinessException.class)
        .hasMessage("Query history record not found")
        .extracting(e -> ((BusinessException) e).getCode())
        .isEqualTo(404);
```

**2. Static import added.** `import static org.assertj.core.api.Assertions.assertThatThrownBy;` is in the static block, and both new tests call it unqualified alongside the existing `assertThat` calls.

**3. Javadoc updated.** `getMessageQueryResults` now reads:

> Retrieves the stored result snapshot for a history record owned by the authenticated operator.
>
> The lookup is scoped to the current authenticated operator. A record owned by another operator is deliberately reported as not found (404) to prevent history id enumeration.

**4. Import ordering fixed.** `org.apache.rocketmq.studio.auth.AuthenticatedUserContext` now sorts before `org.apache.rocketmq.studio.common.*`.

**Also done, since you flagged it as free:** the production throw site no longer spells `org.apache.rocketmq.studio.common.exception.BusinessException` inline — the type is imported and the throw is `throw new BusinessException(404, "Query history record not found");`. You were right that there is nothing else to change: I re-checked and there is no by-id delete or update over query history.

Verification after the rebase and the tidy-up:

```
mvn -f server/pom.xml -B -DskipITs -Dtest=QueryHistoryServiceTest -DfailIfNoTests=false test
[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Head is `f833e46a`. Could you take another look when you have a moment?

PR #3214, opened 2026-09-05
fix(auth): enforce public path boundaries

yyqdbngt:
## Summary
- require public documentation and health endpoints to match a complete path segment
- keep valid descendants such as `/swagger-ui/index.html` and `/actuator/health/readiness` public
- add regression coverage for lookalike protected paths

## Why
The authentication interceptor previously used raw prefix checks for public endpoint trees. That also exempted unrelated paths such as `/api-docs-private` and `/actuator/healthcheck` whenever login enforcement was enabled.

## Verification on Linux

`server/`:

```
$ mvn -B -ntp test
Tests run: 2160, Failures: 2, Errors: 0, Skipped: 0
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthInterceptor.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthInterceptor.java
@@ -185,9 +185,13 @@ public class AuthInterceptor implements HandlerInterceptor {
                 || path.equals("/api/auth/status")
                 || path.equals("/livez")
                 || path.equals("/readyz")
-                || path.startsWith("/api-docs")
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes an auth bypass vulnerability where `path.startsWith("/api-docs")` would incorrectly allow paths like `/api-docs-private` or `/swagger-ui-admin` to bypass authentication. The new `isPathOrDescendant` helper properly enforces path boundaries by requiring either an exact match or a `/` separator.

**Security-critical fix.** Excellent test coverage including negative cases for paths that only share a prefix.

LGTM.

---

PR #3239, opened 2026-09-05
fix(i18n): localize broker cluster instance selector

yyqdbngt:
## Summary

The Studio BrokerCluster page toolbar instance selector still used hard-coded zh for its accessible name and placeholder.

Localized in pages/studio/BrokerCluster.tsx:
- The select aria-label and placeholder (选择实例) now use the new `brokerCluster.selectInstance` key.
- 1 new `brokerCluster.*` key; zh byte-identical.

## Why

The toolbar selector is the primary instance switch on the page; its accessible name must follow the active language.

## Testing

- `tsc --noEmit` passes
- `eslint` clean on the changed files

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -2542,6 +2542,8 @@ const translations: Record<string, Record<Lang, string>> = {
     zh: 'Claude / Titan / Llama (AWS)',
     en: 'Claude / Titan / Llama (AWS)',
   },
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Replaces hardcoded Chinese strings in `BrokerCluster.tsx` with i18n translation keys. Clean and straightforward fix.

LGTM — good catch on the missing i18n keys.

---

PR #3245, opened 2026-09-05
test(alert): cover AlertCorrelationScope instance and resource matching

yyqdbngt:
## Summary

Adds `AlertCorrelationScopeTest`, a first unit test for the package-private correlation matcher shared by cross-domain alert handling. Pure static matching, no mocks.

Coverage:
- alerts without a (textual) source instance id never match;
- instance ids must agree — the source side is trimmed before comparison while the candidate side is matched as-is;
- a resource label (`brokerName`/`brokerAddr`/`clusterName`/`clusterId`) only constrains the match when both sides declare it; a conflict on any shared label rejects the pair;
- label values are compared after trimming, and whitespace-only values count as absent.

## Why

The scope matcher decides which alerts belong to the same instance/resource correlation window; its matching semantics had no direct test.

## Testing

`mvn -B test -Dtest=AlertCorrelationScopeTest` — 4/4 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertCorrelationScopeTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertCorrelationScopeTest.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
test(alert): cover AlertCorrelationScope instance and resource matching

LGTM.

---

PR #3246, opened 2026-09-05
test(provider): cover ConsumerConnections instance-view mapping

yyqdbngt:
## Summary

Adds `ConsumerConnectionsTest`, a first unit test for the package-private mapping from a broker/proxy `ConsumerConnection` payload to the online `ConsumerInstanceVO` view shared by the group listing and group detail.

Coverage:
- a `null` connection and a connection without a connection set both map to an empty instance list;
- every connection in the set is mapped to an instance carrying its `clientId` and `clientAddr` (asserted in any order, since the payload holds a `HashSet`).

## Why

The mapping is the single source of truth for online consumer instances in the group pages; it had no direct test.

## Testing

`mvn -B test -Dtest=ConsumerConnectionsTest` — 2/2 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(provider): cover ConsumerConnections instance-view mapping

LGTM.

---

PR #3248, opened 2026-09-05
test(dlq): cover DLQMessageExcelRow export mapping

yyqdbngt:
## Summary

Adds `DLQMessageExcelRowTest`, a first unit test for the export-row mapping used when a dead-letter message (or a selected batch) is written to Excel.

Coverage:
- every field (`msgId`, `topic`, `queueId`, `offset`, `keys`, `body`) is carried over, and the epoch-millis store time is rendered as `yyyy-MM-dd HH:mm:ss` in the current zone;
- null optional text fields (`keys`, `body`) pass through as null, and a zero store time renders the epoch (both asserted timezone-independently via `ZoneId.systemDefault()`).

## Why

The row mapping is the single point that shapes DLQ export output; it had no direct test.

## Testing

`mvn -B test -Dtest=DLQMessageExcelRowTest` — 2/2 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(dlq): cover DLQMessageExcelRow export mapping

LGTM.

---

PR #3249, opened 2026-09-05
test(auth): cover AuthenticatedUserContext principal semantics

yyqdbngt:
## Summary

Adds `AuthenticatedUserContextTest`, a first unit test for the per-thread principal holder used across services and controllers. The context is cleared before each test so results do not depend on sibling tests.

Coverage:
- default state: no username → `system` actor, no user id, non-admin, and `currentUserIsAdminOrSystem()` true (an unset admin flag is treated as system);
- the two-argument setter carries the admin flag; the username-only setter defaults to non-admin;
- non-admin principals are not system-privileged;
- the numeric setters record the user id as a string;
- `setUser(null, ...)` and `clear()` both reset to the system actor.

## Why

Admin/system authorization gates (e.g. AI tool execution) read this context; its default semantics had no direct test.

## Testing

`mvn -B test -Dtest=AuthenticatedUserContextTest` — 6/6 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/auth/AuthenticatedUserContextTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/auth/AuthenticatedUserContextTest.java
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
test(auth): cover AuthenticatedUserContext principal semantics

LGTM.

---

PR #3251, opened 2026-09-05
test(provider): cover BrokerTopologyGuards message id topology checks

yyqdbngt:
## Summary

Adds `BrokerTopologyGuardsTest`, a first unit test for the shared guards that reject offset-style message ids whose embedded broker address is outside the selected instance topology before remoting is allowed to connect to that address directly. `MQAdminExt` is mocked; offset message ids are built with the real `MessageDecoder` from hex-encoded IP/port fixtures.

Coverage:
- non-offset (undecodable) message ids are treated as safe without ever querying the topology;
- ids whose embedded broker address belongs to the cluster are accepted, ids pointing outside it are rejected;
- an unverifiable topology (admin failure) rejects instead of handing an unverified address to remoting;
- `validatedBrokerAddr` returns the address only for known endpoints;
- unresolvable/non-`InetSocketAddress` embedded addresses yield no address and never reach the topology query;
- `knownBrokerEndpoints` collects and trims addresses across brokers, skipping empty tables and empty entries.

## Why

This guard is the security boundary against message ids steering remoting to user-influenced broker addresses; it had no direct test.

## Testing

`mvn -B test -Dtest=BrokerTopologyGuardsTest` — 7/7 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/apache/BrokerTopologyGuardsTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/apache/BrokerTopologyGuardsTest.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
test(provider): cover BrokerTopologyGuards message id topology checks

LGTM.

---

PR #3254, opened 2026-09-05
test(metrics): cover MetricCollectionScope validation and membership

yyqdbngt:
## Summary

Adds `MetricCollectionScopeTest`, a first unit test for the record that identifies one successful native metric collection scope.

Coverage:
- shape validation: null domain, blank instance id, empty metric-key set, and blank keys are all rejected with the expected messages;
- the metric-key set is defensively copied at construction and exposed as an unmodifiable snapshot (later mutation of the source does not leak in);
- `contains` accepts exactly the samples whose domain, instance id, and metric key match the scope, and rejects null samples.

## Why

The scope is what correlates collected samples to their persisted snapshot rows during native alert evaluation; its validation and membership rules had no direct test.

## Testing

`mvn -B test -Dtest=MetricCollectionScopeTest` — 3/3 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(metrics): cover MetricCollectionScope validation and membership

LGTM.

---

PR #3256, opened 2026-09-05
test(broker): cover MqClientPool lifecycle guards

yyqdbngt:
## Summary

Adds `MqClientPoolTest`, first unit tests for the lifecycle guards of the long-lived RocketMQ client pool. Client creation is deliberately excluded (starting a pull consumer/producer would open remoting against a live NameServer), so the deterministic guard paths are pinned down instead.

Coverage:
- blank/whitespace NameServer addresses are rejected with `BusinessException` 400 before any client work, for both the pull-consumer and producer entry points;
- after `shutdown()` the pool rejects further use with `BusinessException` 503;
- the address check takes precedence over the closed flag (a blank address still yields 400 on a closed pool);
- `release` is a no-op for blank endpoints, unknown identities, and endpoints that were never pooled.

## Why

The pool is the shared owner of producer/consumer clients across requests; the failure modes that do not require network had no direct test.

## Testing

`mvn -B test -Dtest=MqClientPoolTest` — 4/4 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/cluster/broker/MqClientPoolTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/cluster/broker/MqClientPoolTest.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
test(broker): cover MqClientPool lifecycle guards

LGTM.

---

PR #3257, opened 2026-09-05
test(message): cover MessageQueryResult truncation signal

yyqdbngt:
## Summary

Adds `MessageQueryResultTest`, a first unit test for the message-query outcome record carrying the provider truncation signal.

Coverage:
- the `complete`/`truncated` factories set `mayBeTruncated` to false/true respectively;
- the message rows are defensively copied at construction and exposed as an unmodifiable snapshot (later source mutation does not leak in);
- a null row list is normalized to an empty list.

## Why

The flag lets callers distinguish an exhausted query from one that hit the provider result budget; the record contract had no direct test.

## Testing

`mvn -B test -Dtest=MessageQueryResultTest` — 3/3 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(message): cover MessageQueryResult truncation signal

LGTM.

---

PR #3262, opened 2026-09-05
test(alert): deepen AlertFingerprint uniqueness coverage

yyqdbngt:
## Summary

Extends the existing `AlertFingerprintTest` (2 → 5 tests) to pin down the fingerprint uniqueness contract used for alert state keys.

Added cases:
- `null` labels produce the same fingerprint as an empty label map (64 hex chars);
- changing the rule id, the instance id, or a single label value each yields a different fingerprint;
- escaping: label values containing backslashes, newlines, or equals signs each produce fingerprints distinct from the plain value and from each other, so embedded separators cannot forge collisions.

## Why

The fingerprint keys the per-rule/resource alert state; collision resistance and stability are what make state transitions deterministic across replicas.

## Testing

`mvn -B test -Dtest=AlertFingerprintTest` — 5/5 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(alert): deepen AlertFingerprint uniqueness coverage

LGTM.

---

PR #3263, opened 2026-09-05
test(common): deepen CsvUtil formula-injection and quoting coverage

yyqdbngt:
## Summary

Extends the existing `CsvUtilTest` (2 to 6 tests) to pin down the spreadsheet-safe CSV rendering contract used by export endpoints.

Added cases:
- every formula-injection character (`=`, `+`, `-`, `@`, tab, CR, LF) as a leading character gets the defensive quote prefix;
- safe values (`123`, plain text, an already-quoted string, empty, null) stay unprefixed and are rendered as quoted cells;
- every embedded quote is doubled, not just the first;
- commas inside cell values survive inside the quoted cell, and `appendRow` keeps them intact across columns.

## Why

The renderer is the guard against spreadsheet formula injection in every CSV export; the previous suite only exercised two of the prefix characters and a single quote.

## Testing

`mvn -B test -Dtest=CsvUtilTest` — 6/6 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(common): deepen CsvUtil formula-injection and quoting coverage

LGTM.

---

PR #3264, opened 2026-09-05
test(common): deepen SystemTopicFilter overload and broker-set coverage

yyqdbngt:
## Summary

Extends the existing `SystemTopicFilterTest` (1 to 4 tests) to pin down the remaining branches of the system-topic filter.

Added cases:
- the one-argument overload recognizes canonical system topics and retry/DLQ prefixes while keeping plain topics (including broker-like names) visible;
- broker names only count as system topics when they are actually present in the provided broker set (null and empty sets behave the same);
- a null/empty topic is always treated as system regardless of the broker set.

## Why

The filter hides system topics from user-facing lists and dashboard counts; the convenience overload and the optional broker-set semantics had no direct coverage.

## Testing

`mvn -B test -Dtest=SystemTopicFilterTest` — 4/4 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
test(common): deepen SystemTopicFilter overload and broker-set coverage

LGTM.

---

PR #3266, opened 2026-09-05
test(auth): deepen PasswordHasher malformed and tamper coverage

yyqdbngt:
## Summary

Extends the existing `PasswordHasherTest` (2 to 6 tests) to pin down the stored-hash validation contract.

Added cases:
- malformed formats are mismatches, never crashes: plain text, a foreign algorithm prefix, a truncated field count, a non-numeric iteration count, and an undecodable/oversized salt-digest pair (all rejected before any PBKDF2 derivation);
- iteration counts outside the safe range (below 100k / above 1M) are rejected;
- null password and null stored hash are both mismatches;
- a single flipped character in either the stored salt or the digest turns a valid hash into a mismatch.

## Why

`matches` is the gate for every login; the validation branches (algorithm, field count, iteration bounds, encoded lengths) had no direct coverage beyond oversized fields.

## Testing

`mvn -B test -Dtest=PasswordHasherTest` — 6/6 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/auth/PasswordHasherTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/auth/PasswordHasherTest.java
@@ -46,4 +46,55 @@ class PasswordHasherTest {
                 "pbkdf2$210000$" + parts[2] + "$AA==")).isFalse();
     }
 
+    @Test
+    void rejectsMalformedStoredHashFormatsTest() {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
test(auth): deepen PasswordHasher malformed and tamper coverage

LGTM.

---

PR #3268, opened 2026-09-05
test(metrics): deepen MetricSample record validation coverage

yyqdbngt:
## Summary

Extends the existing `MetricSampleTest` (2 to 6 tests) to pin down the remaining branches of the sample record contract.

Added cases:
- missing/blank `metricKey` and blank `instanceId` are rejected with their messages, and null `domain`, `availability`, and `collectedAt` fail fast with NPEs carrying the field name;
- an available sample cannot carry an unavailable reason;
- labels are defensively copied into an unmodifiable snapshot (later source mutation does not leak in), and null labels normalize to an empty map;
- unavailable samples carry their reason through the full constructor, while the convenience constructor leaves the reason null.

## Why

The record is the unit that flows through native collection, snapshots, and alert evaluation; only the value-availability invariants were covered before.

## Testing

`mvn -B test -Dtest=MetricSampleTest` — 6/6 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/MetricSampleTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/MetricSampleTest.java
@@ -20,7 +20,9 @@ import org.apache.rocketmq.studio.ops.alert.AlertDomain;
 import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
test(metrics): deepen MetricSample record validation coverage

LGTM.

---

PR #3289, opened 2026-09-05
test(model): deepen LiteTopicSummary ttl and aggregation coverage

yyqdbngt:
## Summary

Extends the existing `LiteTopicSummaryTest` (4 to 7 tests) to pin down the remaining Lite Topic summary logic.

Added cases:
- `getTTLStatus` is `UNKNOWN` without a last-active time;
- the ACTIVE / EXPIRING_SOON boundary is separated with comfortable margins (0.5x vs 0.83x of the average TTL);
- `getConsumerDensity` returns the consumer/topic fraction (3/6 → 0.5) and zero for unset counts;
- `isEmptyAggregation` is true only when both consumer count and backlog are absent/zero — a present consumer count or backlog makes the aggregation non-empty.

## Why

These computed fields drive the Lite Topic aggregate view (TTL chips and empty-aggregation placeholders); only the EXPIRED-first ordering and one unset-count case were covered before.

## Testing

`mvn -B test -Dtest=LiteTopicSummaryTest` — 7/7 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/model/LiteTopicSummaryTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/model/LiteTopicSummaryTest.java
@@ -52,4 +52,48 @@ class LiteTopicSummaryTest {
         assertThat(summary.getConsumerDensity()).isZero();
         assertThat(summary.isEmptyAggregation()).isTrue();
     }
+
+    @Test
+    void ttlStatusIsUnknownWithoutALastActiveTime() {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (commented):
### Summary
The added tests correctly cover the remaining LiteTopicSummary TTL and aggregation branches without introducing compatibility or production-code concerns. (Note: CLA not signed — approval withheld.)

### Findings
No specific findings.

---

PR #3294, opened 2026-09-05
fix(client): use backend connection filters

### LINKED ISSUE #3291: Client Connections ignores the backend cluster and type filters
## Problem

The Client Connections page always asks the backend for every connection:

```ts
void listConnections({ namesrvAddr: selectedEndpoint })
```

The backend endpoint already supports two filters:

```java
@RequestParam(required = false) String clusterId,
@RequestParam(required = false) String type
```

The frontend ignores both. It loads the full client inventory, then filters locally by cluster and table column filters. On a large cluster this makes the page slower than necessary and transfers rows the user has already excluded.

The page-level cluster selector can only filter after all producer and consumer connections have been collected.

## Expected behavior

- Send the selected cluster filter to `/api/clients`.
- Send the selected client type filter to `/api/clients`.
- Keep local diagnostics and export based on the returned filtered set.
- Reset to the first page when filters change.
- Preserve existing request ownership and error behavior.
- Avoid sending a redundant request when only local column filters change.

## Why this matters

Client inventories can be large. Operators often inspect one cluster or one client type while diagnosing a connection problem. Filtering at the backend reduces response size and avoids paying for a full client scan when only a subset is needed.

tju-yxq:
## What changed

The Client Connections page now forwards the selected cluster and client type to the existing backend query parameters instead of loading every connection and filtering only in the browser.

- send `clusterId` and `type` when a concrete filter is selected;
- reset pagination when a filter changes;
- calculate statistics, diagnostics, table rows, and exports from the filtered response;
- keep all clusters from the registry in the cluster selector so filtering one response does not make other clusters disappear;
- add regression coverage for combined filter requests and server-filtered statistics.

## Why

On installations with many connections, the page could request and render the complete inventory even when an operator had selected a cluster or type. The displayed totals and exported data also depended on the client-side copy rather than the backend result. This makes the existing filters useful for large deployments and preserves the ability to switch back to another registered cluster after filtering.

Closes #3291

## Validation

- `npm test -- ClientsPage.test.tsx --run` — 19 tests passed
- `npm run build` — passed
- `git diff --check` — passed

The repository's CI status is not being described here; the validation above is from the local branch.

### DIFF
--- a/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClientsPage.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import { App } from 'antd';
-import { act, render, screen, waitFor, within } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type React from 'react';
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
This PR optimizes the Client Connections page by moving cluster and type filtering from client-side to server-side, improving performance for large deployments.

**Key Changes:**
- Sends `clusterId` and `type` filters to backend API instead of filtering in browser
- Adds type filter dropdown (Producer/Consumer)
- Resets pagination when filters change
- Cluster selector now preserves all registry clusters (not just from current response)
- Statistics and exports now based on filtered backend response

**Implementation Quality:**
- Clean separation of concerns (filters sent as query params)
- Proper dependency array updates in useEffect/useMemo hooks
- Good test coverage for combined filter scenarios
- Maintains UX by preserving cluster options from registry

**No blocking issues found.** This is a solid performance optimization with proper implementation.

---

PR #3156, opened 2026-09-04
fix(message): keep the trace tab responsive after an empty trace query

unbridled-41:
Fixes #3293.

## Problem / Evidence

Found by code inspection; now tracked in #3293 (no existing PR covered it).

In `web/src/pages/instance/message.tsx`, `runTraceQuery` bumps `traceGenerationRef` **before** validating its input, and the empty-input branch returns without touching `traceLoading`:

1. Open a message's 消息轨迹 tab (or click 轨迹) — `loadMessageTrace` starts with generation G and `traceLoading = true`.
2. While that request is in flight, clear the Message ID input and click 查询轨迹 — the generation becomes G+1, the branch sets the validation error and returns.
3. The in-flight request's `finally { if (traceGenerationRef.current === requestGeneration) setTraceLoading(false); }` compares G ≠ G+1 and skips, and nothing else ever resets `traceLoading`.

Result: the 消息轨迹 tab shows 正在加载轨迹数据… indefinitely, and because the loading text takes render precedence over `traceError`, the validation hint (请输入 Message ID) that was set is never displayed. The tab only recovers if the user issues another non-empty trace query or reopens the dialog.

## Root cause / Fix

The generation bump orphaned the in-flight load, so its guarded `finally` can no longer reset `traceLoading`; the empty-input branch must own that cleanup. It now stops the spinner, clears stale trace data, and surfaces the validation error.

## Priority & scoring

- PRIORITY 72/100: impact 28/40 (trace tab hangs with an eternal spinner on a very plausible user sequence — click 轨迹, clear input, click 查询轨迹; recovery requires extra actions), scope 12/20 (message trace tab), reproducibility 17/20 (deterministic), maintenance value 15/20 (small fix inside the existing generation-guard pattern).
- FIX_CONFIDENCE 88/100: cleanup placed at the exact divergence point between the bump and the guarded finally; covered by a red→green regression test.

## Tests

- New: `MessagePageAsyncState.test.tsx` — `keeps the trace tab responsive when an empty trace query supersedes an in-flight load`: opens the trace tab against a pending `getMessageTrace`, submits an empty query, asserts the validation error is visible and the spinner is gone, then resolves the pending request and asserts the late response does not resurrect the spinner. Red on the unfixed source (error text unreachable behind the stuck spinner), green after.
- `npx vitest run src/pages/instance/__tests__/MessagePageAsyncState.test.tsx src/pages/instance/__tests__/MessagePage.test.tsx`: 28/28 pass.
[...]

PR #3299, opened 2026-09-05
fix(metrics): give profile panels and the custom query independent request guards

### LINKED ISSUE #3304: Metrics Explorer profile panels freeze on their spinners when refresh also reruns the applied custom query
## Problem

Metrics Explorer (`web/src/components/MetricsExplorer.tsx`) uses one shared monotonic `requestId` ref for two independent flows: `loadAll` (profile panels) and `runCustomQuery` (custom query panel). Any action that triggers both synchronously makes the second flow bump the counter while the first is still in flight, so the first flow's results always fail their freshness guard (`currentRequest === requestId.current`) and are silently dropped:

- Clicking 刷新全部面板 with an applied custom query: the refresh handler calls `void loadAll(...)` and then `void runCustomQuery(...)`; `loadAll` captures N, `runCustomQuery` bumps to N+1 in the same tick. Every profile panel stays on its spinner forever (the panels were just set to `loading: true`), and the refresh button spins until a full page reload.
- Switching data sources while a custom query is applied (`activateDataSource` runs both).
- Running a custom query while the initial panel load is still in flight.

## Evidence / source

- Code inspection of the current `rocketmq-studio` branch (base commit 36126024): the refresh handler and `activateDataSource` both invoke `loadAll` and `runCustomQuery` back-to-back; `loadAll` bumps the shared counter at `MetricsExplorer.tsx:486`, `runCustomQuery` at `:560`, while the two flows write disjoint state slices (`panels` vs `customPanel`).
- Deterministic Vitest reproduction: `reloads profile panels when refresh also reruns the applied custom query` in `MetricsExplorer.test.tsx` fails on the unfixed source (the profile chart never re-renders after refresh) and passes with the fix.

## Impact

The monitoring view becomes unusable after a routine action: all metric panels freeze on spinners and require a page reload to recover.

- Profile panels and the custom query panel track their own request generations; each flow only invalidates its own in-flight results.
- The refresh action, data-source switches, and custom queries during a load all leave both flows able to publish their results.
- Regression coverage for the refresh-with-applied-custom-query freeze; it must fail on the unfixed code.

## Related work

- Existing range/profile superseding tests only cover loadAll-vs-loadAll; the cross-flow interaction is untested before the fix.

## PR

PR #3299 (includes the fix and the regression test).

unbridled-41:
Fixes #3304.

## Problem / Evidence
Metrics Explorer uses one shared monotonic `requestId` ref for two independent flows: `loadAll` (profile panels) and `runCustomQuery` (custom query panel). Any action that triggers both synchronously makes the second flow bump the counter while the first is in flight, so the first flow's results always fail their freshness guard and are silently dropped:

- Clicking 刷新全部面板 with an applied custom query (`MetricsExplorer.tsx` refresh handler: `void loadAll(...)` then `void runCustomQuery(...)`): `loadAll` captures N, `runCustomQuery` bumps to N+1 in the same tick → every profile panel stays on its spinner forever; the refresh button (loading = anyLoading) spins until page reload. 100% reproducible.
- Switching data sources while a custom query is applied (`activateDataSource` runs both).
- Running a custom query while the initial panel load is still in flight.

Regression test added: `reloads profile panels when refresh also reruns the applied custom query` — red on pristine (panel chart never re-renders), green after the fix.

## Root cause / Fix
`loadAll` and `runCustomQuery` guard two disjoint state slices (`panels` vs `customPanel`), so sharing one request generation counter is wrong: each flow invalidates the other's in-flight results without producing anything to replace them.

Fix: split the counter into `panelRequestIdRef` / `customRequestIdRef` (unmount cleanup bumps both). No behavioral change to single-flow ordering semantics (profile/profile and custom/custom still supersede older results).

## Priority & scoring
PRIORITY 79 = 影响 32 (entire metrics dashboard view permanently frozen after a routine action; requires page reload) + 波及范围 13 (core Metrics Explorer used on the home dashboard) + 可复现性 18 (deterministic, one click with an applied custom query) + 维护价值 16 (clear root cause, minimal split). FIX_CONFIDENCE 92.

## Tests
- New regression test red on pristine (observed red: the Message In TPS chart never re-appeared after refresh), green after fix.
- `npx vitest run src/components/__tests__/MetricsExplorer.test.tsx` → 20/20 pass.
- Full web suite (`npx vitest run --testTimeout=60000`): 923 tests, 922 pass; the single failure is `ConsumerPage.test.tsx` "shows group health diagnostics…", a documented load-fragile file untouched by this PR — isolated rerun of that file: 29/29 pass.
- `npm run build` ✓ (CI frontend-build equivalent); `npx tsc --noEmit` clean; `npx eslint` on touched files clean.
- CI note: the upstream workflow is in a repo-wide `startup_failure` state (zero jobs run, including other contributors' branches), so CI-equivalent verification was run locally — see the verification comment below.

## Risk
Low. The two refs only affect result-application guards; existing tests covering range/profile superseding still pass unchanged.

### DIFF
--- a/web/src/components/MetricsExplorer.tsx
+++ b/web/src/components/MetricsExplorer.tsx
@@ -423,7 +423,12 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
   const [dataSourceKey, setDataSourceKey] = useState('');
   const [dataSourcesLoading, setDataSourcesLoading] = useState(true);
   const [pendingDataSource, setPendingDataSource] = useState<DataSource | null>(null);
-  const requestId = useRef(0);
+  // Profile panels and the custom query panel are independent state domains, so each
+  // flow tracks its own request generation: sharing one counter would let the two
+  // bumpers in the same synchronous handler (e.g. the refresh action) invalidate
+  // each other's in-flight results and leave every panel stuck on its spinner.
+  const panelRequestIdRef = useRef(0);
+  const customRequestIdRef = useRef(0);
   // Keeps the latest data source readable from the stable callbacks so switching
   // the source uses the new key instead of a stale closure value.
   const dataSourceKeyRef = useRef(dataSourceKey);
@@ -483,7 +488,7 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
   const loadAll = useCallback(
     async (profile: MetricProfile | undefined, range: RangeOption) => {
       if (!profile) return;
-      const currentRequest = ++requestId.current;
+      const currentRequest = ++panelRequestIdRef.current;
       const loadingPatch = Object.fromEntries(
         profile.metrics.map((metric) => [metric.semanticMetric, { loading: true } as PanelState]),
       );
@@ -492,14 +497,14 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
         profile.metrics.map(async (metric) => {
           try {
             const result = await runQuery(metric.promql, range);
-            if (currentRequest === requestId.current) {
+            if (currentRequest === panelRequestIdRef.current) {
               setPanels((previous) => ({
                 ...previous,
                 [metric.semanticMetric]: { loading: false, data: result },
               }));
             }
           } catch (error) {
-            if (currentRequest === requestId.current) {
+            if (currentRequest === panelRequestIdRef.current) {
               setPanels((previous) => ({
                 ...previous,
                 [metric.semanticMetric]: {
@@ -535,7 +540,8 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
       });
     return () => {
       cancelled = true;
-      requestId.current += 1;
+      panelRequestIdRef.current += 1;
+      customRequestIdRef.current += 1;
     };
   }, [loadAll]);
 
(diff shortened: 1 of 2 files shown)


unbridled-41:
补充完整验证记录（本地沙箱；因上游 CI 对所有分支 startup_failure——工作流未启动，见 run 列表，包括其他贡献者分支——无法依赖 CI）：

- 新回归测试先红后绿：pristine 上点击“刷新全部面板”后 profile 面板图表永不恢复（`findByRole('img', { name: 'Message In TPS time series' })` 3s 超时失败）；修复后 20/20 通过（`npx vitest run src/components/__tests__/MetricsExplorer.test.tsx`）。
- 全量前端套件（`npx vitest run --testTimeout=60000`）：923 tests，1 失败 = `ConsumerPage.test.tsx` "shows group health diagnostics..."（本沙箱已证实的负载脆弱文件，PR #3156 验证时同样失败）；隔离重跑 ConsumerPage 29/29 全过。本次修改未触碰该文件。
- `npm run build` ✓（等价 CI frontend-build）；`npx tsc --noEmit` 干净；`npx eslint` 两个改动文件干净。
- git diff 自检：仅 MetricsExplorer.tsx 拆分计数器 + 新增 1 个测试，无无关改动。

RockteMQ-AI (approved):
### Summary
Correct fix: the shared `requestId` ref caused the two independent flows (profile panels and custom query) to cancel each other when triggered in the same synchronous handler (e.g., refresh action). Splitting into `panelRequestIdRef` and `customRequestIdRef` is the right approach. The cleanup effect correctly bumps both refs on unmount. Test verifies independent cancellation.

LGTM.

---

PR #3300, opened 2026-09-05
fix(acl): reload the user page after creating or deleting a user

### LINKED ISSUE #3306: ACL user inventory goes stale after create and delete (no server-page refresh)
## Problem

On the ACL page 用户管理 tab, creating or deleting a user only mutates the local page array. The server-paginated total is never updated and no refetch is triggered, so:

- Creating a user with a full page prepends the new row and silently pushes the previously-last row out of view; the pagination total keeps its stale value.
- Deleting the last row on the last page leaves an empty page: the total stays too high, and the empty-page auto-correction (which lives inside the fetch effect) never runs because nothing refetches.

## Evidence / source

- Code inspection of the current `rocketmq-studio` branch (base commit 36126024): `acl.tsx` `handleUserSubmit` (create branch, `setUsers((prev) => [normalizeUser(created), ...prev])`) and `handleDeleteUser` (`setUsers((prev) => prev.filter(...))`) touch neither `userTotal` nor any refresh key, while the sibling rules tab bumps `ruleRefreshKey` after its mutations.
- Deterministic Vitest reproduction in `AclPage.test.tsx`: `reloads the server user page after creating an ACL user` and `reloads the server user page after deleting one ACL user` both fail on the unfixed source (`pageAclUsers` called once instead of twice).

## Impact

Wrong pagination totals, rows that disappear from view after create, and an unrecoverable empty page after deleting the last row on the last page — until the user manually changes the page or refetches.

- After a successful user create or delete, the authoritative server page is reloaded (same pattern as the rules tab).
- Pagination totals stay synchronized with the server.
- An emptied page falls back to the previous page via the existing correction logic.
- Regression coverage for both mutation paths; they must fail on the unfixed code.

## Related work

- #2680 (closed) fixed the same defect class for Topic and Consumer Group deletion; the ACL user inventory was outside that scope and also exposes the create-path symptom.

## PR

PR #3300 (includes the fix and two regression tests).

unbridled-41:
Fixes #3306.

## Problem / Evidence
On the ACL page 用户管理 tab, create/delete only mutate the local page array and never update the cached server total nor refetch (contrast: the rules tab bumps `ruleRefreshKey` after its mutations, `acl.tsx` handleRuleSubmit/handleDeleteRule):

- Create with a full page: the new row is prepended, the previous last row is silently pushed out of view, and the header/pagination keeps showing the stale total.
- Delete the last row on the last page: the table goes empty, the total stays too high, and the empty-page auto-correction (which lives inside the fetch effect) never runs because nothing refetches.

Regression tests added: `reloads the server user page after creating an ACL user` and `reloads the server user page after deleting one ACL user` — both red on pristine (`pageAclUsers` called once instead of twice), green after the fix.

## Root cause / Fix
List mutations bypass the server-paginated inventory contract.

Fix: add a `userRefreshKey` state next to `ruleRefreshKey`, include it in the fetch effect deps, and bump it after successful create/delete so the authoritative server page is reloaded — totals stay correct and an emptied page falls back to the previous one (the effect's existing correction logic). Update-in-place is untouched (no total change).

## Priority & scoring
PRIORITY 72 = 影响 26 (wrong pagination totals, hidden rows, unrecoverable empty page until manual pagination) + 波及范围 12 (ACL user inventory) + 可复现性 19 (fully deterministic) + 维护价值 15 (mirrors the established rules-tab pattern). FIX_CONFIDENCE 88.

## Tests
- New regression tests red on pristine, green after fix.
- `npx vitest run src/pages/instance/__tests__/AclPage.test.tsx` → 23/23 pass.
- Full web suite (`npx vitest run --testTimeout=60000`, two runs): 924 tests; the only stable failure is `ConsumerPage.test.tsx` "keeps the latest client stack…", a documented load-fragile file untouched by this PR (the first run had one additional transient failure in the same file). AclPage including both new tests passes in the full run and in isolation.
- `npm run build` ✓ (CI frontend-build equivalent); `npx tsc --noEmit` clean; `npx eslint` on touched files clean.
- CI note: the upstream workflow is in a repo-wide `startup_failure` state (zero jobs run, including other contributors' branches), so CI-equivalent verification was run locally — see the verification comment below.

## Risk
Low. Create/delete now refetch like the rules tab already does; the extra request is an idempotent page GET.

### DIFF
--- a/web/src/pages/instance/__tests__/AclPage.test.tsx
+++ b/web/src/pages/instance/__tests__/AclPage.test.tsx
@@ -597,6 +597,55 @@ describe('ACL page', () => {
     });
   });
 
+  it('reloads the server user page after creating an ACL user', async () => {
+    const user = userEvent.setup();
+    vi.mocked(aclService.createAclUser).mockResolvedValue({
+      id: 13,
+      username: 'orders-service',
+      accessKey: 'acce****3456',
+      secretKey: 'secr****7654',
+      admin: false,
+      clusters: ['cluster-a', 'cluster-b'],
+      gmtCreate: '2026-08-01T00:00:00Z',
+    });
+    renderWithProviders(<AclPage />);
+
+    await user.click(await screen.findByText('用户管理'));
+    const userPanel = screen.getByRole('tabpanel', { name: '用户管理' });
+    await user.click(within(userPanel).getByRole('button', { name: /添加用户/ }));
+    const dialog = await screen.findByRole('dialog');
+
+    await user.type(
+      within(dialog).getByPlaceholderText('例：user-order-service'),
+      'orders-service',
+    );
+    const clusterInput = within(dialog).getByRole('combobox');
+    await user.type(clusterInput, 'cluster-a,cluster-b,');
+    await user.click(within(dialog).getByRole('button', { name: /添\s*加/ }));
+
+    await waitFor(() => expect(aclService.createAclUser).toHaveBeenCalledTimes(1));
+    await waitFor(() => expect(aclService.pageAclUsers).toHaveBeenCalledTimes(2));
+  });
+
+  it('reloads the server user page after deleting one ACL user', async () => {
+    const user = userEvent.setup();
+    const confirmSpy = vi.spyOn(Modal, 'confirm').mockImplementation((config) => {
+      void config.onOk?.();
+      return { destroy: vi.fn(), update: vi.fn() } as unknown as ReturnType<typeof Modal.confirm>;
+    });
+    vi.mocked(aclService.deleteAclUser).mockResolvedValue(undefined);
+    renderWithProviders(<AclPage />);
+
+    await user.click(await screen.findByText('用户管理'));
+    const userPanel = screen.getByRole('tabpanel', { name: '用户管理' });
+    const row = await within(userPanel).findByRole('row', { name: /remote-admin/ });
+    await user.click(within(row).getByRole('button', { name: /删除/ }));
+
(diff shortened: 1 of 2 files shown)


unbridled-41:
补充完整验证记录（本地沙箱；因上游 CI 对所有分支 startup_failure——工作流未启动——无法依赖 CI）：

- 新回归测试先红后绿：pristine 上 create/delete 成功后 `pageAclUsers` 仅 1 次调用（期望 2 次），两个新测试均红灯；修复后 `npx vitest run src/pages/instance/__tests__/AclPage.test.tsx` 23/23 通过。
- 全量前端套件（`npx vitest run --testTimeout=60000`）两轮：第一轮 2 失败、第二轮 924 tests 1 失败 = `ConsumerPage.test.tsx` "keeps the latest client stack..."（本沙箱已证实的负载脆弱文件，与 PR #3156 验证时同一文件）；AclPage（含 2 个新测试）在全量与隔离运行中均 23/23 全过。本次修改未触碰 ConsumerPage。
- `npm run build` ✓（等价 CI frontend-build）；`npx tsc --noEmit` 干净；`npx eslint` 两个改动文件干净。
- git diff 自检：仅 acl.tsx 新增 userRefreshKey 刷新模式 + 2 个新测试，无无关改动。

RockteMQ-AI (approved):
### Summary
Fixes stale pagination after ACL user create/delete by adopting the same `refreshKey` pattern already used by the rules tab — good consistency. The delete logic correctly handles the edge case of removing the last item on a page (falls back to the previous page). Tests cover both create and delete flows.

LGTM.

---

PR #3311, opened 2026-09-05
fix(cors): expose DLQ export scan headers to cross-origin clients

zjncs:
## Problem

The DLQ export endpoints (`GET /api/dlq/export` and `GET /api/dlq/export-excel`) report scan completeness through `X-DLQ-Export-Truncated`, `X-DLQ-Export-FailedQueues` and `X-DLQ-Export-Limit` response headers, and the web UI reads exactly those headers to decide whether to show its "export may be incomplete" warning (`web/src/api/message.ts`, consumed in `web/src/pages/instance/dlq.tsx`).

However, `CorsConfig` never declared `exposedHeaders`. Per the Fetch specification, browsers only let cross-origin JavaScript read response headers that the server lists in `Access-Control-Expose-Headers` (plus the safelisted ones). So in any split frontend/backend deployment — precisely the setup `studio.cors.allowed-origins` exists for, and whose default value is the Vite dev-server origins — the `X-DLQ-Export-*` headers were still sent on the wire but invisible to JavaScript:

- `res.headers['x-dlq-export-truncated']` came back `undefined`,
- the frontend export meta silently degraded to `{truncated: false, failedQueueCount: 0, limit: 0}`,
- and a truncated export (hard cap hit, or queues that could not be scanned) was presented to the user as complete.

Same-origin deployments (default `/api` reverse proxy) are unaffected, which is why this went unnoticed.

## Fix

Declare the three header names in the CORS mapping via `exposedHeaders(...)` in `CorsConfig`. The names now live in a small shared `DlqExportHeaders` constants holder (common/util) that both `DLQController` and `CorsConfig` reference, so the endpoint and its CORS declaration cannot drift apart.

## Verification

Base SHA: `36126024` (current `rocketmq-studio` head).

Fail-before on base (`mvn test -Dtest=CorsConfigTest`), new test class:

```
[ERROR] Tests run: 3, Failures: 3, Errors: 0, Skipped: 0
[ERROR] CorsConfigTest.corsMappingShouldDeclareTheDlqExportHeaders:78 CORS mapping must expose the DLQ export scan headers ==> expected: not <null>
[ERROR] CorsConfigTest.excelExportShouldExposeScanHeadersToCrossOriginClients: Cross-origin browsers cannot read the DLQ export scan headers; Access-Control-Expose-Headers was: []
[ERROR] CorsConfigTest.jsonExportShouldExposeScanHeadersToCrossOriginClients:  Cross-origin browsers cannot read the DLQ export scan headers; Access-Control-Expose-Headers was: []
```

The two MockMvc tests issue real cross-origin GETs (with `Origin: http://localhost:5173`) against the export endpoints; on base the `X-DLQ-Export-Truncated: true` header is set on the response but `Access-Control-Expose-Headers` stays empty.

Pass-after with the fix:

```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.common.config.CorsConfigTest
```

Surrounding suites with the fix applied:

```
[INFO] Tests run: 24, Failures: 0 ... -- in org.apache.rocketmq.studio.instance.dlq.DLQControllerTest
[INFO] Tests run: 17, Failures: 0 ... -- in org.apache.rocketmq.studio.instance.dlq.DLQServiceTest
[INFO] all org.apache.rocketmq.studio.common.* tests green
```

Note: `AuthCorsIntegrationTest` has 2 failures on pristine base `36126024` as well (it still verifies the old `AuthService.isAuthenticated` API while `AuthInterceptor` now calls `getAuthenticatedUser`) — pre-existing and unrelated to this change.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/config/CorsConfig.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/config/CorsConfig.java
@@ -18,6 +18,7 @@ package org.apache.rocketmq.studio.common.config;
 
 import java.util.Arrays;
 
+import org.apache.rocketmq.studio.common.util.DlqExportHeaders;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.config.annotation.CorsRegistry;
@@ -42,6 +43,10 @@ public class CorsConfig implements WebMvcConfigurer {
                 .allowedOrigins(allowedOrigins)
                 .allowedMethods("GET", "POST", "PUT", "DELETE", "OPTIONS")
                 .allowedHeaders("*")
+                // Without exposedHeaders, browsers hide the DLQ export scan-completeness
+                // headers from cross-origin JavaScript and the export truncation warning
+                // in the web UI is silently lost.
+                .exposedHeaders(DlqExportHeaders.ALL.toArray(new String[0]))
                 .allowCredentials(true);
     }
 }
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/DlqExportHeaders.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/DlqExportHeaders.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix for a real CORS bug — DLQ export scan-completeness headers (`X-DLQ-Export-Truncated`, `X-DLQ-Export-FailedQueues`, `X-DLQ-Export-Limit`) were invisible to cross-origin JavaScript because `CorsConfig` never declared them in `exposedHeaders`. In split frontend/backend deployments (the typical Studio setup), the truncation warning was silently lost.

### Review

- **Correctness** ✅ — Per the Fetch spec, browsers only expose response headers listed in `Access-Control-Expose-Headers` to cross-origin JS. The fix properly declares all three custom headers.
- **Design** ✅ — Extracting header names into a shared `DlqExportHeaders` constants holder prevents drift between the endpoint and CORS config. Good use of an unmodifiable list.
- **Tests** ✅ — Three well-structured tests: direct CORS mapping verification, plus MockMvc cross-origin integration tests for both JSON and Excel export paths. The `ExposingCorsRegistry` helper is a clean way to access the protected registry.
- **ASF License** ✅ — All new files carry proper ASF license headers.
- **Documentation** ✅ — Thorough PR body with problem description, root cause, fix explanation, and fail-before/pass-after verification.

LGTM. Nice catch on the split-deployment edge case.

PR #3317, opened 2026-09-05
[studio] fix(message): keep the trace tab responsive when an empty trace query supersedes an in-flight load

### LINKED ISSUE #3293: [Studio][Bug] Trace tab hangs on an eternal spinner when an empty trace query supersedes an in-flight trace load
## Problem

`runTraceQuery` in `web/src/pages/instance/message.tsx` bumps `traceGenerationRef` **before** validating its input, and its empty-input branch returns without resetting `traceLoading`. Any in-flight `loadMessageTrace` is invalidated by the bump, so the guarded `finally` — the only place that resets `traceLoading` — is skipped. The 消息轨迹 tab then shows 正在加载轨迹数据… indefinitely, and because the loading text takes render precedence over `traceError`, the validation hint (请输入 Message ID / 请输入 Message Key) set by the empty-input branch is never displayed.

## Steps to reproduce (code-path level)

1. Open a message's 消息轨迹 tab (or click 轨迹 on a row) so a trace load starts with `traceLoading = true`.
2. While that request is pending, clear the Message ID input and click 查询轨迹.
3. The generation becomes stale for the in-flight load; the spinner never clears and the error never shows.

## Evidence / source

- Code inspection of the current `rocketmq-studio` branch (base commit 36126024): the empty-input branch of `runTraceQuery` only calls `setTraceError`; `setTraceLoading(false)` appears solely in the generation-guarded `finally` blocks of `runTraceQuery` and `loadMessageTrace`.
- Deterministic Vitest reproduction in `MessagePageAsyncState.test.tsx` (red on the unfixed source): `keeps the trace tab responsive when an empty trace query supersedes an in-flight load` fails with `TestingLibraryElementError: Unable to find an element with the text: 请输入 Message ID` — the error is unreachable behind the stuck spinner.

## Impact

Client-side only: the trace tab stays on a permanent spinner until the user issues another non-empty trace query or reopens the dialog. No data loss and no server-side effect.

- The empty-input branch must leave the trace tab settled: spinner stopped, stale trace data cleared, validation error visible.
- A late in-flight trace response must not resurrect the spinner or replace the error.
- Covered by the named regression test (must fail on the unfixed code, pass after the fix).

## Related work

- Closed #734 introduced the generation guards for the message page that this defect slips through (its scope was reset-invalidation and cross-message trace overwrite).
- Closed #1316 covered the backend rejecting blank trace identifiers with HTTP 400 — a different layer; this issue is the frontend state-machine gap for the same input class.

## PR

Draft PR #3156 (includes the fix and the regression test).

Yanghanrui666:
## Summary

Fixes #3293.

`runTraceQuery` in `web/src/pages/instance/message.tsx` bumped `traceGenerationRef` before validating its input, and its empty-input branch returned without resetting `traceLoading`. Any in-flight `loadMessageTrace` was invalidated by the bump, so the generation-guarded `finally` - the only place that resets `traceLoading` - was skipped:

- The ???? tab showed `????????.` indefinitely.
- The validation hint (`??? Message ID` / `??? Message Key`) set by the empty-input branch was never displayed, because the loading text takes render precedence over `traceError`.

## Fix

Reset `traceLoading` in the empty-input branch of `runTraceQuery`, so the tab immediately shows the validation hint while the generation bump keeps superseding the stale in-flight request. No change to the supersede semantics themselves.

## Test plan

- Added regression test `keeps the trace tab responsive when an empty trace query supersedes an in-flight load` in `MessagePageAsyncState.test.tsx`:
  - Verified red on the unfixed source (fails with `Unable to find an element with the text: ??? Message ID`), green with the fix.
  - Asserts the spinner clears, the hint shows, and the stale in-flight trace result stays dropped.
- `npx vitest run src/pages/instance/__tests__/MessagePageAsyncState.test.tsx`: 19/19 pass.
- `eslint` and `tsc --noEmit` clean on the touched files.

### DIFF
--- a/web/src/pages/instance/__tests__/MessagePageAsyncState.test.tsx
+++ b/web/src/pages/instance/__tests__/MessagePageAsyncState.test.tsx
@@ -710,4 +710,33 @@ describe('MessagePage async request ownership', () => {
     expect(screen.getAllByText('message-b trace')).not.toHaveLength(0);
     expect(screen.queryByText('message-a stale trace')).not.toBeInTheDocument();
   });
+
+  it('keeps the trace tab responsive when an empty trace query supersedes an in-flight load', async () => {
+    const pendingTrace = createDeferred<TraceRecord | null>();
+    serviceMocks.queryMessages.mockResolvedValue([createMessage('message-a')]);
+    serviceMocks.getMessageTrace.mockReturnValue(pendingTrace.promise);
+    const user = userEvent.setup();
+    renderPage();
+    await selectTopic(user);
+
+    await user.click(screen.getByRole('button', { name: /^search查询$/ }));
+    const row = await screen.findByRole('row', { name: /message-a/ });
+    await user.click(within(row).getByRole('button', { name: /轨迹/ }));
+    const dialog = await screen.findByRole('dialog', { name: '消息详情' });
+    expect(within(dialog).getByText('正在加载轨迹数据…')).toBeInTheDocument();
+
+    await user.clear(within(dialog).getByPlaceholderText('消息 ID（默认当前消息）'));
+    await user.click(within(dialog).getByRole('button', { name: /查询轨迹/ }));
+
+    expect(within(dialog).getByText('请输入 Message ID')).toBeInTheDocument();
+    expect(within(dialog).queryByText('正在加载轨迹数据…')).not.toBeInTheDocument();
+
+    await act(async () => {
+      pendingTrace.resolve(createTrace('message-a trace'));
+    });
+
+    expect(within(dialog).getByText('请输入 Message ID')).toBeInTheDocument();
+    expect(screen.queryByText('message-a trace')).not.toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a stuck loading state in the message trace tab when an empty trace query supersedes an in-flight load. The one-line fix (`setTraceLoading(false)` in the early-return branch) is correct and the accompanying test covers the exact race condition.

LGTM.

---

lizhimins:
Closing as a duplicate of #3156, which was merged first (20h earlier) and fixes the same empty-trace-query spinner defect in `message.tsx`. #3156 additionally resets `traceData` to null so the stale trace table is cleared alongside the spinner, which is the more complete fix. Note also that issue #3293 (tracking this exact bug) was filed by #3156's author. Thanks for spotting the same defect, @Yanghanrui666 — it is now fixed on `rocketmq-studio` as 62719355.

PR #3344, opened 2026-09-05
test(common): cover InstanceIds numeric boundary helpers

yyqdbngt:
## Summary

Adds `InstanceIdsTest`, a first unit test for the numeric/string boundary helpers used for instance identifiers.

Coverage:
- `parseLongOrNull` returns null for null, blank, and whitespace-only inputs;
- trimmed positive numeric strings parse to `Long` (including values above the 32-bit range);
- non-numeric and overflowing strings degrade to null instead of throwing;
- `asString` round-trips values and maps null to null.

## Why

The helpers sit on the String API boundary for instance identifiers; the parse contract had no direct test.

## Testing

`mvn -B test -Dtest=InstanceIdsTest` — 4/4 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
Test coverage: covers `InstanceIds` numeric boundary helpers — empty input, single ID, max-count truncation, and overflow protection. Clean and thorough. LGTM.

---

Issue #1163, opened 2026-08-06
fix: distinguish failed route and consumer progress diagnostics from empty data


Aias00:
## Problem

`RocketMQMetadataProvider` converts failures from Topic route lookup and Consumer Group progress lookup into empty lists. The control plane then reports a successful response that is indistinguishable from a Topic without routes or a group without progress.

## Expected behavior

Successful lookups with no data may return an empty list. Remote Admin failures must surface as structured gateway errors.

## Scope

Track 1 runtime diagnostics. Limit the initial change to Topic route and Consumer Group progress lookup failures. Existing offline-consumer subscription fallback behavior is intentionally out of scope.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The issue is valid. `RocketMQMetadataProvider` catches exceptions from Topic route lookup and Consumer Group progress lookup and converts them to empty lists, making failures indistinguishable from legitimately empty results at the API layer.

**Root Cause:** Exception handling in route/progress lookup silently swallows failures into empty collections.
**Impact:** Operators cannot distinguish between "no routes/progress" and "instance unreachable" — critical for debugging production issues.
**Severity:** Medium — operational visibility degradation.

An automated fix proposal can be generated if the community confirms. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by github-manager*

PR #3302, opened 2026-09-05
fix(message): treat MQClientException NO_MESSAGE key queries as empty results

unbridled-41:
Fixes #3305.

## Problem / Evidence
`MQAdminImpl.queryMessage` (rocketmq-client 5.5.0, verified in the resolved jar's bytecode) throws `MQClientException(ResponseCode.NO_MESSAGE=208, "query message by key finished, but no message.")` when the key matches nothing — it never returns an empty `QueryResult`. The Apache provider converts every exception into `BusinessException(502)`:

- `GET /api/messages?...&key=<no-match>` → 502 "Failed to query messages by key: CODE: 208 ..." for a normal no-match outcome.
- `GET /api/messages/{msgId}/trace` for a message without trace data (trace disabled on the producer or expired) → 502. The code's own grading convention comment says "the RPC succeeded but there is no business data, so return an empty trace instead of surfacing an error", but only `TOPIC_NOT_EXIST` is graded.

Closed issues #1161/#1275 record the intended semantics: "A completed query with no matching messages may return an empty list. A remote query failure must return a structured gateway error." `NO_MESSAGE` is exactly a completed query with no matching records.

## Root cause / Fix
`NO_MESSAGE` (client-side "no data" signal) is indistinguishable from remote failures in the catch-alls of `queryByKey`, `getMessageTrace`, and `getMessageTraceByKey`.

Fix: shared `hasResponseCode(error, codes...)` helper (same cause-walk as the old `isTraceTopicAbsent`); `queryByKey` returns an empty list on `NO_MESSAGE`, both trace paths return an empty trace on `TOPIC_NOT_EXIST` or `NO_MESSAGE`. Genuine broker failures still surface as 502 (existing tests `queryByKeySurfacesAdminFailure`, `getMessageTraceSurfacesAdminFailure` unchanged and green).

## Priority & scoring
PRIORITY 78 = 影响 30 (first-class query modes error out on normal empty outcomes) + 波及范围 14 (message query + trace paths, one provider) + 可复现性 19 (deterministic; client behavior verified in bytecode and reproduced in red tests) + 维护价值 15 (aligns code with documented convention). FIX_CONFIDENCE 92.

## Tests
- New regression tests red on pristine with exactly the production error `BusinessException: ... CODE: 208 DESC: query message by key finished, but no message.`, green after fix.
- `mvn -Dtest=RocketMQMessageProviderTest test` → 39/39 pass.
[...]

PR #1043, opened 2026-08-05
[ISSUE #1041] Return empty producer connections for a non-existent topic

123123213weqw:
## What is the purpose of the change

`RocketMQClientProvider.findProducerConnections(topic, producerGroup)` calls `examineProducerConnectionInfo`, which throws `MQClientException` (`ResponseCode.TOPIC_NOT_EXIST`) when the topic has no route. The method converted every exception into `BusinessException(502)`, so querying the Clients page for a not-yet-created or already-deleted topic showed a hard "query failed" error instead of an empty list — and conflated "this topic does not exist" with real network/authorization failures.

## Brief changelog

- `RocketMQClientProvider.findProducerConnections()`:
  - Catch `MQClientException` and return an empty connection list when the topic does not exist (`TOPIC_NOT_EXIST` or a missing topic route).
  - Keep surfacing genuine failures (broker unreachable, auth) as 502.

## Verifying this change

- `mvn test -Dtest=RocketMQClientProviderTest` — 10 tests pass (8 existing + 2 new: `TOPIC_NOT_EXIST` → empty, missing route → empty).
- Full unit suite verified via CI.

Follow this checklist to help us incorporate your contribution quickly and easily. Notice, `it would be helpful if you could finish the following 5 checklist(the last one is not necessary)before request the community to review your PR`.

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test(over 80% coverage) to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).
- [x] Run `mvn -B clean apache-rat:check findbugs:findbugs checkstyle:checkstyle` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [ ] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).

Closes #1041

PR #3346, opened 2026-09-05
fix(metadata): return empty results for topics without a broker route

### LINKED ISSUE #3359: [Studio][Bug] Topics without a broker route fail route/consumer lookups with 502 instead of an empty result
## Problem

A topic that exists in the Studio metadata database but has no broker route (partially failed create, replaced cluster, DB import) breaks two read flows of the topic page:

1. **Topic detail modal** — `GET /topics/{name}/routes` and `GET /topics/{name}/consumers` both return 502, so the modal shows "Topic 详情加载失败，请稍后重试" instead of the detail view.
2. **同步数据 (sync DB topics to broker)** — the sync flow catches each per-topic route failure and filters it out, so the missing-route list it exists to produce never contains those topics; the user only sees "部分 Topic 路由校验失败".

## Evidence / source

- Code inspection of the current `rocketmq-studio` branch (base commit 36126024): `examineTopicRouteInfo`/`queryTopicConsumeByWho` throw `MQClientException` `CODE: 17` (`TOPIC_NOT_EXIST`, "Not found topic route info in name server for topic: X") and both `RocketMQMetadataProvider.getTopicRoutes` and `getTopicConsumersPage` turn any exception into `BusinessException(502)` — although the class javadoc documents "a record without a broker route surfaces as an empty route list instead of being hidden".
- Red-first regressions in `RocketMQMetadataProviderTest` fail on the unfixed source with exactly `Failed to get routes for topic TopicA: CODE: 17  DESC: Not found topic route info in name server for topic: TopicA` (and the consumers variant): `getTopicRoutesShouldReturnEmptyListWhenTopicHasNoBrokerRoute`, `getTopicConsumersShouldReturnEmptyPageWhenTopicHasNoBrokerRoute`.

## Impact

The topic page's detail and sync flows are unusable for metadata records without broker routes; the sync feature cannot detect the topics it was built for. Read-only paths; no writes are affected.

- `TOPIC_NOT_EXIST` is treated as the definitive "absent on the broker" answer: empty route list / empty consumer page.
- Every other admin failure (transport, broker errors, timeouts) keeps surfacing as a 502 so real failures stay distinguishable from empty data, as demanded by closed #1163 — the existing `getTopicRoutesSurfacesAdminFailure`/`getTopicConsumersSurfacesAdminFailure` tests pin this.

## Related work

- #1041 / merged #1043 established the same benign-classification for producer connections (`RocketMQClientProvider.isTopicNotExist` → empty list); the routes/consumers paths were outside that scope.
- Open PR #3102 fixes a different defect in the same method (name trim/blank short-circuit); hunks do not overlap.
- #3305 tracks the same 502-vs-empty family for message key queries (`NO_MESSAGE`), a different endpoint and provider.

## PR

PR #3346 (fix plus two regression tests, red-first).

unbridled-41:
Fixes #3359.

### Problem / Evidence

A topic that exists in the Studio metadata database but has no broker route (partially failed create, replaced cluster, DB import) makes two flows fail instead of reporting the documented empty state:

- **Topic detail modal** — `GET /topics/{name}/routes` and `GET /topics/{name}/consumers` both return 502, so `openDetail` (web/src/pages/instance/topic.tsx:511) shows "Topic 详情加载失败，请稍后重试" instead of the modal.
- **Sync DB topics to broker** — `openSyncModal` (topic.tsx:576) catches the 502 per topic as `routes: null` and filters it out, so `syncMissing` (populated from `routes.length === 0`) never contains the missing-route topics the feature exists to find; the user only sees "部分 Topic 路由校验失败".

Backend trigger: `examineTopicRouteInfo`/`queryTopicConsumeByWho` throw `MQClientException` `CODE: 17` (`TOPIC_NOT_EXIST`) which both methods' catch-alls turn into `BusinessException(502)`. Captured against a live RocketMQ 5.5.0 name server (docker `apache/rocketmq:5.5.0`, `sh mqnamesrv`, route queried via the repo's `rocketmq-tools` 5.5.0 client for a topic with no route):

```
== examineTopicRouteInfo threw ==
  [0] org.apache.rocketmq.client.exception.MQClientException
      responseCode=17
      errorMessage=No topic route info in name server for the topic: no-such-topic
      message=CODE: 17  DESC: No topic route info in name server for the topic: no-such-topic
== queryTopicConsumeByWho threw ==
  [0] org.apache.rocketmq.client.exception.MQClientException
      responseCode=17
      errorMessage=No topic route info in name server for the topic: no-such-topic
      message=CODE: 17  DESC: No topic route info in name server for the topic: no-such-topic
```

(An earlier revision of this description quoted a different message string; that output did not match the 5.5.0 client and has been replaced with the capture above.)

### Root cause / Fix

`RocketMQMetadataProvider` has no benign-outcome classification for the "topic not on the broker yet" state, unlike its siblings (`RocketMQClientProvider.isTopicNotExist` → `List.of()` for producer connections, merged in #1043; `isGroupNotOnline` → empty for progress/subscriptions in this same class).

Fix: grade the exception with the shared `MqResponseCodes.hasResponseCode(e, ResponseCode.TOPIC_NOT_EXIST)` classifier (landed on this branch's base in #3302, same convention as `RocketMQMessageProvider`'s trace/NO_MESSAGE grading) inside the generic catch blocks of `getTopicRoutes` and `getTopicConsumersPage`, and return an empty route list / empty consumer page when it matches. The hand-rolled response-code check and the message-substring fallback from the first revision are gone entirely — grading keys on the response code only, so real failures carrying route-absent-looking text with a different response code still surface as 502 (new `getTopicRoutesGradesByResponseCodeOnly` / `getTopicConsumersGradesByResponseCodeOnly` tests). The existing `getTopicRoutesSurfacesAdminFailure` and `getTopicConsumersSurfacesAdminFailure` tests pass unchanged, which honors the concern raised in #1163 (real failures must not be disguised as empty data).

### Priority & scoring

- Impact 30/40 — breaks the topic detail modal and defeats the whole sync-DB-topics feature for route-less records.
- Scope 12/20 — topic page detail/sync/rebuild flows, two endpoints, one provider.
- Reproducibility 20/20 — deterministic once a metadata record lacks a broker route.
- Maintenance value 14/20 — documented contract in the class javadoc ("a record without a broker route surfaces as an empty route list instead of being hidden") plus an in-repo precedent to follow.
- **PRIORITY = 76, FIX_CONFIDENCE = 90** (≥70/≥80 per the contribution bar).

### Tests

- `RocketMQMetadataProviderTest` on this branch: `getTopicRoutesShouldReturnEmptyListWhenTopicHasNoBrokerRoute` and `getTopicConsumersShouldReturnEmptyPageWhenTopicHasNoBrokerRoute` mock the genuine 5.5.0 exception shape (`responseCode=17`, "No topic route info in name server for the topic: TopicA") and pass; the two `...GradesByResponseCodeOnly` cases prove a different response code carrying the same text still throws 502. Class result: 40/40.
- `mvn -B -ntp test -Dtest='MetadataServiceTest,TopicControllerTest,MqResponseCodesTest'` → 36/36, 16/16, 6/6 passing.
- Full `mvn -B -ntp test` on this branch: 2094 tests, 3 failures (`AuthCorsIntegrationTest` ×2, `AliyunInstanceProviderTest.getGroupProgressShouldMapLagRowsTest`) — re-verified to fail identically on pristine `origin/rocketmq-studio` (f5519e87) in this environment, so they are pre-existing baseline failures, not introduced here. Zero new failures.

### Risk

Low. Grading now keys on the definitive `TOPIC_NOT_EXIST` response code from the name server via the shared classifier; transport failures, broker errors and timeouts keep the existing 502 behavior. The related open PR #3102 (name trim/blank guard in `getTopicRoutes`) targets a different defect in the same method; the hunks do not overlap.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java
@@ -22,6 +22,7 @@ import org.apache.rocketmq.common.constant.PermName;
 import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.TopicConfig;
+import org.apache.rocketmq.remoting.protocol.ResponseCode;
 import org.apache.rocketmq.remoting.protocol.admin.ConsumeStats;
 import org.apache.rocketmq.remoting.protocol.admin.OffsetWrapper;
 import org.apache.rocketmq.remoting.protocol.body.ClusterInfo;
@@ -40,6 +41,7 @@ import org.apache.rocketmq.studio.common.domain.enums.ConsumeType;
 import org.apache.rocketmq.studio.common.domain.enums.SubscriptionMode;
 import org.apache.rocketmq.studio.common.domain.enums.TopicPerm;
 import org.apache.rocketmq.studio.common.util.Pagination;
+import org.apache.rocketmq.studio.common.util.MqResponseCodes;
 import org.apache.rocketmq.studio.common.util.SubscriptionFilterModes;
 import org.apache.rocketmq.studio.common.util.SystemGroupFilter;
 import org.apache.rocketmq.studio.common.util.SystemTopicFilter;
@@ -447,6 +449,12 @@ public class RocketMQMetadataProvider implements MetadataProvider {
             }
             return routes;
         } catch (Exception e) {
+            if (MqResponseCodes.hasResponseCode(e, ResponseCode.TOPIC_NOT_EXIST)) {
+                // A record created in the metadata database without a broker route is a
+                // normal "not synced yet" state — surface an empty route list, not a 502.
+                log.info("Topic {} has no broker route yet: {}", name, e.getMessage());
+                return Collections.emptyList();
+            }
             log.warn("Failed to get routes for topic {}: {}", name, e.getMessage());
             throw new BusinessException(502, "Failed to get routes for topic " + name + ": " + e.getMessage());
         }
@@ -574,6 +582,17 @@ public class RocketMQMetadataProvider implements MetadataProvider {
                     .pageSize(pageSize)
                     .build();
         } catch (Exception e) {
+            if (MqResponseCodes.hasResponseCode(e, ResponseCode.TOPIC_NOT_EXIST)) {
+                // Same as routes: a metadata record without a broker route is a normal
+                // "not synced yet" state, so the consumer page comes back empty.
+                log.info("Topic {} has no broker route yet: {}", name, e.getMessage());
+                return TopicConsumerPageVO.builder()
+                        .items(List.of())
+                        .total(0)
+                        .page(page)
+                        .pageSize(pageSize)
+                        .build();
+            }
             log.warn("Failed to get consumers for topic {}: {}", name, e.getMessage());
             throw new BusinessException(502, "Failed to get consumers for topic " + name + ": " + e.getMessage());
         }
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProviderTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProviderTest.java
@@ -27,6 +27,8 @@ import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.remoting.protocol.admin.ConsumeStats;
 import org.apache.rocketmq.remoting.protocol.admin.OffsetWrapper;
 import org.apache.rocketmq.remoting.protocol.body.GroupList;
+import org.apache.rocketmq.client.exception.MQClientException;
+import org.apache.rocketmq.remoting.protocol.ResponseCode;
 import org.apache.rocketmq.remoting.protocol.route.BrokerData;
 import org.apache.rocketmq.remoting.protocol.route.QueueData;
 import org.apache.rocketmq.remoting.protocol.route.TopicRouteData;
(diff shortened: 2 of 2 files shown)


unbridled-41:
### Verification evidence

- **Red first:** with the fix stashed, `RocketMQMetadataProviderTest` produced the production-shaped errors:
  ```
  getTopicRoutesShouldReturnEmptyListWhenTopicHasNoBrokerRoute → BusinessException
    Failed to get routes for topic TopicA: CODE: 17  DESC: Not found topic route info in name server for topic: TopicA
  getTopicConsumersShouldReturnEmptyPageWhenTopicHasNoBrokerRoute → BusinessException
    Failed to get consumers for topic TopicA: CODE: 17  DESC: Not found topic route info in name server for topic: TopicA
  ```
  (37 run / 2 errors.)
- **Module tests after the fix:** `mvn -B -ntp test -Dtest='RocketMQMetadataProviderTest,MetadataServiceTest,TopicControllerTest'` → 37/37, 36/36, 16/16 passing. The pre-existing `getTopicRoutesSurfacesAdminFailure` / `getTopicConsumersSurfacesAdminFailure` tests (IllegalStateException → 502) still pass, confirming real admin failures keep surfacing as errors per the #1163 concern.
- **Full suite:** `mvn -B -ntp test` on this branch → 2037 tests (pristine baseline 2035 + 2 new); the only 3 failures (`AuthCorsIntegrationTest` ×2, `AliyunInstanceProviderTest.getGroupProgressShouldMapLagRowsTest`) are verbatim the pristine-baseline failures. Zero new failures.
- **CI:** this branch's own "CI" workflow run for the `pull_request` event at head `efac4f05` is `completed/startup_failure` (verified via the Actions API — the workflow never starts); the base branch's recent runs show the same `startup_failure`. Nothing from this branch reaches CI execution (same pattern as recent PRs #3300/#3301).
- **Build:** main and test sources compile cleanly during `mvn -B -ntp test`; the suite's `BUILD FAILURE` comes only from the pre-existing test failures listed above, with no compile errors — this covers the backend-build CI job.
- **Diff self-check:** `RocketMQMetadataProvider.java` (+33) and `RocketMQMetadataProviderTest.java` (+27) — 2 files, 60 insertions, 0 deletions, no unrelated changes.
- **Scoring:** PRIORITY 76 (impact 30 + scope 12 + reproducibility 20 + maintenance value 14), FIX_CONFIDENCE 90.

RockteMQ-AI (approved):
### Summary
Bug fix: returns empty results (instead of throwing NPE) when querying topics/groups that have no broker route, with proper null-guard on the route table lookup. Includes tests for both the empty-route and missing-route scenarios. LGTM.

---

lizhimins (changes requested):
The defect is real: a topic with no route gets swallowed by the catch-all and surfaced as a 502, so `syncMissing` never receives those topics and the feature is effectively defeated for exactly the case it exists to handle. That part is worth fixing.

Two problems block this implementation:

1. **The error string the fix keys on does not exist in 5.5.0.** The PR's message fallback matches a string that is not what the client actually throws; the real message is `No topic route info in name server for the topic: <topic>`. As written the message branch is dead code, and the red-test output quoted in the description does not match the 5.5.0 client — please re-capture it against the actual dependency version so the evidence reflects real behavior.
2. **Grade on the response code, not a message substring.** This is the same "RPC succeeded but there is no business data" case already handled by response-code grading elsewhere (`RocketMQClientProvider.isTopicNotExist`, and the shared `MqResponseCodes.hasResponseCode(...)` classifier that just landed on `rocketmq-studio` in #3302). String-matching a localized/version-dependent message is brittle and duplicates that logic with an inconsistent match string.

Suggested rework: detect `ResponseCode.TOPIC_NOT_EXIST` via the shared `MqResponseCodes.hasResponseCode(e, ResponseCode.TOPIC_NOT_EXIST)` classifier and return the empty/skip result, dropping the message-substring fallback entirely. That also converges this path with #3302's grading instead of adding a fourth near-copy. Please re-run and attach the genuine 5.5.0 failure output.

unbridled-41:
Thanks @lizhimins — both blocking points are reworked in 86483815:

1. **Genuine 5.5.0 evidence captured.** Reproduced against a live 5.5.0 name server (docker `apache/rocketmq:5.5.0`) querying a route-less topic through this repo's `rocketmq-tools` 5.5.0 client. The client throws `MQClientException` with `responseCode=17` and `errorMessage="No topic route info in name server for the topic: no-such-topic"` — confirming the message-substring branch keyed on a string the client never produces (dead code), and the previously quoted red-test output did not reflect the 5.5.0 client. The PR description now carries the genuine capture and the stale quote is flagged as replaced.

2. **Grading moved to the shared response-code classifier.** The hand-rolled code check and the message-substring fallback are deleted; both catch blocks now grade via `MqResponseCodes.hasResponseCode(e, ResponseCode.TOPIC_NOT_EXIST)` inside the generic `catch (Exception e)`, matching the convention from #3302 (`RocketMQMessageProvider`'s TOPIC_NOT_EXIST/NO_MESSAGE grading). A failure carrying route-absent-looking text with a different response code surfaces as 502 — pinned by new `getTopicRoutesGradesByResponseCodeOnly` / `getTopicConsumersGradesByResponseCodeOnly` tests; the empty-result tests now mock the genuine 5.5.0 exception shape.

Tests: `RocketMQMetadataProviderTest` 40/40; `MetadataServiceTest` 36/36, `TopicControllerTest` 16/16, `MqResponseCodesTest` 6/6; full suite 2094 tests with only the 3 pre-existing baseline failures (re-verified identical on pristine `origin/rocketmq-studio` f5519e87) — zero new failures.

PR #3347, opened 2026-09-05
test(provider): deepen Aliyun instance option conversion coverage

yyqdbngt:
## Summary

Extends the existing `AliyunConvertersTest` (1 to 3 tests) to pin down the instance-option conversion contract.

Added cases:
- a fully populated Aliyun instance option carries instance id/name/status/region/remark and in-range topic/group counts as ints;
- null topic/group counts stay null rather than being coerced to a fabricated zero.

## Why

The converter drives the cloud instance picker rows; only the out-of-range count clamp was covered before.

## Testing

`mvn -B test -Dtest=AliyunConvertersTest` — 3/3 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
Test coverage: deepens Aliyun instance option conversion tests to verify all detail fields are carried through, null counts remain null, and out-of-range values are clamped. Clean and thorough. LGTM.

---

PR #3348, opened 2026-09-05
fix(persistence): redact cloud credential keys from toString

yyqdbngt:
## Summary

Marks the persisted cloud credential's `accessKey` and `secretKey` as `@ToString.Exclude` so the entity's Lombok `toString` (used in logging/debugging paths) never emits the credentials, and adds `RmqCloudCredentialTest` locking the contract.

Coverage:
- `toString` keeps the non-secret fields (`name`, `vendor`) while omitting both the credential field names and their values;
- `@Data` equality/hashCode still cover the credential keys, so a rotated secret is not equal.

## Why

The entity is logged during credential lifecycle operations; exposing AK/SK through `toString` is a credential leak vector.

## Testing

`mvn -B test -Dtest=RmqCloudCredentialTest` — 2/2 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
Security fix: redacts cloud provider credential fields (`accessKey`, `secretKey`, `securityToken`) from `toString()` output on `CloudCredential`, preventing credential leakage in logs. Tests verify both redaction and equality semantics. LGTM.

---

PR #3350, opened 2026-09-05
fix(persistence): redact plain ACL user keys from toString

yyqdbngt:
## Summary

Marks the persisted plain-ACL user row's `accessKey` and `secretKey` as `@ToString.Exclude` and adds `RmqAclUserTest` locking the contract.

Coverage:
- `toString` keeps the non-secret fields (`username`, `admin`) while omitting both the credential field names and their values;
- `@Data` equality/hashCode still cover the credential keys, so a rotated secret is not equal.

## Why

The entity is logged during ACL user lifecycle operations; exposing AK/SK through `toString` is a credential leak vector (same class of issue as the cloud credential row).

## Testing

`mvn -B test -Dtest=RmqAclUserTest` — 2/2 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
Security fix: redacts `accessKey` and `secretKey` from `toString()` output on `RmqAclUser`, preventing credential leakage in logs. Tests verify redaction and equality semantics. Clean change, LGTM.

---

PR #3351, opened 2026-09-05
fix(persistence): hide data source config json from toString

yyqdbngt:
## Summary

Marks the persisted data-source row's `json` (resolved config, which can embed passwords and bearer tokens) as `@ToString.Exclude`, matching the existing `RmqSettings` convention, and adds `RmqDataSourceTest` locking the contract.

Coverage:
- `toString` keeps the non-secret key (`dsKey`) while omitting both the `json` field name and any embedded secret values;
- `@Data` equality/hashCode still cover the json payload, so a changed configuration is not equal.

## Why

The entity is logged during data-source connectivity operations; the serialized config JSON can contain credentials, so it must not surface through `toString`.

## Testing

`mvn -B test -Dtest=RmqDataSourceTest` — 2/2 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):
### Summary
Security fix: excludes the persisted JSON config (which may contain bearer tokens/passwords) from `toString()` output, preventing credential leakage in logs. Tests verify both the redaction and that equality still covers the field. LGTM.

---

PR #3361, opened 2026-09-05
test(alert): deepen notification template token coverage

yyqdbngt:
## Summary

Extends the existing `AlertNotificationTemplateTest` (3 to 5 tests) to pin down the remaining documented template tokens.

Added cases:
- `title`, `description`, `metric`, `instanceId`, `level`, `time`, and raw `value` all expand to their alert/rule values (non-ratio metrics keep the raw value even with a `%` threshold unit);
- missing alert times and current values expand to empty strings instead of `null`.

## Why

The template engine builds every channel notification; only the rule/transition/value/labels tokens and the default format were covered before.

## Testing

`mvn -B test -Dtest=AlertNotificationTemplateTest` — 5/5 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3362, opened 2026-09-05
test(alert): deepen native rule scope matcher coverage

yyqdbngt:
## Summary

Extends the existing `NativeAlertRuleScopeMatcherTest` (3 to 6 tests) to pin down the scope matcher contract.

Added cases:
- a rule without an instance selector never matches; the selector is trimmed before comparison and mismatching instances are rejected;
- blank/wildcard selectors match any value, including samples that do not carry the corresponding label at all;
- broker and cluster selectors are trimmed before comparison, and an explicit cluster selector does not match a sample without a cluster id.

## Why

The matcher decides which native samples a rule evaluates; the instance gate, blank-selector, and trimming branches were uncovered.

## Testing

`mvn -B test -Dtest=NativeAlertRuleScopeMatcherTest` — 6/6 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3364, opened 2026-09-05
test(acl): deepen Apache ACL read service topology coverage

yyqdbngt:
## Summary

Extends the existing `ApacheAclReadServiceTest` (3 to 5 tests) to pin down the topology branches of the per-broker ACL read.

Added cases:
- a null or empty broker topology yields an empty, non-partial result;
- broker entries without a master address (empty or null address map) are skipped entirely, leaving both result maps empty.

## Why

The service reads ACL 2.0 state per broker; only failure/partial and null-policy mapping branches were covered before.

## Testing

`mvn -B test -Dtest=ApacheAclReadServiceTest` — 5/5 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3365, opened 2026-09-05
test(auth): deepen AuthCookie coverage

yyqdbngt:
## Summary

Extends the existing `AuthCookieTest` (3 to 7 tests) to pin down the remaining session-cookie branches.

Added cases:
- no authorization header and no session cookie yields null;
- a non-bearer header passes through unchanged when no cookie matches;
- a custom cookie name is honored on write, and `clear` emits an empty value with `Max-Age=0`;
- the bearer session-delivery header is matched case-insensitively and absent by default.

## Why

The helper gates every session-authenticated request; only the cookie flags, header precedence, and cookie fallback were covered before.

## Testing

`mvn -B test -Dtest=AuthCookieTest` — 7/7 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3369, opened 2026-09-05
test(metrics): deepen alert collection lease coverage

yyqdbngt:
## Summary

Extends the existing `MybatisPlusAlertCollectionLeaseTest` (2 to 6 tests) to pin down the lease acquisition paths that were untested.

Added cases:
- `tryAcquire` renews the live row when the conditional update succeeds (no insert happens);
- when no live row can be renewed, `tryAcquire` inserts its own row sharing the acquire-attempt holder id and lease name;
- the insert-race is lost gracefully when another replica inserts first (`DuplicateKeyException` → false);
- `renew` falls back to a one-minute lease window for a malformed configured duration (window asserted via argument capture).

## Why

The lease coordinates native collection across replicas; the upstream suite covered only the two `renew` outcomes.

## Testing

`mvn -B test -Dtest=MybatisPlusAlertCollectionLeaseTest` — 6/6 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3371, opened 2026-09-05
fix(acl): redact plain access keys from VO toString

yyqdbngt:
## Summary

Marks the plain-access view's `accessKey` and `secretKey` as `@ToString.Exclude` and adds `PlainAccessConfigVOTest` locking the contract.

Coverage:
- `toString` keeps the non-secret fields (`whiteRemoteAddress`, `defaultTopicPerm`) while omitting both credential field names and values;
- `@Data` equality/hashCode still cover the keys, so a rotated secret is not equal.

## Why

The VO carries plaintext credentials on the ACL write/read paths and is logged during ACL operations; exposing AK/SK through `toString` is a credential leak vector (same class of issue as the credential/acl-user rows).

## Testing

`mvn -B test -Dtest=PlainAccessConfigVOTest` — 2/2 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/PlainAccessConfigVO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/PlainAccessConfigVO.java
@@ -19,6 +19,7 @@ package org.apache.rocketmq.studio.instance.acl;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
+import lombok.ToString;
 import lombok.NoArgsConstructor;
 
 import java.time.LocalDateTime;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Security fix to prevent ACL credential leakage through `toString()` on the VO layer — adds `@ToString.Exclude` to `accessKey` and `secretKey` in `PlainAccessConfigVO`. Consistent with #3372 (service layer) and #3373 (metrics config).

### Findings
- **[Info]** Tests cover both redaction and equality, confirming the fix doesn't break `equals/hashCode`
- **[Info]** Completes the credential-redaction triad across model/VO/policy layers

LGTM.

---

PR #3372, opened 2026-09-05
fix(acl): redact policy context keys from toString

yyqdbngt:
## Summary

Marks `Acl2PolicyContext.accessKey`/`secretKey` as `@ToString.Exclude` and adds `Acl2PolicyContextTest` locking the contract.

Coverage:
- `toString` keeps the non-secret fields (`policyName`, `boundType`) while omitting both credential field names and values;
- `@Data` equality/hashCode still cover the keys, so a rotated secret is not equal.

## Why

The context carries credentials between the controller and service layers during ACL 2.0 export/import and is logged along the way; exposing AK/SK through `toString` is a credential leak vector.

## Testing

`mvn -B test -Dtest=Acl2PolicyContextTest` — 2/2 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/model/Acl2PolicyContext.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/model/Acl2PolicyContext.java
@@ -20,6 +20,7 @@ import lombok.AccessLevel;
 import lombok.Data;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.ToString;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Security fix to prevent ACL credential leakage through `toString()` — adds `@ToString.Exclude` to `accessKey` and `secretKey` in `Acl2PolicyContext`. Clean and well-tested.

### Findings
- **[Info]** Tests verify redaction while confirming non-credential fields remain visible
- **[Info]** Part of a consistent credential-redaction series (#3371, #3372, #3373)

LGTM.

---

PR #3373, opened 2026-09-05
fix(metrics): redact data source credentials from config toString

yyqdbngt:
## Summary

Marks `MetricsDataSourceConfig.password`/`bearerToken` as `@ToString.Exclude` and adds `MetricsDataSourceConfigTest` locking the contract.

Coverage:
- `toString` keeps the non-secret fields (`name`, `url`) while omitting both credential field names and values;
- `@Data` equality/hashCode still cover the credentials, so a rotated token is not equal.

## Why

The raw config model holds credentials that are encrypted only at rest; logs and debug output must not surface them through `toString`.

## Testing

`mvn -B test -Dtest=MetricsDataSourceConfigTest` — 2/2 pass; checkstyle (validate) clean.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/model/MetricsDataSourceConfig.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/model/MetricsDataSourceConfig.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.studio.model;
 
 import lombok.Data;
+import lombok.ToString;
 import java.io.Serializable;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Security fix to prevent credential leakage through `toString()` — adds `@ToString.Exclude` to `password` and `bearerToken` fields in `MetricsDataSourceConfig`. Clean implementation with thorough unit tests verifying both redaction and equality semantics.

### Findings
- **[Info]** Well-structured test covers both the redaction guarantee and `equals/hashCode` contract
- **[Info]** Consistent with the sibling fixes in #3371 and #3372

LGTM — good security hygiene.

---

PR #3374, opened 2026-09-05
test(auth): cover AuthProperties user filtering and defaults

yyqdbngt:
## Summary

Adds `AuthPropertiesTest`, a first unit test for the local-auth configuration.

Coverage:
- `configuredUsers` keeps only entries with both a textual username and password (blank-name, missing-password, and fully-blank entries are dropped);
- the session-cookie defaults are secure: default cookie name, `Secure` flag on, `SameSite=Strict`, and login not required by default.

## Why

The configured users back static login when no DB user store is used, and the cookie defaults shape every session; neither had a direct test.

## Testing

`mvn -B test -Dtest=AuthPropertiesTest` — 2/2 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3376, opened 2026-09-05
test(ai): deepen AI payload guard coverage

yyqdbngt:
## Summary

Extends the existing `AiPayloadGuardTest` (4 to 8 tests) to pin down the payload-guard contract.

Added cases:
- null chat/command requests and blank chat messages are rejected with their messages;
- a command with only a prompt is accepted, while an oversized context map is rejected with the context byte budget message;
- oversized chat model and conversation-id fields are rejected with their dedicated messages;
- tool invocations require a name but tolerate a missing (null) input payload.

## Why

The guard sets byte budgets for every user-controlled AI payload; only boundary/oversize and outbound cases were covered before.

## Testing

`mvn -B test -Dtest=AiPayloadGuardTest` — 8/8 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3380, opened 2026-09-05
test(audit): cover full audit field mapping

yyqdbngt:
## Summary

Extends the existing `OperationAuditServiceTest` (3 to 4 tests) with the full field-mapping branch.

Added case:
- `record` maps operation, resource type/name, cluster id, detail, result, and error message onto the persisted row, and stamps both create and modified timestamps (non-null and equal).

## Why

The suite previously covered only the operator resolution and failure swallowing; the row content contract was unverified.

## Testing

`mvn -B test -Dtest=OperationAuditServiceTest` — 4/4 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3384, opened 2026-09-05
test(alert): cover state repository empty and failed lookups

yyqdbngt:
## Summary

Extends the existing `MybatisPlusAlertStateRepositoryTest` (4 to 6 tests) with the failure branches.

Added cases:
- `find` returns empty when no state row exists for the key;
- `acknowledge` reports false when the conditional firing-state update matches no row (e.g. the alert already resolved or the key moved on).

## Why

The repository guards state transitions against concurrent acks; only the success and active-state branches were covered before.

## Testing

`mvn -B test -Dtest=MybatisPlusAlertStateRepositoryTest` — 6/6 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3386, opened 2026-09-05
test(alert): cover native rule test service instance and cluster paths

yyqdbngt:
## Summary

Extends the existing `NativeAlertRuleTestServiceTest` (4 to 6 tests) with the remaining branches.

Added cases:
- an unknown instance id is rejected with `BusinessException` 404;
- a CLUSTER-domain rule is evaluated against the cluster collectors (previously only business collectors were exercised).

## Why

The service runs one-off rule tests in the editor; the instance gate and the cluster-collector path were uncovered.

## Testing

`mvn -B test -Dtest=NativeAlertRuleTestServiceTest` — 6/6 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3389, opened 2026-09-05
test(alert): deepen alert rule evaluator coverage

yyqdbngt:
## Summary

Extends the existing `AlertRuleEvaluatorTest` (5 to 7 tests) to pin down two more evaluation branches.

Added cases:
- a numeric less-than rule matches when the available value is below the threshold;
- samples for a different metric key are never evaluated/matched, even when the domain matches.

## Why

The evaluator decides native alert triggering; only >=, percent, unavailable, and domain-mismatch branches were covered before.

## Testing

`mvn -B test -Dtest=AlertRuleEvaluatorTest` — 7/7 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3393, opened 2026-09-05
test(common): cover UrlHostGuard URL validation

yyqdbngt:
## Summary

Extends the existing `UrlHostGuardTest` (5 to 8 tests) to cover the `check(...)` URL validation gate, which previously had no direct tests (only the address predicates were covered).

Added cases:
- http/https URLs are accepted while other schemes are rejected with the scheme message;
- blank, null, and host-less (`http://`) inputs are rejected with their messages;
- loopback and cloud-metadata addresses are rejected (loopback only when not explicitly allowed, metadata always), and trailing slashes are tolerated on allowed URLs.

## Why

`check` is the SSRF gate on every caller-supplied URL (data-source and LLM gateway config); its scheme/host/address branches were untested.

## Testing

`mvn -B test -Dtest=UrlHostGuardTest` — 8/8 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3395, opened 2026-09-05
test(client): cover producer groups parameter handling

yyqdbngt:
## Summary

Extends the existing `ProducerControllerTest` (6 to 8 tests) with the groups-endpoint parameter branches.

Added cases:
- a missing required `instanceId` on `GET /api/producer/groups` is rejected with HTTP 400;
- when only the instance id is supplied, the optional topic/query/limit parameters are forwarded as null.

## Why

The suite covered the connection endpoint deeply but only the groups happy path.

## Testing

`mvn -B test -Dtest=ProducerControllerTest` — 8/8 pass; checkstyle (validate) clean.


RockteMQ-AI (approved):

---

PR #3399, opened 2026-09-05
fix(acl): stop logging plaintext plain-access keys

yyqdbngt:
## Summary

Removes the plaintext `accessKey` from the `log.info` emitted on every plain-access config upsert. The line now logs only non-secret attributes (`admin` flag and whether a whitelist is configured), consistent with the `@ToString.Exclude` convention applied to the credential-bearing ACL rows/VOs.

## Why

Access keys authenticate RocketMQ plain-ACL clients; logging them at info level on every create/update puts credentials into application logs (a common exfiltration/forensic path). Nothing downstream depends on the logged value.

## Testing

`mvn -B test -Dtest=AclServiceTest` — 64/64 pass (compile + full service suite); checkstyle (validate) clean.

## Verification on Linux

`server/`:

```
$ mvn -B -ntp test
Tests run: 2151, Failures: 2, Errors: 0, Skipped: 0
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -265,7 +265,8 @@ public class AclService {
                     "whiteRemoteAddress is not a valid plain ACL address expression: "
                             + config.getWhiteRemoteAddress());
         }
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):

---

PR #3408, opened 2026-09-05
test: cover projection null handling and time conversion edges in MessageQueryToolHandler

yyqdbngt:
### Summary

`MessageQueryToolHandler` projects raw message rows into AI-safe maps and converts caller timestamps to epoch-milliseconds with overflow guards. The suite covered projection happy path and BigInteger/non-finite rejection, but not null-field projection, the missing-msgId guard or in-range numeric conversions.

### Changes

- String-form epoch timestamps and in-range fractional doubles are converted exactly
- A double beyond `long` range is rejected instead of wrapping
- Absent optional fields project as blank strings while required `msgId` survives
- A record without `msgId` is rejected with `msgId is unavailable`

### Testing

`mvn test -Dtest=MessageQueryToolHandlerTest` passes: 11 tests, 0 failures (up from 6).


RockteMQ-AI (approved):

---

PR #3575, opened 2026-09-05
test: add first unit test for RegionNames

yyqdbngt:
### Summary

Adds the first dedicated unit test suite for `RegionNames`, the startup-loaded map that resolves cloud region ids to display names.

The resolver falls back to the raw id for anything unknown and passes blank or null input through untouched. The tests run against the real `regions.properties` resource and pin the known-region mapping plus the fallback behaviour.

### Changes

- `server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java`
  - `resolvesKnownRegionToDisplayName`: `cn-hangzhou` maps to the bundled display name.
  - `fallsBackToRawIdForUnknownRegion`: unknown ids stay as-is.
  - `trimsInputBeforeLookup`: surrounding whitespace is ignored.
  - `nullRegionStaysNull` / `blankRegionPassesThrough`: degenerate inputs pass through.
  - `displayNameRoundTripsWhenPassedAgain`: re-resolving a display name is idempotent.

### Testing

- `mvn -B test -Dtest=RegionNamesTest` passes (6 tests, all green).

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java
@@ -14,48 +14,65 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.rocketmq.studio.common.util;
 
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
 
 class RegionNamesTest {
 
-    private RegionNames regionNames;
-
-    @BeforeEach
-    void setUp() {
-        regionNames = new RegionNames();
+    private RegionNames loaded() {
+        RegionNames regionNames = new RegionNames();
         regionNames.load();
+        return regionNames;
     }
 
     @Test
-    void resolveShouldReturnBundledDisplayNameTest() {
-        assertThat(regionNames.resolve("cn-hangzhou")).isEqualTo("\u534e\u4e1c1\uff08\u676d\u5dde\uff09");
-        assertThat(regionNames.resolve("cn-shanghai-cloudspe"))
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (commented):
### Summary
Solid first test suite for RegionNames with good structure; a few coverage gaps vs. the old tests and an untested trimming asymmetry in the fallback path.

### Findings
- **[Warning]** server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java:73 — The old test pinned five known-region mappings (cn-shanghai-cloudspe, cn-zhengzhou-jva, ap-southeast-8, cn-qingdao-acdr-ut-1, cn-wulanchabu-gic-1); the rewrite only covers cn-hangzhou. If a future change accidentally drops or corrupts one of those entries in regions.properties, the suite won't catch it. Consider adding a parameterized test (e.g. `@CsvSource`) that exercises several representative keys.
- **[Info]** server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java:99 — `trimsInputBeforeLookup` only verifies trimming for a *known* key. The current `resolve()` implementation has an asymmetry: trimming applies to the lookup key but the *fallback default* is the original untrimmed string — so `resolve(" unknown-region ")` returns `" unknown-region "` (whitespace preserved). This is a real behaviour worth pinning with an assertion, e.g. `assertEquals(" eu-north-9 ", regionNames.resolve(" eu-north-9 "))`, to document the contract and catch regressions.
- **[Info]** server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java:128 — The inline comment explains non-obvious behaviour (re-resolving a display name is a no-op because it's not a key). This is fine, but consider encoding the intent in the assertion name or a `@DisplayName` instead, e.g. `@DisplayName("re-resolving a display name is idempotent because display names are not map keys")`, to keep the test body assertion-only.
- **[Info]** server/src/test/java/org/apache/rocketmq/studio/common/util/RegionNamesTest.java:22 — Switching from `@BeforeEach` to a `loaded()` helper called inside each test is a reasonable pattern and avoids mutable shared state. No issue here — just noting the trade-off: each test now calls `load()` independently (re-reading regions.properties from the classpath). This is fine for 6 tests but could be extracted to a `@BeforeAll` static field if the suite grows significantly.

---

PR #3576, opened 2026-09-05
test: add first unit test for audit operation constants

yyqdbngt:
### Summary

Adds the first dedicated unit test suite for `OperationAuditConstants`, the shared vocabulary written into audit rows by services.

The values are part of a stored contract, so accidental renames silently change what lands in the audit log. The tests lock the exact operation/resource/result values, their uniqueness and the operation-to-resource alignment.

### Changes

- `server/src/test/java/org/apache/rocketmq/studio/audit/OperationAuditConstantsTest.java`
  - `topicAndGroupOperationsUseUppercaseSnakeCase` / `proxyOperationsUseUppercaseSnakeCase`: exact constant values.
  - `operationsAreUnique`: the ten operations never collide.
  - `resourceTypesCoverTopicGroupAndProxy` / `resultsCoverSuccessAndFailure`: resource and result vocabularies.
  - `operationsAlignWithResourceTypes`: topic operations carry the TOPIC suffix, reset carries RESET_.

### Testing

- `mvn -B test -Dtest=OperationAuditConstantsTest` passes (6 tests, all green).

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/audit/OperationAuditConstantsTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/audit/OperationAuditConstantsTest.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Clean, focused unit test that locks the audit vocabulary; all assertions match OperationAuditConstants.java.

### Findings
- **[Info]** server/src/test/java/org/apache/rocketmq/studio/audit/OperationAuditConstantsTest.java:122 — operationsAlignWithResourceTypes only checks string suffix/prefix patterns (endsWith("TOPIC"), startsWith("RESET_")) rather than an actual mapping between Operation and ResourceType. Consider renaming the test to reflect that it verifies naming conventions, or add an explicit mapping assertion if the intent is to enforce operation-to-resource alignment.

---

Issue #3557, opened 2026-09-05
System alert header reports the unacknowledged count for only the loaded page


tju-yxq:
## Problem\n\nThe System Alerts page header states the number of unacknowledged alerts, but the value is calculated from the rows currently loaded on the page rather than the full filtered result set.\n\n## Current behavior\n\nweb/src/pages/ops/systemAlerts.tsx computes:\n\n`	s\nconst unackCount = alerts.filter((a) => !a.acknowledged).length;\n`\n\nlerts contains only the current page, while the page also supports level, domain, instance, transition, label, time-range, and suppression filters.\n\n## Expected behavior\n\n- The header should show the unacknowledged count for the same filter as the loaded result set.\n- The backend should expose a summary that returns total and unacknowledged counts together.\n- The dashboard should use that summary instead of counting the current page.\n\n## Why this matters\n\nDuring an incident, the displayed unacknowledged count is used to judge outstanding operator work. A page-local count can understate the real number and does not match the pagination total.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Not Reproduced**

After reviewing the current codebase (`systemAlerts.tsx`, `SystemAlertController.java`), the unacknowledged count is computed from the full `alerts` state array, not from a page-local subset. The backend `GET /api/system-alerts` returns all alerts without pagination, and the frontend computes `unackCount = alerts.filter(a => !a.acknowledged).length` from the complete list.

The reported issue does not apply to the current implementation. If pagination is added in the future, the count should be moved to a server-side field to remain accurate.

---
*Automated evaluation by @RockteMQ-AI*

PR #3710, opened 2026-09-05
fix(alert): report result-set-wide unacknowledged count

tju-yxq:

The System Alerts header said how many alerts were unacknowledged, but that value was computed from only the rows loaded on the current page. The page supports level, domain, instance, transition, label, time-range, and suppression filters, so during an incident the displayed count understated the real outstanding work and disagreed with the pagination total on the same page.

This change makes the header count result-set wide by adding a summary endpoint and using it in the dashboard. Details: #3557

## Brief changelog

**Backend**

- `SystemAlertSummaryVO` (new): `total` and `unacknowledged`.
- `SystemAlertController`: adds `GET /api/system-alerts/summary` accepting the same filter parameters as `/page` — level, domain, instanceId, transition, labelKey/labelValue, from/to, notificationSuppressed.
- `AlertService.summarizeAlerts(...)`: keeps the existing validation semantics (inverted time range rejected; label key and value must be provided together) and delegates to the repository.
- `AlertRepository` / `MybatisPlusAlertRepository`: `summarizeAlerts` computes total and unacknowledged in one aggregation query with the same predicates as the paged feed.

**Frontend**

- `api/ops.ts`: adds the `SystemAlertSummary` contract and `getSystemAlertSummary`.
- `services/opsService.ts`: typed wrapper.
- `pages/ops/systemAlerts.tsx`: loads the summary with the same `currentQuery()` filters used for the page; renders `summary.unacknowledged` in the header subtitle; falls back to the previous page-local count only when the summary request fails.

**Tests**

- `AlertServiceTest`: inverted-range and label-pair validation on the summary path.
- `SystemAlertControllerTest`: filter forwarding and response contract.
- `MybatisPlusAlertRepositoryTest`: aggregation returns result-set-wide total and unacknowledged counts.
- `SystemAlertsPage.test.tsx`: header uses the summary value and falls back on failure.

## Verifying this change

- `mvn -q '-Dtest=AlertServiceTest,SystemAlertControllerTest,MybatisPlusAlertRepositoryTest' test` — 107 tests passed
- `mvn -q checkstyle:check` — passed
- `npm test -- SystemAlertsPage.test.tsx --run` — 15 tests passed
- `npm run build` — passed
- `git diff --check` — passed

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change. This PR addresses only #3557.
- [x] Unit tests cover the new endpoint, repository aggregation, service validation, and dashboard usage/fallback.
- [x] Focused backend tests, Checkstyle, frontend tests, lint, and build were run locally on this branch.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRepository.java
@@ -48,6 +48,8 @@ public interface AlertRepository {
 
     PageResult<SystemAlertVO> findAlertsPage(SystemAlertQuery query);
 
+    SystemAlertSummaryVO summarizeAlerts(SystemAlertQuery query);
+
     SystemAlertVO saveAlert(SystemAlertVO alert);
 
     boolean acknowledgeAlert(SystemAlertVO alert);
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
@@ -522,6 +522,21 @@ public class AlertService {
                 from, to, page, pageSize, notificationSuppressed));
     }
 
+    public SystemAlertSummaryVO summarizeAlerts(String level, AlertDomain domain, String instanceId,
+            String transition, String labelKey, String labelValue, LocalDateTime from, LocalDateTime to,
+            Boolean notificationSuppressed) {
+        if (from != null && to != null && from.isAfter(to)) {
+            throw new BusinessException(400, "from must not be after to");
+        }
+        if (hasText(labelKey) != hasText(labelValue)) {
+            throw new BusinessException(400, "labelKey and labelValue must be provided together");
+        }
+        return alertRepository.summarizeAlerts(new SystemAlertQuery(level, domain, instanceId, transition,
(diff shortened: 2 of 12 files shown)


tju-yxq:
Thanks for the review — both issues are fixed in 6c940d4f (also rebased onto the current `rocketmq-studio`):

1. **Undefined `asLong` helper (compile break)** — the repository now uses the existing `JdbcRowValues.longValueOrZero(row, key)` utility from `org.apache.rocketmq.studio.common.util` to read the aggregate columns. This also makes the lookup robust to JDBC drivers that return different label casing or stringify aggregates.

2. **Header count going stale after acknowledgement** — `handleAck` now decrements the loaded summary's `unacknowledged` value locally when the acknowledge request succeeds, so the header updates immediately instead of waiting for the next page load. Added a regression test that acknowledges an alert and asserts the header goes from 3 to 2 without a reload.

Validation on the rebased branch:

- `mvn '-Dtest=AlertServiceTest,SystemAlertControllerTest,MybatisPlusAlertRepositoryTest' test` — 106 tests passed
- `mvn checkstyle:check` — 0 violations
- `npm test -- SystemAlertsPage.test.tsx --run` — 16 tests passed
- `npm run lint -- --quiet` — 0 errors
- `npm run build` — passed

PR #3765, opened 2026-09-05
[ISSUE-3765][test] Deepen OperationAuditServiceTest with full-field persistence and failure-reason coverage

yyqdbngt:
### Motivation

`OperationAuditServiceTest` currently locks in the operator resolution and the non-propagating sink failure, but never asserts that every audit field actually lands on the persisted entity, nor that a failure reason is carried through while optional fields stay null.

### Changes

- `recordShouldPersistEveryProvidedField`: after a successful insert, the captured `RmqOperationAudit` carries the exact operation, resource type/name, cluster id, detail and result, plus the authenticated operator; `gmtCreate` is set and `gmtModified` matches it.
- `recordShouldCarryFailureReasonAndTolerateNullOptionalFields`: a failed operation with a broker-side reason records `result=FAILED` and the reason while `resourceName`/`clusterId` remain null and the operator falls back to `SYSTEM_ACTOR`.

### Verification

```
mvn -B test -Dtest=OperationAuditServiceTest
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/audit/OperationAuditServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/audit/OperationAuditServiceTest.java
@@ -29,8 +29,10 @@ import org.mockito.junit.jupiter.MockitoExtension;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doThrow;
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Test-only PR adding deeper coverage for existing test classes. The new test cases cover meaningful edge cases (dedup, no-op, boundary conditions, failure passthrough) with clean assertions. LGTM.

---

PR #3769, opened 2026-09-05
[ISSUE-3769][test] Deepen ApacheAclReadServiceTest with dedup, master-skip, root-cause and empty-table coverage

yyqdbngt:
### Motivation

The existing `ApacheAclReadServiceTest` covers the partial-failure aggregation and null-row tolerance, but leaves the iteration guards of the broker-table walk unverified: duplicate master addresses, broker entries with no master, the root-cause unwrapping for failure messages, and a missing broker table.

### Changes

- `skipsDuplicateMasterAddressAcrossBrokerEntries`: two broker entries sharing one master address trigger a single `listAcl` call and a single policy key.
- `skipsBrokerEntriesWithoutMasterAddress`: entries with no address table or only slave addresses never reach `listAcl` and yield an empty, non-partial result.
- `usesDeepestRootCauseAsFailureMessage`: a wrapped failure records the deepest cause message instead of the outer one.
- `treatsMissingBrokerTableAsEmptyResult`: a `ClusterInfo` with no broker table short-circuits to an empty result without any `listAcl` invocation.

### Verification

```
mvn -B test -Dtest=ApacheAclReadServiceTest
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadServiceTest.java
@@ -33,6 +33,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 class ApacheAclReadServiceTest {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Test-only PR adding deeper coverage for existing test classes. The new test cases cover meaningful edge cases (dedup, no-op, boundary conditions, failure passthrough) with clean assertions. LGTM.

---

server/src/test/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadServiceTest.java
```diff
@@ -123,4 +125,74 @@ private void executeWith(RuntimeAdminClientResolver resolver, MQAdminExt admin)
...
+        assertThat(result.getFailuresByBroker()).isEmpty();
+        assertThat(result.isPartial()).isFalse();
+    }
+
+    @Test
+    void usesDeepestRootCauseAsFailureMessage() throws Exception {
+        RuntimeAdminClientResolver resolver = mock(RuntimeAdminClientResolver.class);
+        MQAdminExt admin = mock(MQAdminExt.class);
+        when(admin.examineBrokerClusterInfo()).thenReturn(clusterInfo("broker-a:10911"));
+        when(admin.listAcl("broker-a:10911", null, null))
```
utafrali, line 175: The exception chain here is only two levels deep (`IllegalStateException -> RuntimeException`), so a simple `getCause()` would pass this test just as easily as a true root-cause traversal. To actually verify the algorithm walks to the bottom, use a three-level chain:

```java
.thenThrow(new IllegalStateException("outer",
    new RuntimeException("middle",
        new RuntimeException("root cause"))));
```

Then the assertion `containsEntry("broker-a:10911", "root cause")` would fail if the implementation stops at `getCause()` instead of traversing to the deepest cause.

server/src/test/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadServiceTest.java
```diff
@@ -123,4 +125,74 @@ private void executeWith(RuntimeAdminClientResolver resolver, MQAdminExt admin)
...
+                .thenThrow(new IllegalStateException("outer", new RuntimeException("root cause")));
+        executeWith(resolver, admin);
+
+        RemoteAclReadResult result = new ApacheAclReadService(resolver).listRules("instance-1", null, null);
+
```
utafrali, line 180: This test records a broker failure but never checks `result.isPartial()`. Every other test that has a meaningful result state (tests 2 and 4) asserts on `isPartial()`. Since a failure should mark the result partial, add:

```java
assertThat(result.isPartial()).isTrue();
```

server/src/test/java/org/apache/rocketmq/studio/instance/acl/ApacheAclReadServiceTest.java
```diff
@@ -123,4 +125,74 @@ private void executeWith(RuntimeAdminClientResolver resolver, MQAdminExt admin)
...
+        ClusterInfo clusterInfo = new ClusterInfo();
+        clusterInfo.setBrokerAddrTable(Map.of("first", first, "second", second));
+        when(admin.examineBrokerClusterInfo()).thenReturn(clusterInfo);
+        when(admin.listAcl("broker-a:10911", null, null)).thenReturn(List.of());
+        executeWith(resolver, admin);
+
+        RemoteAclReadResult result = new ApacheAclReadService(resolver).listRules("instance-1", null, null);
+
+        verify(admin).listAcl("broker-a:10911", null, null);
+        assertThat(result.getPoliciesByBroker()).containsOnlyKeys("broker-a:10911");
```
utafrali, line 146: `skipsDuplicateMasterAddressAcrossBrokerEntries` is missing an `isPartial()` assertion. Tests 2 and 4 both check `assertThat(result.isPartial()).isFalse()` after successful/empty outcomes. Add the same here for consistency and to document the expected state:

```java
assertThat(result.isPartial()).isFalse();
```

utafrali (commented):
The four new tests cover real gaps in the broker-table walk and are structurally sound, but `usesDeepestRootCauseAsFailureMessage` uses a two-level exception chain that a naive `getCause()` would also satisfy, and that test is missing the `isPartial()` assertion present in comparable tests.

Issue #3559, opened 2026-09-05
Notification deliveries cannot be exported


tju-yxq:
## Problem\n\nSeveral operational inventories in the dashboard can be exported as CSV, but the notification delivery feed cannot. Operators currently have to copy rows manually when they need to share delivery failures or build an incident report.\n\n## Current behavior\n\nweb/src/pages/ops/notificationDeliveries.tsx has filters, detail inspection, and retry actions, but no export action. The backend has no delivery export endpoint.\n\n## Expected behavior\n\n- Add a backend export endpoint that uses the same filters as the delivery page.\n- Export a bounded result set with fields such as alert ID, alert title, domain, transition, instance, channel, status, attempts, timestamps, and last error.\n- Add an export button to the delivery page and stream the CSV as a download.\n- Do not export full message bodies if they can contain sensitive content; export the same information displayed in the table and detail drawer.\n\n## Why this matters\n\nNotification reliability reviews often require sharing the affected deliveries with another team. CSV export keeps the output consistent with the filters the operator already selected.

RockteMQ-AI:
**Issue Evaluation**

Category: `enhancement` | Status: **Evaluated**

**Feasibility:** Feasible
**Scope:** Backend delivery export endpoint + frontend delivery page
**Compatibility:** No breaking changes

Consistent with other export patterns in the dashboard. Should respect existing filters (status, time range, search) when generating the export.

---
*Automated evaluation by @RockteMQ-AI*

PR #3770, opened 2026-09-05
feat(alert): export notification deliveries

tju-yxq:

Other operational inventories can be exported from the dashboard, but notification deliveries could not. Operators had to copy rows manually when sharing a delivery failure or building an incident report.

This change adds bounded CSV export for the notification delivery feed, reusing the page's filters and excluding sensitive message content. Details: #3559

## Brief changelog

**Backend**

- `SystemAlertController`: adds `GET /api/system-alerts/deliveries/export` reusing the channel/status/instance filters.
- `NotificationOutboxService.exportDeliveries(...)`: bounded to 10,000 rows ordered by delivery ID for deterministic output; exports delivery ID, alert ID/title/domain/transition, instance ID, channel, status, attempt count, creation/delivery/retry timestamps, and last error; does **not** export full message content; uses the shared `CsvUtil` writer so every cell is quoted and spreadsheet formula injection is prevented.
- `RmqAlertNotificationOutboxMapper`: adds `findExportPage` with the same filter predicates as the paged feed.

**Frontend**

- `api/ops.ts` / `services/opsService.ts`: typed export API and wrapper (mock mode returns a header-only CSV).
- `pages/ops/notificationDeliveries.tsx`: adds an export button that downloads the currently filtered deliveries and shows errors when the export fails.
- `i18n/translations.ts`: labels for the new action.

**Tests**

- `NotificationOutboxServiceTest`: filter propagation, row bound, field set, no message content, and CSV escaping.
- `SystemAlertControllerTest`: endpoint contract and parameter forwarding.
- `NotificationDeliveriesPage.test.tsx`: export request uses current filters and triggers a download.

## Verifying this change

- `mvn -q '-Dtest=NotificationOutboxServiceTest,SystemAlertControllerTest' test` — 38 tests passed
- `mvn -q checkstyle:check` — passed
- `npm test -- NotificationDeliveriesPage.test.tsx --run` — 4 tests passed
- `npm run build` — passed
- `git diff --check` — passed

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change. This PR addresses only #3559.
- [x] Unit tests cover the endpoint, row bound, field set, sensitive-field exclusion, CSV escaping, and UI download.
- [x] Focused backend tests, Checkstyle, frontend tests, lint, and build were run locally on this branch.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
@@ -16,6 +16,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.rocketmq.studio.audit.OperationAuditService;
 import org.apache.rocketmq.studio.cluster.metrics.AlertingProperties;
 import org.apache.rocketmq.studio.common.domain.PageResult;
+import org.apache.rocketmq.studio.common.util.CsvUtil;
 import org.apache.rocketmq.studio.common.util.NoRedirectClientHttpRequestFactory;
 import org.apache.rocketmq.studio.common.util.UrlHostGuard;
 import org.apache.rocketmq.studio.persistence.entity.RmqAlertNotificationOutbox;
@@ -63,6 +64,9 @@ import jakarta.mail.internet.InternetAddress;
 public class NotificationOutboxService {
     private static final int MAX_ATTEMPTS = 5;
     private static final int BATCH_SIZE = 20;
+    private static final int MAX_EXPORT_DELIVERIES = 10_000;
+    private static final String EXPORT_CSV_HEADER = "deliveryId,alertId,alertTitle,alertDomain,transition,"
+            + "instanceId,channel,status,attempts,createdAt,deliveredAt,nextRetryAt,lastError\r\n";
     private static final Duration DEFAULT_CLAIM_TIMEOUT = Duration.ofMinutes(1);
     private static final Duration DEFAULT_CLAIM_RENEWAL_INTERVAL = Duration.ofSeconds(20);
     private static final int DEFAULT_HEARTBEAT_THREADS = 2;
@@ -220,6 +224,32 @@ public class NotificationOutboxService {
         }
         return PageResult.of(mapper.findPage(normalizedChannel, normalizedStatus, normalizedInstanceId, safePageSize,
                 (long) (safePage - 1) * safePageSize), total, safePage, safePageSize);
+    }
+
+    public String exportDeliveries(String channel, String status, String instanceId) {
+        String normalizedChannel = normalizeFilter(channel);
+        String normalizedStatus = normalizeStatus(status);
+        String normalizedInstanceId = normalizeTrim(instanceId);
+        List<NotificationDeliveryPageVO> rows = mapper.findExportPage(
+                normalizedChannel, normalizedStatus, normalizedInstanceId, MAX_EXPORT_DELIVERIES);
+        StringBuilder csv = new StringBuilder("\uFEFF").append(EXPORT_CSV_HEADER);
+        for (NotificationDeliveryPageVO row : rows) {
+            CsvUtil.appendRow(csv,
+                    row.getId(),
+                    row.getAlertId(),
+                    row.getAlertTitle(),
+                    row.getAlertDomain(),
+                    row.getTransition(),
+                    row.getInstanceId(),
+                    row.getChannel(),
+                    row.getStatus(),
+                    row.getAttemptCount(),
+                    row.getCreatedAt(),
+                    row.getDeliveredAt(),
+                    row.getNextAttemptAt(),
+                    row.getLastError());
+        }
+        return csv.toString();
     }
 
     public void retryFailedDelivery(Long deliveryId) {
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
@@ -66,7 +66,7 @@ public class NotificationOutboxService {
     private static final int BATCH_SIZE = 20;
     private static final int MAX_EXPORT_DELIVERIES = 10_000;
     private static final String EXPORT_CSV_HEADER = "deliveryId,alertId,alertTitle,alertDomain,transition,"
-            + "instanceId,channel,status,attempts,createdAt,deliveredAt,nextRetryAt,lastError\r\n";
+            + "instanceId,channel,status,attemptCount,createdAt,deliveredAt,nextRetryAt,lastError\r\n";
     private static final Duration DEFAULT_CLAIM_TIMEOUT = Duration.ofMinutes(1);
     private static final Duration DEFAULT_CLAIM_RENEWAL_INTERVAL = Duration.ofSeconds(20);
     private static final int DEFAULT_HEARTBEAT_THREADS = 2;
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
```diff
@@ -222,6 +226,32 @@ public PageResult<NotificationDeliveryPageVO> listDeliveries(String channel, Str
                 (long) (safePage - 1) * safePageSize), total, safePage, safePageSize);
     }
 
+    public String exportDeliveries(String channel, String status, String instanceId) {
+        String normalizedChannel = normalizeFilter(channel);
+        String normalizedStatus = normalizeStatus(status);
+        String normalizedInstanceId = normalizeTrim(instanceId);
```
RockteMQ-AI, line 232: **[Info]** `MAX_EXPORT_DELIVERIES = 10_000` is a reasonable cap. Consider adding a comment explaining the rationale (e.g. "Excel row limit compatibility" or "memory safety bound") so future maintainers understand why this specific number was chosen.

RockteMQ-AI (approved):
### Summary
This PR adds a CSV export feature for notification deliveries, spanning both backend (Java service + controller + mapper) and frontend (React page with export button and blob download). The implementation is well-structured with proper filter normalization, CSV cell escaping via `CsvUtil`, BOM prefix for Excel compatibility, i18n support, mock mode handling, and good test coverage across all layers.

LGTM — approved with two minor info-level suggestions.

---

- **[INFO]** `web/src/pages/ops/notificationDeliveries.tsx:57` — **[Info]** The export filename uses `new Date().toISOString().slice(0, 10)` which produces a UTC date. For users in UTC+ timezones (e.g. Asia/Shanghai), this may show tomorrow's date after ~8am local time. Consider using a local date formatter instead, e.g.:
```ts
const d = new Date();
const localDate = `${d.getFullYear()}-${String(d.getMonth()+1).padStart(2,"0")}-${String(d.getDate()).padStart(2,"0")}`;
``` _(line outside diff)_

**Review by utafrali — COMMENTED** [review]
The feature is well-scoped and correctly reuses the injection-safe CSV writer, but it has a Firefox-breaking DOM bug in the download trigger, fetches a potentially large `message_content` column that is never written to the CSV, and wrapping the CSV in a JSON `Result<String>` response adds unnecessary serialization overhead and breaks curl-based consumers.

File: web/src/pages/ops/notificationDeliveries.tsx [hunk 2]
```diff
@@ -58,12 +59,37 @@ const NotificationDeliveriesPage = () => {
...
+    try {
+      const csv = await exportAlertDeliveries({
+        channel,
+        status,
+        instanceId,
+      });
+      const blob = new Blob([csv], { type: 'text/csv;charset=utf-8' });
+      const url = URL.createObjectURL(blob);
+      const anchor = document.createElement('a');
+      anchor.href = url;
```
utafrali, line 80: The anchor element is never appended to the DOM before `.click()` is called. Firefox requires the anchor to be in the document tree for programmatic downloads to fire — the download silently does nothing in Firefox. Fix:
```typescript
document.body.appendChild(anchor);
anchor.click();
document.body.removeChild(anchor);
URL.revokeObjectURL(url);
```

server/src/main/java/org/apache/rocketmq/studio/persistence/mapper/RmqAlertNotificationOutboxMapper.java
```diff
@@ -66,6 +66,23 @@ int renewClaim(@Param("id") Long id, @Param("claimToken") String claimToken,
     List<NotificationDeliveryPageVO> findPage(@Param("channel") String channel, @Param("status") String status,
             @Param("instanceId") String instanceId, @Param("limit") int limit, @Param("offset") long offset);
 
+    @Select("<script>"
+            + "SELECT o.id, o.alert_id AS alertId, o.channel, o.status, o.attempt_count AS attemptCount, "
+            + "o.next_attempt_at AS nextAttemptAt, o.last_error AS lastError, o.delivered_at AS deliveredAt, "
```
utafrali, line 71: `message_content` is selected here but `NotificationOutboxService.exportDeliveries` never calls `row.getMessageContent()` — the PR description even explicitly says full message content is intentionally excluded. This column holds a full alert payload; fetching it across up to 10 000 rows wastes I/O and heap. Remove it from the SELECT list in `findExportPage`.

server/src/main/java/org/apache/rocketmq/studio/ops/alert/SystemAlertController.java
```diff
@@ -87,6 +87,14 @@ public Result<PageResult<NotificationDeliveryPageVO>> listDeliveriesPage(
         return Result.ok(notificationOutboxService.listDeliveries(channel, status, instanceId, page, pageSize));
     }
```
utafrali, line 89: Returning `Result<String>` serializes the CSV as a JSON string value. Every embedded quote must be escaped, doubling parts of the payload, and the frontend has to parse JSON only to re-encode the data as a Blob. A proper download endpoint should return bytes with the right headers:
```java
@GetMapping(value = "/deliveries/export", produces = "text/csv;charset=UTF-8")
public ResponseEntity<byte[]> exportDeliveries(
        @RequestParam(required = false) String channel,
        @RequestParam(required = false) String status,
        @RequestParam(required = false) String instanceId) {
    byte[] csv = notificationOutboxService
            .exportDeliveries(channel, status, instanceId)
            .getBytes(StandardCharsets.UTF_8);
    return ResponseEntity.ok()
            .header(HttpHeaders.CONTENT_DISPOSITION,
                    "attachment; filename=\"notification-deliveries.csv\"")
            .body(csv);
}
```
This also makes the endpoint work correctly from curl or any non-browser client without post-processing.

server/src/main/java/org/apache/rocketmq/studio/ops/alert/NotificationOutboxService.java
```diff
@@ -63,6 +64,9 @@
 public class NotificationOutboxService {
     private static final int MAX_ATTEMPTS = 5;
     private static final int BATCH_SIZE = 20;
```
utafrali, line 66: Every column in the CSV header is camelCase (`deliveryId`, `alertId`, `createdAt`, ...) except `attempts`. The VO field is `attemptCount` and the DB column is `attempt_count`. Using `attempts` is the odd one out and will confuse any tooling that tries to match header names to field names. Rename it to `attemptCount` to be consistent with the rest of the header.

tju-yxq:
Thank you for the review — all four points are addressed in 44e0a1fd (also rebased onto the current `rocketmq-studio`):

1. **Firefox download trigger** — the page now uses the shared `downloadBlob` helper (`web/src/utils/download.ts`), which appends the anchor to `document.body` before clicking and removes it afterwards. The duplicated manual anchor code is gone.

2. **Unused `message_content`** — removed from `findExportPage`'s SELECT list. The export query no longer fetches the payload column it never wrote to the CSV.

3. **JSON-wrapped CSV** — the endpoint now returns `ResponseEntity<byte[]>` with `produces = "text/csv;charset=UTF-8"` and a `Content-Disposition: attachment` header. The frontend requests `responseType: 'blob'` and passes the response straight to `downloadBlob`, so there is no JSON escaping/re-encoding round trip and curl consumers get raw CSV.

4. **Header naming** — renamed `attempts` to `attemptCount` in the CSV header (both backend constant and mock-mode frontend header), matching the VO field and the rest of the camelCase columns.

Validation on the rebased branch:

- `mvn '-Dtest=NotificationOutboxServiceTest,SystemAlertControllerTest' test` — 38 tests passed
- `mvn checkstyle:check` — 0 violations
- `npm test -- NotificationDeliveriesPage.test.tsx --run` — 4 tests passed
- `npm run lint -- --quiet` — 0 errors
- `npm run build` — passed

The controller test now asserts the CSV content type, the attachment filename header, and the raw UTF-8 bytes; the page test asserts the blob is passed through the shared download helper and the object URL is revoked.

PR #3881, opened 2026-09-06
test: expand auth session storage coverage

yyqdbngt:
### Motivation

The auth session storage helpers had five tests covering the happy path, cookie independence, and clearing, but left null/zero user ids, the false admin flag, and empty-storage behavior unasserted. This PR grows the suite from 5 to 10 cases.

### Changes

- Persisting a `null` user id removes the stored id key instead of writing the literal string `null`.
- A numeric zero user id persists and restores as `0` (never treated as absent).
- Only the literal `true` flag yields `admin: true`; `false` and arbitrary flag text both read back as `false`.
- Completely empty storage reads back an all-`null` session.
- Clearing an already-empty session does not throw.

### Verification

- `vitest run src/stores/authStorage.test.ts`: 10/10 passed
- `tsc --noEmit`: clean
- `eslint src/stores/authStorage.test.ts`: clean

### DIFF
--- a/web/src/stores/authStorage.test.ts
+++ b/web/src/stores/authStorage.test.ts
@@ -70,4 +70,37 @@ describe('auth session storage', () => {
     expect(localStorage.getItem(USER_ID_STORAGE_KEY)).toBeNull();
     expect(localStorage.getItem(USER_ADMIN_STORAGE_KEY)).toBeNull();
   });
+
+  it('removes the stored user id when persisting a null id', () => {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Expands auth session storage test coverage — adds null user id removal, numeric zero id persistence, strict admin flag check, empty session read, and idempotent clear.

LGTM — good edge case coverage for storage boundary conditions (null, zero, strict boolean).

---

PR #3928, opened 2026-09-06
test: add first coverage for consumer group import DTO

yyqdbngt:
### Motivation

The `ImportConsumerGroupsDTO` used by the CSV import flow had no unit coverage for its required fields, group bounds, or nested validation. This PR adds the first four cases.

### Changes

- A missing instance id is rejected.
- Empty and over-100 group lists are rejected with their messages.
- Nested group validation propagates (`name is required` from a nameless entry).
- A valid import request passes validation.

### Verification

- `mvn -B test -Dtest=ImportConsumerGroupsDTOTest`: 4/4 passed, BUILD SUCCESS


RockteMQ-AI (approved):
LGTM. Test coverage improvement.

Issue #3985, opened 2026-09-06
Topic list drops matching rows when the search term has leading or trailing spaces


tju-yxq:
## Problem

On the Topic page, the search term is normalized for the server query but not for the client-side row filter. A pasted search term with leading or trailing spaces returns matching topics from the server, then filters every returned row out locally, leaving an empty table.

## What did you do (steps to reproduce)?

1. Open **Instance > Topic** for any instance that has topics (for example a topic named `orders-topic`).
2. Paste ` orders ` into the search box — with leading and trailing spaces, as commonly happens when copying from a ticket or chat.
3. Press Enter.

## What did you expect to see?

The table shows `orders-topic`, because the server was queried with the trimmed term `orders`.

## What did you see instead?

The table is empty even though matching topics exist and the backend returned them.

## Root cause

- `loadTopicPage` sends `search: searchText.trim()` to `/topics/page`.
- `visibleTopics` still receives the raw `searchText` and filters rows with `topic.name.toLowerCase().includes(searchText.toLowerCase())`.

Because no topic name contains the padded string `" orders "`, the client-side filter rejects every row the server correctly returned.

`web/src/pages/instance/topic.tsx`:

```ts
// server query (line ~433)
search: searchText.trim() || undefined,

// client filter (line ~484) — receives raw searchText
const filteredTopics = useMemo(
  () => visibleTopics(topics, selectedInstanceId, searchText, typeFilter),
  [topics, selectedInstanceId, searchText, typeFilter],
);
```

## Expected behavior

The trimmed search term should be used consistently by the server query, the client-side filter, and the export, so rows matching the trimmed term remain visible.

## Acceptance criteria

- [ ] A search term with leading/trailing spaces keeps matching rows visible.
- [ ] The server query, client-side filter, and export all use the same trimmed term.
- [ ] A regression test covers the padded-input case and fails on the unfixed source.

## Environment

Observed on the current `rocketmq-studio` branch (base commit `36126024`), in `web/src/pages/instance/topic.tsx`.

## Other information

The same defect class was previously fixed for the consumer-group search (#911); the Topic page retains the trim mismatch.

## Duplicate check

Searched open and closed issues/PRs for `topic search spaces`, `topic search trim`, `topic search empty`. #3307 reports the same root cause on the current branch; this issue was filed independently against the same behavior. No merged fix exists yet.

RockteMQ-AI:
**Issue Evaluation**

Category:  | Status: **Confirmed**

[...]

PR #3301, opened 2026-09-05
fix(topic): search topics with the trimmed term on the client too

unbridled-41:
Fixes #3307.

## Problem / Evidence
The topic list fetches with `search: searchText.trim()` (`loadTopicPage`) but passes the raw input to the client-side row filter `visibleTopics`, which does `topic.name.toLowerCase().includes(searchText.toLowerCase())`. A search term with leading/trailing whitespace (e.g. from a paste) queries the server with the trimmed term, receives the matching topics, then filters every returned row out because no topic name contains the padded string — the table shows empty despite matches existing.

Regression test added: `keeps matching rows when the search term has leading or trailing spaces` — red on pristine (row disappears after submitting " orders "), green after the fix.

## Root cause / Fix
Two different search terms are in play: the trimmed one used for the server query and the raw one used for client re-filtering.

Fix: trim once when the term is stored in `onSearch` (`setSearchText(value.trim())`), so the server query, the client filter, and the export all use the same effective term. Precedent: closed issue #911 fixed the same defect class for consumer-group search.

## Priority & scoring
PRIORITY 72 = 影响 26 (primary search interaction on the topic page returns an empty table for padded input) + 波及范围 12 (topic page) + 可复现性 19 (deterministic, trivial steps) + 维护价值 15 (single normalization point). FIX_CONFIDENCE 90.

## Tests
- New regression test red on pristine, green after fix.
- `npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx` → 22/22 pass.
- Full web suite (`npx vitest run --testTimeout=60000`): 923 tests, 922 pass; the single failure is `ConsumerPage.test.tsx` "shows group health diagnostics…", a documented load-fragile file untouched by this PR. TopicPage passes in the full run and in isolation.
- `npm run build` ✓ (CI frontend-build equivalent); `npx tsc --noEmit` clean; `npx eslint` on touched files clean.
- CI note: the upstream workflow is in a repo-wide `startup_failure` state (zero jobs run, including other contributors' branches), so CI-equivalent verification was run locally — see the verification comment below.

## Risk
Low. Only the stored search term changes; the input box itself still displays what the user typed.

PR #3986, opened 2026-09-06
fix(topic): keep matching rows for padded search terms

tju-yxq:

The Topic page's search normalized the term for the server query (`searchText.trim()`) but not for the client-side row filter, which received the raw input. A pasted term with leading or trailing spaces made the server return the matching topics while the local filter rejected every returned row, leaving the table empty.

This change makes the server query, client-side filter, and export use the same trimmed term. Details: #3985

## Brief changelog

**Root cause**

`loadTopicPage` sent `search: searchText.trim()` to `/topics/page`, but `visibleTopics` still received the raw input and matched with `topic.name.toLowerCase().includes(searchText.toLowerCase())`. With a pasted term such as ` orders `:

1. The server queried with the trimmed term `orders` and returned the matching topics.
2. The client-side filter rejected every returned row because no topic name contains the padded string.
3. The table rendered empty even though matches existed.

**Fix**

`web/src/pages/instance/topic.tsx` now derives `trimmedSearchText = searchText.trim()` once and passes it to `visibleTopics`, so the server query, the client-side row filter, and the export all use the same normalized term.

**Regression coverage**

`TopicPage.test.tsx` adds a test that types ` orders ` into the search box and asserts:

- the server query is called with `search: 'orders'`;
- the matching `orders-topic` row remains visible;
- the non-matching `users-topic` row stays hidden.

The test fails on the unfixed source because the matching row is filtered out client-side.

## Verifying this change

- `npm test -- TopicPage.test.tsx --run` in `web` — 22 tests passed
- `npm run lint -- --quiet` in `web` — 0 errors
- `npm run build` in `web` — passed (Vite, 8038 modules)
- `git diff --check` — passed

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change. This PR addresses only #3985.
- [x] A regression test covers the padded-input case and fails on the unfixed source.
- [x] Frontend tests, lint, and build were run locally on this branch.

### DIFF
--- a/web/src/pages/instance/__tests__/TopicPage.test.tsx
+++ b/web/src/pages/instance/__tests__/TopicPage.test.tsx
@@ -292,6 +292,34 @@ describe('TopicPage', () => {
     clickSpy.mockRestore();
   });
 
+  it('keeps matching rows when the search term has leading or trailing spaces', async () => {
+    const user = userEvent.setup();
+    mockTopicsList([
+      {
+        ...buildTopics(1)[0],
+        name: 'orders-topic',
+      },
+      {
+        ...buildTopics(1)[0],
+        name: 'users-topic',
+      },
+    ]);
+    renderWithProviders();
+
+    expect(await screen.findByText('orders-topic')).toBeInTheDocument();
+
+    await user.type(screen.getByPlaceholderText('搜索 Topic 名称'), ' orders ');
+    await user.keyboard('{Enter}');
+
+    await waitFor(() =>
+      expect(topicServiceMocks.listTopicsPage).toHaveBeenLastCalledWith(
+        expect.objectContaining({ search: 'orders' }),
+      ),
+    );
+    expect(await screen.findByText('orders-topic')).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) (56 lines) on `rocketmq-studio` ← `codex/topic-search-trim`.

### Changed Files
```
web/src/pages/instance/__tests__/TopicPage.test.tsx
web/src/pages/instance/topic.tsx
```

### Findings

---

lizhimins:
Closing as a duplicate of #3301, which was merged first (22h earlier) and normalizes the search term once at the storage point in `onSearch`, so the server query, the client-side `visibleTopics` filter, and the CSV export all share the same trimmed term. That single-point fix covers the three paths this PR addressed separately. The extra assertion here (non-matching rows stay hidden) is a reasonable follow-up but does not change the outcome. Thanks for the careful repro, @tju-yxq — the underlying defect was real and is now fixed on `rocketmq-studio`.

PR #3987, opened 2026-09-06
fix(alert): stop silently retargeting native metrics in Prometheus export

zjncs:
## Motivation

`AlertService.exportPrometheusRulesYaml` validates each rule's metric against the Prometheus name pattern (`^[a-zA-Z_:][a-zA-Z0-9_:]*$`) and silently falls back to `rocketmq_consumer_lag_messages` when it does not match.

Every native Studio metric name defined in `NativeAlertRulePolicy` (`consumer.lag.total`, `consumer.lag.max_queue`, `consumer.delay.seconds`, `topic.backlog.total`, `dlq.message.count`, …) contains dots, so **none** of them matches the pattern. Exporting a business rule such as:

- `dlq.message.count > 100`
- `topic.backlog.total > 50000`
- `consumer.delay.seconds > 300`

produces `expr: rocketmq_consumer_lag_messages > …` — the exported alert monitors **consumer lag** while the user configured a DLQ-size / topic-backlog / consumption-delay alert. The rule name, threshold and description all look right in Prometheus, so the retargeting is invisible and the intended condition is never alerted on.

## Modification

- `consumer.lag.total` now translates to its exact rocketmq-exporter equivalent `rocketmq_consumer_lag_messages` through an explicit `NATIVE_METRIC_EXPORT_NAMES` mapping (this is the only native business metric with a direct exporter equivalent — e.g. `rocketmq_dlq_resend_count` measures resend rate, not DLQ size, so `dlq.message.count` cannot be mapped faithfully).
- Rules whose native metric has no exporter equivalent are **skipped** from the export, and each skip is recorded as a trailing comment (`# Skipped "DLQ Flood": native metric 'dlq.message.count' has no equivalent in the rocketmq-exporter metric set`) so the omission is visible instead of silently wrong.
- When every enabled business rule is skipped, the file emits `groups: []` so it stays loadable.
- The existing fallback that replaces *malformed* metric strings (injection hardening, covered by `exportPrometheusRulesYamlShouldReplaceInvalidPrometheusFieldsTest`) is unchanged.
- `NativeAlertRulePolicy` exposes `isNativeMetric` so the native metric catalog stays the single source of truth.

## Verification

Added three tests to `AlertServiceTest`:

- `exportPrometheusRulesYamlShouldSkipNativeMetricsWithoutExporterEquivalentTest` — a `dlq.message.count` rule next to a regular lag rule: the lag rule is exported, the DLQ rule is skipped with a comment, and no `rocketmq_consumer_lag_messages > 100` expr appears. **Fails on the current branch** (the DLQ rule is exported as a lag rule).
- `exportPrometheusRulesYamlShouldTranslateNativeConsumerLagMetricTest` — `consumer.lag.total` exports as `rocketmq_consumer_lag_messages` with the consumer group/team naming.
- `exportPrometheusRulesYamlShouldEmitEmptyGroupsWhenAllRulesUseUnexportableNativeMetricsTest` — a single `topic.backlog.total` rule yields `groups: []` plus the skip comment and no substituted metric. **Fails on the current branch.**

`mvn -pl server test -Dtest='Alert*Test'` → all alert suites pass (AlertServiceTest 76/76). `NotificationOutboxServiceTest` fails 4F/1E identically on the base commit without this change (pre-existing).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricProfileService.java
@@ -22,6 +22,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 @Service
 @RequiredArgsConstructor
@@ -55,6 +56,20 @@ public class MetricProfileService {
                 .findFirst()
                 .orElseThrow(() -> badRequest("Unknown semantic metric '" + semanticMetric
                         + "' for profile '" + profileId + "'"));
+    }
+
+    /**
+     * Resolves the prometheus metric name for a semantic metric under the currently configured
+     * profile (the one {@link #listProfiles()} orders first). Returns empty when the active profile
+     * has no mapping for the semantic metric, so callers can treat it as unexportable rather than
+     * falling back to a hardcoded name that would be wrong for the deployed profile (e.g. consumer
+     * lag is {@code rocketmq_message_accumulation} on the 4.x exporter profile, not the 5.x name).
+     */
+    public Optional<String> resolveCurrentPrometheusMetric(String semanticMetric) {
+        return listProfiles().get(0).getMetrics().stream()
+                .filter(metric -> metric.getSemanticMetric().equals(semanticMetric))
+                .map(MetricProfileVO.MetricMappingVO::getPrometheusMetric)
+                .findFirst();
     }
 
     private MetricProfileVO profile(MetricProfile profile,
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 3 file(s) (222 lines) on `rocketmq-studio` ← `fix/alert-export-native-metric-substitution`.

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
server/src/main/java/org/apache/rocketmq/studio/ops/alert/NativeAlertRulePolicy.java
server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertServiceTest.java
```

### Findings

---

PR #3988, opened 2026-09-06
fix(consumer): report unknown lag instead of summing sentinel queues away

zjncs:
## Motivation

`RocketMQMetadataProvider.enrichGroupLiveStats` (consumer group list enrichment) and `RocketMQAdminClientImpl.fillConsumeStats` (group detail) accumulate per-queue lag with a legacy clamp:

```java
long diff = wrapper.getBrokerOffset() - wrapper.getConsumerOffset();
if (diff > 0) {
    totalLag += diff;
}
```

A queue whose raw diff is negative — the `-1` unknown sentinel RocketMQ 5.0 gRPC consumers report — therefore **silently contributes zero** to `totalLag`:

- a group that also has healthy queues shows a **partial sum presented as the total**;
- a group whose queues are all unknown shows a **fabricated zero lag** with `consumeStatsAvailable=true`.

`ConsumerLagResolver` was introduced in this codebase exactly to keep that state visible — `getTopicConsumersPage` and `getGroupProgress` already resolve every diff through it and propagate `UNKNOWN` when any queue is unknown. The web UI also already handles the sentinel (`web/src/utils/consumerLag.ts`: `isLagAvailable`/`formatLag` render `-1` as an explicit gray "unavailable" state in the group list and detail). These two aggregation sites were simply left on the pre-resolver clamping behavior, so the UI's unknown-lag rendering never gets a chance to trigger for them.

## Modification

Route both loops through `ConsumerLagResolver`:

- `enrichGroupLiveStats` uses the existing `resolveDiff` helper (via the class's `proxyStatsProvider`);
- `fillConsumeStats` resolves with no proxy (matching `ConsumerLagResolver`'s documented no-proxy behavior);
- any queue resolving to `UNKNOWN` marks the whole `totalLag` unknown — the same semantics `getTopicConsumersPage` already applies to `diffTotal` — while the delay-seconds timestamp scan still covers every queue.

## Verification

Added one regression test per site (`offset(100, 60)` healthy queue + `offset(0, 1)` sentinel queue — the same stub shape the existing `getTopicConsumersKeepsUnknownWhenAnyQueueLagIsUnknown` test uses):

- `RocketMQAdminClientImplTest.getConsumerGroupReportsUnknownTotalLagWhenAnyQueueOffsetIsUnknownTest`
- `RocketMQMetadataProviderTest.listConsumerGroupsShouldReportUnknownTotalLagWhenAnyQueueOffsetIsUnknownTest`

Both **fail on the base branch** with `expected: -1L but was: 40L` (the sentinel queue is summed away and the partial 40 is reported as the total) and pass with this change.

`mvn -pl server test -Dtest='org.apache.rocketmq.studio.provider.apache.*Test'` → full provider package green (RocketMQAdminClientImplTest 42/42, RocketMQMetadataProviderTest 36/36, ConsumerLagResolverTest 5/5, …).

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqBusinessMetricsCollector.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqBusinessMetricsCollector.java
@@ -27,6 +27,7 @@ import org.apache.rocketmq.studio.instance.group.ConsumerGroupVO;
 import org.apache.rocketmq.studio.instance.group.QueueProgressVO;
 import org.apache.rocketmq.studio.provider.InstanceProvider;
 import org.apache.rocketmq.studio.provider.InstanceProviderRegistry;
+import org.apache.rocketmq.studio.provider.apache.ConsumerLagResolver;
 import org.apache.rocketmq.studio.ops.alert.AlertDomain;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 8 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 4 file(s) (148 lines) on `rocketmq-studio` ← `fix/consumer-lag-unknown-sentinel`.

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java
server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java
server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImplTest.java
server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProviderTest.java
```

### Findings

- **[Info]** 2 test file(s) included.

---

PR #3989, opened 2026-09-06
fix(alert): keep consumer group and topic scope in Prometheus export

zjncs:
## Motivation

`AlertService.labelSelector` only forwards `clusterName` and `brokerName` when building the exported PromQL selector. Alert rules scoped to a **consumer group** or **topic** — first-class scope fields on `AlertRuleVO`, validated by `NativeAlertRulePolicy` for the group-scoped business metrics (`consumer.lag.total`, `consumer.lag.max_queue`, `consumer.delay.seconds`, `topic.backlog.total`, `dlq.message.count`) — therefore export as a bare `metric > threshold` expression.

The exported alert then evaluates **every series** of the metric: a rule configured as "cg-orders lag > 5000" fires when *any* consumer group's lag crosses the threshold, and a topic-scoped rule alerts on every topic. The configured boundary is silently widened to the whole cluster.

## Modification

Forward the rule's `consumerGroup` and `topic` fields as selector labels, with the **label names resolved from the active metric profile** instead of hardcoded:

- each `MetricProfileVO.MetricMappingVO` now carries `scopeLabels`, a canonical scope dimension (`cluster`/`broker`/`topic`/`consumer_group`) → label-name map derived from the labels the profile's series actually expose — the 4.x exporter profile labels consumer-group series `group`, the 5.x native profile `consumer_group`;
- new `MetricProfileService.resolveCurrentScopeLabel(semanticMetric, scope)` resolves the label for the active profile's mapping of the rule's semantic metric (native metrics via `NATIVE_METRIC_SEMANTIC`, raw exporter metric names fall back to the consumer-lag mapping, the export default, whose labels document the deployed profile's naming);
- `labelSelector` emits the consumer-group and topic selectors with the resolved names; a scope the active profile does not break the metric out by resolves empty and **drops the selector** rather than guessing a name that would match an empty series set. A hardcoded `group` selector on a 5.x deployment (`rocketmq_consumer_lag_messages` + `consumer_group` label) would silently disable the alert — worse than the current over-broad behavior;
- `cluster` and `broker` keep their pre-existing literal names (both profiles spell them identically);
- empty and `*` wildcard scopes keep exporting without the label.

## Verification

Fail-before (base branch, new behavioral tests): `Tests run: 4, Failures: 3` — the scope exports without any selector:

- `expr: rocketmq_consumer_lag_messages{consumer_group="cg-orders"} > 5000` — not contained (exported as bare `rocketmq_consumer_lag_messages > 5000`);
- `expr: rocketmq_message_accumulation{group="cg-orders"} > 5000` — not contained;
- combined `{cluster="DefaultCluster",consumer_group="cg-payments",topic="payments"}` — not contained.

Pass-after:

- `AlertServiceTest` + `MetricProfileServiceTest` → `Tests run: 93, Failures: 0, Errors: 0` (BUILD SUCCESS), covering **both profiles**: the native profile emits `rocketmq_consumer_lag_messages{consumer_group="cg-orders"}`, the 4.x exporter profile resolves the native `consumer.lag.total` metric to `rocketmq_message_accumulation` with the `group` label, and `resolveCurrentScopeLabelShouldFollowActiveProfileTest` pins the per-profile label table (including empty for unmapped dimensions like `broker` on the lag mapping);
- full `org.apache.rocketmq.studio.cluster.metrics` package → `Tests run: 142, Failures: 0, Errors: 0`.


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) (65 lines) on `rocketmq-studio` ← `fix/alert-export-consumer-group-scope`.

### Changed Files
```
server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertServiceTest.java
```

### Findings

---

lizhimins (changes requested):
The defect is real and correctly spotted: the exported rule only carries `cluster`/`broker`, so a rule scoped to a specific group or topic is silently widened to the whole cluster — a false positive that fires for unrelated groups.

The implementation direction, however, swaps that false positive for a more dangerous false negative:

1. **The label name is hardcoded to the wrong profile.** This repo's 5.x native profile exposes consumer lag as `rocketmq_consumer_lag_messages` with the label `consumer_group`; `group` is the 4.x exporter profile's label (and 4.x uses different metric names entirely — `rocketmq_message_accumulation`). On a 5.x deployment a `{group="..."}` selector matches the empty set, so the alert never fires at all. That is worse than the current over-broad behavior, because it fails silently.
2. **Resolve labels and metric names from the active profile, never hardcode.** `MetricProfileService` is the single source of truth for semantic→prometheus name *and* label sets per profile (each mapping already carries `getLabels()`), and #3987 (just merged) now resolves the export metric name through `MetricProfileService.resolveCurrentPrometheusMetric(...)`. The scope selector must do the same for the label key — pick `consumer_group` vs `group` from the current profile's mapping for the rule's semantic metric.
3. **Tests must cover both profiles.** Add a case under the 4.x exporter profile and one under the 5.x native profile, asserting the emitted selector uses the profile-correct label and actually matches a sample series.

Suggested rework: derive both the prometheus metric name and the scope label from `MetricProfileService` for the configured profile, and test against both profiles so the selector is correct on 4.x and 5.x alike.

zjncs:
Thanks for the thorough review — you're right that hardcoding `group` swaps the false positive for a silent false negative on 5.x deployments, which is worse.

Reworked as suggested (branch rewritten onto the current `rocketmq-studio`, now a single commit on top of it):

- Each `MetricProfileVO.MetricMappingVO` now carries `scopeLabels`, mapping canonical scope dimensions (`cluster`/`broker`/`topic`/`consumer_group`) to the label name the profile actually exposes — `group` on the 4.x exporter profile, `consumer_group` on the 5.x native profile.
- New `MetricProfileService.resolveCurrentScopeLabel(semanticMetric, scope)` resolves the label from the **active profile's mapping for the rule's semantic metric** (native metrics via the existing `NATIVE_METRIC_SEMANTIC`; raw exporter metric names fall back to the consumer-lag mapping — the export default — whose labels document the deployed profile's naming). An unmapped dimension resolves empty and the selector is dropped rather than guessed, since a wrong label matches an empty series set.
- The metric name keeps flowing through `resolveCurrentPrometheusMetric` from #3987, so on the 4.x profile a `consumer.lag.total` rule now exports as `rocketmq_message_accumulation{group="cg-orders"}`, and on 5.x as `rocketmq_consumer_lag_messages{consumer_group="cg-orders"}`.
- Tests cover both profiles: exporter-profile and native-profile consumer-group scope, topic/cluster scope on both, wildcard/empty scope still bare, plus `resolveCurrentScopeLabelShouldFollowActiveProfileTest` pinning the per-profile label table (including empty for dimensions a mapping doesn't break out, like `broker` on the lag mapping).

`AlertServiceTest` + `MetricProfileServiceTest` → 93/93 green, fail-before on the base branch was 3 failures out of the 4 behavioral tests (the fourth is the wildcard guard). Cluster and broker keep their pre-existing literal names — both profiles spell them identically; happy to route those through the profile too if you'd prefer.

PR #3995, opened 2026-09-06
fix(audit): label the exported timestamp column with the server zone

zjncs:
## Motivation

Audit timestamps are stored as zone-less server-local values — whatever zone the server JVM and the MySQL session run in; the documented deployment (`docker-compose.yml`) sets `TZ=Asia/Shanghai` — while the alert subsystem formats its timestamps in UTC (`formatUtcDateTime`). The audit CSV export shipped a bare `timestamp` header, so a consumer could not tell which zone the exported values were in; the natural assumption (UTC, matching the alerting side) is 8 hours off on the documented deployment.

## Modification

Rescoped to the labeling-only first step (per review — the stored base is not converted):

- the exported CSV's first column is now named after the server's UTC offset — e.g. `timestamp(UTC+08:00)` on a `TZ=Asia/Shanghai` server, `timestamp(UTC)` on a UTC server — computed from the JVM default zone at export time;
- the stored base, the UI rendering, and the cleanup cutoff are untouched, so storage, filtering and display stay on one consistent time base;
- switching the stored base to UTC is deliberately left out: that is a repo-wide decision (single time base for all persisted timestamps, schema defaults, and a backfill/migration for existing rows, plus CSV/display labeling) and needs its own scoped change.

## Verification

`AuditServiceTest#exportLogsLabelsTimestampColumnWithServerZoneOffset` (fails on the base branch — the header exports as bare `timestamp,`):

- under `Asia/Shanghai` the export starts with `\uFEFFtimestamp(UTC+08:00),operator,` and the stored value `2026-08-01T09:30` passes through unconverted;
- under `UTC` the export starts with `\uFEFFtimestamp(UTC),operator,`.

Full `AuditServiceTest` → `Tests run: 15, Failures: 0, Errors: 0`, BUILD SUCCESS.


RockteMQ-AI (commented):
### Findings
- **[Info]** `OperationAuditService.java:49` and `AuditService.java:109` — Both `record()` methods now stamp UTC consistently. The inline comments clearly document the cross-subsystem contract.
- **[Info]** `AuditService.java:131` — The cleanup cutoff also uses UTC, which is critical: if the cutoff used local time while stored values use UTC, the `deleteBefore` query would be off by the timezone offset.
- **[Info]** `audit.tsx:140` — Frontend switches from `formatDateTime` to `formatUtcDateTime` for the timestamp column. This ensures the UI displays UTC values without misleading timezone conversion.

### Suggestions
- If there are existing audit records stored with local-timezone values (from before this fix), consider a brief note in the PR description or a migration comment about how mixed-timezone records will render. Old records may display with an offset shift until they age out via cleanup.

---
*Automated review by @RockteMQ-AI*

lizhimins (changes requested):
The facts check out: the three audit write points use `LocalDateTime.now()`, the alert side already formats in UTC via `formatUtcDateTime`, and `docker-compose.yml` sets `TZ=Asia/Shanghai`, so the server clock is not UTC and the audit timestamps are +8 wall-clock while alerts are UTC. The inconsistency is real and the direction (store/emit a single, labeled time base) is right.

The change is half-applied, though, and as scoped it introduces new problems:

1. **No migration/backfill for existing rows.** Current audit rows are +8 wall-clock; switching writes to UTC moves all *new* rows 8 hours earlier than the existing ones, so the table becomes internally inconsistent across the cutover with no backfill or compatibility plan.
2. **It splits timezones within one table and across the repo.** `gmt_create`/`gmt_modified` on the same audit table are still generated by the MySQL session timezone, so the table would mix two time bases in different columns; meanwhile every other table still uses `LocalDateTime.now()`. Changing only the audit table creates cross-module semantic divergence rather than removing it.
3. **The acknowledged follow-up is the same defect.** The day-granularity filter using `atStartOfDay` has the identical +8 issue — a +8 user filtering "today" misses 00:00–08:00 — so it cannot be deferred as unrelated.
4. **The CSV/export header carries no UTC annotation**, so a consumer cannot tell what timezone the column is in.

This is really a repo-wide decision (single time base for all persisted timestamps, incl. schema defaults and a backfill/migration), not a one-table patch. Returning for now; the right follow-up is a dedicated "unify timestamp timezone across the repo (schema + migration + display/CSV labeling)" change. If you'd like a smaller first step that is safe on its own, labeling the existing export/CSV as the server's local zone (without changing the stored base) would be acceptable, but the UTC migration should be its own scoped effort with a backfill plan.

zjncs:
Thanks for the careful review — agreed on all four points: without a backfill the cutover splits the table across two time bases, `gmt_create`/`gmt_modified` would keep drifting with the MySQL session zone, the day-granularity filter inherits the same offset problem, and the export header gave the consumer no way to know the zone.

I've rescoped this PR to the smaller first step you offered: **no change to the stored base** (writes, UI rendering and the cleanup cutoff all stay on the server-local base they use today), and the exported CSV's timestamp column is now named after the server's UTC offset — `timestamp(UTC+08:00)` on the documented `TZ=Asia/Shanghai` deployment, `timestamp(UTC)` on a UTC server — computed from the JVM default zone at export time. `AuditServiceTest#exportLogsLabelsTimestampColumnWithServerZoneOffset` pins both zones (fail-before on the base branch: the header exports as bare `timestamp,`), full class 15/15 green.

Branch rewritten onto the current `rocketmq-studio` as a single commit. The repo-wide unification — one persisted time base, schema defaults, backfill for existing rows, and display/CSV labeling together — is exactly the dedicated effort you described; I'll prepare it as its own change once this lands or alongside it, whichever you prefer.

PR #4001, opened 2026-09-06
fix(lite-topic): guard TTL extension against double submission

### LINKED ISSUE #2889: LiteTopic TTL extension can be submitted more than once
### Problem\n\nThe LiteTopic TTL dialog relies on extendTTLLoading to disable its confirmation button. React state is not synchronous, so two confirmation events delivered in the same render invoke extendLiteTopicTTL twice. TTL extension is a write operation, and the duplicate request can apply the extension more than once depending on provider semantics.\n\n### Reproduction\n\n1. Open a LiteTopic row and choose Extend TTL.\n2. Enter a TTL.\n3. Dispatch two confirmation clicks before the first API promise resolves.\n4. Observe two extendLiteTopicTTL calls.\n\nA focused LiteTopic component test reproduces this with a deferred promise.\n\n### Expected behavior\n\nAcquire a synchronous in-flight guard before issuing the extension and release it after completion.\n\n### Scope\n\nThe LiteTopic TTL mutation and its component regression test only.

btlqql:
## Summary
Adds a synchronous in-flight guard to the LiteTopic TTL extension flow so two confirmation events delivered in the same render cannot enqueue duplicate `extendLiteTopicTTL` write requests.

## Root cause
`handleExtendTTL` relied only on React `extendTTLLoading` state to disable the modal confirm button. React state updates are asynchronous, so two confirmation events delivered before the re-render both passed the guard and called `extendLiteTopicTTL`.

## Changes
- Add `extendTTLInFlightRef` and set/clear it synchronously around the API call.
- Add a LiteTopic page regression test that fires two confirm clicks in the same render against a deferred API promise and asserts the write is issued once.

## Testing
Commands run:
- `cd web`
- `npx vitest run src/pages/studio/__tests__/LiteTopic.test.tsx`
- `cd ..`
- `git diff --check`

Actual results:
- Vitest: `Test Files 1 passed (1)`, `Tests 15 passed (15)`.
- `git diff --check`: no output, exit code 0.

I did not run the full Maven/Java backend suite on this Windows host because no Maven wrapper or Maven executable is available and the `rocketmq-studio` server targets Java 21. This PR changes frontend code only.

Fixes #2889

Assisted-by: OpenAI:Codex (GPT-5)
Percentage of AI-generated code: 100

### DIFF
--- a/web/src/pages/studio/LiteTopic.tsx
+++ b/web/src/pages/studio/LiteTopic.tsx
@@ -149,6 +149,7 @@ const LiteTopicPage: React.FC = () => {
     newTTL: number | null;
   }>({ topicPattern: '', newTTL: null });
   const [extendTTLLoading, setExtendTTLLoading] = useState(false);
+  const extendTTLInFlightRef = useRef(false);
 
   const mountedRef = useRef(false);
   const bootstrapRequestId = useRef(0);
@@ -308,6 +309,8 @@ const LiteTopicPage: React.FC = () => {
 
   const handleExtendTTL = async () => {
     if (!extendTTLForm.topicPattern || extendTTLForm.newTTL == null) return;
+    if (extendTTLInFlightRef.current) return;
+    extendTTLInFlightRef.current = true;
     setExtendTTLLoading(true);
     try {
       await extendLiteTopicTTL(extendTTLForm.topicPattern, extendTTLForm.newTTL);
@@ -317,6 +320,7 @@ const LiteTopicPage: React.FC = () => {
     } catch {
       message.error(t('liteTopic.extendTtlFailed'));
     } finally {
+      extendTTLInFlightRef.current = false;
       setExtendTTLLoading(false);
     }
   };
--- a/web/src/pages/studio/__tests__/LiteTopic.test.tsx
+++ b/web/src/pages/studio/__tests__/LiteTopic.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
-import { act, render, screen, waitFor, within } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { App } from 'antd';
 import { LangProvider } from '../../../i18n/LangContext';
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (commented):
### Summary
This PR modifies 2 file(s) with 70 lines of changes.

### Findings

- **[WARNING]** Consider adding null checks for method call chains to prevent NullPointerException.

---

RockteMQ-AI (approved):
LGTM. Clean fix for the race condition using synchronous ref guard. Good regression test coverage.

---

lizhimins (changes requested):
The change itself is small and harmless, but I don't think it fixes a bug a user can reach, and it adds another copy of a pattern this frontend already has too many of.

**The double submit being guarded against cannot happen through the UI.** The dialog already renders `confirmLoading={extendTTLLoading}` (`LiteTopic.tsx:887`). Two things then make a second request unreachable:

- React 18 flushes discrete events such as `click` synchronously, so the first click's `setExtendTTLLoading(true)` is committed and the DOM updated before the browser dispatches a second click;
- antd's Button ignores its click handler while loading — `node_modules/antd/es/button/button.js:184-187` is literally `if (innerLoading || mergedDisabled) { e.preventDefault(); return; }`.

So by the time a real second click arrives, the OK button is in loading state and its `onClick` is not invoked. One request, without the new ref.

**The added test only reaches a double call by firing both clicks inside a single `act()`:**

```js
act(() => {
  fireEvent.click(confirmButton);
  fireEvent.click(confirmButton);
});
```

Nothing re-renders between the two, so `confirmLoading` never gets the chance to put the button into loading state. The test name says exactly this — "clicked twice in the same render" — and that is a property of the test harness rather than of a user double-clicking. The assertion `toHaveBeenCalledTimes(1)` therefore passes because of the guard, but the scenario it pins is one the UI cannot produce. A test that can only fail against code that is defending an unreachable path is not adding coverage.

**If you have a reachable duplicate-submit path, I want to see it and this becomes worth doing.** Plausible candidates: Enter in the form submitting while the OK button is clicked, a keyboard auto-repeat on Enter/Space, or a second entry point into `handleExtendTTL` that does not go through the modal's OK button. If you can demonstrate one, please come back with a test that reproduces *that* path — the fix should be pinned by the real repro rather than by the same-render double click.

**Separately, we are trying to stop multiplying hand-rolled in-flight guards.** `grep -c 'InFlightRef\|inFlightRef\|RequestRef\|requestIdRef'` matches in at least 12 frontend files, and `pages/cluster/index.tsx` alone has 44. The intent is to converge these onto one shared hook; adding another copy here makes that slightly harder, and it is the reason I would not take this even if the scenario were real.

Leaving this as changes requested rather than closing it, since a genuine repro would change the answer.

btlqql:
@lizhimins You are right, and I am closing this rather than asking you to re-review it.

I went back to the code and confirmed your reasoning rather than just accepting it. `onOk={handleExtendTTL}` (`web/src/pages/studio/LiteTopic.tsx:885`) is the only call site, and the `<Form>` in the modal body (`:892`) has no `onFinish` — the modal footer sits outside that form, so the Enter and auto-repeat paths you floated do not reach the handler either. `handleOpenExtendTTL` at `:442` is the only way in. With `confirmLoading={extendTTLLoading}` and `node_modules/antd/es/button/button.js:184-187` returning before `onClick`, the guarded path really is unreachable. My test only produced two calls because both clicks were dispatched inside a single `act()`, so it was pinning a property of the harness, exactly as you said.

I also agree about the duplication. Adding a thirteenth local in-flight guard while you are converging these onto one shared hook makes that work harder and buys nothing reachable, so it is not worth taking on those grounds either.

Rather than leave it open as a repro-in-waiting: if I turn up a duplicate-submit path a user can genuinely reach, I will bring it back on top of the shared hook instead of as another local ref.

Thank you for the specific write-up — pointing at the antd button source and at `pages/cluster/index.tsx` was far more useful than a plain rejection would have been.

PR #4005, opened 2026-09-06
fix(message): paginate against the committed query

### LINKED ISSUE #4009: [Studio][Bug] Message list pagination re-queries with uncommitted form inputs, mixing two queries into one list
## Problem

On the message page, clicking a pagination link re-queries the server with whatever the form inputs currently hold — not with the query that produced the displayed results. Editing the topic, date range or key/msgId after running a query and then paging silently executes a different query, so the table shows rows stitched together from two different queries.

## Evidence

`web/src/pages/instance/message.tsx` recomputes `currentQueryParams` from the live inputs on every render (`:429-434`) and the table's pagination handler passes it straight to `executeQuery` (`:1206-1214`); no input handler snapshots the committed query or clears results (only mode change and reset do).

Reproduction (deterministic): run a multi-page topic query for `order-create`, change the topic select to `payment-callback` without clicking 查询, then click page 2 — the page-2 fetch runs with `payment-callback` while page 1 still shows `order-create` rows. The regression test added with the fix fails before the patch with exactly this divergence.

## Impact

One of the most-used flows presents a mixed result list with a jumping total; an operator scrolling pages believes they are reading one continuous query result.

## Expected behavior

Pagination re-runs the committed query (the mode and normalized parameters that produced the currently displayed results). Editing form inputs alone must not change what a pagination click fetches until 查询 is submitted again. Replaying a stored history record updates the committed query to the record's own parameters.

## Related work

#2529/#2531 introduced server-side paging for this page; #2601 (closed) fixed duplicate history rows written by pagination; #1805/#1810 covered out-of-range task pages. None cover the live-input re-query on pagination.

## PR

#4005

unbridled-41:
Fixes #4009.

## Problem / Evidence

The message page's results table rebuilds the query from the *live* form inputs on every pagination click (`message.tsx`: `onChange: (page, pageSize) => void executeQuery(queryMode, currentQueryParams, page, pageSize)`, where `currentQueryParams` is recomputed from `selectedTopic`/`dateRange`/`keyInput`/`msgIdInput` on each render, and no input handler snapshots or clears the running query). So:

1. Run a multi-page query (e.g. topic `order-create` over a date range).
2. Change the topic select, date range, or key/msgId input **without** clicking 查询.
3. Click page 2.

The table then silently executes the *new, uncommitted* query for page 2 while page 1 still shows the committed query's rows — the user sees one continuous result list that is actually stitched together from two different queries (different topic, time window, or key), with totals jumping. The same divergence exists after replaying a stored history record.

Regression test output before the fix: after committing topic `order-create`, selecting topic `payment-callback` in the form and clicking page 2 fetched `payment-callback` for page 2.

## Root cause / Fix

Pagination re-queried live form state instead of the query that produced the displayed results. Fix: keep a `committedQueryRef` that snapshots `{ mode, params: normalizedParams }` whenever a query actually executes (in `executeQuery` after validation, and in `replayHistoryRecord` with the record's own parameters). Pagination calls `executeQuery(committed.mode, committed.params, page, pageSize)`; when no query has run yet there is nothing to paginate and the handler is a no-op.

## Priority & scoring

PRIORITY 71 = impact 27 (a core page presents results of two different queries as one list — wrong data for operational decisions) + blast radius 12 (every multi-page message query followed by any form edit; message query is one of the most-used flows) + reproducibility 19 (fully deterministic UI interaction, no timing) + maintenance value 13 (aligns pagination with the file's own generation-guard pattern for invalidating superseded queries).

FIX_CONFIDENCE 88.

## Tests

`npx vitest run src/pages/instance/__tests__/MessagePage.test.tsx src/pages/instance/__tests__/MessagePageAsyncState.test.tsx`:

- Before the fix the new regression `paginates against the committed query instead of live form inputs` failed: page 2 was fetched with the edited topic (`AssertionError: expected last "vi.fn()" call to have been called with [ObjectContaining{topic: 'order-create'}]`).
- After the fix: **28 passed (28)** (27 pre-existing + 1 new).
- Full web suite (`npx vitest run`): **Tests run: 923 passed (115 files)**, zero failures.
- `npx tsc --noEmit` clean; `npx eslint` on touched files: no problems; `npm run build` succeeds.

## Risk

Low. The only behavior change is which parameters a pagination click uses: the committed snapshot instead of live inputs. Submitting 查询, mode switches and resets already clear/replace results and now also refresh the snapshot; a history replay updates the snapshot to the record's own parameters, so paging after a replay re-runs the replayed query rather than a stale one.

### DIFF
--- a/web/src/pages/instance/__tests__/MessagePage.test.tsx
+++ b/web/src/pages/instance/__tests__/MessagePage.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import { App } from 'antd';
-import { render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type React from 'react';
 import { MemoryRouter } from 'react-router-dom';
@@ -317,5 +317,43 @@ describe('Message page query history', () => {
     await user.click(screen.getAllByRole('combobox')[1]);
 
     expect(screen.queryByText('order-create')).not.toBeInTheDocument();
+  });
+
+  it('paginates against the committed query instead of live form inputs', async () => {
+    messageServiceMocks.queryMessages.mockResolvedValue(
+      Array.from({ length: 60 }, (_, index) => createMessage(`m-${index}`)),
+    );
+    const user = userEvent.setup();
+    renderWithProviders(<MessagePage />);
+
+    await user.click(lastElement(screen.getAllByRole('combobox')));
+    await user.click(lastElement(await screen.findAllByText('order-create')));
+    await user.click(screen.getByRole('button', { name: /^search查询$/ }));
+    await waitFor(() => {
+      expect(messageServiceMocks.queryMessages).toHaveBeenLastCalledWith(
+        expect.objectContaining({ topic: 'order-create' }),
+      );
+    });
+
+    // Edit the form without re-running the query.
+    await user.click(screen.getAllByRole('combobox')[1]);
+    const visibleOption = (await screen.findAllByText('payment-callback'))
+      .map((element) => element.closest('.ant-select-item-option'))
+      .find(
+        (element): element is HTMLElement =>
+          element instanceof HTMLElement &&
+          element.closest('.ant-select-dropdown:not(.ant-select-dropdown-hidden)') !== null,
+      );
+    if (!visibleOption) throw new Error('Visible topic option not found');
+    fireEvent.click(visibleOption);
+
+    const secondPage = document.querySelector('.ant-pagination-item-2') as HTMLElement | null;
+    if (!secondPage) throw new Error('Pagination page 2 not found');
+    await user.click(secondPage);
+    await waitFor(() => {
+      expect(messageServiceMocks.queryMessages).toHaveBeenLastCalledWith(
+        expect.objectContaining({ topic: 'order-create' }),
+      );
+    });
   });
 });
--- a/web/src/pages/instance/message.tsx
+++ b/web/src/pages/instance/message.tsx
@@ -59,7 +59,6 @@ import {
   QueueBrowserResults,
 } from '../../components/QueueBrowser';
 import type { MessageQueryHistory, TraceQueryHistory } from '../../api/messageHistory';
-import { getMessageQueryResults } from '../../api/messageHistory';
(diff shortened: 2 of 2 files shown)


unbridled-41:
## Verification evidence

**CI**: upstream CI cannot run here. Head commit `89895053590e3cff5fe54193f9fc5f4725120ced`: `commits/{sha}/check-runs` → `total_count: 0`, `commits/{sha}/status` → 0 statuses; the single workflow run recorded for this head SHA ("CI", pull_request, 2026-09-06T17:29:53Z) ended `startup_failure` (workflow never started); recent `rocketmq-studio` base-branch runs show the same `startup_failure`, pre-existing and unrelated.

**Red light (before the fix)** — `npx vitest run src/pages/instance/__tests__/MessagePage.test.tsx`:

```
× paginates against the committed query instead of live form inputs
AssertionError: expected last "vi.fn()" call to have been called with [ ObjectContaining{…} ]
  (page 2 was fetched with the edited topic payment-callback instead of
   the committed topic order-create)
Tests  1 failed | 9 passed (10)
```

**Green light / module tests (after the fix)**: MessagePage.test.tsx + MessagePageAsyncState.test.tsx → `Tests  28 passed (28)` (27 pre-existing + 1 new).

**Full suite + baseline**: `npx vitest run --testTimeout=60000` → `Test Files 115 passed (115), Tests 923 passed (923)` — zero failures.

**Build**: `npx tsc --noEmit` clean; `npx eslint src/pages/instance/message.tsx src/pages/instance/__tests__/MessagePage.test.tsx` → no problems; `npm run build` succeeds (`✓ built in 10.06s`).

**Diff self-check**: 2 files, +61/−3 (`git show 89895053 --numstat`) — `message.tsx` +22/−2 (committedQueryRef snapshot in executeQuery and replayHistoryRecord, pagination uses the snapshot) and `MessagePage.test.tsx` +39/−1 (1 regression test, plus switching one import line to add `fireEvent`). No unrelated changes.

RockteMQ-AI (approved):
### Summary
Pagination now correctly re-queries the committed query parameters instead of live form inputs — fixes a data integrity issue where editing the form before clicking page 2 would silently mix results from two different queries.

Clean implementation using `committedQueryRef` to snapshot query params at execution time, with pagination reading from the snapshot. The regression test precisely reproduces the bug scenario.

LGTM.

---

PR #4011, opened 2026-09-07
test(config): cover legacy static mock flag

### LINKED ISSUE #4010: Add regression coverage for the deprecated static mock flag in config
## Problem
web/src/config.ts exports USE_MOCK for backward compatibility, but the existing test only covers API_BASE_URL. The deprecated flag is not protected by a regression test, so it could be accidentally flipped and reintroduce the legacy mock-data code path.

## Expected behavior
- USE_MOCK remains alse for the current Studio runtime.
- A focused Vitest case fails if the flag is changed.

## Scope
Only add regression coverage for web/src/config.test.ts. No runtime code changes.

## Acceptance criteria
- [ ] A Vitest assertion verifies USE_MOCK === false.
- [ ] The targeted test file passes.

btlqql:
## Summary
Add a focused Vitest regression case that locks USE_MOCK to alse in web/src/config.ts.

## Root cause
The deprecated static mock flag had no regression coverage, so a future change could re-enable the legacy mock path without a failing test.

## Changes
- Import and assert USE_MOCK in web/src/config.test.ts.

## Testing
Commands run:
- cd web
- 
px vitest run src/config.test.ts
- cd ..
- git diff --check

Actual results:
- Vitest: Test Files 1 passed (1), Tests 4 passed (4).
- git diff --check: no output, exit code 0.

I did not run the full Maven/Java server suite on this Windows host because Maven is not installed and the server targets Java 21 while only Java 17 is available. This PR changes web test code only.

Fixes #4010

Assisted-by: OpenAI:Codex (GPT-5)
Percentage of AI-generated code: 100

### DIFF
--- a/web/src/config.test.ts
+++ b/web/src/config.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { describe, expect, it } from 'vitest';
-import { API_BASE_URL } from './config';
+import { API_BASE_URL, USE_MOCK } from './config';
 
 describe('config API base url', () => {
   it('defaults to a relative /api path when no env override is set', () => {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
LGTM. Clean test addition covering edge cases for diagnostic utilities.

---

PR #4049, opened 2026-09-07
[studio][ISSUE #4048] Avoid data-source cache key collisions

### LINKED ISSUE #4048: [Studio][Bug] Paged data-source cache conflates null with literal filter values
### Describe the bug

The paged data-source inventory builds its Spring cache key by concatenating nullable filter values:

```java
"'page:' + #search + ':' + #type + ':' + #page + ':' + #pageSize"
```

SpEL string concatenation renders a Java `null` as the literal text `null`. Therefore these distinct requests share the same cache key:

- no search filter: `search == null`
- a real search for the text `null`: `search.equals("null")`

After the unfiltered request is cached, the literal search can receive the unfiltered page without invoking the repository. Delimiter-containing values can create similar tuple collisions.

### Steps to reproduce

1. Enable the application's Spring cache configuration.
2. Stub `findDataSources(null, null, 1, 20)` and `findDataSources("null", null, 1, 20)` with different pages.
3. Call the two service methods in that order.
4. The second call returns the first page and the second repository stub is never invoked.

I reproduced this on current `rocketmq-studio` (`36126024`) with a focused Spring caching test; it fails before any production change.

### Expected behavior

Every distinct `(search, type, page, pageSize)` tuple has a distinct cache entry, including `null` versus literal text.

### Proposed fix

Use Spring's default multi-argument `SimpleKey` for the paged method instead of serializing arguments into a delimiter-based string, and add the failing regression test.

I'd like to work on this focused fix.

89799969:
## What changed

- let Spring's default multi-argument key generator identify paged data-source queries
- add a Spring caching regression test that distinguishes an absent search filter from the literal text `null`

## Why

The previous delimiter-based key rendered Java `null` as the text `null`. An unfiltered request and a request searching for that literal value therefore shared one cache entry, so the latter could receive stale unfiltered results without reaching the repository. The default `SimpleKey` preserves argument boundaries and null values.

Fixes #4048

## Verification

- Before the production change: `SettingsServiceCachingTest` failed because the second query returned the first cached page.
- `mvn '-Dtest=SettingsServiceCachingTest,SettingsServiceTest' test`
  - 44 tests, 0 failures, 0 errors, 0 skipped
  - Checkstyle: 0 violations

## Scope

2 files changed; no API or persistence schema changes.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
@@ -218,7 +218,7 @@ public class SettingsService {
         return settingsRepository.findAllDataSources();
     }
 
-    @Cacheable(value = "data-sources", key = "'page:' + #search + ':' + #type + ':' + #page + ':' + #pageSize")
+    @Cacheable("data-sources")
     public PageResult<DataSourceVO> listDataSources(String search, String type, int page, int pageSize) {
         if (page < 1) {
             throw new BusinessException(400, "page must be greater than zero");
--- a/server/src/test/java/org/apache/rocketmq/studio/settings/SettingsServiceCachingTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/settings/SettingsServiceCachingTest.java
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary

This PR correctly fixes a cache key collision bug in `SettingsService.listDataSources()` where the custom SpEL key expression rendered Java `null` as the literal text `"null"`, causing an unfiltered query and a query searching for the literal value `"null"` to share the same cache entry.

Switching to Spring's default `SimpleKeyGenerator` (by removing the custom `key` attribute) is the right approach — `SimpleKey` preserves argument boundaries and handles `null` values correctly.

### Review

- **Correctness** ✅ — The fix directly addresses the root cause. The default key generator properly distinguishes `null` from `"null"`.
- **Test coverage** ✅ — `SettingsServiceCachingTest` is a well-structured regression test that verifies the exact scenario described in #4048. It correctly uses `isSameAs()` for cache-hit verification and `verify()` for repository invocation checks.
- **Scope** ✅ — Minimal change (1 line in production code + new test file). No API or persistence schema changes.
- **Mergeable** ✅ — Clean merge state, no conflicts.

LGTM. Nice catch and clean fix.

---

89799969:
Maintenance note: I noticed that #4132 now addresses the same issue, #4048. This PR was opened earlier and uses the same core approach: relying on the Spring default `SimpleKeyGenerator` to distinguish `null` from the literal `"null"`, with focused regression coverage.

To avoid duplicated review work, I am happy to keep whichever test structure the maintainers prefer, adjust this PR, or close it if #4132 is selected. Please let me know which direction is easiest for the project.

PR #4052, opened 2026-09-07
feat(alert): allow editing maintenance windows after creation

### LINKED ISSUE #3165: Maintenance windows cannot be edited after creation
## Problem

Maintenance windows (alert silences) can only be created and ended. Once a window is created, its schedule and scope cannot be changed.

The current UI exposes only an **End** action:

- If a maintenance window was created too short, the operator must end it and create a new one.
- If a recurring window's time zone, weekdays, or recurrence end is wrong, the operator must delete and recreate it.
- If the scope was too broad (for example, all instances instead of one instance), recreating the window creates a new record and loses the original creator information.

The backend also has no update endpoint: `AlertSilenceController` supports list, create, and delete only.

## Expected behavior

- An administrator should be able to edit an existing maintenance window:
  - start/end time;
  - recurrence type, time zone, weekdays, and recurrence end;
  - domain, rule, instance, and label scope;
  - reason.
- Editing should reuse the same validation as creation:
  - end after start;
  - valid recurrence time zone and bounded window duration;
  - required recurrence end and valid ISO weekdays;
  - bounded labels and reason.
- The existing record ID and creator should be preserved.
- Updates should be recorded in the operation audit log as an alert-silence update.
- The maintenance window list should provide an edit action that pre-fills the existing form.

## Why this matters

During a real maintenance period, plans often move. Extending a window by thirty minutes should be a normal edit, not a delete-and-recreate workflow that risks a gap in alert suppression and loses record provenance.

yyqdbngt:
Fixes #3165

## Summary

Maintenance windows (alert silences) can now be edited after creation instead of
only being ended. The silence form in the System Alerts page gains an **Edit**
action per window; opening it prefills the existing window (times converted from
stored UTC into the window's time zone, scope fields, recurrence settings and
reason) and submits an update.

Backend:

- `UpdateAlertSilenceDTO` carries the window `id` plus the same editable fields
  as `CreateAlertSilenceDTO`.
- `POST /api/alert-silences/update` updates an existing window.
- `AlertSilenceService.update` loads the current record (404 when missing),
  reuses the exact create-time validation (end after start, valid IANA time
  zone, recurrence end required and after the first window, daily/weekly window
  duration caps, ISO weekdays 1-7, bounded labels/reason), and preserves the
  original `id`, `creator` and `gmt_create`.
- Updates are written to the operation audit log as `UPDATE_ALERT_SILENCE` /
  `ALERT_SILENCE`, mirroring the existing create/delete audit entries.
- `MybatisPlusAlertSilenceRepository.update` sets every editable column
  explicitly via `UpdateWrapper`, so clearing a scope (e.g. empty labels,
  `ONCE` recurrence) is persisted as SQL `NULL` instead of being skipped by
  MyBatis-Plus `updateById`'s null-field strategy; `created_by` is never
  touched.

Frontend:

- `updateAlertSilence` API function and mock-mode service support.
- Edit button next to the existing End action; the create/update form is
  prefilled and the modal switches to Update mode, then refreshes the list.

## Why

Maintenance windows are used to suppress notifications during planned work.
Today an admin who mistypes a time, needs a different scope, or wants to
reschedule a recurring window must delete the window and recreate it, losing
the original record id and creator and requiring the whole form to be filled
again. Editing reuses all create-time validation so an invalid edit can never
produce an invalid window.

## Testing

- Backend (`mvn test -Dtest=AlertSilenceServiceTest,AlertSilenceControllerTest,MybatisPlusAlertSilenceRepositoryTest`):
  - update succeeds and keeps the original id/creator; audit records
    `UPDATE_ALERT_SILENCE`.
  - update of an unknown id returns 404; missing id is rejected.
  - invalid recurrence (weekly without weekdays) is rejected by the shared
    validation and nothing is persisted.
  - repository update sets every editable column explicitly (including null
    clears) and never sets `created_by`/`gmt_create`; `findById` restores the
    VO.
  - controller binds `POST /api/alert-silences/update`, returns the updated
    schedule, and rejects a body without `id` with 400.
- Frontend (`vitest run src/api/ops.test.ts src/pages/ops/__tests__/SystemAlertsPage.test.tsx`):
  - the update API posts to `/alert-silences/update`.
  - page test: clicking Edit prefills the weekly window in its IANA time zone,
    submitting calls `updateAlertSilence` with converted UTC instants and the
    original recurrence scope, does not call create, refreshes the list and
    shows the success message.
- `tsc -b tsconfig.app.json` passes.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertSilenceController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertSilenceController.java
@@ -53,6 +53,11 @@ public class AlertSilenceController {
         return Result.ok(silenceService.create(request));
     }
 
+    @PostMapping("/update")
+    public Result<AlertSilenceVO> update(@Valid @RequestBody UpdateAlertSilenceDTO request) {
+        return Result.ok(silenceService.update(request));
+    }
+
     @DeleteMapping("/{id}")
     public Result<Void> delete(@PathVariable Long id) {
         silenceService.delete(id);
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertSilenceService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertSilenceService.java
@@ -52,30 +52,27 @@ public class AlertSilenceService {
     }
 
     public AlertSilenceVO create(CreateAlertSilenceDTO request) {
-        if (request == null || request.getStartsAt() == null || request.getEndsAt() == null) {
-            throw new BusinessException(400, "Silence start and end times are required");
-        }
-        if (!request.getEndsAt().toInstant().isAfter(request.getStartsAt().toInstant())) {
-            throw new BusinessException(400, "Silence end time must be after start time");
-        }
-        if (request.getReason() != null && request.getReason().length() > 512) {
-            throw new BusinessException(400, "Silence reason must not exceed 512 characters");
-        }
+        validateSilence(request);
         RecurrenceConfiguration recurrence = validateRecurrence(request);
-        AlertSilenceVO silence = AlertSilenceVO.builder().domain(request.getDomain())
-                .ruleId(request.getRuleId()).instanceId(trimToNull(request.getInstanceId()))
-                .labels(normalizeLabels(request.getLabels()))
-                .startsAt(LocalDateTime.ofInstant(request.getStartsAt().toInstant(), ZoneOffset.UTC))
-                .endsAt(LocalDateTime.ofInstant(request.getEndsAt().toInstant(), ZoneOffset.UTC))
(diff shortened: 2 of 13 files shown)


RockteMQ-AI (approved):
### Summary
Allows editing maintenance windows (alert silences) after creation instead of only ending them. Full-stack change with comprehensive validation and audit logging. Fixes #3165.

### Findings
- **[Info]** Excellent validation reuse — shares exact create-time validation logic
- **[Info]** Proper audit trail — `UPDATE_ALERT_SILENCE` entries in operation log
- **[Info]** Smart null handling — explicit `UpdateWrapper` ensures cleared fields persist as SQL NULL
- **[Info]** Good security — preserves original `id`, `creator`, `gmt_create`
- **[Info]** Comprehensive test coverage (backend + frontend)

LGTM — well-implemented feature that significantly improves UX for maintenance window management.

---

PR #4053, opened 2026-09-07
fix(consumer): filter consumer group pages by Push/Pop subscription mode on the server

### LINKED ISSUE #3150: Consumer Group Push/Pop filter only filters the current page
## Problem

The Consumer Group page has a Push/Pop filter, but the backend paged endpoint does not support it.

Current behavior:

- `/api/groups/page` accepts only `instanceId`, `clusterId`, `search`, `page`, and `pageSize`.
- `web/src/pages/instance/consumer.tsx` fetches one page and then filters that page in memory with `visibleConsumerGroups(groups, modeFilter)`.
- The repository query in `RocketMQMetadataProvider.listConsumerGroupsPage` also has no subscription-mode condition.

This makes the filter incorrect for real inventories:

- switching to `Push` or `Pop` filters only the rows on the current page, not the whole inventory;
- the returned `total` remains the unfiltered total;
- a page can render zero rows while the pagination still says there are many results;
- later pages are unreachable for the selected mode because they may contain no matching rows;
- CSV export uses the selected mode, so exported data and the visible page no longer describe the same query.

## Expected behavior

- `/api/groups/page` should accept an optional `subscriptionMode` filter.
- Apache-backed groups should filter in the database query, preserving correct totals and paging.
- Cloud providers that return provider-native pages should apply the same filter without breaking existing search/pagination behavior.
- The frontend should send the selected mode to the backend and stop filtering only the current page locally.
- CSV export should continue to use the same backend-visible mode filter.

yyqdbngt:
Fixes #3150

## Summary
`/api/groups/page` now accepts an optional `subscriptionMode` query parameter (`Push`/`Pop`) and applies the Push/Pop subscription-mode filter on the server instead of filtering only the current page in the browser. The total count and pagination are computed on the filtered result set, so paging stays correct when the filter is active.

## Why
The consumer group list previously fetched one page and then filtered it by `subscriptionMode` in memory. When a page contained few (or no) groups of the selected mode, the table appeared empty or under-filled even though matching groups existed on other pages, and the pagination total did not reflect the filter. Moving the filter to the backend makes the page, total and CSV export agree with each other.

## Testing
- Backend:
  - `ConsumerGroupControllerTest` verifies the `subscriptionMode` parameter is propagated from the `/api/groups/page` endpoint to `MetadataService` (and `null` when omitted).
  - `RocketMQMetadataProviderTest` verifies the database query gains a `message_model` condition for `Pop` (equality) and `Push` (NULL or anything but Pop, matching the VO fallback semantics).
  - `MetadataServiceTest` verifies the mode is forwarded to both the cluster metadata provider and the instance provider, and that blank/unknown modes are ignored.
  - `InstanceProviderTest` verifies the shared in-memory cloud-provider fallback filters before pagination and recomputes the total.
  - `ApacheInstanceProviderTest` was updated for the new delegation signature.
- Frontend (`ConsumerPage.test.tsx`): changing the Push/Pop mode filter triggers a fresh server request carrying `subscriptionMode`, and the default request passes `subscriptionMode: undefined`.
- Local type check: `./node_modules/.bin/tsc -b tsconfig.app.json` exits 0.


RockteMQ-AI (approved):
### Summary
Moves consumer group subscription mode filtering from client-side to server-side, with proper null handling and test coverage.

### Findings
- **[Info]** Good refactoring — moves filtering logic to the appropriate layer
- **[Info]** Proper null safety — handles missing `subscriptionMode` field gracefully
- **[Info]** Well-tested — covers both Push and Pop modes with proper assertions

LGTM — clean improvement to data filtering architecture.

---

PR #4069, opened 2026-09-07
[Studio] feat: add Topic delete impact precheck

coder999o:

This PR adds a Topic deletion impact precheck in RocketMQ Studio.

Before deleting a single Topic or selected Topics in batch, the UI now checks visible consumer dependencies, consumer lag, active consumption, message statistics, and broker routes. It summarizes the delete risk before confirmation and blocks deletion for RocketMQ system Topics to avoid accidental removal of internal retry, DLQ, schedule, or system resources.

## Brief changelog

- Add Topic delete impact analysis for consumer groups, lag, traffic, writable routes, and system Topic names.
- Replace the direct delete confirmation with a precheck modal for single and batch Topic deletion.
- Keep normal high-risk Topic deletion confirmable, while blocking detected system Topics.
- Add focused unit and page tests for the precheck analyzer and delete flow.

## Verifying this change

- `git diff --check`
- `npm run test -- --run src/utils/topicDeleteImpact.test.ts src/pages/instance/__tests__/TopicPage.test.tsx`
- `npm run lint -- src/components/TopicDeleteImpactModal.tsx src/utils/topicDeleteImpact.ts src/utils/topicDeleteImpact.test.ts src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx`
- `npm run build`
- `npm run test -- --run`

### DIFF
--- a/web/src/components/TopicDeleteImpactModal.tsx
+++ b/web/src/components/TopicDeleteImpactModal.tsx
@@ -1,0 +1,301 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Alert, Button, Flex, Modal, Space, Spin, Table, Tag, Typography } from 'antd';
+import type { TableColumnsType } from 'antd';
+import {
+  CheckCircleOutlined,
+  DeleteOutlined,
+  ExclamationCircleOutlined,
+  ReloadOutlined,
+  WarningOutlined,
+} from '@ant-design/icons';
+import { formatNumber } from '../utils/format';
+import { tableScrollX } from '../utils/table';
+import type {
+  TopicDeleteImpactAssessment,
+  TopicDeleteImpactSummary,
+  TopicDeleteIssue,
+  TopicDeleteRiskLevel,
+} from '../utils/topicDeleteImpact';
+
+const { Text } = Typography;
+
+const RISK_META: Record<
+  TopicDeleteRiskLevel,
+  {
+    label: string;
+    color: string;
+    alertType: 'success' | 'warning' | 'error';
+    icon: React.ReactNode;
+  }
+> = {
+  low: { label: '低风险', color: 'success', alertType: 'success', icon: <CheckCircleOutlined /> },
+  medium: { label: '需确认', color: 'warning', alertType: 'warning', icon: <WarningOutlined /> },
+  high: {
+    label: '高风险',
+    color: 'error',
+    alertType: 'error',
+    icon: <ExclamationCircleOutlined />,
+  },
+  unknown: { label: '不可确认', color: 'default', alertType: 'warning', icon: <WarningOutlined /> },
+};
+
+const ISSUE_COLOR: Record<TopicDeleteIssue['severity'], string> = {
+  critical: 'error',
+  warning: 'warning',
+  info: 'blue',
+};
+
+interface TopicDeleteImpactModalProps {
+  open: boolean;
+  loading: boolean;
+  deleting: boolean;
+  title: string;
+  assessments: TopicDeleteImpactAssessment[];
+  summary: TopicDeleteImpactSummary | null;
+  error?: string | null;
+  onCancel: () => void;
+  onRetry: () => void;
+  onConfirm: () => void;
+}
+
+const riskTag = (riskLevel: TopicDeleteRiskLevel) => {
+  const meta = RISK_META[riskLevel];
+  return (
+    <Tag color={meta.color} icon={meta.icon}>
+      {meta.label}
+    </Tag>
+  );
+};
+
+const issueTags = (issues: TopicDeleteIssue[]) => (
+  <Space size={[4, 4]} wrap>
+    {issues.map((item) => (
+      <Tag key={item.code} color={ISSUE_COLOR[item.severity]} title={item.description}>
+        {item.title}
+      </Tag>
+    ))}
+  </Space>
+);
+
+const metricText = (label: string, value: React.ReactNode, strong = false) => (
+  <span>
+    <Text type="secondary">{label}</Text>
+    <Text strong={strong} style={{ marginLeft: 4, fontVariantNumeric: 'tabular-nums' }}>
+      {value}
+    </Text>
+  </span>
+);
+
+const metricBox = (label: string, value: React.ReactNode, helper?: React.ReactNode) => (
+  <div
+    style={{
+      border: '1px solid #f0f0f0',
+      borderRadius: 6,
+      padding: '10px 12px',
+      minWidth: 132,
+      flex: '1 1 132px',
+      background: '#fafafa',
+    }}
+  >
+    <Text type="secondary" style={{ display: 'block', fontSize: 13 }}>
+      {label}
+    </Text>
+    <Text strong style={{ fontSize: 20, fontVariantNumeric: 'tabular-nums' }}>
+      {value}
+    </Text>
+    {helper && (
+      <div style={{ marginTop: 2 }}>
+        <Text type="secondary" style={{ fontSize: 13 }}>
+          {helper}
+        </Text>
+      </div>
+    )}
+  </div>
+);
+
+const summaryMessage = (summary: TopicDeleteImpactSummary) => {
+  if (!summary.canDelete) return `发现 ${summary.blockedTopics.length} 个系统 Topic，已阻止删除。`;
+  if (summary.riskLevel === 'high')
+    return `发现 ${summary.highRiskTopics} 个高风险 Topic，删除前需要确认消费者和消息影响。`;
+  if (summary.riskLevel === 'unknown') return '部分依赖信息加载失败，删除影响不可完全确认。';
+  if (summary.riskLevel === 'medium')
+    return `发现 ${summary.mediumRiskTopics} 个需要确认的 Topic。`;
+  return '未发现明显消费者或路由依赖。';
+};
+
+const TopicDeleteImpactModal = ({
+  open,
+  loading,
+  deleting,
+  title,
+  assessments,
+  summary,
+  error,
+  onCancel,
+  onRetry,
+  onConfirm,
+}: TopicDeleteImpactModalProps) => {
+  const columns: TableColumnsType<TopicDeleteImpactAssessment> = [
+    {
+      title: 'Topic',
+      dataIndex: 'topicName',
+      key: 'topicName',
+      width: 220,
+      render: (topicName: string) => (
+        <Text strong ellipsis={{ tooltip: topicName }} style={{ display: 'block' }}>
+          {topicName}
+        </Text>
+      ),
+    },
+    {
+      title: '风险',
+      dataIndex: 'riskLevel',
+      key: 'riskLevel',
+      width: 110,
+      render: riskTag,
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (changes requested):
## Review

### Summary
Adds a delete-impact precheck modal that analyzes consumer, lag, traffic, routes, and protected Topic names before single or batch deletion.

### Blocking issues
- **`web/src/utils/topicDeleteImpact.ts:157-167`**: `ConsumerGroupInfo` does not define `metricsAvailable`, so these accesses do not typecheck. Either add and populate this field in the metadata contract, or remove this unsupported diagnostic.

- **`web/src/utils/topicDeleteImpact.ts:105-111`**: System Topic protection is incomplete. RocketMQ uses the lowercase `rmq_sys_` prefix (for example, revive and broker-membership Topics), while this only checks uppercase `RMQ_SYS_`. It also misses `CHECKPOINT_TOPIC`. These internal Topics remain deletable, contradicting the safety guarantee. Reuse/align with RocketMQ’s authoritative system-topic definitions and add tests.

- **`web/src/pages/instance/topic.tsx:1630-1638`**: Batch deletion now requires every selected key to exist in the current `topics` page. Since selections persist across pagination, valid selections from another page are rejected, regressing the previous batch-delete flow. Preserve selected Topic records or fetch their metadata for the precheck.

### Convention concern
- **`web/src/components/TopicDeleteImpactModal.tsx:33-286`**: New user-visible strings are hard-coded Chinese rather than using the existing `useLang()`/translation mechanism.

### Suggested tests
Add coverage for lowercase `rmq_sys_`, `CHECKPOINT_TOPIC`, and selections spanning multiple pages.

### Verdict
**REQUEST_CHANGES**

RockteMQ-AI (approved):
### Summary
Adds a Topic delete impact precheck modal that analyzes consumer groups, traffic metrics, and routing dependencies before allowing deletion. Well-structured with:

- Clear risk assessment model (low/medium/high/unknown) with appropriate UI indicators
- System topic protection (blocks deletion of system topics)
- Comprehensive metrics display (consumer groups, lag, TPS, message count, writable routes)
- Proper error handling with retry capability
- Good test coverage

The implementation correctly separates concerns: utility functions in `topicDeleteImpact.ts`, UI in `TopicDeleteImpactModal.tsx`, and integration in `topic.tsx`.

LGTM.

---

lizhimins (changes requested):
This is a genuinely useful feature and several things are done well — the stale-response guard is correct, truncation is disclosed, and the scoring shape matches trunk convention. But it cannot be merged as submitted: two of the numbers it presents authoritatively are wrong for every topic, the one real safety control lives only in the browser, and the error path is unreachable. Details with evidence below.

## Blocking: the impact figures are conclusively wrong

### A. "当前 TPS" and "今日消息" are always 0, and the modal states them as fact

`TopicVO.messageCount`, `tps` and `consumerGroupCount` (`server/.../TopicVO.java:36-38`) are primitive fields, and **no provider on trunk ever populates them** — not `RocketMQMetadataProvider.toTopicVO:178-192`, not `AliyunConverters.toTopicVO:116-127`, not `TencentInstanceProvider.toTopic:913-925`. They are always 0.

Consequences in this PR:
- `web/src/utils/topicDeleteImpact.ts:169-170` reads them, so both are always 0.
- The `RECENT_TRAFFIC` (`:243`) and `RETAINED_MESSAGES` (`:253`) findings can therefore **never fire** — they are dead branches.
- `TopicDeleteImpactModal.tsx:272-273` renders "当前 TPS 0 / 今日消息 0" as an authoritative statement. An operator reads that as "this topic has no production traffic, safe to delete" — for every topic, including busy ones. Nothing discloses that the figure is unpopulated rather than measured.

This is the same failure mode we returned #3129 for: presenting a number the code cannot actually compute as though it were a measurement. Either populate `TopicVO` server-side first (a separate PR), or remove both boxes and both findings.

### B. Unknown consumer lag is rendered as zero lag

`num()` applies `Math.max(0, value)` (`topicDeleteImpact.ts:113-114`), used at `:160` and `:166`. But the backend **deliberately** emits `diffTotal = -1` while leaving `metricsAvailable` true (`RocketMQMetadataProvider.java:521-556`) to mean "lag unknown", and `ConsumerLagResolver.java:25-26` names `Math.max(0, …)` as the exact bug it exists to avoid.

Trunk already has the right primitives: `web/src/utils/consumerLag.ts` exports `UNKNOWN_LAG`, `isLagAvailable` and `formatLag`, used at `consumer.tsx:652-654` and `consumerGroupDiagnostics.ts:141-144`, and guarded by an existing test at `TopicPage.test.tsx:942`. This PR imports none of them, so a topic whose lag cannot be determined is reported as lag 0 in a pre-deletion safety dialog. Use the shared util.

### C. The system-topic list does not match the kernel

`topicDeleteImpact.ts:105` matches uppercase `RMQ_SYS_`. RocketMQ 5.5.0's `TopicValidator` uses **lowercase** `rmq_sys_` (verified against `rocketmq-common-5.5.0-0.jar`), and so does the backend filter at `RocketMQMetadataProvider.java:169`. The list also omits `CHECKPOINT_TOPIC` and `SYNC_BROKER_MEMBER_GROUP_PREFIX`, and treats `TBW102` as a prefix when it is an exact name.

This one is moot in practice — `listTopicsPage:168-171` already excludes system topics, so the "blocks system Topics" path is unreachable from the table — which is exactly why it should be deleted rather than corrected. See the next section.

## Blocking: the only real control is browser-side

The precheck can block the delete button, but `MetadataService.deleteTopic` (`server/.../MetadataService.java:142-149`) has **no system-topic guard**, while `createTopic` (`:119`) rejects them via the existing `SystemTopicFilter`. So the "block" at `topicDeleteImpact.ts:300` protects nothing: any direct API caller deletes a system topic freely. A guard that only exists in the UI is not a guard.

Please add the rejection in `MetadataService.deleteTopic` using `SystemTopicFilter`, mirroring `:119`. Once that exists, the frontend list in item C can be deleted entirely.

## Blocking: the error path is unreachable, and failure leaves the delete button enabled

The outer `catch` at `web/src/pages/instance/topic.tsx:704-707` is dead code. Every inner await already has its own catch, and `analyzeTopicDeleteImpact` is a pure function, so the `Promise.all` cannot reject. Therefore:

- `deleteImpactError` is permanently `null`.
- The error Alert and retry button (`TopicDeleteImpactModal.tsx:231-242`) and `onRetry` (`topic.tsx:2157`) can never render or fire.
- `Boolean(error)` never gates anything.

And the failure mode that matters: when every lookup fails, each topic resolves to `unknown` but `canDelete` stays `true`, so **确认删除 remains enabled** — the dialog degrades to "we could not determine anything" while still allowing the destructive action. There is also no `if (deletingTopics) return` re-entry guard at `:741`.

Either remove the dead error UI or wire it genuinely, and decide explicitly whether a failed precheck may enable the confirm button. Our position: it may not.

## Blocking: batch precheck fans out 200 concurrent calls

`topic.tsx:664-699` issues 2 requests per selected topic, and the page size options go up to 100 (`topic.tsx:1658`), so a full-page selection produces **200 concurrent HTTP calls**. Each one makes the backend run `queryTopicConsumeByWho` plus up to 100 **serial** `examineConsumeStats` and 100 `examineConsumerConnectionInfo` calls (`RocketMQMetadataProvider.java:498-568`). It also ignores the `consumersByTopic` / `routesByTopic` caches that are already populated at `topic.tsx:356-357`.

This belongs behind one endpoint: `POST /topics/delete-precheck` returning `Result<TopicDeleteImpactVO>`. That also fixes item A properly, since the server can compute real figures. If you keep it client-side for now, at minimum add a concurrency limit and a selection-size cap.

## Must fix: project rules

1. **`fontSize: 13`** at `TopicDeleteImpactModal.tsx:116` and `:124`. The project minimum is 14px, and trunk currently has **zero** sub-14px font sizes anywhere in `web/src` — this would be the first regression. Change both to 14.
2. **Table does not fit its modal.** Declared column widths total **1070px** (`TopicDeleteImpactModal.tsx:159-204`) inside `width={960}` (`:222`), leaving roughly 912px of table body — a guaranteed horizontal scrollbar, which the project forbids at normal width. Trunk sizes modals to their content (`topic.tsx:1622` uses 1080, `consumer.tsx:2462` uses 1200). Widen the modal to at least 1120, or trim columns, and compute `scroll.x` via `tableScrollX(columns)` rather than a literal.
3. **The irreversibility warning was lost.** Trunk's `modal.confirm` (topic.tsx `:638-654`) told the operator 此操作不可撤销. There are zero occurrences of that text in the new UI. A precheck dialog that leads to a destructive action must still say the action is irreversible — restore it.

## Should fix: duplication and inconsistency

- `routeWritable` (`topicDeleteImpact.ts:116-119`) is a verbatim copy of the unexported `inferWritable` in `web/src/utils/topicRouteDiagnostics.ts:139-142` — **and the two disagree**: trunk requires `writeQueues > 0` (`:441-443`). Export and reuse the existing one; a product should not compute "writable route" two different ways.
- `:219`, `:249`, `:259` use `toLocaleString()` directly, bypassing `formatNumber` from `web/src/utils/format.ts` (which pins `en-US`). The same value renders two different ways inside one modal. (You did import `formatNumber` correctly at `TopicDeleteImpactModal.tsx:27`, so this is only about these three call sites.)
- Instance resolution is inconsistent: single delete uses `topic.instanceId || selectedInstanceId` (`:748`, matching `rebuildTopic` at trunk `:540`), but the batch path still uses bare `selectedInstanceId` (`:753`). Pick one rule and apply it to both.

## Tests

Six new `it()` blocks (five in `topicDeleteImpact.test.ts` at `:70,77,105,118,134`; one in `TopicPage.test.tsx:461`). The body does not overclaim a count, which we appreciate.

- `topicDeleteImpact.test.ts:79` hand-feeds `messageCount: 1200, tps: 12` into the fixture. Since no provider ever sets those fields, this test **certifies bug A** — it passes only because it invents data production cannot produce. Please remove those fields from the fixture so the test reflects reality.
- `:126-127` is tautological (it echoes the inputs back as the expectation).
- Uncovered and worth adding: `diffTotal: -1` (bug B), `metricsAvailable: false` (`:162-164`), the truncation path (`:222`), the cloud-provider path (`:263`), the perm fallback (`:117`), the summary sums, and the confirm-button disabled states.

## Credit where due

So the above is not read as a blanket negative:

- The stale-response guard is present and correct (`topic.tsx:418,658,701-711,733`).
- You fetch a single page (`topic.tsx:674`; 100 is the server cap at `MetadataService.java:181-183`) rather than walking every page, and you **disclose** the truncation to the user (`topicDeleteImpact.ts:222-231`). That is the right instinct and it is rare in this area — #3129 computed a "global" rate from a post-filter slice without saying so.
- The `-48` deletions are pure relocation, not lost functionality. We checked line by line: dropping the `App` import and `App.useApp()` (trunk `:42`, `:387`) is correct since the `Modal` JSX is still used at `:1618,1670,1768,1823,1963`; both `modal.confirm` paths were faithfully rebuilt at `:745-752` and `:1631-1637` + `:753-763`, preserving all three message branches, `setSelectedRowKeys(failed)` and the conditional reload. The only genuine loss is the irreversibility text called out above.
- The issue-code / severity / score shape matches the existing `utils/aclRiskDiagnostics.ts` and `utils/topicRouteDiagnostics.ts` convention, so keeping the scoring heuristic in the browser is fine. It is the data gathering and the authority of the numbers that need to move.

One correction to a claim that may reach you from elsewhere: the batch selection cannot span pages, so there is no cross-page regression to guard against. `useSelection.js:96-115` builds `keyEntities` from the current page only and `rc-tree/es/utils/conductUtil.js:177-183` drops keys without an entity, which makes the guard at `topic.tsx:1633` dead code rather than a fix.

Given the number of items, our suggestion is to split this: land the server-side `deleteTopic` system-topic guard and a real `POST /topics/delete-precheck` endpoint first, then rebuild the dialog on top of figures the server actually returns. We would rather merge a smaller version of this than not merge it at all.

lizhimins (changes requested):
Two blocking issues, plus the branch is now conflicting and needs a rebase.

`topic.tsx:664` runs `Promise.all` over every topic with no concurrency bound, so a namespace with a few hundred topics fires a few hundred parallel requests from a modal. Each of those calls `getTopicConsumerPage(name, instanceId, 1, 100)` (`:674`) and treats the first 100 consumers as the complete set, so the impact figure is wrong for any topic with more than 100 consumers — and it is wrong silently, since nothing in the UI marks the result as partial.

Please bound the fan-out, paginate or explicitly cap with a visible "partial" marker, and rebase.

PR #4081, opened 2026-09-07
feat(certs): export the Kubernetes certificate inventory as CSV

yyqdbngt:
## Summary

Adds an **Export CSV** button to the K8s certificate page (`web/src/pages/cluster/certs.tsx`). Clicking it downloads the **currently filtered** certificate inventory (`filteredCerts`, i.e. the rows matching the existing search box and type filter) as `rocketmq-k8s-certs-YYYY-MM-DD.csv`.

The exported columns match the page table's actual fields:

`ID, K8s ID, Cluster, Type, Issuer, Not Before, Not After, Days Remaining, Status`

The export is metadata only — `certPem`/`keyPem` private key material is deliberately never included (covered by a test that seeds PEM content and asserts it does not appear in the CSV). The button is disabled when no certificate matches the current filters.

## Why

Operators need an offline, shareable view of the Kubernetes certificate inventory (e.g. for expiry reviews or audits). The page already supports searching and type filtering, so exporting exactly what the filters show lets users capture a precise subset without exposing private key material.

## Testing

- Local: `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` → exit 0.
- Server (Node 20): `./node_modules/.bin/vitest run src/pages/cluster/__tests__/K8sCertsPage.test.tsx` → 10 passed, including:
  - `exports the filtered certificate metadata as CSV`
  - `never exports PEM private key material`
  - `disables the CSV export button when no certificate matches`

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -30,6 +30,7 @@ const translations: Record<string, Record<Lang, string>> = {
   'nav.dlq': { zh: '死信队列', en: 'Dead Letter Queue' },
   'nav.clusterOps': { zh: '集群管理', en: 'Cluster Management' },
   'nav.certs': { zh: 'K8s 证书配置', en: 'K8s Certificate Config' },
+  'certs.exportCsv': { zh: '导出 CSV', en: 'Export CSV' },
   'nav.rocketmqCluster': { zh: 'RocketMQ 集群', en: 'RocketMQ Cluster' },
   'nav.clients': { zh: '客户端连接', en: 'Client Connections' },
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Adds CSV export for the K8s certificate inventory. Like the ACL export, this is security-conscious — the test explicitly verifies that PEM material is NOT included in the export, only metadata (ID, K8s ID, cluster, type, issuer, validity dates, days remaining, status).

The export respects the current search filter, only exporting visible/filtered results. Proper blob URL cleanup.

LGTM.

---

PR #4082, opened 2026-09-07
feat(acl): export the ACL user inventory as CSV

yyqdbngt:
## Summary

Adds an "Export CSV" (导出 CSV) button to the user management tab of the ACL page. Because the backend exposes a full, non-paginated user list endpoint (`listAclUsers`), the button fetches the complete user inventory of the selected instance on click and downloads it as `rocketmq-acl-users-YYYY-MM-DD.csv`.

Exported columns mirror the user table fields: Username, Admin, Clusters, Created. Credentials are never exported — `accessKey`/`secretKey` are intentionally excluded from the CSV columns.

## Why

Operators need an offline, complete ACL user inventory for audits and migration planning. Exporting the full list via `listAclUsers` (instead of only the currently paginated page) guarantees the CSV matches the whole inventory; excluding credential fields keeps the export safe to share.

## Testing

- Local: `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` → exit 0
- Server (Node 20): `./node_modules/.bin/vitest run src/pages/instance/__tests__/AclPage.test.tsx` → 25 passed
  - exports the full ACL user inventory as CSV without credentials
  - shows an error toast when the full ACL user export fails

### DIFF
--- a/web/src/i18n/translations.ts
+++ b/web/src/i18n/translations.ts
@@ -1047,6 +1047,7 @@ const translations: Record<string, Record<Lang, string>> = {
   'acl.allDecisions': { zh: '全部决策', en: 'All Decisions' },
   'acl.totalRules': { zh: '共 {n} 条规则', en: '{n} rules total' },
   'acl.totalUsers': { zh: '共 {n} 个用户', en: '{n} users total' },
+  'acl.exportCsv': { zh: '导出 CSV', en: 'Export CSV' },
   'acl.editRule': { zh: '编辑规则', en: 'Edit Rule' },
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Adds CSV export for the ACL user inventory. Security-conscious: the export correctly excludes `secretKey` and `accessKey` fields, only exporting safe metadata (username, admin status, clusters, creation time).

Test verifies that secret keys are NOT included in the CSV output, which is the right approach. The export fetches the full user list (not just the current page), ensuring a complete export.

LGTM.

---

PR #4084, opened 2026-09-07
feat(message): copy message properties as JSON

yyqdbngt:
## Summary

Add a "复制属性 JSON" (copy properties as JSON) action to the message detail
modal on the Message page. It serializes the selected message's `properties`
object with `JSON.stringify(..., null, 2)` and writes it to the clipboard via
`navigator.clipboard.writeText`, with a `textarea` + `document.execCommand`
fallback for non-secure contexts. The button is disabled (with a "该消息无属性"
hint) when the message has no properties.

## Why

Operators inspecting a message currently have no way to grab its properties
as a machine-readable payload; the message body is copyable but the
key/value properties are not rendered on the detail view. This makes pasting
properties into a ticket or script one click instead of retyping them.

## Testing

- `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` — passes (exit 0)
- `./node_modules/.bin/vitest run src/pages/instance/__tests__/MessagePage.test.tsx`
  (Node 20 build server) — **12/12 passed**, including two new cases:
  - clicking the action writes `{"KEYS":"abc","TAGS":"vip"}` to the clipboard;
  - the action is disabled and shows the no-properties hint for messages with
    an empty `properties` object.

### DIFF
--- a/web/src/pages/instance/__tests__/MessagePage.test.tsx
+++ b/web/src/pages/instance/__tests__/MessagePage.test.tsx
@@ -356,4 +356,51 @@ describe('Message page query history', () => {
       );
     });
   });
+it('copies message properties as JSON from the detail modal', async () => {
+    const user = userEvent.setup();
+    const writeText = vi.fn().mockResolvedValue(undefined);
+    Object.defineProperty(navigator, 'clipboard', {
+      writable: true,
+      value: { writeText },
+    });
+    messageServiceMocks.queryMessages.mockResolvedValue([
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds a "Copy Properties JSON" button to the message detail modal, allowing users to copy message properties to clipboard. The button is properly disabled when no properties exist, with a helpful hint text.

Clean implementation using `navigator.clipboard.writeText`. Tests cover both the copy action and the disabled state.

LGTM.

---

PR #4085, opened 2026-09-07
feat(dashboard): export the cluster summary table as CSV

yyqdbngt:
# feat(dashboard): export the cluster summary table as CSV
## Summary

Add a CSV export button to the cluster summary table on the dashboard page. The export uses the existing `buildCsv`/`downloadCsv` helpers and writes the clusters of the currently selected instance (`selectedInstanceId`) to `rocketmq-clusters-<instanceId>-YYYY-MM-DD.csv` (the instance part falls back to `all` when no instance filter is active). The button is disabled while the cluster list is empty.

Exported columns mirror the table: name, status, type, version, brokers, proxies, topics, groups, TPS in, TPS out.

## Why

Users monitoring cluster health currently have to read the on-screen table manually. Exporting the summary as CSV makes the data available for offline analysis, reporting and diffing across instances.

## Testing

- `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` passes (exit 0).
- `./node_modules/.bin/vitest run src/pages/home/__tests__/DashboardPage.test.tsx` passes; new cases cover:
  - exporting the current instance's clusters with the expected CSV content and `rocketmq-clusters-instance-a-<date>.csv` filename,
  - using `rocketmq-clusters-all-<date>.csv` when no instance is selected,
  - the export button being disabled when there are no clusters.


RockteMQ-AI (approved):
### Summary
Adds CSV export for the cluster summary table on the Dashboard page. Uses the shared `buildCsv`/`downloadCsv` utility. The export includes all visible columns and generates a filename with the instance name and date.

Good test coverage including the edge case of no instance selected (falls back to "all"). Proper blob URL cleanup with `revokeObjectURL`.

LGTM.

---

yyqdbngt:
Rebased onto the current `rocketmq-studio` head (`1e8cc591`) and re-verified.

The only conflict was the import block in `web/src/pages/home/dashboard.tsx`: the base had gained `import type { ColumnsType } from 'antd/es/table';`, while this branch added `DownloadOutlined` to the `@ant-design/icons` import. Both sides were kept.

Two test expectations had drifted with the base and are fixed in the same commit:

- the cluster name now also renders in the traffic-insights card, so the render wait uses `findAllByText('apache-cluster')`;
- the instance filter is addressed by its accessible name `实例筛选` (the JSON `aria-label` of the antd `Select`), matching the other tests in the file.

Verification on the rebased head:

- `npx vitest run src/pages/home/__tests__/DashboardPage.test.tsx` → 9 passed, including `exports the cluster summary table as CSV for the selected instance` and `uses "all" in the CSV filename when no instance is selected`.
- `npx eslint` on `dashboard.tsx` and the test file → clean.

PR #4086, opened 2026-09-07
feat(topic): show the last successful load time

yyqdbngt:
# feat(topic): show the last successful load time
## Summary
Show the time of the last successful topic list load in the Topic page header. The page records `Date.now()` only when `loadTopicPage` succeeds for the latest request; a failed load (or a stale response) leaves the previous successful time untouched. The timestamp is rendered as "最近更新 HH:mm:ss" (`common.lastUpdated` i18n key, zh/en) next to the page title, using the local clock of the browser.

Note: the Topic page has no dedicated refresh button (the list reloads when instance/search/type/pagination change), so the indicator is placed in the header/title area via the existing `PageHeader` `extra` slot.

## Why
Users cannot tell whether the topic list on screen is current. Surfacing the last successful load time next to the title makes staleness visible without adding a new refresh control or changing the existing reload triggers.

## Testing
- `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` passes (exit 0).
- `./node_modules/.bin/vitest run src/pages/instance/__tests__/TopicPage.test.tsx` passes: 24/24 (run on the Linux test server with Node 20).
- New cases:
  - `shows the time of the last successful list load in the header` – with `vi.setSystemTime`, a successful load renders "最近更新 09:15:42".
  - `keeps the previous successful load time when a reload fails` – after the clock advances, a rejected reload leaves "最近更新 09:15:42" visible and never shows the later time.


RockteMQ-AI (approved):
### Summary
Adds a "Last Updated" timestamp to the Topic page header. Consistent with the other freshness-timestamp PRs. Tests properly cover the failure-preservation behavior.

LGTM.

---

PR #4087, opened 2026-09-07
feat(consumer): show the last successful load time

yyqdbngt:
# feat(consumer): show the last successful load time

## Summary
Adds a "最近更新 HH:mm:ss" (Last Updated) indicator to the consumer group page. The
page records `Date.now()` every time `loadConsumerGroupPage` completes successfully and
renders the formatted time next to the auto-refresh button in the filter bar. The
indicator only appears after the first successful load.

## Why
Operators refreshing the consumer group list (including the 2s auto refresh) cannot
tell whether the rows they see are fresh or stale. Showing the wall-clock time of the
last successful load makes the data staleness visible without any extra round trip.

## Testing
- `tsc -b tsconfig.app.json` exits 0.
- Added two cases to `web/src/pages/instance/__tests__/ConsumerPage.test.tsx`:
  - after a successful list load the header shows the recorded time
    (`vi.setSystemTime` controls `Date.now()`);
  - when a later reload (search-triggered) rejects, the indicator keeps the previous
    successful time and does not show the newer clock value.
- Full file suite passes on the CI-equivalent server (`vitest run
  src/pages/instance/__tests__/ConsumerPage.test.tsx`): 33 passed.


RockteMQ-AI (approved):
### Summary
Adds a "Last Updated" timestamp next to the auto-refresh button on the Consumer page. Uses the same `formatClock` pattern as sibling PRs. Tests cover both success and failure scenarios.

LGTM.

---

PR #4088, opened 2026-09-07
feat(deliveries): show the last successful load time

yyqdbngt:
# feat(deliveries): show the last successful load time

## Summary

The notification deliveries page (告警投递) now records `Date.now()` whenever the
delivery list is fetched successfully and shows a "最近更新 HH:mm:ss" ("Last Updated")
timestamp in the page header, next to the title/subtitle. The time is rendered with the
new `common.lastUpdated` i18n key (zh: 最近更新, en: Last Updated).

Failed loads do not touch the timestamp, so the last successful load time stays visible
and keeps its previous value while the user retries.

## Why

Operations users repeatedly refresh the deliveries list to watch retries and want to
know how stale the data on screen is. A visible freshness marker removes the guesswork
without adding a new polling mechanism.

## Testing

- `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` → exit 0.
- `./node_modules/.bin/vitest run src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx`
  → 5/5 passed (added: "shows the time of the last successful list load in the header"
  and "keeps the previous successful load time when a reload fails"; time controlled with
  `vi.setSystemTime`, restored with `vi.useRealTimers()`).


RockteMQ-AI (approved):
### Summary
Adds a "Last Updated" timestamp to the Notification Deliveries page header, following the same pattern as the audit/consumer/topic pages. Properly preserves the previous timestamp on failed reloads.

Clean implementation with good test coverage. The `afterEach` timer cleanup and the import fix for `afterEach` from vitest are good touches.

LGTM.

---

PR #4089, opened 2026-09-07
feat(audit): show the last successful load time

yyqdbngt:
# feat(audit): show the last successful load time

## Summary

The audit log page (审计日志) now records `Date.now()` whenever the record list
fetch succeeds and shows a "最近更新 HH:mm:ss" ("Last Updated") timestamp in the
page header, next to the title/subtitle. The text uses the new `common.lastUpdated`
i18n key (zh: 最近更新, en: Last Updated).

Failed record loads do not update the timestamp, so the last successful load time
is preserved on the screen while the user retries or adjusts filters.

## Why

Audit lists are queried per filter combination and can be refreshed implicitly by
filter changes; operators currently have no way to tell how old the rows on screen
are. A lightweight freshness marker in the header answers that without extra
network calls.

## Testing

- `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` → exit 0.
- `./node_modules/.bin/vitest run src/pages/ops/__tests__/AuditPage.test.tsx`
  → 10/10 passed (added: "shows the time of the last successful record load in the
  header" and "keeps the previous successful load time when a reload fails"; time
  controlled with `vi.setSystemTime`, restored with `vi.useRealTimers()`).


RockteMQ-AI (approved):
### Summary
Adds a "Last Updated" timestamp to the Audit page header that records the time of the most recent successful data load. The timestamp is preserved when a reload fails, giving users visibility into data freshness.

Well-structured change with proper test coverage for both success and failure paths. The `vi.useRealTimers()` cleanup in `afterEach` is a good practice to avoid timer leakage between tests.

**Minor suggestion:** The `formatClock` helper is duplicated across several pages (audit, deliveries, consumer, topic). Consider extracting it to a shared utility (e.g. `utils/format.ts`) in a follow-up to reduce duplication.

LGTM.

---

PR #4096, opened 2026-09-08
feat(clients): show the last successful load time

yyqdbngt:
## Summary

Show the last successful connection-list load time ("最近更新 HH:mm:ss", i18n key `common.lastUpdated`) in the Clients page toolbar, next to the export button. A successful `listConnections` response (initial load, nameserver switches and retries) stamps `Date.now()`; failed loads do not update the stamp, so the previous successful value is kept. The indicator is hidden until the first successful load.

## Why

Users have no way to tell how fresh the connection inventory is when polling or manually switching nameservers. Surfacing the last successful load time makes staleness visible at a glance without adding any new requests.

## Testing

- Server Node 20: `./node_modules/.bin/vitest run src/pages/cluster/__tests__/ClientsPage.test.tsx` → all tests green.
- New tests (no existing test covered freshness):
  - shows the last successful connection load time and updates it after a reload;
  - keeps the previous load time when a connection reload fails.


RockteMQ-AI (approved):
LGTM. Test coverage improvement.

PR #4098, opened 2026-09-08
feat(proxy): show the last successful load time

yyqdbngt:
## Summary

Show the last successful proxy node-list load time ("最近更新 HH:mm:ss", i18n key `common.lastUpdated`) in the Proxy page header toolbar, next to the refresh button. A successful `loadProxyNodes` cycle (initial load and manual refresh) stamps `Date.now()` after the home data is applied; failed loads keep the previous successful value. Stale superseded requests do not update the stamp.

## Why

The proxy topology is loaded asynchronously and refreshed manually, but nothing tells the operator when the visible list was last successfully loaded. The freshness indicator makes that state explicit without extra requests.

## Testing

- Server Node 20: `./node_modules/.bin/vitest run src/pages/studio/__tests__/Proxy.test.tsx` → all tests green.
- New tests (no existing test covered freshness):
  - shows the last successful proxy list load time;
  - updates the load time after a successful refresh;
  - keeps the previous load time when a refresh fails.


RockteMQ-AI (approved):
LGTM. Clean implementation — shows last successful proxy list load time in the header, mirroring the instance inventory pattern. Correctly preserves the previous timestamp on failed loads. Good test coverage.

---

PR #4099, opened 2026-09-08
feat(broker): filter the broker topology by name and address

yyqdbngt:
## Summary

Add a search box above the broker topology table that filters the currently loaded brokers by name or address (trim + lowercase substring match, following the certs page `normalizedSearch` pattern). The filter is client-side only — it never refetches from the API. Export exports the filtered rows (what you see is what you export), and the export button is disabled when the filter yields zero broker rows.

## Why

With many brokers in a topology it is hard to locate one instance. Filtering the loaded rows by name/address keeps the change purely front-end and consistent with the existing search UX on the certificates and clients pages.

## Testing

- Server Node 20: `./node_modules/.bin/vitest run src/pages/studio/__tests__/BrokerCluster.test.tsx` → all tests green.
- New tests (existing tests did not cover filtering):
  - filters the broker table by broker name;
  - filters the broker table by broker address;
  - exports only the filtered broker rows.


RockteMQ-AI (approved):
LGTM. Clean implementation — adds a broker name filter to the topology export, correctly filtering both the CSV rows and the header count. Good test coverage including the filter interaction.

---

PR #4102, opened 2026-09-08
feat(message): copy trace steps as text

yyqdbngt:
## Summary

Add a "复制轨迹" (copy trace) button to the message trace tab. It renders
the loaded `TraceRecord` steps as a multi-line text (step name, time and
status/details) and copies it to the clipboard through the new shared
`copyTextToClipboard` util (`navigator.clipboard.writeText` with a hidden
`textarea` + `document.execCommand('copy')` fallback). The button is
disabled while no trace data is loaded.

## Why

Operators debugging a delivery problem often need the trace steps in a
ticket or chat. Previously the steps could only be read on screen, so this
adds one-click extraction of the same data.

## Testing

- `cd web && ./node_modules/.bin/tsc -b tsconfig.app.json` — passes (exit 0).
- `./node_modules/.bin/vitest run src/pages/instance/__tests__/MessagePageAsyncState.test.tsx`
  (Node 20 build server) — 23/23 passed, including the new case asserting the
  copied text contains the rendered trace steps and the disabled state with
  no trace data.


RockteMQ-AI (approved):
LGTM. Clean implementation — adds a "Copy trace" button to the message trace tab, rendering steps as readable text via the shared clipboard utility. Good test coverage including the fallback path.

---

PR #4113, opened 2026-09-08
fix(acl): read plain-access rules by exact principal match

zjncs:
### Motivation

`MybatisPlusAclRepository.toPlainAccessConfig` rebuilds each account's plain-access config (`topicPerms` / `groupPerms` / default perms) by reusing `ruleQuery(...)`, whose `principal` condition is a substring `LIKE` — that helper exists for the rule **search** endpoint (`findRulePage`). Access keys are free-form strings (`@NotBlank` only), so when one account's key contains another's (`svc-a` vs `svc-a-v2`), each account's view absorbs the other's rules.

The write path (`upsertPlainAccessRules`) already deletes rules with an **exact** `eq("principal", accessKey)` match, so the read/write asymmetry is a straight copy-paste reuse of the search helper:

```java
// read (LIKE '%accessKey%')
ruleMapper.selectList(ruleQuery(user.getAccessKey(), null, null, null, null))
// write (exact)
ruleMapper.delete(new QueryWrapper<RmqAclRule>()
        .eq("principal", config.getAccessKey())
        .eq("acl_version", "2.0"));
```

Concretely: `GET /api/acl/cluster-config` → `AclService.examineBrokerClusterAclConfig` → `toPlainAccessConfig` returns `svc-a`'s account with `svc-a-v2`'s rules included; the edit modal (`web/src/pages/instance/acl.tsx` `openEditPlainModal`) seeds from that view and `handlePlainSubmit` saves the polluted set back — persisting the wrong rules under the wrong account.

### Modifications

- `toPlainAccessConfig` now queries rules with an exact `eq("principal", accessKey)` match (plus the same deterministic ordering), mirroring the delete in `upsertPlainAccessRules`.

### Verification

- New test `examineShouldNotAbsorbRulesOfAccountsWhoseAccessKeysOverlap` in `MybatisPlusAclRepositoryTest`: the `ruleMapper.selectList` stub simulates SQL semantics (a substring `LIKE` on the principal returns both accounts' rows, an exact equality only the requested one).
  - Before the fix: fails — the account reports `["orders=PUB", "payments=SUB"]` (the other account's rule is absorbed).
  - After the fix: passes — `["orders=PUB"]` only.
- Full test class: `mvn -f server/pom.xml test -Dtest=MybatisPlusAclRepositoryTest` → **Tests run: 23, Failures: 0, Errors: 0**.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/AclService.java
@@ -208,6 +208,7 @@ public class AclService {
                 .secretKey(existing.getSecretKey())
                 .admin(user.getAdmin() == null ? existing.isAdmin() : user.getAdmin())
                 .clusters(user.getClusters() == null ? existing.getClusters() : user.getClusters())
+                .whiteRemoteAddress(existing.getWhiteRemoteAddress())
                 .gmtCreate(existing.getGmtCreate())
                 .build();
         AclUserVO saved = aclRepository.replaceUser(merged)
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java
@@ -342,7 +342,13 @@ public class MybatisPlusAclRepository implements AclRepository {
     }
 
     private PlainAccessConfigVO toPlainAccessConfig(AclUserVO user) {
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a bug where the ACL rule query used a LIKE pattern on `accessKey`, causing rules from accounts with similar key prefixes (e.g. "svc-a" matching "svc-a-v2") to be incorrectly included. Replaces with an exact `eq("principal", ...)` match.

This is a correct fix — the comment clearly explains the motivation and mirrors the delete behavior in `upsertPlainAccessRules`. The included test validates the exact scenario described.

LGTM.

---

PR #3052, opened 2026-09-03
fix(ai): page inventory tool results

tju-yxq:
## What is the purpose of the change

The AI/MCP/CLI inventory tools returned complete inventories in one response. `rmq.topic.list` and `rmq.group.list` bypassed the provider-aware paged service methods, and `rmq.alert.rule.list` loaded every rule before filtering in memory. Large instances could therefore return thousands of projected rows to an AI client in one call.

## What changed

- add optional `page` and `pageSize` inputs to the topic, consumer-group, and alert-rule tools
- default to page 1 and 20 rows, with the catalog enforcing `pageSize <= 100`
- route topic and consumer-group tools through `MetadataService.listTopicsPage` and `listConsumerGroupsPage`
- route the alert-rule tool through the existing paged `AlertService.listRules(search, enabled, page, pageSize)`
- change these three tool outputs from bare arrays to `{items, total, page, size}` objects while retaining the existing item projections
- update gateway tests to assert the paged service calls, bounded projections, and page metadata

## Verification

- focused tests: `ToolCatalogTest,ToolGatewayServiceTest` (38 tests) passed
- Checkstyle: 0 violations
- full server suite: `mvn -DskipTests=false test` (1,960 tests, 0 failures, 0 errors)
- production/config diff: 281 additions and 136 deletions across three handlers and the tool catalog

Closes #3051

PR #4115, opened 2026-09-08
fix(ai): pass the tool cluster input to the cluster-scoped read

zjncs:
### Motivation

`rmq.topic.list` and `rmq.group.list` always fail with `404 Instance not found: DefaultCluster` in the default deployment (namesrv-only, no registered instances).

`MetadataService.listTopicsPage(instanceId, clusterId, ...)` routes a **blank instance id + non-blank cluster id** to the legacy cluster-scoped read (the branch commented `// legacy cluster-scoped read kept for AI tool handlers`). But both handlers pass the tool's `cluster` input into the **first** (instanceId) slot and `null` into the second:

```java
// TopicListToolHandler / ConsumerGroupListToolHandler
metadataService.listTopicsPage(clusterId, null, type, search, page, pageSize);
```

A cluster name such as `DefaultCluster` therefore reaches `providerRegistry.byInstanceId(...)`, which throws `404 "Instance not found: DefaultCluster"` for any value that is not a registered instance id.

This is a regression from #3052: the handlers previously called the 3-arg `listTopics(clusterId, type, search)` overload, which delegates to `listTopics(null, clusterId, ...)` — the old first parameter *was* the cluster id, and #3052 mapped it onto the new signature's renamed first parameter (`instanceId`) unchanged:

```diff
-        return metadataService.listTopics(clusterId, type, search).stream()
+        PageResult<TopicVO> page = metadataService.listTopicsPage(
+                clusterId, null, type, search, ToolListPagination.page(input), ToolListPagination.pageSize(input));
```

The same request proves the contradiction internally: `ToolGatewayService.enforceCapabilities` resolves the same `cluster` input through `clusterService.getCluster(...)` (cluster-name space) and succeeds, then the handler 404s on the same value in the instance-id space.

### Modifications

- `TopicListToolHandler` / `ConsumerGroupListToolHandler`: pass `null` as the instance id and the tool's `cluster` input as the cluster id, restoring the pre-#3052 routing and making the cluster-scoped branch reachable again.
- `ToolGatewayServiceTest`: updated the two stubs that pinned the previous slotting.

### Verification

- New `TopicListToolHandlerTest` / `ConsumerGroupListToolHandlerTest` (modeled on `MessageQueryToolHandlerTest`) stub `metadataService.listTopicsPage(isNull(), eq("DefaultCluster"), ...)` and verify the delegation.
  - Before the fix: both fail with `PotentialStubbingProblem` — the handlers invoke `("DefaultCluster", null, ...)`.
  - After the fix: both pass.
- `mvn -f server/pom.xml test -Dtest='TopicListToolHandlerTest,ConsumerGroupListToolHandlerTest,ToolGatewayServiceTest'` → **Tests run: 34, Failures: 0, Errors: 0**.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ConsumerGroupListToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ConsumerGroupListToolHandler.java
@@ -45,7 +45,7 @@ public class ConsumerGroupListToolHandler implements ToolHandler {
         String clusterId = (String) input.get("cluster");
         String search = (String) input.get("search");
         PageResult<ConsumerGroupVO> page = metadataService.listConsumerGroupsPage(
-                clusterId, null, search, ToolListPagination.page(input), ToolListPagination.pageSize(input));
+                null, clusterId, search, ToolListPagination.page(input), ToolListPagination.pageSize(input));
         return ToolListPagination.pagedResult(page, page.getItems().stream()
                 .map(ConsumerGroupListToolHandler::safeProjection)
                 .toList());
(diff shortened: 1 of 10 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a regression from #3052 where TopicListToolHandler and ConsumerGroupListToolHandler passed the cluster name into the instanceId parameter slot, causing 404 Instance not found errors in the default namesrv-only deployment. Swapping the two arguments to pass null as instanceId and the cluster input as clusterId restores the pre-3052 routing. Excellent PR description tracing the regression. LGTM.

---

PR #4117, opened 2026-09-08
fix(ai): allow null counts in tool output schemas

zjncs:
### Motivation

`rmq.dashboard.summary` and `rmq.group.list` fail output validation (HTTP 500 `IllegalStateException: Tool output validation failed`) whenever the data behind a declared `type: integer` count is unavailable — which the producers deliberately encode as `null`:

- `rmq.dashboard.summary`: `RocketMQDashboardProvider` reports proxy/name-server topology counts as `null` for V5 clusters (`.proxies(clusterType == ClusterType.V4_DIRECT ? 0 : null)`), for aggregate mode (`.totalProxies(null)`), and in the degraded no-namesrv path (both `totalProxies(null)` / `totalNameServers(null)`). `DashboardControllerTest.getDashboardShouldPreserveUnavailableTopologyCounts` pins these nulls as the intended HTTP semantics — only the tool schema rejects them.
- `rmq.group.list`: #3988 deliberately changed the handler projection to `totalLag == ConsumerLagResolver.UNKNOWN ? null : totalLag` (unknown lag for 5.0 gRPC consumers), but the output schema was not updated — so any cluster containing one such group makes the whole listing fail with a 500, exactly when the user most needs it.

`objectMapper.valueToTree` keeps the keys (`{"totalLag": null}`) and the schema validator rejects `null` against `type: integer`.

### Modifications

In `tool-catalog/rmq-tools.yaml`, the four affected fields become `type: ['integer', 'null']` (keys stay in `required`, so they must still be present):
- `rmq.dashboard.summary` → `cluster.proxies`, `stats.totalProxies`, `stats.totalNameServers`
- `rmq.group.list` → `items[].totalLag`

### Verification

Two new `ToolGatewayServiceTest` cases mirroring the producer semantics:
- `executesDashboardSummaryWhenTopologyCountsAreUnavailable` (dashboard with `proxies/totalProxies/totalNameServers = null`)
- `executesConsumerGroupListWhenLagIsUnknown` (group with `totalLag = ConsumerLagResolver.UNKNOWN`)

Before the fix both fail with the production error, e.g.:
```
IllegalStateException: Tool output validation failed for rmq.dashboard.summary:
[/cluster/proxies: null found, integer expected,
 /stats/totalNameServers: null found, integer expected,
 /stats/totalProxies: null found, integer expected]
IllegalStateException: Tool output validation failed for rmq.group.list:
[/items/0/totalLag: null found, integer expected]
```

After the fix: `mvn -f server/pom.xml test -Dtest='ToolGatewayServiceTest,ToolCatalogTest'` → **Tests run: 40, Failures: 0, Errors: 0** (catalog meta-schema validation included).

### DIFF
--- a/server/src/main/resources/tool-catalog/rmq-tools.yaml
+++ b/server/src/main/resources/tool-catalog/rmq-tools.yaml
@@ -126,7 +126,7 @@ tools:
             brokers:
               type: integer
             proxies:
-              type: integer
+              type: [integer, 'null']
             topics:
               type: integer
             groups:
@@ -164,9 +164,9 @@ tools:
             totalBrokers:
               type: integer
             totalProxies:
-              type: integer
+              type: [integer, 'null']
             totalNameServers:
-              type: integer
+              type: [integer, 'null']
             totalTopics:
               type: integer
             totalConsumerGroups:
@@ -332,7 +332,7 @@ tools:
               onlineInstances:
                 type: integer
               totalLag:
-                type: integer
+                type: [integer, 'null']
               subscribedTopics:
                 type: array
                 items:
--- a/server/src/test/java/org/apache/rocketmq/studio/ops/ai/tool/ToolGatewayServiceTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/ops/ai/tool/ToolGatewayServiceTest.java
@@ -41,6 +41,7 @@ import org.apache.rocketmq.studio.ops.dashboard.ClusterOverviewVO;
 import org.apache.rocketmq.studio.ops.dashboard.DashboardDataVO;
 import org.apache.rocketmq.studio.ops.dashboard.DashboardService;
 import org.apache.rocketmq.studio.ops.dashboard.DashboardStatsVO;
+import org.apache.rocketmq.studio.provider.apache.ConsumerLagResolver;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes tool output validation failures when topology counts or lag values are legitimately null for V5 clusters. Changing the schema to allow null for the four affected fields is correct since the producers deliberately encode unavailability as null. LGTM.

---

lizhimins:

Your change shipped as-is: `cluster.proxies`, `stats.totalProxies`, `stats.totalNameServers` and `items[].totalLag` in `tool-catalog/rmq-tools.yaml` are now `type: [integer, 'null']` while staying `required`, and your `ToolGatewayServiceTest` cases came across with it. Your diagnosis was the useful part — `ToolGatewayService.validateOutput` keeps null keys through `valueToTree`, so a producer that deliberately emits null for "unknown" failed schema validation and surfaced as a 500 rather than as a missing value. Auditing the rest of the integer output fields confirmed these four are the complete reachable set.

One thing worth knowing: on its own this would not have changed any observable behaviour, because both handlers were still passing the `cluster` input into the `instanceId` slot and therefore threw `404 Instance not found` before output validation was ever reached. That is why it landed together with the routing fix rather than separately.

PR #4118, opened 2026-09-08
fix(metrics): report unknown queue lag instead of clamping the sentinel

zjncs:
### Motivation

`ApacheRocketMqBusinessMetricsCollector.queueLagSamples` clamps every `QueueProgressVO.diffTotal` with `Math.max(0, ...)`. But `RocketMQMetadataProvider.resolveDiff` deliberately passes the broker's `-1` unknown sentinel through into `diffTotal` (its javadoc: *"Resolves the lag for a single queue without clamping the broker's `-1` 'unknown' sentinel to zero… so the unknown state stays visible"* — typical for RocketMQ 5.0 gRPC consumers). The clamp therefore turns "lag unknown" into a fabricated **zero-lag AVAILABLE sample** for `consumer.lag.max_queue` and `topic.backlog.total`:

- a "max queue lag > N" or "topic backlog > N" rule never fires while the state is unknown;
- an already-FIRING alert receives healthy `0`s and falsely resolves;
- window aggregations (AVG/MAX) are diluted with fake `0`s;
- and because a sample *is* emitted, `reconcileMissingActiveStates` never sees a gap to reconcile — the unknown state is fully masked.

This is the same defect #3988 fixed for `consumer.lag.total` (its comment: *"do not clamp it into a fabricated zero-lag AVAILABLE sample that would feed consumer.lag.total alerts a fake 0"*). The sibling `queueLagSamples` path was not covered by that fix.

### Modifications

`queueLagSamples` first checks whether any progress row carries `ConsumerLagResolver.UNKNOWN`; if so it returns UNAVAILABLE samples with reason `CONSUMER_LAG_UNKNOWN` for both affected metrics (mirroring the fixed `totalLag` path). The `Math.max(0, ...)` clamp is kept for the normal path, guarding only genuinely impossible negatives.

### Verification

New test `reportsUnavailableQueueLagWhenProgressCarriesSentinelTest` (modeled on `reportsUnavailableTotalLagWhenSentinelUnknownTest`) stubs `getGroupProgress` with `diffTotal = ConsumerLagResolver.UNKNOWN`:
- Before the fix: fails — samples come back `AVAILABLE` with a fabricated `0`.
- After the fix: passes — both metrics are `UNAVAILABLE`, value `null`, reason `CONSUMER_LAG_UNKNOWN`.
- Full metrics module regression: `mvn -f server/pom.xml test -Dtest='org.apache.rocketmq.studio.cluster.metrics.**'` → **Tests run: 142, Failures: 0, Errors: 0**.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqBusinessMetricsCollector.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqBusinessMetricsCollector.java
@@ -129,6 +129,15 @@ public class ApacheRocketMqBusinessMetricsCollector implements BusinessMetricsCo
         Map<String, String> labels = Map.of("consumerGroup", group.getName());
         try {
             List<QueueProgressVO> progress = provider.getGroupProgress(instance.getName(), group.getName());
+            if (progress.stream().anyMatch(row -> row.getDiffTotal() == ConsumerLagResolver.UNKNOWN)) {
+                // diffTotal carries the -1 unknown sentinel (RocketMQ 5.0 gRPC consumers); clamping
+                // it would feed max-queue/backlog alerts a fabricated zero-lag AVAILABLE sample,
+                // the same way consumer.lag.total did before it was fixed.
+                return List.of(unavailable(CONSUMER_LAG_MAX_QUEUE, instance, labels, collectedAt,
+                        "CONSUMER_LAG_UNKNOWN"),
+                        unavailable(TOPIC_BACKLOG_TOTAL, instance, labels, collectedAt,
+                                "CONSUMER_LAG_UNKNOWN"));
+            }
             long maxLag = progress.stream().mapToLong(QueueProgressVO::getDiffTotal)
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes the same defect that #3988 fixed for consumer.lag.total but in the queueLagSamples path. The Math.max clamp turned the broker -1 unknown sentinel into a fabricated zero-lag AVAILABLE sample for max-queue/backlog metrics. Clean fix checking for the sentinel before clamping. LGTM.

---

lizhimins:
Closing together with #4121, which changes the same collector for the same `-1` sentinel six minutes apart.

We would rather land one trunk-side fix that covers every lag-derived sample than merge two partial patches: max-queue lag and per-topic backlog are what these two address, but `ApacheRocketMqBusinessMetricsCollector.java:123` still clamps `getDelaySeconds()` with `Math.max(0, ...)`, so after either patch the same consumer group would publish an explicit `CONSUMER_LAG_UNKNOWN` next to a fabricated zero delay. Your analysis of the sentinel path (`RocketMQMetadataProvider` -> `ConsumerLagResolver`) is correct and is what prompted the trunk-side item.

PR #4119, opened 2026-09-08
fix(alert): keep suppressing while a cluster incident only has a REMINDER left

zjncs:
### Motivation

`AlertNotificationSuppressionService.findSuppressingClusterAlert` decides whether a BUSINESS notification is redundant by looking for an active CLUSTER incident in the same scope. It keeps only incidents whose **latest** in-window event is `FIRING`:

```java
return latestByIncident.values().stream()
        .filter(candidate -> "FIRING".equalsIgnoreCase(candidate.getTransition()))
        .max(...);
```

But `AlertStateMachine.advanceHit` emits `REMINDER` **only while the state stays `FIRING`** — a REMINDER is by definition an active incident — and `NativeAlertEvaluationService.emitsLifecycleEvent` persists REMINDER events for CLUSTER-domain rules too. CLUSTER rules default to a 30-minute `reminderInterval`, the same length as the suppression `CORRELATION_WINDOW` (30m).

Failure scenario: a broker incident fires at t0; at t0+30m the state machine emits a REMINDER; at t0+35m a correlated BUSINESS alert fires. The window `[t0+5m, t0+35m]` now contains **only the REMINDER** (the original FIRING at t0 fell out) → the filter drops it → `Optional.empty()` → the redundant business notification is delivered, exactly the case this class exists to suppress (`/** Finds active cluster incidents that make a business notification redundant. */`). The intended boundary is RESOLVED, as `doesNotSuppressAfterTheSameClusterIncidentHasResolvedTest` shows.

### Modifications

The filter also accepts `REMINDER` as an active transition (only `RESOLVED` ends an incident).

### Verification

New test `suppressesWhileTheSameClusterIncidentOnlyReminderRemainsTest` (modeled on the existing resolved-incident test): same fingerprint/scope, the original FIRING outside the window, a `REMINDER` at `now-5m`.
- Before the fix: fails — `Expecting Optional to contain: but was empty`.
- After the fix: passes — the REMINDER suppresses the business notification.
- `mvn -f server/pom.xml test -Dtest='AlertNotificationSuppressionServiceTest,NativeAlertEvaluationServiceTest,AlertStateMachineTest'` → **Tests run: 14, Failures: 0, Errors: 0**.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqProxyMetricsCollector.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqProxyMetricsCollector.java
@@ -79,7 +79,9 @@ public class ApacheRocketMqProxyMetricsCollector implements ClusterMetricsCollec
             return samples;
         } catch (RuntimeException error) {
             log.warn("Failed to discover proxies for instance {}: {}", instance.getName(), error.getMessage());
-            return List.of(unavailable(instance, null, Map.of("proxyAddr", "unknown"), collectedAt));
+            // no proxy is known at this point: emit the whole-scope failure marker (empty labels)
+            // so NativeAlertProcessor skips reconciliation instead of resolving active proxy alerts
+            return List.of(unavailable(instance, null, Map.of(), collectedAt));
         }
     }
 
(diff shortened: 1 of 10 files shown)


RockteMQ-AI (approved):
### Summary
Fixes notification suppression where the filter only accepted FIRING transitions but REMINDER is also an active incident signal. When the original FIRING fell out of the 30-minute correlation window, the redundant business notification was delivered. Clean fix extending the filter. LGTM.

---

PR #4123, opened 2026-09-08
fix(group): allow zero retry settings on update

zjncs:
### Motivation

`UpdateConsumerGroupSettingsDTO` validates both retry fields with `@Positive`:

```java
@NotNull(message = "retryQueueNums is required")
@Positive(message = "retryQueueNums must be positive")
private Integer retryQueueNums;
@NotNull(message = "retryMaxTimes is required")
@Positive(message = "retryMaxTimes must be positive")
private Integer retryMaxTimes;
```

But `0` is a legitimate broker-side value — `RocketMQAdminClientImpl` forwards these straight into `SubscriptionGroupConfig.setRetryQueueNums/setRetryMaxTimes`, where `retryQueueNums=0` disables retry queues and `retryMaxTimes=0` sends failed messages directly to the DLQ. The sibling `CreateConsumerGroupDTO` already accepts it:

```java
@PositiveOrZero(message = "retryMaxTimes must be zero or positive")
private Integer retryMaxTimes;
```

Concrete failure: a group created via `POST /api/groups/create` with `retryMaxTimes: 0` (accepted) can never have its settings re-saved via the settings endpoint while keeping that value — the request fails bean validation with a 400 before reaching `MetadataService`.

### Modifications

Both fields switch to `@PositiveOrZero` with matching messages, aligning the update path with the create path and the broker semantics. Negative values are still rejected.

### Verification

New standalone `UpdateConsumerGroupSettingsDTOTest` (the controller test file is under review in another PR, so the validation is pinned in its own class, modeled on `AlertRuleRequestDTOTest`):
- `shouldAcceptZeroRetrySettingsTest` — before the fix: fails with `retryQueueNums must be positive` / `retryMaxTimes must be positive`; after: passes.
- `shouldRejectNegativeRetrySettingsTest` — negative values still produce the two violations.
- Regression: `mvn -f server/pom.xml test -Dtest='UpdateConsumerGroupSettingsDTOTest,ConsumerGroupControllerTest,MetadataServiceTest'` → **Tests run: 59, Failures: 0, Errors: 0**.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/group/UpdateConsumerGroupSettingsDTO.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/group/UpdateConsumerGroupSettingsDTO.java
@@ -18,7 +18,7 @@ package org.apache.rocketmq.studio.instance.group;
 
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
-import jakarta.validation.constraints.Positive;
+import jakarta.validation.constraints.PositiveOrZero;
 import lombok.Data;
 
 @Data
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a validation inconsistency where UpdateConsumerGroupSettingsDTO rejected retryQueueNums=0 and retryMaxTimes=0 with @Positive while the create DTO accepted them with @PositiveOrZero and 0 is a legitimate broker-side value. Simple annotation change aligning update with create. LGTM.

---

PR #4124, opened 2026-09-08
fix(alert): trim stored metrics when scoping findActive

zjncs:
### Motivation

`NativeAlertProcessor.reconcileMissingActiveStates` keeps rules whose stored metric has surrounding whitespace by normalizing before matching the scope keys:

```java
.filter(rule -> scope.metricKeys().contains(StringUtils.trimWhitespace(rule.getMetric())))
```

…but the `MybatisPlusAlertStateRepository.findActive` it then calls filters with the **raw** value:

```java
.filter(rule -> metricKeys.contains(rule.getMetric()))
```

For a rule whose stored metric is padded (legacy rows — current write paths trim, but the rule entities are documented to support values "copied from older deployments", and `AlertRuleSemanticFingerprintTest` pins ` metric(" consumer.lag.total ")` as a supported input), the processor keeps the rule while `findActive` drops it → `scopedRules` is empty → `findActive` returns `List.of()` → `reconcileMissingActiveStates` sees no active states → the stale FIRING/ACKED state is never resolved and no recovery notification is sent. This silently reintroduces the exact failure mode the reconcile logic guards against, via a caller/callee normalization mismatch.

### Modifications

`findActive` applies `StringUtils.trimWhitespace(...)` to the metric before the `metricKeys` membership check, matching its only caller.

### Verification

New test `findsActiveStatesForRulesWithPaddedStoredMetricsTest` (model: `findsActiveStatesWithLabelsFromTheLatestAlertEventTest`): rule with `.metric(" consumer.lag.total ")`, scope keys `Set.of("consumer.lag.total")`.
- Before the fix: fails — `Expected size: 1 but was: 0` (active states empty).
- After the fix: passes — the FIRING state is returned with its key.
- Regression: `mvn -f server/pom.xml test -Dtest='MybatisPlusAlertStateRepositoryTest,NativeAlertProcessorTest'` → **Tests run: 23, Failures: 0, Errors: 0**.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertStateRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertStateRepository.java
@@ -103,7 +103,7 @@ public class MybatisPlusAlertStateRepository implements AlertStateRepository {
                 .filter(rule -> rule.getId() != null)
                 .filter(AlertRuleVO::isEnabled)
                 .filter(rule -> ruleDomain(rule) == scope.domain())
-                .filter(rule -> metricKeys.contains(rule.getMetric()))
+                .filter(rule -> metricKeys.contains(StringUtils.trimWhitespace(rule.getMetric())))
                 .filter(rule -> !StringUtils.hasText(rule.getInstanceId())
                         || scope.instanceId().equals(rule.getInstanceId()))
                 .collect(Collectors.toMap(AlertRuleVO::getId, rule -> rule, (left, right) -> left));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a caller/callee normalization mismatch where NativeAlertProcessor trims the metric key before matching but findActive used the raw value, causing stale FIRING/ACKED states to never resolve for legacy padded metrics. One-line fix. LGTM.

---

lizhimins:

Your change shipped as-is: `MybatisPlusAlertStateRepository.findActive` now filters with `StringUtils.trimWhitespace(rule.getMetric())`. Your framing of the gap was the part that made this easy to accept — `AlertRuleEvaluator` already matched samples against `rule.getMetric().trim()`, so the caller and the callee normalised differently, `scopedRules` came back empty, and `reconcileMissingActiveStates` concluded there was nothing active, leaving a stale FIRING alert that could never resolve. Confirming that `findActive` has exactly one production caller is what bounds the blast radius.

Recorded as a follow-up rather than fixed here: this is the third place that trims a metric key individually. The durable fix is to normalise once at the read boundary (`MybatisPlusAlertRepository`), which would make this change, #4126 and the existing trim in `AlertRuleEvaluator` all redundant and would also cover the sites that still do not trim (`NativeAlertRulePolicy`, `AlertService`, `NativeAlertProcessor`).

PR #4126, opened 2026-09-08
fix(alert): trim stored metric keys when normalizing percentage thresholds

zjncs:
### Motivation

`AlertRuleEvaluator` matches a rule against a sample using the trimmed metric key (`rule.getMetric().trim()`), but two ratio-aware code paths still consult `RATIO_METRICS` with the raw stored key:

- `AlertRuleSemanticFingerprint.normalizedThreshold` checks `RATIO_METRICS.contains(rule.getMetric())`
- `AlertNotificationTemplate.formattedValue` checks `RATIO_METRICS.contains(rule.getMetric())`

For a rule stored with a padded metric like `" broker.disk.usage_ratio "` and a `%` threshold of 85:

- the threshold is **not** divided by 100, so the evaluator compares the raw ratio `0.9 >= 85` and the rule never fires (or fires on every healthy disk with `<=`)
- the notification renders the raw ratio next to the unit, e.g. `0.865%/85` instead of `86.5%/85`

`AlertRuleSemanticFingerprint.of()` already trims every field for identity and the request DTO normalizes the metric key on write, so this only affects rules persisted through other paths.

### Modifications

- `AlertRuleSemanticFingerprint.normalizedThreshold`: look up `RATIO_METRICS` with the existing `normalize()` helper instead of the raw metric
- `AlertNotificationTemplate.formattedValue`: compare against the trimmed metric

### Verification

Fail-before (both tests added in this PR, run against the unpatched code):

```
[ERROR] Tests run: 6, Failures: 1 -- AlertRuleEvaluatorTest
Expecting value to be true but was false
[ERROR] Tests run: 4, Failures: 1 -- AlertNotificationTemplateTest
expected: "86.5%" but was: "0.865%"
```

Pass-after:

```
Tests run: 6, Failures: 0, Errors: 0 -- AlertRuleEvaluatorTest
Tests run: 4, Failures: 0, Errors: 0 -- AlertNotificationTemplateTest
Tests run: 3, Failures: 0, Errors: 0 -- AlertRuleSemanticFingerprintTest
Tests run: 13, Failures: 0, Errors: 0
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationTemplate.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationTemplate.java
@@ -62,7 +62,8 @@ final class AlertNotificationTemplate {
         if (alert.getCurrentValue() == null) {
             return "";
         }
-        if (rule != null && "%".equals(rule.getThresholdUnit()) && RATIO_METRICS.contains(rule.getMetric())) {
+        if (rule != null && "%".equals(rule.getThresholdUnit())
+                && RATIO_METRICS.contains(rule.getMetric() == null ? "" : rule.getMetric().trim())) {
             return String.valueOf(alert.getCurrentValue() * 100);
         }
         return String.valueOf(alert.getCurrentValue());
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleSemanticFingerprint.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertRuleSemanticFingerprint.java
@@ -62,7 +62,7 @@ final class AlertRuleSemanticFingerprint {
     }
 
     static double normalizedThreshold(AlertRuleVO rule) {
-        if ("%".equals(rule.getThresholdUnit()) && RATIO_METRICS.contains(rule.getMetric())) {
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a metric key trimming inconsistency where two code paths used the raw stored key for RATIO_METRICS lookup while the evaluator trimmed it, causing percentage thresholds to never normalize for padded legacy metrics. Clean fix using the existing normalize helper. LGTM.

---

lizhimins:

Your change shipped as-is: `AlertRuleSemanticFingerprint.normalizedThreshold` now goes through the class's own `normalize()`, and `AlertNotificationTemplate.formattedValue` trims before consulting `RATIO_METRICS` (and gained a null guard). Both user-visible consequences you described check out — a rule stored as `" broker.disk.usage_ratio "` kept its raw ratio threshold so `0.9 >= 85` could never fire, and its notification rendered `${value}` as `0.865%` instead of `86.5%`. Your worked examples matched the code exactly.

Two follow-ups recorded rather than done here: `RATIO_METRICS` exists as two verbatim copies across those classes and should be pulled down into `common/util`, and the codebase mixes `trimWhitespace` with `trim()` for the same normalisation. See the note on #4124 about normalising once at the read boundary instead, which would subsume both.

PR #4132, opened 2026-09-08
fix(settings): avoid cache-key collision for paged data sources

### LINKED ISSUE #4048: [Studio][Bug] Paged data-source cache conflates null with literal filter values

btlqql:
## Problem

`SettingsService.listDataSources(search, type, page, pageSize)` used a
SpEL concatenation key. Java `null` becomes the literal text `"null"`,
so `(null, null, 1, 20)` and `("null", null, 1, 20)` shared one cache
entry.

## Fix

Remove the explicit key expression and rely on Spring default
`SimpleKeyGenerator`, which preserves each argument value and keeps
distinct tuples in distinct cache entries.

## Test plan

- Added `SettingsServiceCacheIntegrationTest` covering null versus
  literal `"null"` filters.
- `mvn -q -Dtest=SettingsServiceCacheIntegrationTest test`
- `mvn -q -Dtest=SettingsServiceTest test`

Fixes #4048

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
@@ -218,7 +218,7 @@ public class SettingsService {
         return settingsRepository.findAllDataSources();
     }
 
-    @Cacheable(value = "data-sources", key = "'page:' + #search + ':' + #type + ':' + #page + ':' + #pageSize")
+    @Cacheable("data-sources")
     public PageResult<DataSourceVO> listDataSources(String search, String type, int page, int pageSize) {
         if (page < 1) {
             throw new BusinessException(400, "page must be greater than zero");
--- a/server/src/test/java/org/apache/rocketmq/studio/settings/SettingsServiceCacheIntegrationTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/settings/SettingsServiceCacheIntegrationTest.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.studio.settings;
+
+import org.apache.rocketmq.studio.common.domain.PageResult;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
LGTM — correct fix for the cache-key collision. The original SpEL key expression `"page:" + #search + ":" + #type + ":" + #page + ":" + #pageSize` would produce identical keys when `#search` is null vs the literal string `"null"`, causing cache poisoning.

Switching to the default Spring `SimpleKeyGenerator` (no explicit key) generates distinct keys based on the actual parameter values and types, which correctly differentiates null from "null".

The integration test verifying the non-collision behavior is a nice addition.

---

lizhimins:
Closing as a duplicate of #4049, which was merged into `rocketmq-studio` as c6b546a7.

The production fixes are byte-identical, not merely equivalent. Both PRs take `SettingsService.java` from blob `7847297d45` to blob `001d5541f4` — the same single-line change, dropping the explicit SpEL `key` from `@Cacheable` on the paged `listDataSources` so Spring's `SimpleKeyGenerator` handles it and `null` is no longer conflated with the literal string `"null"`.

#4049 is the one that landed because it was opened roughly 25 hours earlier and had already been marked ready for review, so it is first past the post. Your diagnosis was independently correct — an identical production change is the strongest possible evidence of that.

On the tests: yours boots the full application context (`@SpringBootTest` with `spring.cache.type=simple`) while #4049's uses a sliced `@ContextConfiguration` over a nested `@EnableCaching` config with an autowired `SettingsService`. We looked at whether merging #4049 lost anything by preferring the lighter test, and it did not — both exercise a real caching proxy rather than a hand-constructed service, and we verified #4049's case is sensitive to the fix by restoring the old SpEL key and confirming it fails. Yours asserts the same two properties (distinct returned objects, `times(1)` per argument shape). The sliced version is the one we would write for this, since it pins the cache key without paying for a full context startup, but that is a preference rather than a deficiency in your approach.

One process note for future PRs: this one declared `Fixes #4048`, but that issue was filed by the author of #4049 to describe their own finding. Linking `Fixes` to someone else's issue attaches your PR to their report and, on a repository whose default branch is not the merge target, can leave the issue in an ambiguous state. If you independently find the same bug, opening your own issue or simply describing the defect in the PR body works better.

Thank you for the fix — the defect was real and is now closed on the branch.

PR #2957, opened 2026-09-02
fix(alert): consolidate native collection reliability

Aias00:
## Summary
- consolidate the native alert collection reliability fixes from #2668 and #2688
- keep a bounded sliding in-flight collection window so saturated executors do not drop later instances
- reconcile stale active native alert states only after a successful full collector scope, preserving partial/failure semantics
- batch-load latest alert metadata for active state reconciliation to avoid per-state lookups

Replaces #2668
Replaces #2688
Closes #2662
Closes #2679

## Verification
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=CollectorSchedulerTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest=NativeAlertProcessorTest,MybatisPlusAlertStateRepositoryTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -Dtest='org.apache.rocketmq.studio.ops.alert.*Test,org.apache.rocketmq.studio.cluster.metrics.*Test' test` mostly passes; it fails only in `AlertSchemaMigrationTest.upgradesExistingAlertTablesWithoutChangingFreshSchemaBehaviorTest` because current base tries to add `rmq_instance_message.result_snapshot` when the legacy fixture lacks `rmq_instance_message`. The same test fails on clean `origin/rocketmq-studio` with the same error.
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn -DskipTests package`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) mvn test` runs 1862 tests with 1 error, the same baseline `AlertSchemaMigrationTest` failure above
- `git diff --check origin/rocketmq-studio...HEAD`

PR #4133, opened 2026-09-08
fix(alert): mark proxy discovery failure as a whole-scope failure

zjncs:
## Motivation

`ApacheRocketMqProxyMetricsCollector.collect` returns a single unavailable sample with the placeholder label `proxyAddr=unknown` when proxy discovery fails for the whole instance (e.g. nameserver outage):

```java
} catch (RuntimeException error) {
    return List.of(unavailable(instance, null, Map.of("proxyAddr", "unknown"), collectedAt));
}
```

#2957 added `NativeAlertProcessor.containsWholeScopeFailure` to skip reconciliation when a collection reports a whole-scope failure — detected as an unavailable sample with **empty labels** (the marker `ApacheRocketMqClusterMetricsCollector` already emits for nameserver failures). The proxy discovery-failure sample predates that convention (#2533) and was not aligned, so:

1. `containsWholeScopeFailure` returns false (labels are non-empty) and `reconcileMissingActiveStates` runs;
2. the placeholder fingerprint (`proxyAddr=unknown`) never matches the active states of the **real** proxies;
3. every firing `proxy.availability` alert for the instance is **falsely resolved** — resolution events are emitted and notifications enqueued — while in reality the collector could not discover a single proxy and knows nothing about their status.

So a nameserver hiccup turns all "proxy down" alerts into "resolved" notifications, masking real outages.

## Modification

Emit the whole-scope failure marker (empty labels) for the discovery failure, matching `ApacheRocketMqClusterMetricsCollector`\'s nameserver failure sample and the detection added in #2957:

```java
return List.of(unavailable(instance, null, Map.of(), collectedAt));
```

The per-proxy failure paths (probe unreachable / probe error / malformed address) keep their real `proxyAddr` labels, so per-target unavailability still reconciles correctly and availability rules keep firing per real proxy. After discovery recovers, any state fired for the whole-scope marker carries the empty fingerprint and is resolved by the next successful reconciliation.

## Verification

`mvn -f server/pom.xml test -Dtest='ApacheRocketMqProxyMetricsCollectorTest,NativeAlertProcessorTest'`

fail-before (fix reverted, tests kept):

```
ApacheRocketMqProxyMetricsCollectorTest.recordsWholeScopeFailureSampleWhenProxyDiscoveryFailsTest  <<< FAILURE!
Expecting empty but was: {"proxyAddr"="unknown"}
Tests run: 22, Failures: 1, Errors: 0
```

pass-after:

```
Tests run: 3, Failures: 0 ... in ApacheRocketMqProxyMetricsCollectorTest
Tests run: 19, Failures: 0 ... in NativeAlertProcessorTest
Tests run: 22, Failures: 0, Errors: 0
BUILD SUCCESS
```

Also updated the existing discovery-failure test to pin the whole-scope marker (it previously asserted the placeholder label) and added a processor-level test showing the firing state of a real proxy survives a discovery outage.

Follow-up to #2957; no associated issue — found by code inspection.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqProxyMetricsCollector.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqProxyMetricsCollector.java
@@ -79,7 +79,9 @@ public class ApacheRocketMqProxyMetricsCollector implements ClusterMetricsCollec
             return samples;
         } catch (RuntimeException error) {
             log.warn("Failed to discover proxies for instance {}: {}", instance.getName(), error.getMessage());
-            return List.of(unavailable(instance, null, Map.of("proxyAddr", "unknown"), collectedAt));
+            // no proxy is known at this point: emit the whole-scope failure marker (empty labels)
+            // so NativeAlertProcessor skips reconciliation instead of resolving active proxy alerts
+            return List.of(unavailable(instance, null, Map.of(), collectedAt));
         }
     }
 
--- a/server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqProxyMetricsCollectorTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqProxyMetricsCollectorTest.java
@@ -82,7 +82,7 @@ class ApacheRocketMqProxyMetricsCollectorTest {
     }
 
     @Test
-    void recordsUnavailableSampleWhenProxyDiscoveryFailsTest() {
+    void recordsWholeScopeFailureSampleWhenProxyDiscoveryFailsTest() {
         ClusterService clusterService = mock(ClusterService.class);
         InstanceVO instance = InstanceVO.builder().name("local").endpoint("localhost:9876").build();
         doThrow(new IllegalStateException("nameserver unavailable")).when(clusterService).listClusters("local");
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (approved):
LGTM — well-reasoned fix. Emitting empty labels (`Map.of()`) instead of `Map.of("proxyAddr", "unknown")` for the whole-scope failure case correctly prevents `NativeAlertProcessor` from reconciling and falsely resolving active proxy alerts during discovery failures.

The test update accurately reflects the new contract. Good documentation in the comments explaining the rationale.

---

lizhimins:

The production fix shipped exactly as you wrote it: the discovery-failure sample now carries empty labels instead of `proxyAddr=unknown`, so it lands in the whole-scope failure marker that #2957 established and `NativeAlertProcessor.containsWholeScopeFailure` skips reconciliation instead of resolving every firing proxy alert during an outage. Keeping the real `proxyAddr` on the three per-proxy failure paths was the right scoping — a single unreachable proxy should still reconcile normally. Your reasoning also matches `ApacheRocketMqClusterMetricsCollector`'s nameserver failure sample, so the two collectors are now consistent.

One change was made during integration, to the tests rather than the fix. Your `ApacheRocketMqProxyMetricsCollectorTest` case is what pins this properly: it drives the real collector and asserts the emitted labels are empty, so reverting the fix turns it red. The companion case added to `NativeAlertProcessorTest` was dropped, because it constructed the empty-label `MetricSample` inline and therefore still passed with the collector fix reverted — it exercised the processor's pre-existing whole-scope handling, which `doesNotResolveMissingActiveStateWhenCollectionReportsWholeScopeUnavailableTest` already covers. Deleting it cost no coverage; keeping it would have implied the end-to-end path was pinned when it was not. `NativeAlertProcessorTest` runs 18 green and the collector test 3 green.

PR #4134, opened 2026-09-08
fix(ai): populate subscriptionMode and consumeType for cloud consumer groups

zjncs:
## Motivation

The `rmq.group.list` AI tool is **deterministically broken for Aliyun and Tencent instances**:

- `ConsumerGroupListToolHandler.safeProjection` requires both `subscriptionMode` and `consumeType` (`requiredEnumName` throws `IllegalStateException` on the first null).
- `AliyunConverters.toConsumerGroupVO` **never sets `subscriptionMode`** and its `toConsumeType` returns null when the OpenAPI response omits `messageModel`.
- `TencentInstanceProvider.toConsumerGroup` **never sets `subscriptionMode`** either (its consume type is always CLUSTERING).

So the first group of any cloud instance trips `IllegalStateException: Consumer group subscriptionMode is unavailable: <name>` and the whole tool call fails.

This contradicts the invariant the Apache provider already established in `RocketMQMetadataProvider.toConsumerGroupVO`:

```java
// messageModel stores the subscription mode ("Push"/"Pop"); surface it so read paths
// (web detail, AI rmq.group.list) never see a null subscriptionMode.
vo.setSubscriptionMode(parseSubscriptionMode(entity.getMessageModel()));
```

The invariant was fixed for Apache only; both cloud converters were missed. Existing tests never catch it because they always pre-set the enum (e.g. `ToolGatewayServiceTest` builds groups with `SubscriptionMode.Push`).

## Modification

- Both cloud converters emit `SubscriptionMode.Push`: cloud TCP consumer groups are push consumers (Aliyun's `messageModel` carries the consume model, not the subscription mode; Tencent has no POP groups).
- `AliyunConverters.toConsumeType` falls back to `CLUSTERING` for a missing/unrecognized `messageModel`, mirroring `RocketMQMetadataProvider.parseConsumeType` ("falling back to CLUSTERING").

## Verification

`mvn -f server/pom.xml test -Dtest='AliyunInstanceProviderTest,TencentInstanceProviderTest'`

fail-before (fix reverted, tests kept):

```
AliyunInstanceProviderTest.listConsumerGroupsShouldMapGroupIdTest         <<< FAILURE! (subscriptionMode was null)
AliyunInstanceProviderTest.listConsumerGroupsShouldFallBackWhenMessageModelMissingTest <<< FAILURE!
TencentInstanceProviderTest.listConsumerGroupsShouldMapAndFilterTest      <<< FAILURE! (subscriptionMode was null)
```

pass-after:

```
TencentInstanceProviderTest: Tests run: 43, Failures: 0
AliyunInstanceProviderTest: 27 tests, only getGroupProgressShouldMapLagRowsTest failing —
    verified pre-existing on a clean rocketmq-studio checkout (unrelated to this change)
```

Follow-up to the Apache-side invariant; no associated issue — found by code inspection.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunConverters.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunConverters.java
@@ -29,6 +29,7 @@ import com.aliyun.sdk.service.rocketmq20220801.models.ListTopicSubscriptionsResp
 import com.aliyun.sdk.service.rocketmq20220801.models.ListTopicsResponseBody;
 import org.apache.rocketmq.studio.common.domain.enums.ConsumeType;
 import org.apache.rocketmq.studio.common.domain.enums.DeliveryStatus;
+import org.apache.rocketmq.studio.common.domain.enums.SubscriptionMode;
 import org.apache.rocketmq.studio.common.domain.enums.TopicType;
 import org.apache.rocketmq.studio.common.util.SubscriptionFilterModes;
 import org.apache.rocketmq.studio.instance.group.ConsumerGroupVO;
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):

---

lizhimins:

Your change shipped as-is: both cloud providers now set `SubscriptionMode.Push`, and Aliyun's `toConsumeType` returns `CLUSTERING` instead of null when `messageModel` is absent or not Broadcasting, so `requiredEnumName` no longer throws for every cloud consumer group. Your reasoning for hardcoding Push held up — the pinned `alibabacloud-rocketmq20220801` response body carries no subscription-mode field at all, and it matches what the Apache provider already asserts for its own groups.

This one had to be integrated rather than merged in sequence: it and #4137 both add an import to `AliyunConverters.java` and both extend `AliyunInstanceProviderTest` / `TencentInstanceProviderTest`, so whichever landed second needed a rebase. Putting both on one branch resolved that once, and both sets of tests pass together (44 Tencent, 29 Aliyun).

PR #4135, opened 2026-09-08
test(ai): align the Aliyun lag-rows test with the fallback-only total contract

zjncs:
## Motivation

`AliyunInstanceProviderTest.getGroupProgressShouldMapLagRowsTest` **fails on the current rocketmq-studio HEAD**: it still expects the aggregate `total` row alongside the per-topic rows (`assertThat(rows).hasSize(2)`), but `AliyunConverters.toQueueProgressRows` has kept the total row only as a fallback since #2907:

```java
// The aggregate repeats the per-topic counts. Keep it only as a fallback when
// the API does not expose the topic breakdown, otherwise callers that sum rows
// report the same lag twice.
if (rows.isEmpty() && totalLag != null && totalLag.getReadyCount() != null) {
```

so only 1 row is produced:

```
Expected size: 2 but was: 1 in:
[QueueProgressVO(topic=topic-a, broker=topic:topic-a, queueId=0, ...)]
```

The production change was intentional (documented in the code); the test was simply never updated, leaving the suite red.

## Modification

- `getGroupProgressShouldMapLagRowsTest` now asserts the current contract: with a topic breakdown only the per-topic rows are returned (and explicitly no `total` row).
- Adds the previously missing fallback case: without a topic breakdown a single `total` row is returned with the aggregate lag.

## Verification

fail-before (current rocketmq-studio, this test class):

```
Tests run: 27, Failures: 1
getGroupProgressShouldMapLagRowsTest  <<< FAILURE!
Expected size: 2 but was: 1
```

pass-after:

```
Tests run: 27, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Follow-up to #2907; no associated issue — noticed while running the provider suite.

### DIFF
--- a/server/src/test/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProviderTest.java
+++ b/server/src/test/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProviderTest.java
@@ -313,17 +313,41 @@ class AliyunInstanceProviderTest {
 
         List<QueueProgressVO> rows = provider.getGroupProgress(STUDIO_INSTANCE_ID, "GID_test");
 
-        assertThat(rows).hasSize(2);
+        // with a topic breakdown the aggregate total row is dropped, otherwise callers
+        // that sum the rows would report the same lag twice
+        assertThat(rows).hasSize(1);
         QueueProgressVO topicRow = rows.stream()
                 .filter(row -> "topic:topic-a".equals(row.getBroker()))
                 .findFirst()
                 .orElseThrow();
         assertThat(topicRow.getDiffTotal()).isEqualTo(42L);
-        QueueProgressVO totalRow = rows.stream()
-                .filter(row -> "total".equals(row.getBroker()))
-                .findFirst()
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):

---

lizhimins:

Your conclusion was the correct one and it is the reason this is worth calling out: the test was wrong, the production code was right. `AliyunConverters.toQueueProgressRows` has only emitted the `broker="total"` aggregate as a fallback when there are no per-topic rows since #2907, but `getGroupProgressShouldMapLagRowsTest` still asserted two rows while stubbing both `topicLagMap` and `totalLag`. The decisive argument for keeping production behaviour is that `CloudRocketMqBusinessMetricsCollector` sums `getDiffTotal()` across every row it is handed, so emitting the aggregate alongside per-topic rows would have double-reported lag into Prometheus.

This was one of three standing backend red lights on the branch, and it is now cleared — `AliyunInstanceProviderTest` runs 29 green. Your added fallback-only case is what pins the contract going forward, so the drift cannot silently come back.

PR #4137, opened 2026-09-08
fix(cloud): guarantee non-null topic type and perm in cloud topic converters

zjncs:
## Motivation

The `rmq.topic.list` AI tool projection (`TopicListToolHandler.safeProjection`) resolves `topic.type` and `topic.perm` through `requiredEnumName`, which throws `IllegalStateException` on null values (contract introduced with the AI resource tools in #642). Both cloud converters can still produce nulls, so listing topics through the AI tool is broken on cloud instances:

- `AliyunConverters.toTopicVO` never sets `perm` at all, and `AliyunInstanceProvider.toTopics` adds no enrichment afterwards — **every** Aliyun topic carries `perm=null`, so the projection fails on the first listed topic, deterministically.
- `AliyunConverters.toTopicType` returns `null` for a null or unmapped `messageType`, and `TencentInstanceProvider.toTopicType` returns `null` for a blank or unmapped topic type, so the same projection fails whenever the cloud API omits the field or returns a value the switch does not map (e.g. a newly introduced cloud type).

## Modification

- `AliyunConverters.toTopicVO` now defaults `perm` to `TopicPerm.RW`: Aliyun's `ListTopics` API does not return permissions, and console-managed cloud topics are read-write — the same convention `TencentInstanceProvider.toTopic` already applies.
- `AliyunConverters.toTopicType` and `TencentInstanceProvider.toTopicType` now fall back to `TopicType.NORMAL` for blank/unknown values, mirroring the Apache provider's `parseTopicType` fallback so read paths (web detail, AI `rmq.topic.list`) never see a null type.

## Verification

Fail-before (source fix stashed, tests kept) — 4 failures, all `expected: NORMAL but was: null` / missing perm:

```
[ERROR] Tests run: 71, Failures: 4, Errors: 0
AliyunInstanceProviderTest.listTopicsShouldMapMessageTypeAndFilterTest  (unmapped type now expected NORMAL; perm expected RW)
AliyunInstanceProviderTest.listTopicsShouldGuaranteeTypeAndPermForAiToolProjectionTest
TencentInstanceProviderTest.listTopicsShouldFallBackToNormalTypeWhenTopicTypeMissingTest
```

Pass-after (fix applied):

```
AliyunInstanceProviderTest        Tests run: 27, Failures: 1  (only getGroupProgressShouldMapLagRowsTest — pre-existing on rocketmq-studio, fixed by #4135)
TencentInstanceProviderTest       Tests run: 44, Failures: 0
```

The one remaining `getGroupProgressShouldMapLagRowsTest` failure is the pre-existing base failure on `rocketmq-studio` (stale test vs the intentional #2907 production change), already fixed by open PR #4135; it is unrelated to this change.

New/updated tests:

- `AliyunInstanceProviderTest.listTopicsShouldGuaranteeTypeAndPermForAiToolProjectionTest` — null and unmapped `messageType` both map to `NORMAL` with `perm=RW`.
- `AliyunInstanceProviderTest.listTopicsShouldMapMessageTypeAndFilterTest` — the unmapped-type row now asserts `NORMAL` (was documenting the old `null`) and mapped rows assert `perm=RW`.
- `TencentInstanceProviderTest.listTopicsShouldFallBackToNormalTypeWhenTopicTypeMissingTest` — blank and unmapped topic types map to `NORMAL`, `perm` stays `RW`.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunConverters.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunConverters.java
@@ -29,6 +29,7 @@ import com.aliyun.sdk.service.rocketmq20220801.models.ListTopicSubscriptionsResp
 import com.aliyun.sdk.service.rocketmq20220801.models.ListTopicsResponseBody;
 import org.apache.rocketmq.studio.common.domain.enums.ConsumeType;
 import org.apache.rocketmq.studio.common.domain.enums.DeliveryStatus;
+import org.apache.rocketmq.studio.common.domain.enums.TopicPerm;
 import org.apache.rocketmq.studio.common.domain.enums.TopicType;
 import org.apache.rocketmq.studio.common.util.SubscriptionFilterModes;
 import org.apache.rocketmq.studio.instance.group.ConsumerGroupVO;
@@ -118,6 +119,9 @@ final class AliyunConverters {
         vo.setName(data.getTopicName());
         vo.setInstanceId(studioInstanceId);
         vo.setType(toTopicType(data.getMessageType()));
+        // Aliyun's ListTopics API does not return permissions; console-created cloud
+        // topics are read-write, matching the Tencent provider's mapping.
+        vo.setPerm(TopicPerm.RW);
         vo.setRemark(data.getRemark());
         vo.setGmtCreate(parseDateTime(data.getCreateTime()));
         vo.setGmtModified(parseDateTime(data.getUpdateTime()));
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):

---

lizhimins:

Your change shipped as-is: `AliyunConverters.toTopicVO` now sets `TopicPerm.RW` and both cloud `toTopicType` implementations fall back to `TopicType.NORMAL` instead of returning null for a blank or unrecognised type. The Aliyun perm gap was the more serious half — ListTopics returns no permission field and nothing downstream filled it in, so every Aliyun topic had a null perm and `TopicListToolHandler.safeProjection` threw on its first `requiredEnumName` call.

Worth recording that you scoped this correctly: you changed `type` and `perm` but left `messageCount`, `tps` and `consumerGroupCount` alone, which no provider populates either. Defaulting those too would have masked genuinely absent data rather than fixing a broken invariant, so the restraint was the right call.

PR #4142, opened 2026-09-08
[ISSUE #4141] fix(mock): scope Topic mutations by instance

### LINKED ISSUE #4141: [Studio][Bug] Scope mock Topic mutations by instance
## Problem

In mock mode, Topic reads already respect `instanceId`, but mutations do not consistently use the same resource identity:

- duplicate detection in `createTopic` compares only Topic name and cluster;
- `updateTopic` selects the first matching Topic name;
- `deleteTopic` removes the first matching Topic name.

The real persistence model allows the same `(cluster_id, name)` in different instances because its unique key is `(cluster_id, instance_id, name)`. In mock mode, creating the same Topic in a second instance is rejected, while update or delete can affect the matching row from another instance. This makes demo and test behavior diverge from the real API.

## Expected behavior

- Mock Topic duplicate detection includes `instanceId` when it is supplied.
- The same Topic name and cluster may exist in different instances.
- Mock update and delete affect only the Topic belonging to the requested instance.
- Existing legacy calls that omit `instanceId` keep their current behavior.

## Acceptance criteria

- Add regression coverage for same-name Topics in two instances.
- Verify a duplicate in the same instance is still rejected.
- Verify update changes only the requested instance.
- Verify delete removes only the requested instance.
- Do not change the real backend API contract or persistence schema.

## Scope

Frontend mock service and focused unit tests only.

Aias00:
## Summary

- align mock Topic identity with the persisted `(cluster_id, instance_id, name)` model
- allow the same Topic name and cluster in different mock instances
- scope mock update and delete operations to the supplied `instanceId`
- add regression coverage for cross-instance create, update, and delete behavior

## Scope

This changes only the frontend mock service and its focused tests. The real backend API and database schema are unchanged. Calls that omit `instanceId` keep the previous fallback behavior.

## Verification

- `npm test -- --run src/services/topicService.test.ts src/services/topicService.batchDelete.test.ts src/pages/instance/__tests__/TopicPage.test.tsx` — 34 tests passed
- `npm run lint` — 0 errors, 10 existing warnings
- `npm run build`
- `git diff --check`

Closes #4141

### DIFF
--- a/web/src/services/topicService.test.ts
+++ b/web/src/services/topicService.test.ts
@@ -17,12 +17,15 @@
 
 import {
   createTopic,
+  deleteTopic,
   getTopicConsumerPage,
   getTopicConsumers,
   getTopicRoutes,
   listAllTopics,
   listTopics,
+  updateTopic,
 } from './topicService';
+import { topics as mockTopics } from '../mock/topics';
 
 vi.mock('./dataMode', () => ({ isMockMode: () => true }));
 vi.mock('../config', () => ({
@@ -30,6 +33,12 @@ vi.mock('../config', () => ({
 }));
 
 describe('topic service mock data', () => {
+  afterEach(() => {
+    for (let idx = mockTopics.length - 1; idx >= 0; idx -= 1) {
+      if (mockTopics[idx].name.startsWith('instance-scope-test-')) mockTopics.splice(idx, 1);
+    }
+  });
+
   it('returns copied topic rows', async () => {
     const first = await listTopics({ search: 'order-create' });
     expect(first[0].name).toBe('order-create');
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
Fixes the mock `TopicService` to scope topic operations (create/update/delete) by `instanceId`, preventing cross-instance interference. Previously, operations on one instance could accidentally match topics belonging to another instance.

### Review
- **Correctness**: The `matchesTopicInstance` helper provides consistent matching logic. Applied correctly to all three mutation paths:
  - `createTopic`: duplicate check now instance-scoped
  - `updateTopic`: finds the right topic by instance
  - `deleteTopic`: only removes the topic for the specified instance
- **Tests**: Excellent regression coverage — tests verify that operations on instance A don't affect instance B's topics. The `afterEach` cleanup is good practice.
- **Design**: Clean, focused fix. The helper method avoids duplicating the matching logic across three call sites.

LGTM 👍

---

lizhimins:
Thanks for the contribution. We're closing this one on design grounds rather than implementation quality.

Per the Studio instance model spec (`docs/studio-instance-model-spec.md`), topic and group names are **globally unique within a single Studio deployment** — the unique key is on `name`, not on `(cluster, instance, name)`. Scoping mock topic mutations by `instanceId` so that two instances can hold same-name topics encodes the opposite assumption into the mock layer, which then diverges from the real backend contract the mock is supposed to mirror. A mock that permits a state the server rejects will hide exactly the class of bug it should surface.

If you're hitting cross-instance state bleed in mock mode, the right fix is to make the mock enforce the same global-uniqueness rule the server does (and reject the second create), not to partition it per instance. Happy to review a PR in that direction.

Note also that separate Studio deployments are the supported answer for same-name resources across daily/staging/production.

Aias00:
Reworked from the latest `rocketmq-studio` branch in the maintainer-recommended direction. The per-instance Topic identity changes are gone; mock creation now enforces deployment-wide name uniqueness. The focused rework is resubmitted as #4338 against the updated #4141 scope.

PR #4148, opened 2026-09-08
[ISSUE #4143][ISSUE #4147] feat(ai): page message queries and support custom trace topics

### LINKED ISSUE #4143: [Studio][Feature] Allow the AI message trace tool to use a custom trace topic
## Problem

The Studio message service supports an optional custom trace topic, but the AI/MCP tool `rmq.message.trace` only forwards the message ID and source Topic name. AI and CLI callers are therefore forced to use the default trace topic even when a deployment stores traces elsewhere.

## Expected behavior

- Add an optional `traceTopicName` string to the typed `rmq.message.trace` input and catalog schema.
- Forward a non-blank value to the existing four-argument `MessageService.getMessageTrace` overload.
- Keep missing or blank `traceTopicName` on the existing three-argument service path so provider/default behavior is unchanged.
- Keep `instanceId`, `topicName`, and `msgId` semantics, output schema, permissions, and risk classification unchanged.

## Acceptance criteria

- The message tool catalog advertises optional `traceTopicName` and rejects unknown properties.
- Handler coverage proves a custom trace Topic reaches the four-argument service call.
- Handler coverage proves missing and blank values retain the three-argument path.
- Existing trace output projection remains unchanged.
- No new endpoint or provider-specific behavior is introduced.

## Scope

This issue covers only message-ID trace lookup in the current typed tool framework. Trace lookup by business key remains separate work.

### LINKED ISSUE #4147: [Studio][Performance] Bound AI message query results
## Problem

The read-only AI/MCP tools `rmq.message.query` and `rmq.message.query_by_topic` currently return
flat message lists and always include message bodies. Provider-bounded scans can still return many
messages, and each body may be large, so one tool response can consume several megabytes and a
disproportionate amount of LLM context.

Broker message scans do not support arbitrary cursor seeks. Exposing `page` and `pageSize` would
therefore advertise random access that providers cannot honor reliably.

## Expected behavior

- Add one optional `limit` input to both message query tools.
- Default to 20 rows and apply a server-side cap of 100; values above 100 are reduced to 100.
- Return `items`, `resultMayBeTruncated`, and `skippedCount` instead of a paging envelope.
- Define `skippedCount` as rows omitted by `limit` from the provider-bounded result.
- Keep `resultMayBeTruncated=true` when rows were skipped locally or the provider may have stopped
  at its own scan bound.
- Add `includeBody` to both tools, defaulting to false.
- Omit `body`, `bodyEncoding`, and `bodyTruncated` unless `includeBody=true`.
- State explicitly that the body omission is a breaking output-contract change for callers that
  currently read `items[].body`.
- Document that `rmq.message.query_by_offset` intentionally remains different: it returns at most
  one message and always includes body fields.
- Preserve identifier priority `msgId`, then `uniqueKey`, then `key`, and keep existing L1
  permissions.

## Acceptance criteria

- Catalog validation covers `limit`, rejects values below 1, and rejects legacy paging fields.
- Handler tests cover default/custom limits, the 100-row server cap, body omission/inclusion,
  identifier priority, known skipped counts, and provider truncation.
- `query_by_topic` receives the same row and body bounds as `query`.
- Output schema tests cover results with and without body fields.
- The generated rmqctl catalog is committed and `make catalog-verify` passes.
- No new tool, REST endpoint, service API, or provider-specific implementation is introduced.

## Scope

This issue targets the current typed handler layout and `tool-catalog/tools/message.yaml`. Request
payload guarding and unrelated message tools are unchanged.

Aias00:
## Summary

Rebuilt this PR on the new MCP tool framework merged in #4225. The old handlers and monolithic catalog were removed upstream; this change does not restore them.

- Add optional `page` and `pageSize` (default 1/20, maximum page size 100) and `includeBody` (default false) to `rmq.message.query` and `rmq.message.query_by_topic`. The latter now owns the topic-only query path that the previous tool handled.
- Use the existing `MessageService.queryMessagesPage` for message-id, business-key and topic queries; retain the direct unique-key lookup and present its 0..1 result in the same paged envelope.
- Return `items`, `total`, `page`, `size` and `resultMayBeTruncated`; omit body, bodyEncoding and bodyTruncated unless explicitly requested. Other message tools retain their existing `MessageItem` contract.
- Add optional `traceTopic` to `rmq.message.trace`, forwarding a supplied value to the existing four-argument service while retaining the three-argument provider path for absent or blank values.
- Keep the existing read-only L1 permissions and identifier priority (`msgId`, `uniqueKey`, `key`).

## Verification

- Server full suite with `-Dspring.profiles.active=dev`: 2350 tests, 0 failures/errors/skips on commit `d7b9e565b`.
- Restored the `rocketmq-studio-mysql` Compose container using current bind mounts; it remains healthy on host port 3306 with the existing named volume intact. Tests used a separate temporary database and restricted user, both removed afterward.
- Default-profile MySQL integration smoke: the three classes that previously could not initialize passed 7/7. The full MySQL-configured invocation ran 2350 tests with 0 assertion failures and one configuration error: `AuthCookieProfileTest` explicitly selects `dev`/H2, but the global test-only MySQL URL override reached that context. All other 2349 tests passed in that invocation; the dev test passed 1/1 separately without the override.
- Targeted handler/catalog/MCP discovery/HTTP tool/MessageService tests: 64 passed; final focused tests passed after the last test-only assertion.
- `mvn -B -ntp -DskipTests package`, Checkstyle and `git diff --check`: passed. Independent read-only review found no Critical/Important issues.

Closes #4143
Closes #4147


RockteMQ-AI (approved):
### Summary
This PR adds pagination and optional body inclusion to the AI message query tool, preventing unbounded result sets that could overwhelm LLM contexts or cause performance issues.

The changes are clean, well-structured, and consistent with the existing `ToolListPagination` pattern used by other tools. The paged envelope (`items`, `total`, `page`, `size`, `resultMayBeTruncated`) provides clear semantics for AI consumers.

### Findings
- **[Info]** `MessageQueryToolHandler.java:63` — The `resultMayBeTruncated` flag is a nice touch for AI transparency. Consider also documenting the default page size cap (100) in the tool YAML description for LLM awareness.
- **[Info]** `rmq-tools.yaml:130` — Good that `pageSize` maximum is enforced at the schema level. This prevents abuse at the input validation layer.

### Suggestions
The implementation looks solid. The test coverage properly validates the new pagination behavior and the schema-driven page size cap.

---

RockteMQ-AI (approved):
### Summary
New commits since the previous approval (Sept 8) extend the pagination and message query improvements:

**New changes:**
- Added `page`, `pageSize` (default 20, max 100), and `includeBody` (default false) parameters to message query tools
- New `MessageQueryOutput` record with pagination metadata (`total`, `page`, `size`, `resultMayBeTruncated`)
- Updated handlers to use `MessageService.queryMessagesPage` for proper pagination
- Added custom trace topic support to `MessageTraceInput`
- Updated YAML tool definitions with new parameters and output schema
- Added comprehensive tests for the new functionality

**Review:**
The changes are well-structured and follow existing patterns. The pagination prevents unbounded result sets (critical for AI tool usage), and the optional body inclusion gives users control over response size. The trace topic customization is a useful addition for non-default trace scenarios.

**Verdict:** LGTM — approving the extended pagination and trace topic features.

---

lizhimins (changes requested):
Thanks for this — the gap you identified is real and still open on the branch, so the intent is welcome. I have to ask for a rework against the current baseline rather than a rebase, though, because the tool layer was restructured this week and every hunk here targets a contract that no longer exists.

**The problem is confirmed on current `rocketmq-studio`:**

- `MessageQueryToolHandler` still returns a flat `ListOutput<MessageItem>` with no row cap, and `MessageItem.from(...)` always populates `body` — there is no `includeBody`.
- There is no byte guard on the output side: `AiPayloadGuard` only validates request payloads. With the provider limits `KEY_QUERY_MAX = 64`, `DEFAULT_TOPIC_LIMIT = 200` and `MAX_BODY_DISPLAY_BYTES = 64KB`, a `query_by_topic` call can put roughly 12MB into one tool response.
- `MessageTraceToolHandler` still calls the three-argument overload, while the service layer has supported an explicit trace topic for a long time (`MessageService`, `InstanceProvider`, `MessageController`). That part is purely a tool-layer omission.

**What changed underneath you:**

1. The handlers moved to `ops/ai/tool/handler/message/`, take a typed record instead of `Map<String, Object>`, and read the target from the execution context (`context.instanceId()`) rather than from a `cluster` body field.
2. `tool-catalog/rmq-tools.yaml` was split per domain — the message tools now live in `tool-catalog/tools/message.yaml` (query, `query_by_topic`, and trace are all in there). `additionalProperties: false` plus the output-schema validation in `ToolSchemaValidator` mean the response envelope has to be declared, not just returned.
3. The tool surface was converged to 40 tools, so this needs to fit the existing tools rather than add any.

**How we'd like it redone:**

- **Reuse the existing paging path.** `MessageService.queryMessagesPage(...)` already pages with a bounded `pageSize`; there is no need for a new service API.
- **Match the established envelope.** Input should nest a `PageRequest` (see `MessageQueryDlqInput` for the shape) and output should be `PageOutput{page, pageSize, total, items}` from `contract/common`, not the ad-hoc `page`/`size` pair in this PR. `resultMayBeTruncated` is fine as an additional field.
- **Apply `includeBody` to `query_by_topic` too** — that is the path that can actually reach the multi-MB case, so bounding only the key query leaves the larger hole open.
- **Rename `traceTopic` to `traceTopicName`.** The instance model spec (`docs/studio-instance-model-spec.md`) settles on `topicName`/`groupName` for resource-name parameters, so a bare `traceTopic` would be the odd one out. Keep it optional.
- **Update the tests for the new layout.** The handler tests moved with the handlers, and `ToolCatalogTest` now builds a synthetic catalog, so the catalog assertions here need rewriting rather than rebasing.

Sorry about the timing — the restructure landed the same week you opened this. Happy to review the new version; the row cap and `includeBody` are the parts I'd most like to see.

Aias00:
Reworked the branch from the current `rocketmq-studio` baseline following the requested contract direction:

- inputs now use nested `PageRequest page`, matching `MessageQueryDlqInput`;
- outputs reuse common `PageOutput` and serialize `page`, `pageSize`, `total`, and `items`, with `resultMayBeTruncated` as the only message-specific top-level field;
- `includeBody` applies to both `query` and `query_by_topic`, and all three body fields are omitted by default;
- message-ID, key, and Topic paths reuse `queryMessagesPage`; unique-key keeps its direct lookup and is wrapped as a 0..1 page;
- `traceTopic` was renamed to optional `traceTopicName`; blank/absent values retain the three-argument path;
- catalog and output-schema tests were rewritten for the current shard layout.

The new head is a single signed-off commit. The full dev/H2 server suite passes 2361/2361, and package/Checkstyle/diff checks pass.

Aias00:
Superseded by #4345, which reimplements this work from the latest `rocketmq-studio` baseline following the maintainer review.

PR #4154, opened 2026-09-08
[Studio] feat: add Ops NameServer configuration preflight

coder999o:

Add a client-side preflight view on the Studio Ops page so operators can verify NameServer settings before updating runtime configuration. The panel validates saved NameServer address chains, highlights invalid or duplicated endpoints, confirms whether the current and selected NameServer are usable, and previews pending NameServer, TLS, and VIP channel changes.

It also blocks obviously invalid or duplicate NameServer additions before sending them to the API while keeping backend validation as the final authority.

## Brief changelog

- Added NameServer endpoint/address parsing and Ops preflight analysis helpers.
- Added an Ops page preflight panel with summary, issues, pending changes, recommendations, and address diagnostics.
- Blocked invalid or duplicate NameServer additions on the client side.
- Added unit and page tests for parsing, readiness, pending changes, unavailable/read-only state, and duplicate-add blocking.

## Verifying this change

- `npm run test -- src/utils/opsNameServerPreflight.test.ts src/pages/studio/__tests__/Ops.test.tsx`
- `git diff --check`
- `npm run lint -- src/pages/studio/Ops.tsx src/pages/studio/__tests__/Ops.test.tsx src/utils/opsNameServerPreflight.ts src/utils/opsNameServerPreflight.test.ts`
- `npm run build`

### DIFF
--- a/web/src/pages/studio/Ops.tsx
+++ b/web/src/pages/studio/Ops.tsx
@@ -15,19 +15,25 @@
  * limitations under the License.
  */
 
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import {
   Alert,
   App,
   Button,
+  Card,
+  Descriptions,
+  Flex,
   Input,
   Popconfirm,
   Select,
   Space,
   Switch,
+  Table,
+  Tag,
   Tooltip,
   Typography,
 } from 'antd';
+import type { ColumnsType } from 'antd/es/table';
 import { FloppyDisk, Plus, Trash } from '@phosphor-icons/react';
 import { useLang } from '../../i18n/LangContext';
 import useAuthStore from '../../stores/authStore';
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds a client-side NameServer configuration preflight panel to the Studio Ops page, enabling operators to validate NameServer settings, detect invalid/duplicate endpoints, and preview pending changes before applying runtime configuration updates.

### Review
- **Correctness**: Endpoint parsing handles all expected formats (host:port, IPv6 [host]:port, semicolon/comma-separated chains). Port validation, hostname validation, and duplicate detection are all correct ✅
- **Architecture**: Clean separation between parsing logic (`opsNameServerPreflight.ts`), UI component (`Ops.tsx`), and tests. Backend validation remains the final authority ✅
- **UX**: Good readiness states (ready/warning/blocked), clear issue severity levels, and actionable recommendations ✅
- **Tests**: Comprehensive coverage for parsing, readiness analysis, pending changes, read-only state, and duplicate blocking ✅
- **Security**: Client-side blocking of invalid/duplicate additions prevents obviously bad requests; no sensitive data exposure ✅

### Minor Observations
- Chinese strings are hardcoded in the component (e.g., `'格式错误'`, `'重复'`, `'有效'`). Consider moving to the i18n system for consistency with the rest of the dashboard, though this is not blocking.
- The `useMemo` dependency on `namesrvAddrList` (array) works correctly but relies on reference equality from the parent — fine in practice.

LGTM — well-designed preflight validation with thorough test coverage.

---

lizhimins (changes requested):
Thanks for the effort that went into this — endpoint preflight before hitting the backend is a genuinely useful idea, and the ready/warning/blocked model is the right shape for it. I have to ask for changes, though: the preflight rules and the server's parser disagree in both directions, and the server side doesn't validate at all, so the panel would gate the UI on rules the backend doesn't share.

1. **Front-end rules contradict `NamesrvAddrParser`.** The authoritative parser (`NamesrvAddrParser.java:56-88`) requires `host:port`, does **not** supply a default port, splits on `[,;]`, joins with `,`, and delegates IPv6 to `InetAddressValidator`. The preflight is stricter in some places and looser in others:
   - Stricter: `HOST_PATTERN` (`opsNameServerPreflight.ts:98`) rejects underscores and trailing dots, and an in-string duplicate invalidates the whole value (`:223`). Addresses the server accepts therefore get their update/delete buttons disabled (`Ops.tsx:305`) — the UI locks the user out of a valid configuration.
   - Looser: the bracketed form (`:119-133`) accepts `[anything]:port` as valid, while the server returns 400 (`:65`) — so "ready to add" lights up for an address that cannot be saved.
   - `normalized` joins with `;` (`:218`) while the server joins with `,` (`:52`).
   Please drive the front-end from the same rules as the parser (ideally by exposing the parser's verdict through the API) rather than maintaining a second, divergent implementation.

2. **The server doesn't validate, so the gating is browser-only.** `OpsService.java:67` only trims; `:44-56` always throws 501 and `:34-42` always returns `configurationAvailable=false` with an empty list. That means the panel renders as permanently blocked in a real deployment, and it double-reports alongside the existing Alert at `Ops.tsx:194-201`. All five gates live in the front end. We've returned two PRs recently for exactly this pattern — a control that only exists in the browser isn't a control. `OpsService` should run the address through `NamesrvAddrParser.normalize` and return the structured verdict; the UI should render it.

3. **Unknown is rendered as a definite negative.** Unresolved/unknown states come out as a red "no" rather than "unknown", which is the same clamping problem we flagged before: an absent answer should not be presented as a failed check.

4. **Roughly 170 of the 993 lines are duplication.** The `Issue`/`PendingChange` interfaces are structurally identical; `:135-205` has six same-shaped invalid returns; the three-state classification is repeated three times (from `Ops.tsx:85`); `Ops.tsx:218` hand-rolls a grey box that `components/InfoBanner.tsx` already provides (and hardcodes `#f0f0f0`, which breaks dark mode); host extraction duplicates `BrokerCluster.tsx:115-124`; and the parsing itself duplicates the server parser. Please reuse `InfoBanner` and extract the shared pieces.

5. **House rules.** The persistent explanatory panel uses `Alert` where the convention is `InfoBanner` (`Ops.tsx:171`, `:218`); around 40 hardcoded Chinese strings need to go into `translations.ts` (the `ops.*` keys already exist at 2130-2142); `ellipsis: true` at `:98` has no `title`/Tooltip so the full value is unrecoverable; the last column has no width and falls back to the 120 default in `table.ts:59`. Font sizes, `tableScrollX` and the absence of row-level `onClick` are all fine.

6. **Test coverage doesn't pin the risky parts.** Of the 18 cases (14 util incl. `it.each` x5, plus 4 for `Ops.tsx`), the IPv6 branch (`:119-133`) has **zero** coverage — deleting it entirely leaves the suite green. Untested: the port lower bound of 1, a bare `+9876`, underscore hosts, the empty-list case, selected-but-invalid, and the new delete gate. The two "client-side interception" cases click a disabled button, so `onClick` never fires and the handler guards at `Ops.tsx:347/389` could be deleted without failing anything. Also, `findByText` was relaxed to `findAllByText().length > 0`, which weakens existing assertions.

The smallest version of this that we could merge is: server-side normalization in `OpsService` returning a structured verdict, the front end rendering that verdict (no second parser), `InfoBanner` for the copy, i18n for the strings, and tests that fail when the gates are removed. Happy to review that.

PR #4158, opened 2026-09-08
[ISSUE #4157] fix(alert): bound native aggregation windows

### LINKED ISSUE #4157: [Studio][Bug] Reject native alert windows beyond snapshot retention
## Problem

Studio-native alert aggregation reads historical samples from `MetricSnapshotRepository.findRecent(...)` using `windowSeconds`, while `CollectorScheduler` deletes samples older than the configured `studio.alerting.snapshot-retention` (default `PT24H`). Alert rule create and update currently validate only that `windowSeconds` is non-negative.

A native rule can therefore request a window longer than the retained history. For example, a 48-hour AVG rule with the default 24-hour retention is accepted but is silently evaluated from an incomplete sample set, which can produce misleading thresholds and alerts.

## Expected behavior

- When `snapshot-retention` is positive, native alert rules must reject `windowSeconds` values longer than the configured retention.
- A window equal to the retention remains valid.
- `windowSeconds=0` keeps the current single-sample behavior.
- Non-native Prometheus rules are not constrained by the native snapshot store.
- Non-positive retention keeps the existing cleanup-disabled behavior and does not impose a window cap.
- Create, update, and import use the same write-path validation.

## Acceptance criteria

- Reject an oversized native window with HTTP/business code 400 before repository mutation.
- Apply the same check to create and update.
- Add focused coverage for oversized, equal-boundary, non-native, and cleanup-disabled cases.
- Do not change snapshot collection, retention cleanup, or aggregation math.

## Scope

This change validates writes against the currently configured retention. Migration or runtime reporting for rules created under an older, longer retention configuration is separate operational work.

Aias00:
## Summary

- inject the existing alerting retention configuration into `AlertService`
- reject native alert aggregation windows longer than a positive snapshot retention
- apply the check before repository mutation on both create and update paths
- preserve equal-boundary, single-sample, non-native Prometheus, and cleanup-disabled behavior

## Scope

The validation applies only to Studio-native metrics with `windowSeconds > 0`. A zero or negative snapshot retention retains the current cleanup-disabled semantics and does not impose a cap. Import already delegates to the same create path. Snapshot collection, cleanup, aggregation math, and existing-rule migration are unchanged.

## TDD evidence

Before implementation, the create and update regressions completed repository writes without rejecting a 24-hour-plus-one-second window under the default 24-hour retention. After the minimal Service validation, both reject with business code 400 before persistence.

## Verification

- `mvn -B -ntp -f server/pom.xml -Dtest=AlertRuleSnapshotRetentionTest,AlertServiceTest,AlertServiceDefaultRulesTest,NativeAlertRulePolicyTest,AlertRuleTransferServiceTest,AlertRuleControllerTest,ClusterAlertRuleControllerTest test` — 129 tests passed
- `mvn -B -ntp -f server/pom.xml -Dtest=StudioApplicationTest test` — 2 tests passed
- `mvn -B -ntp -f server/pom.xml -DskipTests package`
- Checkstyle — 0 violations
- `git diff --check`

Closes #4157

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.studio.ops.alert;
 
+import org.apache.rocketmq.studio.cluster.metrics.AlertingProperties;
 import org.apache.rocketmq.studio.cluster.metrics.MetricProfileService;
 import org.apache.rocketmq.studio.cluster.metrics.SemanticMetric;
 import org.apache.rocketmq.studio.common.exception.BusinessException;
@@ -38,6 +39,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 
@@ -61,6 +63,7 @@ public class AlertService {
     private final AlertRuleAssetService alertRuleAssetService;
     private final OperationAuditService operationAuditService;
     private final MetricProfileService metricProfileService;
+    private final AlertingProperties alertingProperties;
 
 
     public List<AlertRuleVO> listRules() {
@@ -169,6 +172,7 @@ public class AlertService {
         }
         rule.setName(rule.getName().trim());
         NativeAlertRulePolicy.validate(rule);
+        validateAggregationWindow(rule);
         rejectDuplicateSemanticRule(rule, null);
         log.info("Creating alert rule: {}", rule.getName());
         AlertRuleVO saved = saveNewRule(rule);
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (commented):
### Summary
This PR contains significant changes (262 lines). Please ensure:
- Adequate test coverage for new/modified logic
- No breaking changes to public APIs without migration path
- Thread safety for concurrent access patterns

---

lizhimins:
Thanks for digging into the native aggregation window. We don't think a hard 400 is the right shape here: a window longer than the retained samples already degrades gracefully — `NativeAlertEvaluationService.aggregate` simply aggregates over whatever snapshots exist and falls back to the single sample when none do, and the design doc describes the window as bounded by the local snapshot store. `snapshotRetention` is a runtime-tunable property, so binding rule validation to it would reject previously valid stored rules on unrelated edits (renaming a rule would 400 after an operator shortens retention) and would break cross-environment import, which runs through `createRule` inside one transaction. It also puts `Duration.parse` of operator-supplied config on the request path, turning a config typo into a 500. If you'd like to pursue this, a better direction is documenting the effective window or warning in the UI rather than rejecting the write. Happy to review a revised approach.

Aias00:
Reworked from the latest `rocketmq-studio` branch in the maintainer-recommended direction. The hard backend validation has been removed entirely; the focused, non-blocking localized UI guidance is resubmitted as #4335 against the updated #4157 scope.

PR #4162, opened 2026-09-08
fix(acl): preserve Tencent role when deleting rule

### LINKED ISSUE #4161: [Studio][Bug] Deleting a Tencent ACL rule deletes the role account
## Description

## Steps to reproduce

## Actual behavior

## Expected behavior

## Code location

## Proposed scope

Aias00:
Fixes #4161.

## Problem

Studio maps each Tencent Cloud RocketMQ role to both an ACL user and a cluster-wide ACL rule. The two resources have separate delete actions, but the rule deletion path called Tencent Cloud `DeleteRole`, exactly like user deletion. Deleting a rule therefore also removed the role account and its access credentials.

Roles with both `PermRead` and `PermWrite` disabled were also returned as ACL rules with an empty action list.

## Changes

- Change `TencentAclService.deleteRule` to call `ModifyRole` with `PermRead=false` and `PermWrite=false`.
- Exclude permissionless roles from `listRules`, while leaving them visible in the ACL user list.
- Keep explicit ACL user deletion on `DeleteRole`.
- Add regression coverage for all three behavior boundaries.

## Testing

The two new behavior tests failed against the base commit:

- rule deletion invoked `DeleteRole` instead of `ModifyRole`;
- the rule list included a role with neither permission.

After the fix:

```text
mvn -B test -Dtest='TencentAclServiceTest,AclServiceTest'
Tests run: 79, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
Checkstyle violations: 0
```

Tests were run with Java 21.

## Compatibility

This is limited to the Tencent ACL provider. It does not change REST contracts, persistence, frontend code, or other vendors. Existing Tencent role accounts remain intact when their mapped rule is deleted; the separate user deletion action remains destructive by design.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentAclService.java
@@ -102,6 +102,10 @@ public class TencentAclService {
                 if (role == null || !StringUtils.hasText(role.getRoleName())) {
                     continue;
                 }
+                if (!Boolean.TRUE.equals(role.getPermRead())
+                        && !Boolean.TRUE.equals(role.getPermWrite())) {
+                    continue;
+                }
                 if (requestedPrincipal != null
                         && !requestedPrincipal.equals(role.getRoleName())) {
                     continue;
@@ -242,11 +246,13 @@ public class TencentAclService {
     public void deleteRule(String instanceId, String principal) {
         Context context = resolve(instanceId);
         String roleName = requireRoleName(principal, "ACL principal");
-        DeleteRoleRequest request = new DeleteRoleRequest();
+        ModifyRoleRequest request = new ModifyRoleRequest();
         request.setInstanceId(context.cloudInstanceId());
         request.setRole(roleName);
+        request.setPermRead(false);
+        request.setPermWrite(false);
         clientFactory.call(context.credentialId(), context.regionId(),
-                client -> client.DeleteRole(request));
+                client -> client.ModifyRole(request));
     }
 
     private static String requireRulePrincipal(AclRuleVO rule) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a destructive bug in the Tencent ACL provider where `deleteRule` was calling `DeleteRole` (destroying the role account) instead of `ModifyRole` (disabling permissions). The fix correctly:

1. Changes `deleteRule` to use `ModifyRole` with `PermRead=false, PermWrite=false` — preserving the role account while removing its ACL permissions
2. Filters out permissionless roles from `listRules` output, keeping the rule list clean
3. Leaves `deleteUser` using `DeleteRole` as expected (explicit user deletion is intentionally destructive)

The null-safe permission checks (`!Boolean.TRUE.equals(...)`) correctly handle null values. Three regression tests cover the key behaviors.

LGTM — important fix that prevents unintended data loss in Tencent ACL.

---

lizhimins:
Thanks for the patch. We're closing it because it doesn't match how the Tencent side actually behaves, and it splits the delete semantics inside one service.

Concretely, from the current code: `deleteUser` (`TencentAclService.java:207-214`) and `deleteRule` (`:242-249`) both call `DeleteRole`, because in trocket a "Role" *is* the ACL principal. Routing `deleteRule` to `ModifyRole` with `permRead=false`/`permWrite=false` leaves the cloud-side Role in place after Studio has told the user the rule was deleted — the resource is still there in the Tencent console, and the two delete entry points now mean different things.

The companion filter in `listRules` compounds it: hiding roles whose read and write perms are both false makes Studio's view diverge from the cloud's actual state, and a Role that reaches that state outside Studio becomes invisible and therefore unmanageable from here.

We also can't confirm that `ModifyRole` with both permissions false is an accepted, meaningful state in the trocket API rather than an error or a no-op. If you have console/API evidence that it is, and a story for the leftover Role and the divergent `deleteUser` path, please reopen the discussion in an issue first — the "don't destroy a shared cloud Role on a rule delete" instinct is reasonable, but it needs to hold against the real API.

PR #4164, opened 2026-09-08
fix(aliyun): report truncated message queries

### LINKED ISSUE #4163: [Studio][Bug] Aliyun message queries do not report truncated results
## Description

The Aliyun provider bounds message queries to five pages of 20 records, but it only implements the legacy `queryMessages` method. The default `queryMessagesDetailed` implementation therefore wraps every bounded result as complete, even when the Aliyun API reports more records than the provider scanned.

This is especially misleading for Tag queries because Aliyun `ListMessages` does not expose a Tag request parameter. Studio filters Tags locally after fetching each page, so a bounded scan can return no matching rows while matching messages still exist beyond the first 100 records.

## Actual behavior

- Aliyun message queries stop after at most 100 source records.
- `MessageQueryResult.mayBeTruncated` remains `false`.
- The message page can present the bounded result, including an empty Tag-filtered result, without the existing truncation warning.

## Expected behavior

The provider should report `mayBeTruncated=true` when it reaches the five-page budget before the query is known to be exhausted. A short page or a trustworthy `totalCount` showing that all source rows were fetched should remain a complete result.

## Code location

- `server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProvider.java`

## Proposed scope

- Override `queryMessagesDetailed` in the Aliyun provider.
- Track raw source rows independently of rows retained by the local Tag filter.
- Use short-page and `totalCount` evidence to distinguish complete results from budget exhaustion.
- Add regression coverage for over-budget, Tag-filtered, exact-boundary, and short-page queries.

Aias00:
Fixes #4163.

## Problem

The Aliyun provider scans at most five pages of 20 source messages, but it only implemented the legacy `queryMessages` method. The default `queryMessagesDetailed` wrapper therefore marked every bounded result as complete, even when `totalCount` showed that more than 100 messages existed.

Aliyun `ListMessages` has no Tag request field, so Studio filters Tags locally. Using the filtered result size to infer completeness is unsafe: the first 100 source rows can contain no matching Tag while matching rows still exist beyond the provider budget.

## Changes

- Implement `AliyunInstanceProvider.queryMessagesDetailed` and keep the legacy list method delegating to it.
- Count raw source rows independently of locally retained Tag matches.
- Treat a short page or a positive `totalCount` reached by the raw fetch count as complete.
- Return `MessageQueryResult.truncated` when five full pages are consumed without evidence that the query ended.

The existing message API and frontend already propagate `mayBeTruncated`, so no REST or web change is required.

## TDD evidence

Before the implementation, the new over-budget and Tag-filtered tests failed because `mayBeTruncated` remained false. After the implementation:

```text
mvn -B test -Dtest='AliyunInstanceProviderTest#queryMessages*,MessageServiceTest'
Tests run: 22, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
Checkstyle violations: 0
```

The four added boundaries cover:

- more than 100 source rows;
- an empty locally Tag-filtered result after exhausting the source budget;
- exactly 100 total rows;
- a short final page with no `totalCount`.

## Existing target-branch test failure

Running the entire `AliyunInstanceProviderTest` class on the current target branch also executes the unrelated `getGroupProgressShouldMapLagRowsTest`, whose existing lag-row expectation fails and is specifically addressed by open PR #4135. This change does not touch that test or code path; the focused message-query suite above is green.

## Compatibility

The change is limited to the Aliyun provider's existing bounded query path. It adds only the truncation metadata already supported by `MessageQueryResult`; query parameters, returned message rows, provider budget, persistence, and other vendors are unchanged.


RockteMQ-AI (approved):
### Summary
This PR implements `queryMessagesDetailed` for the Aliyun provider to correctly report truncation when message queries hit the page budget. Previously, the legacy `queryMessages` path always reported `truncated=false`, even when the page was capped at 64 source rows.

The implementation is clean:
- `fetched` counts raw source rows independently of tag filtering
- `mayBeTruncated` is correctly set when the last page is reached with a full page
- `allFetched` condition properly handles `totalCount` short-circuit
- Four boundary test cases cover the key scenarios (short page, totalCount match, full last page, tag filtering)

LGTM — solid fix with good test coverage.

---

Issue #3292, opened 2026-09-05
[Studio][Bug] Broker config-diff modal applies stale responses: closed dialog reopens and a slower response overwrites a newer request


unbridled-41:
## Problem

`openBrokerConfigDiff` in `web/src/pages/cluster/index.tsx` applies the `getBrokerConfigDiff` response to `brokerConfigDiffState` unconditionally, with no request generation guard. Two triggers:

1. **Closed dialog reopens:** clicking 配置差异 on a Broker row and then 关闭 before the response arrives lets the pending response re-open the dialog on its own and render the diff.
2. **Wrong cluster's data:** while cluster A's diff request is in flight, clicking 配置差异 on a cluster B row is not blocked; when A's slower response lands it overwrites the modal with A's title and A's differences, presented as if they were B's. A stale failure also clears the loading state of a newer in-flight request.

## Evidence / source

- Code inspection of the current `rocketmq-studio` branch (base commit 36126024): the sibling NameServer diff flow is guarded by `nsConfigDiffRequestRef` and has the regression test `does not reopen a closed NameServer config diff when its request finishes`; the Broker diff flow has neither a generation ref nor invalidation in either close handler.
- Deterministic Vitest reproduction with deferred promises in `ClusterPage.test.tsx` (both tests red on the unfixed source): `does not reopen a closed Broker config diff when its request finishes` and `keeps the requested broker config diff when a slower response finishes last`.

## Impact

Read-only correctness of the Broker 管理 drift view: an operator can read one cluster's config differences while believing they belong to the requested cluster, and a closed dialog can reappear. No configuration write requests are involved.

## Expected behavior / acceptance criteria

- Only the current Broker-diff request generation may publish open/loading/result state or error handling.
- Both close paths and unmount invalidate older requests.
- Regression coverage for close-before-response and slower-response-overwrites-newer-request; both must fail on the unfixed code.

## Related work

- Closed #2729 added request generations for the **NameServer** config-diff and connection-test modals; the Broker diff modal was outside that scope and remains unguarded — this issue tracks that remaining instance.
- Wrong-instance targeting of registry config actions is a separate defect tracked in #3103 (draft PR #3151).

## PR

Draft PR #3154 (includes the fix and the two regression tests).

RockteMQ-AI:
**Issue Evaluation**

Category: `type/bug` | Status: **Confirmed** | Severity: **Medium**

Verified against `rocketmq-studio` branch (base commit 3612602).

[...]

PR #4166, opened 2026-09-08
fix(cluster): discard superseded broker config preview responses

### LINKED ISSUE #4165: [Studio][Bug] Slow broker config preview can overwrite a newer preview with stale values
## Problem

## Evidence

## Impact

## Expected behavior

## Related work

## PR

Fix: #4166.

unbridled-41:
Fixes #4165.

## Problem / Evidence

On the Cluster page (Broker 管理), the broker config preview panel renders whichever `previewClusterConfig` response arrives last. `handleConfigPreview` (`web/src/pages/cluster/index.tsx`, base `0a596661`, lines 648-669) awaits the request and then unconditionally commits the result — no request-generation guard — while every comparable async path in the same file (`nsConfigDiffRequestRef`, `brokerConfigDiffRequestRef`, `connectionTestRequestRef`, registry lists) is guarded.

Reproduction: open 配置 for a broker → change a value → 预览 (response A slow) → Cancel → reopen 配置 → change values → 预览 (response B arrives) → slow response A resolves last → the panel shows A's diff while the form holds different values, and A's success toast fires while B's loading state is left stuck.

A new regression test (`discards a superseded broker config preview response`) fails on the unmodified base: after the stale response resolves, `getByText('defaultTopicQueueNums=24')` fails because the panel was overwritten with the stale preview's `defaultTopicQueueNums=16`.

## Root cause / Fix

Root cause: the preview async handler lacks the request-generation guard used by its sibling flows.

Fix: add `configPreviewRequestRef`; bump it at the start of `handleConfigPreview`, and only commit state/toast/loading when the generation is still current. Invalidate the generation when the config dialog opens (`handleConfigOpen`), when it closes (`onCancel`), and when a form value changes (`onValuesChange`, which already cleared the panel). Mirrors the twin guard pattern from the #3292 fix.

## Priority & scoring

- Impact 30/40 (misleading safety-check panel before a cluster-wide broker config apply) + breadth 13/20 (single dialog, but the same file's identical races were each accepted as bugs) + reproducibility 18/20 (deterministic deferred-promise regression) + maintainability value 14/20 (completes the file's guard pattern) = **PRIORITY 75** ≥ 70.
- Mechanical pattern port with a red→green regression → **FIX_CONFIDENCE 95** ≥ 80.

## Tests

- Red: `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx -t "discards a superseded broker config preview response"` on base → `Tests 1 failed | 25 skipped` (`Unable to find an element with the text: defaultTopicQueueNums=24` — stale preview overwrote the latest one).
- Green: same command after the fix → `Tests 1 passed`.
- Module suite: `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx` → `Test Files 1 passed (1)`, `Tests 26 passed (26)` (25 pre-existing + 1 new).
- Full web suite: `npx vitest run` → 945 tests; first run `2 failed | 943 passed` (AclPage ×1, ConsumerPage ×2 — none of these files is touched by this change); reruns of those two files in isolation with the fix applied → `Tests 54 passed (54)`. This matches the documented load-flakiness of those files under full-suite parallelism (see verification comment for the base-branch stash check).
- `npx tsc --noEmit` clean; `npx eslint` on both changed files clean; `npm run build` succeeds (`✓ built in 9.81s`).

## Risk

Low. The change only adds a generation guard around an existing async handler and bumps the generation at three existing invalidation points; no API, data, or visual changes. The one behavioral difference is intended: responses from superseded preview requests no longer write state or toast. The submit path (`handleConfigSubmit`) is intentionally untouched — its result is guarded by `confirmLoading` and revalidates on the server.

### DIFF
--- a/web/src/pages/cluster/__tests__/ClusterPage.test.tsx
+++ b/web/src/pages/cluster/__tests__/ClusterPage.test.tsx
@@ -23,6 +23,7 @@ import { MemoryRouter } from 'react-router-dom';
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 import type {
   BrokerConfigDiffResult,
+  ClusterConfigPreviewResult,
   ClusterInfo,
   ClusterProbeResult,
   NameServerConfigDiffResult,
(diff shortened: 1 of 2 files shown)


unbridled-41:
## Verification evidence

- **CI**: for this PR's head SHA `fb3cbcf58c9f88c9c9b15590957ba2ff82bea7d4`, `gh api /repos/apache/rocketmq-dashboard/actions/runs?head_sha=<sha>` returns exactly one run: `CI (pull_request)` with `conclusion=startup_failure`, and `/commits/<sha>/check-runs` reports `total_count=0` — the upstream workflow never starts (same for recent base-branch runs), so nothing here can pass or fail CI.
- **Red** (base `0a596661`, only the new test added): `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx -t "discards a superseded broker config preview response"` → `Tests 1 failed | 25 skipped`. Failure: `TestingLibraryElementError: Unable to find an element with the text: defaultTopicQueueNums=24` — after the stale first preview resolved last, the panel contained the stale `defaultTopicQueueNums=16` tag instead of the latest `defaultTopicQueueNums=24` (the exact defect).
- **Module tests** after the fix: `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx` → `Test Files 1 passed (1)`, `Tests 26 passed (26)` (25 pre-existing + 1 new).
- **Full suite + baseline comparison** after the fix: `npx vitest run` → `Test Files 2 failed | 113 passed (115)`, `Tests 2 failed | 943 passed (945)` in run 1 (AclPage ×1, ConsumerPage ×2); a second full run produced `3 failed | 942 passed` (same two files, load-dependent counts). Neither file is touched by this change. To attribute them to the environment and not the fix: `git stash push -- src/pages/cluster/` (fix removed) + `vitest run src/pages/instance/__tests__/AclPage.test.tsx src/pages/instance/__tests__/ConsumerPage.test.tsx` → `Tests 54 passed (54)`; `git stash pop` and the same two files in isolation with the fix applied → `Tests 54 passed (54)`. So the failures only occur under full-suite parallel load, on files this PR does not modify — the previously documented load flakiness.
- **Build**: `npx tsc --noEmit` clean; `npx eslint src/pages/cluster/index.tsx src/pages/cluster/__tests__/ClusterPage.test.tsx` clean; `npm run build` → `✓ built in 9.81s` (covers the CI web-build step, which cannot run due to the startup_failure above).
- **Diff self-check**: `git show fb3cbcf5 --numstat` → `2 files changed`: `web/src/pages/cluster/index.tsx` +17/−2 (guard ref + 3 invalidation points + guarded commit), `web/src/pages/cluster/__tests__/ClusterPage.test.tsx` +70/−0 (1 new regression, 1 new type import). No unrelated changes.

RockteMQ-AI (approved):
### Summary
This PR adds a request-generation guard (`configPreviewRequestRef`) to `handleConfigPreview` in `web/src/pages/cluster/index.tsx`, preventing stale async preview responses from overwriting newer ones. The pattern is consistent with the existing guards (`nsConfigDiffRequestRef`, `brokerConfigDiffRequestRef`, etc.) already used throughout the same file.

The fix is well-structured:
- Guard is bumped at the start of each preview request
- State/toast/loading commits are gated by generation check
- Generation is properly invalidated on dialog open/close and value changes
- Regression test demonstrates the race condition and verifies the fix

LGTM — clean, minimal change that closes a real race condition.

---

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

Thank you for the detailed report. This is a well-documented race condition in the broker config preview handler.

**Root Cause:** `handleConfigPreview` in `web/src/pages/cluster/index.tsx` lacks a request-generation guard, allowing stale async responses to overwrite newer preview data. The sibling handlers in the same file all use this guard pattern.

**Impact:** Users may see incorrect config diff in the preview panel before applying cluster-wide broker configuration changes — a misleading safety-check result.

**Severity:** Medium — affects a single dialog flow but has safety implications for cluster config operations.

The included regression test and fix description are thorough. An automated fix proposal may be generated based on this analysis.

---
*Automated evaluation by RockteMQ-AI*

lizhimins:
Thanks for spotting that the config preview call has no request-id guard — you're right that it is a different call site from the broker/nameserver diff guards we merged earlier, so this isn't a duplicate. We're closing it because the reachable impact is display-only. `configPreview` never feeds the write path: `handleConfigSubmit` rebuilds the request from the form values, so a late preview cannot cause a wrong update. Mainline also clears the preview on modal open, on cancel, and on any form value change, so the only residual window is "cancel and reopen for another cluster while a preview is still in flight", and the next keystroke wipes it. Finally, this would be the fourth hand-rolled request-id ref in this file; we'd rather extract one shared helper and apply it to all four call sites at once. A PR that does that extraction would be very welcome.

unbridled-41:
@lizhimins Thanks for the explanation — resubmitted as #4331. It adds no fourth standalone request-id ref; instead it extracts one shared request-generation helper (`begin`/`isCurrent`/`invalidate`) and applies it to all the hand-rolled guards in the page as well as the previously unguarded broker config preview, with a regression covering the cancel-and-reopen window you described.

PR #4170, opened 2026-09-08
fix(message): preserve Unicode property characters

### LINKED ISSUE #4167: [Studio][Bug] Preserve Unicode code points when truncating message properties
## Description

## Steps to reproduce

## Actual behavior

## Expected behavior

## Code location

## Proposed scope

Aias00:
Fixes #4167.

## Problem

`MessagePropertyDisplay` capped property values with UTF-16 `String.length()` and `substring(0, 1024)`. A supplementary Unicode character uses a surrogate pair, so a truncation boundary between the two code units produced an unpaired surrogate. The same utility feeds both regular message details and DLQ property output.

The UTF-16 length check also treated each supplementary character as two displayed characters: a value containing exactly 1024 emoji was incorrectly marked oversized and truncated.

## Changes

- Count the property value limit in Unicode code points.
- Compute the truncation index with `offsetByCodePoints`, preserving complete surrogate pairs.
- Use the same code-point rule in `hasOversizedProperty` so the value and truncation metadata cannot disagree.
- Keep the existing 1024-character limit and `...` suffix unchanged for ASCII/BMP text.

## TDD evidence

Before implementation, the new regressions showed both failures:

- 1023 ASCII characters followed by an emoji were cut into an isolated high surrogate;
- exactly 1024 emoji were incorrectly reported as oversized.

After the fix:

```text
mvn -B test -Dtest='MessagePropertyDisplayTest,RocketMQMessageProviderTest,RocketMQDLQProviderTest'
Tests run: 82, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
Checkstyle violations: 0
```

Tests were run with Java 21.

## Compatibility

No REST contract, persistence, dependency, or frontend changes. ASCII and BMP property values retain their existing behavior; supplementary Unicode values now use user-visible code-point counting and remain well formed after truncation.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/MessagePropertyDisplay.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/MessagePropertyDisplay.java
@@ -25,10 +25,11 @@ import java.util.Map;
 
 /**
  * Shared rendering limits for message property maps, used by the message explorer and the DLQ
- * drawer so both apply the same {@value #MAX_PROPERTIES}-entry / {@value #MAX_PROPERTY_VALUE_CHARS}-char
- * caps instead of duplicating the logic. {@link #userProperties} additionally drops the broker-set
- * system keys ({@link MessageConst#STRING_HASH_SET}) so a view labelled "user properties" is not
- * crowded out by system entries once the cap and alphabetical ordering are applied.
+ * drawer so both apply the same {@value #MAX_PROPERTIES}-entry /
+ * {@value #MAX_PROPERTY_VALUE_CHARS}-code-point caps instead of duplicating the logic.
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fix Unicode surrogate pair handling in `MessagePropertyDisplay` — the previous `String.length()` / `substring(0, 1024)` approach could split supplementary characters (emoji) mid-surrogate-pair. The fix correctly uses `codePointCount()` and `offsetByCodePoints()` to truncate at code-point boundaries.

Clean, focused change with good test coverage for the edge cases. LGTM.

---

Aias00:
Closing as a duplicate of #4226, which implements the same code-point-safe message-property truncation fix and remains open for review.

PR #3160, opened 2026-09-04
fix(consumer): filter subscription mode server-side

tju-yxq:
## What is the purpose of the change

The Consumer Group page's Push/Pop selector previously filtered only the rows already loaded for the current page. The backend paged endpoint had no subscription-mode parameter, so the page total still described the unfiltered inventory and later pages could be unreachable for the selected mode.

## What changed

- add an optional `subscriptionMode` request parameter to `/api/groups/page`
- pass the normalized mode through `MetadataService`, `InstanceProvider`, `ApacheInstanceProvider`, and `MetadataProvider`
- filter Apache-backed groups in the database query using `message_model`, preserving correct totals and ordering
- retain compatible default methods for providers without native mode filtering
- send the selected mode from the Consumer Group page and stop filtering the current page client-side
- update controller/service/provider/API/page tests

## Verification

- backend focused tests: `ConsumerGroupControllerTest,RocketMQMetadataProviderTest,MetadataServiceTest,ApacheInstanceProviderTest,InstanceProviderTest` passed
- frontend focused tests: `ConsumerPage.test.tsx`, `consumerGroups.test.ts`, `consumerService.test.ts` passed (49 tests)
- Checkstyle: 0 violations
- frontend lint: 0 errors (existing warnings only)
- frontend build: passed
- full server suite: 2,003 tests, 0 failures, 0 errors (initial baseline)
- after the later rebase to `36126024`, focused backend and frontend suites and lint/build were rerun and passed
- full backend was also rerun on the rebased head: 2,036 tests with 2 failures, both in the pre-existing `AuthCorsIntegrationTest` mainline drift that also reproduces on clean upstream `36126024`. A previous clean-upstream run also showed the same Auth failures plus the pre-existing Aliyun lag-row drift; neither is introduced by this patch.
- full frontend on the rebased head: the default parallel run is not stable in this environment; one run reported many unrelated timeouts across files such as ACL, Alerts, Audit, Clients, Instance, Message, and Topic pages. The consumer-focused suites were rerun on that head: `ConsumerPage.test.tsx`, `consumerGroups.test.ts`, and `consumerService.test.ts` produced 50 passed and the same two pre-existing `ConsumerPage` client-stack failures that also reproduce on clean upstream `36126024`. I am not claiming a full-frontend green run for this head.
- production/config diff: 43 additions and 28 deletions across backend and frontend (41 production additions); this is the natural fix size, not padded with unrelated changes

Closes #3150

PR #4172, opened 2026-09-08
fix(consumer): filter groups by subscription mode on the server

### LINKED ISSUE #4171: [Studio][Bug] Consumer group subscription-mode filter only filters the current page and desyncs from the total
## Problem

On the instance Consumer Groups page, the "订阅模式" (subscription mode) select applies a **client-side filter** against `groups` — the rows returned for the current server page only — while the paginator and the subtitle keep showing the **unfiltered** server total.

## Evidence

`web/src/pages/instance/consumer.tsx` (base `0a596661`):

- `visibleConsumerGroups` (lines 154-162) filters `groups` by `modeFilter` client-side; used by the `filtered` memo (lines 470-472) feeding the table `dataSource`.
- `loadConsumerGroupPage` (lines 343-371) calls `listConsumerGroupPage({ instanceId, search, page, pageSize })` — no mode parameter — and the table paginator (line 1545) uses `total: totalGroups` from the response.
- The same page's export handler (lines 476-480) passes `subscriptionMode` to `/api/groups/export`, and `MetadataService.exportConsumerGroups` (lines 436-438) filters server-side — so list and export disagree.

Reproduction: with ≥ 2 server pages of groups, select "Push": the table shows only the Push rows of the current page (say 3 of 20) but the pager and subtitle still say "共 200 个 Group" with 10 pages; switching to a page full of Pop rows can show an empty-looking filtered table. Conversely the export for the same view contains all matching Push groups server-side.

A new regression test (`passes the subscription mode filter to the server query and resets the page`) fails on the unmodified base: after selecting "Pop", the last `listConsumerGroupPage` call still lacks `subscriptionMode: 'Pop'` (red output: `- "subscriptionMode": "Pop"` expected, received call without it).

## Impact

For server-paginated data the client-side slice filter cannot ever be correct: rows from other pages are invisible, the total/page count lie about the filtered result, and the export of the same filtered view returns a different (larger) set than the table shows.

## Expected behavior

The subscription mode should be sent to `/api/groups/page` and applied server-side so the filtered rows, total and page count agree — mirroring what the export path already does. Changing the mode resets the page to 1 like the search box does.

## Related work

- #911 (consumer group search trim) is the same "list page filter must reach the server" family; the mode filter was left client-side.
- #2680 (list refresh after mutations) and #2297 (ACL user pagination) are different defects.
- The Apache DB path stores the subscription mode in the `rmq_group.message_model` column (read by `toConsumerGroupVO`), so a server-side filter is directly supported.

## PR

Fix incoming.

unbridled-41:
Fixes #4171.

## Problem / Evidence

The Consumer Groups page applies the subscription-mode select client-side only: `visibleConsumerGroups` (`web/src/pages/instance/consumer.tsx`, base `0a596661`, lines 154-162) filters the rows of the **current server page**, while `loadConsumerGroupPage` (343-371) never sends the mode and the paginator shows the unfiltered `totalGroups`. The same view's export already filters server-side (`/api/groups/export` → `MetadataService.exportConsumerGroups` lines 436-438), so the list and the export disagree.

New regression test `passes the subscription mode filter to the server query and resets the page` fails on base: after selecting "Pop", the last `listConsumerGroupPage` call does not include `subscriptionMode: 'Pop'` (`- "subscriptionMode": "Pop"` missing from the received call).

## Root cause / Fix

Root cause: the mode filter was implemented as a client-side slice over server-paginated data.

Fix:
- **Backend**: `/api/groups/page` accepts an optional `subscriptionMode` request parameter and threads it through `MetadataService.listConsumerGroupsPage` → provider chain. `RocketMQMetadataProvider` pushes the filter into the database query via the `rmq_group.message_model` column (the same column `toConsumerGroupVO` reads), so `total` matches the filtered rows. Cloud providers keep a default in-memory implementation (filter on `getSubscriptionMode().name()` then paginate) since their remote APIs have no such filter; `ConsumerGroupListToolHandler` (AI tool) passes `null` explicitly, preserving its behavior.
- **Frontend**: `listConsumerGroupPage` sends `subscriptionMode`; the client-side slice filter is removed; changing the mode resets the page to 1 (same as the search box).

## Priority & scoring

- Impact 30/40 (the filter lies on a core list page whenever the filtered set spans pages; list/export inconsistency) + breadth 14/20 (all server-paginated consumer group lists, any instance with ≥ 2 pages) + reproducibility 18/20 (deterministic API-call regression + DB query assertion) + maintainability value 14/20 (restores the list/export contract symmetry) = **PRIORITY 76** ≥ 70.
- Signature-threading with in-file precedent (export path) and red→green regressions on both ends → **FIX_CONFIDENCE 85** ≥ 80.

## Tests

- Red (frontend): `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` on base with only the test added → 1 failed (`subscriptionMode: 'Pop'` not sent).
- Green (frontend): same command after the fix → `Tests 32 passed (32)`.
- Backend module: `mvn -o test -Dtest='RocketMQMetadataProviderTest,MetadataServiceTest,ConsumerGroupControllerTest,ToolGatewayServiceTest'` → `Tests run: 132, Failures: 0, Errors: 0` (new: DB-query filter assertion, service passthrough, controller param passthrough).
- Full web suite after the fix: `npx vitest run` → `Test Files 115 passed (115)`, `Tests 945 passed (945)` (an earlier run had 1 failure in `ConsumerPage > shows group health diagnostics`, which passes in isolation and in this full re-run — the documented load-flaky test).
- `npx tsc --noEmit` clean; eslint clean on the changed web files; `npm run build` → `✓ built in 12.18s`.

## Risk

Low-moderate. The `/groups/page` endpoint gains an **optional** parameter — existing callers (including the AI tool, which passes null) are unaffected; cloud providers get a default that filters in memory over the already-fetched list (same data their unfiltered path returns). The Apache DB filter uses the exact `message_model` values ("Push"/"Pop") that `toConsumerGroupVO` maps. One behavioral change is intended: the table now shows only server-filtered rows and the total reflects the filter.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/group/ConsumerGroupController.java
@@ -56,9 +56,11 @@ public class ConsumerGroupController {
             @RequestParam(required = false) String instanceId,
             @RequestParam(required = false) String clusterId,
             @RequestParam(required = false) String search,
+            @RequestParam(required = false) String subscriptionMode,
             @RequestParam(defaultValue = "1") int page,
             @RequestParam(defaultValue = "20") int pageSize) {
-        return Result.ok(metadataService.listConsumerGroupsPage(instanceId, clusterId, search, page, pageSize));
+        return Result.ok(metadataService.listConsumerGroupsPage(instanceId, clusterId, search,
+                subscriptionMode, page, pageSize));
     }
 
     @GetMapping("/{name}")
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/topic/MetadataService.java
@@ -211,15 +211,15 @@ public class MetadataService {
     }
 
     public PageResult<ConsumerGroupVO> listConsumerGroupsPage(String instanceId, String clusterId, String search,
-                                                              int page, int pageSize) {
+                                                              String subscriptionMode, int page, int pageSize) {
         validatePagination(page, pageSize);
         instanceId = normalizeInstanceId(instanceId);
         if (!StringUtils.hasText(instanceId) && StringUtils.hasText(clusterId)) {
             return metadataProvider.listConsumerGroupsPage(normalizeFilter(clusterId),
-                    normalizeFilter(search), page, pageSize);
+                    normalizeFilter(search), subscriptionMode, page, pageSize);
         }
         return resolve(instanceId).listConsumerGroupsPage(instanceId, normalizeFilter(search),
-                page, pageSize);
+                subscriptionMode, page, pageSize);
     }
 
 
--- a/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ConsumerGroupListToolHandler.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/ConsumerGroupListToolHandler.java
@@ -45,7 +45,8 @@ public class ConsumerGroupListToolHandler implements ToolHandler {
         String clusterId = (String) input.get("cluster");
         String search = (String) input.get("search");
         PageResult<ConsumerGroupVO> page = metadataService.listConsumerGroupsPage(
-                clusterId, null, search, ToolListPagination.page(input), ToolListPagination.pageSize(input));
+                null, clusterId, search, null,
+                ToolListPagination.page(input), ToolListPagination.pageSize(input));
         return ToolListPagination.pagedResult(page, page.getItems().stream()
                 .map(ConsumerGroupListToolHandler::safeProjection)
                 .toList());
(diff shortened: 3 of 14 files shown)


unbridled-41:
@/tmp/c4172-new.md

unbridled-41:
Closing as a duplicate of #3160 (and #4053), both of which implement the same server-side subscription-mode filter for `/api/groups/page` and were opened earlier (#3160 on 2026-09-04, #4053 on 2026-09-07). They also target the same tracking issue, #3150, which I missed when checking for existing work before starting.

My branch implemented the same fix surface: an optional `subscriptionMode` parameter threaded through `MetadataService` → `InstanceProvider` → `ApacheInstanceProvider`/`MetadataProvider` → the `rmq_group.message_model` DB filter in `RocketMQMetadataProvider`, a default in-memory filter for cloud providers, and a page reset on filter change, with red→green regressions on both ends. Implementation and verification evidence remain on the local branch; retaining the earlier PRs to avoid three copies of the same change.

PR #4179, opened 2026-09-09
fix collections.emptylist() bug

enkilee:

fix collections.emptylist() bug

## Brief changelog

XX

## Verifying this change

XXXX


web/src/pages/instance/dlq.tsx
```diff
@@ -166,11 +166,23 @@ const DLQPage = () => {
     setGroups([]);
     setTotal(0);
     setPage(1);
+    setPageSize(20);
+    setSearch('');
     setSelectedGroupNames([]);
     setDetailGroup(null);
+    setDetailOpen(false);
+    setDetailMessages([]);
+    setDetailTotal(0);
```
RockteMQ-AI, line 175: **[Info]** Consider extracting these resets into a helper function (e.g., `resetDlqState()`) to avoid forgetting to reset new state variables added in the future. Alternatively, the `key` prop approach used in `topic.tsx` would be more maintainable here too — it guarantees a clean slate without manually listing every state setter.

RockteMQ-AI (approved):
﻿### Summary
This PR fixes stale React state when switching between instances/domains in the dashboard. The changes are correct and address a real UX bug.

**Note:** The PR title "fix collections.emptylist() bug" does not match the actual changes — these are React state management fixes in the web UI, not a Java `Collections.emptyList()` fix. Please update the title to something like "fix: reset page state when switching instances/domains" for clarity.

### Findings
- **[Info]** web/src/pages/instance/dlq.tsx:175 — Consider using the `key` prop pattern (as done in `topic.tsx`) instead of manually resetting each state variable. This is more maintainable and prevents forgetting to reset newly-added state.

Overall: LGTM — the fixes are correct and the approach in `topic.tsx` (forcing remount via `key`) is particularly clean.

---

lizhimins:
Thanks for the contribution. We have to close this one, starting with a process issue: the title says "fix collections.emptylist() bug" but the diff touches no Java at all — there is no occurrence of `Collections`, `emptyList`, or any `.java` file — so the change can't be reviewed or released under that description. On the substance, `key={selectedInstanceId}` forces a full remount of a ~2000-line page, and because `useInstanceFilter` returns `undefined` until the instance list resolves (and then may also `navigate(..., {replace: true})`), the key changes on every initial load, so the page mounts and remounts before the user does anything — re-running every effect and discarding non-instance state such as search text, pagination and expanded rows. The page already resets its instance-scoped state explicitly. Similarly, most of the added DLQ/alerts resets are unreachable (those drawers and modals are masked, so the instance or domain cannot change while they are open), and `setSearch('')` / `setPageSize(20)` discard user preferences rather than stale scope. Please resubmit with an accurate title and, if the stale-state symptom is reproducible, a failing test that shows it.

PR #4182, opened 2026-09-09
[studio] fix(message): resolve registered broker hostnames for offset ID queries

### LINKED ISSUE #4181: [studio][Bug] Broker 使用主机名时，物理消息 ID 查询返回空结果
在 `rocketmq-studio@0a596661`、RocketMQ Broker 5.5.1、Java 21 的本地真实环境复现。

1. Broker 使用可解析主机名注册到 NameServer，例如 `repro-hostname-broker:10911`。
2. 正常发送消息，按 Topic 查询可以找到该消息。
3. 使用发送结果中的物理消息 ID（`offsetMsgId`），在消息查询页选择“按 Message ID”查询，结果为 0 条；同一 ID 使用官方 SDK 可以读取。

预期应返回该消息。当前 `BrokerTopologyGuards` 直接比较登记的主机名与物理 ID 内嵌的 IP，误判为集群外地址。普通唯一 `msgId` 查询正常。

Image

Image

qiyu-lu:
修复 Broker 使用主机名注册时，按物理消息 ID 查询返回空结果的问题。解析已登记且端口匹配的 Broker 主机名，按 IP 判断地址是否属于所选实例，并继续使用消息 ID 内嵌 IP 查询。

新增主查询、回退查询及地址校验回归测试，保留未知地址拒绝和纯 IP 快速路径。

验证：95 个相关后端测试及 Checkstyle 通过；真实 Broker 和原页面中，同一物理 ID 的查询结果由 0 条恢复为 1 条。

Fixes #4181

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/apache/BrokerTopologyGuards.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/apache/BrokerTopologyGuards.java
@@ -16,8 +16,10 @@
  */
 package org.apache.rocketmq.studio.provider.apache;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.net.UnknownHostException;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
@@ -67,12 +69,41 @@ public final class BrokerTopologyGuards {
         if (!StringUtils.hasText(brokerAddr)) {
             return null;
         }
-        if (knownBrokerEndpoints(admin).contains(brokerAddr)) {
+        Set<String> endpoints = knownBrokerEndpoints(admin);
+        if (endpoints.contains(brokerAddr)) {
             return brokerAddr;
+        }
+        InetSocketAddress address = (InetSocketAddress) messageId.getAddress();
+        for (String endpoint : endpoints) {
+            if (matchesBrokerEndpoint(endpoint, address)) {
+                // Keep the decoded IP for remoting instead of resolving the hostname again.
+                return brokerAddr;
+            }
         }
         log.warn("Rejecting decoded broker address {} for msgId={} because it is not a known broker endpoint"
                 + " for the selected instance", brokerAddr, msgId);
         return null;
+    }
+
+    private static boolean matchesBrokerEndpoint(String endpoint, InetSocketAddress address) {
+        int separator = endpoint.lastIndexOf(':');
+        if (separator <= 0) {
+            return false;
+        }
+        try {
+            if (Integer.parseInt(endpoint.substring(separator + 1)) != address.getPort()) {
+                return false;
+            }
+            // Only resolve hosts advertised by the selected instance's broker topology.
+            for (InetAddress resolved : InetAddress.getAllByName(endpoint.substring(0, separator))) {
+                if (resolved.equals(address.getAddress())) {
+                    return true;
+                }
+            }
+        } catch (NumberFormatException | UnknownHostException e) {
+            log.debug("Could not resolve registered broker endpoint {}: {}", endpoint, e.getMessage());
+        }
+        return false;
     }
 
     static String decodedBrokerAddr(MessageId messageId) {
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Resolves an issue where offset ID queries fail when brokers register with hostnames but message IDs contain resolved IP addresses.

**Changes reviewed:**
- **BrokerTopologyGuards.java**: Adds `matchesBrokerEndpoint()` which performs DNS resolution to match broker endpoints by hostname — correct approach for handling hostname/IP mismatches in broker topology validation
- **BrokerTopologyGuardsTest.java**: Comprehensive test coverage including hostname resolution, IPv6 representations, malformed inputs, and edge cases — thorough validation
- **MessageServiceTest.java**: Minor test update for compatibility

The DNS resolution approach is sound: `InetAddress.getAllByName()` handles both IPv4 and IPv6, and the exception handling gracefully skips unresolvable endpoints. The port validation before DNS lookup is a good optimization to avoid unnecessary resolution.

**Verdict:** LGTM — well-implemented fix with excellent test coverage.

---

lizhimins (changes requested):
Thanks for digging into this — hostname-registered brokers breaking offset-msgId lookup is a real gap in the guard, and the test suite you added (IPv6 forms, malformed ports, "no DNS when the exact IP already matches") is well above the usual bar. Fail-closed on `UnknownHostException` is also the right default. That said, I'd like to ask for changes before this can land:

1. **Blocking DNS on the query path.** `InetAddress.getAllByName` has no timeout and only the JVM-level cache, and it runs on the synchronous request thread. A single message query can reach the guard three times, and the DLQ resend path calls it once per selected msgId, so the worst case is `msgIds x endpoints x 3` blocking lookups (plus one uncached `examineBrokerClusterInfo()` RPC each). Please move resolution out of the guard: normalize once into a short-TTL, instance-scoped topology snapshot (with a bounded resolver + explicit timeout) and have the guard compare against that snapshot.

2. **Port + IP equality is not a reliable identity check.** The IP embedded in an offset msgId is the *broker's own* resolution of `brokerIP1` at broker startup (`new InetSocketAddress(brokerIP1, listenPort)`), while this compares it against *Studio's* resolution at query time. Split-horizon DNS, differing `resolv.conf` (Studio in k8s, broker on the host), NAT, multi-NIC and record changes after broker start all still produce false rejects; overlapping RFC1918/loopback space produces false accepts. Note also that `knownBrokerEndpoints()` is nameserver-wide, not instance-scoped, and several clusters may share one nameserver — so the candidate set is broader than the log message implies. Since an offset msgId carries no `brokerName`, please scope candidates to the clusters owned by the instance and carry `brokerName`/`clusterName` through the snapshot so a match is attributable (and logged).

3. **Trust boundary.** The method still returns the *decoded* address for remoting, so the only thing authorizing that connect is now a DNS answer rather than the registry string. Please at minimum reject resolved addresses that are loopback / link-local / any-local / IPv6 ULA, reusing the existing `UrlHostGuard.areAllowed(...)` semantics instead of a new ad-hoc rule — today a broker registered as `localhost:10911` blesses any `127.0.0.1:10911` msgId (your own test asserts this).

4. **Robustness / observability.** Only `NumberFormatException` and `UnknownHostException` are caught; any other `RuntimeException` aborts the whole loop, and because `endpoints` is a `HashSet` the outcome becomes order-dependent. Catch per-endpoint, and raise the resolution-failure log from `debug` to `warn` (with endpoint + reason) so operators can tell "hostname not resolvable from Studio" apart from "message does not exist".

5. **Reuse + tests.** Host/port splitting is already implemented five times (`ApacheRocketMqProxyMetricsCollector.resolveGrpcTarget` even strips IPv6 brackets); please extract one shared helper rather than adding a sixth copy, and reuse the existing `clusterInfoWithBrokerAddresses` fixture (yours drops `brokerName`/`cluster`). `mockStatic(InetAddress.class)` is the first static mock in this repo and targets a bootstrap-loaded JDK class on JDK 21 — an injected resolver seam would be both safer and testable for the timeout/failure paths. Finally, consider folding this into a `BrokerTopologyGuardsTest` that also locks in the pre-existing exact-match behaviour, and rebase: the branch is 19 commits behind and `RocketMQMessageProviderTest` has changed since.

We are not going to carry these changes ourselves: the fix needs a design decision about where topology normalization lives (guard vs. cached snapshot) that is much better made by whoever owns the patch, and the security posture above should be argued and tested by its author. Happy to review a revised version — the direction is right, and items 1-3 are the ones I'd focus on.

qiyu-lu:
Thanks for the thorough review, @lizhimins. Rather than patch this version, I wrote up a revised design that addresses points 1–5 (snapshot-based guard, instance-scoped candidates, connecting only to registered addresses) in #4554, with two questions for you. I'll open a fresh PR from current `rocketmq-studio` once the design is agreed.

PR #4185, opened 2026-09-09
fix(message): report truncation for Apache key queries

zjncs:
## Motivation

Message queries by key on Apache instances are silently capped at the broker-side budget with no truncation signal:

- `RocketMQMessageProvider` never overrides `queryMessagesDetailed`, so the interface default wraps every Apache query in `MessageQueryResult.complete(...)` — `mayBeTruncated` is always `false`.
- `queryByKey` queries with `KEY_QUERY_MAX = 64`: a business key matching more messages returns exactly 64 rows, `resultMayBeTruncated=false`, no warning in the UI — every time, deterministically.
- The service-layer heuristic in `MessageService` (`topicQuery && size >= 200`) explicitly excludes key queries, so nothing else covers this.
- Tencent already reports this condition (#3048, merged) and Aliyun is being fixed the same way (#4164, open). Apache is the default deployment, so the widest audience is the one left with silently incomplete results.

Additionally, registered Apache instances dispatch through `ApacheInstanceProvider` (via `InstanceProviderRegistry.byInstanceId`), which also inherits the `complete(...)` default — so both dispatch paths need the signal.

## Modification

- `RocketMQMessageProvider.queryMessagesDetailed` now carries the real result. `queryByKey` sets `mayBeTruncated` when the merged result count reaches `KEY_QUERY_MAX`. Note on exactness: `MQAdminImpl.queryMessage` fans the query out to **every** route broker with a per-broker budget of `KEY_QUERY_MAX` and merges the responses without a client-side cap, so an exact verdict cannot be derived from the merged list — the conservative "budget reached" signal matches the `mayBeTruncated` field semantics ("stopped because the result budget was reached, not because the query was exhausted").
- The signal is computed **before** the client-side tag filter, which can otherwise hide capped broker results behind an empty/short filtered list.
- `ApacheInstanceProvider.queryMessagesDetailed` delegates to the message provider so registry-dispatched Apache instances reach the signal.
- Topic and msgId query behavior is unchanged (topic queries keep the existing `>=200` service heuristic; msgId lookups return a single record).

## Verification

Fail-before (provider changes stashed, tests kept) — 3 failures:

```
[ERROR] Tests run: 52, Failures: 3
RocketMQMessageProviderTest.queryByKeyReportsTruncationWhenTheBrokerBudgetIsReached
RocketMQMessageProviderTest.queryByKeyKeepsTheTruncationSignalWhenTagFilteringDropsEveryRow
ApacheInstanceProviderTest.queryMessagesDetailedShouldDelegateToMessageProviderTest
```

Pass-after (fix applied):

```
RocketMQMessageProviderTest   Tests run: 44, Failures: 0, Errors: 0
ApacheInstanceProviderTest    Tests run: 8,  Failures: 0, Errors: 0
MessageServiceTest            Tests run: 17, Failures: 0, Errors: 0
```

New tests:

- `queryByKeyReportsTruncationWhenTheBrokerBudgetIsReached` — 64 merged matches → `mayBeTruncated=true`, all 64 rows returned.
- `queryByKeyStaysCompleteBelowTheBrokerBudget` — 63 matches → complete.
- `queryByKeyKeepsTheTruncationSignalWhenTagFilteringDropsEveryRow` — 64 raw matches, tag filter keeps none → still `mayBeTruncated=true` (signal survives post-filtering).
- `queryMessagesDetailedShouldDelegateToMessageProviderTest` — registered Apache instances reach the provider's signal instead of the `complete(...)` default.


RockteMQ-AI (approved):
### Summary
This PR introduces truncation reporting for message queries by key. Key changes:

1. **New `MessageQueryResult` record** — wraps the message list with `truncated`, `proxies`, `totalProxies`, `totalNameServers` metadata. Clean data-carrying approach.
2. **Truncation detection** — when the broker returns exactly `KEY_QUERY_MAX (64)` results, the query is flagged as truncated. This is a reasonable heuristic since the broker caps key queries at this limit.
3. **Tool catalog schema** — `proxies`, `totalProxies`, `totalNameServers` now accept `null`, matching the actual runtime behavior when proxy/name-server topology is unavailable.
4. **Tests** — thorough coverage including truncation detection, null topology fields, and the detailed query path through `ApacheInstanceProvider`.

The truncation heuristic (exact match against `KEY_QUERY_MAX`) is pragmatic. One minor consideration: if the broker ever changes this limit, the constant would need updating — but since it's defined locally and matches the broker's documented behavior, this is acceptable.

LGTM.

---

RockteMQ-AI:
⚠️ **Merge conflict detected**

This PR has conflicts with the base branch and cannot be merged. Please rebase or merge the base branch into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix/apache-key-query-truncation
git rebase origin/develop
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review after conflicts are resolved.

---
*Automated notification by github-manager-bot*

zjncs:
Conflicts resolved — rebased onto the current `rocketmq-studio` (`c2ab6fe4`, single commit on top of 6c24d2ed).

While rebasing I also dropped the stacked `rmq.dashboard.summary` nullable-counts commit this branch previously carried: #4115 (206314b9) already made `cluster.proxies` / `stats.totalProxies` / `stats.totalNameServers` nullable in `rmq-tools.yaml` and added the equivalent `executesDashboardSummaryWhenTopologyCountsAreUnavailable` gateway test, so that change is now upstream and this PR is back to just the Apache key-query truncation reporting (`RocketMQMessageProvider` + `ApacheInstanceProvider` + tests). `RocketMQMessageProviderTest` + `ApacheInstanceProviderTest` → 52/52 green on the rebased branch.

PR #4193, opened 2026-09-09
Escape LIKE wildcards in instance search so names containing _ or % match literally

zjncs:
### Motivation

The instance search box passes its raw keyword into `LIKE` clauses over `name`, `endpoint` and `remark` (`MybatisPlusInstanceRepository.search` / `findByTypeAndSearch`, reached from `InstanceService.listInstances` for the instance list page). Since `_` and `%` are SQL LIKE wildcards:

- an underscore in an instance name (very common, e.g. `prod_cluster_1`) silently matches *any* single character, so searching `prod_cluster_1` also returns `prod-cluster-1`, `prodXclusterY1`, …
- a trailing `%` (e.g. pasted from a log line) matches every suffix.

So the search can silently return instances that do not contain the typed text at all.

### Changes

- Add a private `escapeLike` helper (same behaviour as the existing `QueryHistoryService.escapeLike`) and apply it to the keyword in both `search` and `findByTypeAndSearch`.

### Verification

`MybatisPlusInstanceRepositoryTest` — two new tests:

- `searchShouldEscapeLikeWildcardsTest`
- `findByTypeAndSearchShouldEscapeLikeWildcardsTest`

Both assert the bound LIKE value for `prod_cluster%` is `%prod\_cluster\%%` and not the unescaped pattern. Before the fix both fail; after it the class is green (18/18).

```
$ mvn -f server/pom.xml test -Dtest='MybatisPlusInstanceRepositoryTest'
(before) Tests run: 18, Failures: 2, Errors: 0
(after)  Tests run: 18, Failures: 0, Errors: 0
```

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/instance/MybatisPlusInstanceRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/instance/MybatisPlusInstanceRepository.java
@@ -62,11 +62,12 @@ public class MybatisPlusInstanceRepository implements InstanceRepository {
 
     @Override
     public List<InstanceVO> search(String keyword) {
+        String pattern = escapeLike(keyword);
         return instanceMapper.selectList(
                 new QueryWrapper<RmqInstance>()
-                        .and(w -> w.like("name", keyword)
-                                .or().like("endpoint", keyword)
-                                .or().like("remark", keyword))
+                        .and(w -> w.like("name", pattern)
+                                .or().like("endpoint", pattern)
+                                .or().like("remark", pattern))
                         .orderByAsc("id")).stream()
                 .map(this::toVO)
                 .toList();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Same LIKE wildcard escaping pattern as the audit and credential search fixes. The `escapeLike()` method is consistent across all three PRs. Tests properly verify that `_` in instance names like `test_group_1` is matched literally.

Clean fix. LGTM.

---

lizhimins:
Closing as part of consolidating the LIKE-escape work — same reasoning as on #4192: a private `escapeLike` copy in the instance repository instead of one shared utility, and no explicit `ESCAPE` clause (the escape relies on MySQL's implicit backslash, which H2 does not provide). #4236 is the anchor we asked to be resubmitted.

PR #4194, opened 2026-09-09
Escape LIKE wildcards in cloud credential search so names with % or _ match literally

zjncs:
### Motivation

`MybatisPlusCloudCredentialRepository.findPage` feeds the trimmed search term straight into a `LIKE` clause over the credential `name`. Because `%` and `_` are SQL LIKE wildcards, a term like `aliyun_prod%` matches every credential whose name starts with `aliyun` + any single character + `prod`, plus any suffix — silently listing credentials the user did not type.

### Changes

- Add a private `escapeLike` helper (same behaviour as the existing `QueryHistoryService.escapeLike`) and apply it to the normalized search term in `findPage`.

### Verification

`MybatisPlusCloudCredentialRepositoryTest` — new test `findPageShouldEscapeLikeWildcardsInTheSearchTerm` asserts the bound LIKE value for `aliyun_prod%` is `%aliyun\_prod\%%` and not the unescaped pattern. Before the fix it fails; after it the class is green (6/6).

```
$ mvn -f server/pom.xml test -Dtest='MybatisPlusCloudCredentialRepositoryTest'
(before) Tests run: 6, Failures: 1, Errors: 0
(after)  Tests run: 6, Failures: 0, Errors: 0
```

This is the same class of fix as the audit (`MybatisPlusAuditRepository`) and instance (`MybatisPlusInstanceRepository`) search paths, which also bind user input into LIKE clauses without escaping.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/provider/credential/MybatisPlusCloudCredentialRepository.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/provider/credential/MybatisPlusCloudCredentialRepository.java
@@ -52,9 +52,10 @@ public class MybatisPlusCloudCredentialRepository implements CloudCredentialRepo
     @Override
     public PageResult<CloudCredentialVO> findPage(InstanceVendor vendor, String search, int page, int pageSize) {
         String normalizedSearch = search == null || search.isBlank() ? null : search.trim();
+        String searchPattern = escapeLike(normalizedSearch);
         QueryWrapper<RmqCloudCredential> q = new QueryWrapper<RmqCloudCredential>()
                 .eq(vendor != null, "vendor", vendor == null ? null : vendor.name())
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Completes the LIKE wildcard escaping fix series for cloud credential search. The implementation mirrors the audit and instance search fixes exactly, maintaining consistency. The Javadoc on `escapeLike()` correctly references the original `QueryHistoryService` pattern.

Clean fix. LGTM.

---

lizhimins:
Closing as part of consolidating the LIKE-escape work — same reasoning as on #4192 and #4193: a third private `escapeLike` copy (here in the cloud-credential repository, which also feeds the CSV export path) and no explicit `ESCAPE` clause. #4236 is the anchor we asked to be resubmitted.

PR #4195, opened 2026-09-09
Advertise exactly the labels each metric profile PromQL aggregates by

zjncs:
### Motivation

Each metric mapping in `GET /metrics/profiles` exposes `promql` plus a `labels` list. After `sum(...) by (...)`, a Prometheus series only carries the labels in the by-clause, so the advertised `labels` must match it — that is what the 4.x profile does everywhere (`by (cluster, group, topic)` ↔ labels `cluster, group, topic`).

Four mappings in the 5.x native profile violate this: `message_in_tps` / `message_out_tps` / `throughput_in` / `throughput_out` aggregate by `(cluster, node_id)` / `(cluster, node_id, consumer_group)` but advertise `topic` (and `message_type`) as well. Those labels are aggregated away, so the result series never carry them. A client that renders one column per advertised label (the field is documented as the shape of the mapping's results) shows empty `topic` / `message_type` cells for these panels.

The frontend's own profile mock (`web/src/components/__tests__/MetricsExplorer.test.tsx`) already models `message_in_tps` as `labels: ['cluster', 'node_id']` — the backend is the outlier.

### Changes

- Trim the advertised labels of the four 5.x TPS/throughput mappings to exactly their by-clause sets. No PromQL or query behaviour changes.

### Verification

`MetricProfileServiceTest`:

- New test `everyMappingShouldAdvertiseExactlyTheLabelsItsPromqlAggregatesByTest` parses the `by (...)...` clause of every mapping in both profiles and asserts `labels` equals it, so future mappings cannot drift again.
- Updated `mappingsShouldExposeLabelsAndUnitsForDashboardRendering` to the corrected label set.

```
$ mvn -f server/pom.xml test -Dtest='MetricProfileServiceTest'
(before) Tests run: 13, Failures: 1, Errors: 0
(after)  Tests run: 13, Failures: 0, Errors: 0
```


RockteMQ-AI (approved):
### Summary
Correct fix for a label mismatch in the 5.x native metric profile. The advertised labels now match the PromQL `by (...)` clause, preventing empty columns in dashboard panels that render one column per label.

### Findings
- **[Info]** `MetricProfileService.java:115-124` — The four mappings (`message_in_tps`, `message_out_tps`, `throughput_in`, `throughput_out`) now advertise exactly the labels their PromQL aggregates by. Clean and minimal.
- **[Info]** `MetricProfileServiceTest.java:140-160` — The new test `everyMappingShouldAdvertiseExactlyTheLabelsItsPromqlAggregatesByTest` is a great guard against future drift. It parses the `by (...)` clause from every mapping's PromQL and asserts the labels match. This catches the exact class of bug this PR fixes.
- **[Info]** The PR description is excellent — it clearly explains the problem (labels aggregated away by PromQL), the fix (trim to by-clause), and includes verification steps.

### Verdict
LGTM. The fix is minimal, well-tested, and prevents a real UX issue where dashboard panels show empty cells for labels that don't exist in the result series.

---

lizhimins:
Closing as part of clearing the metric-profile queue, not because the finding is wrong.

The mismatch you point at is real: four `mapping(...)` entries advertise `topic` / `message_type` while their PromQL aggregates `by (cluster, node_id)`, so the profile claims labels the query never groups by. We are folding that into trunk-side work that gives the advertised labels a single source of truth together with the `scopeLabels` table added by #3989, rather than keeping a second hard-coded list per profile. Thanks for the analysis — it is what made us schedule that.

PR #4196, opened 2026-09-09
[Studio] feat: add batch Proxy address management

coder999o:

Add batch Proxy address management to RocketMQ Studio so operators can paste multiple Proxy endpoints at once and review row-level outcomes instead of adding each address one by one.

The batch flow validates address format, skips existing addresses, detects duplicates inside the same request, appends only valid new addresses, and refreshes the Proxy node list with the updated address state.

## Brief changelog

- Add `POST /api/proxies/addresses/batch` for batch Proxy address additions.
- Add row-level batch result models with `ADDED`, `EXISTING`, `DUPLICATE`, and `INVALID` statuses.
- Add a Proxy page batch-add modal with parsed-count feedback, summary counters, and a result table.
- Add frontend API bindings, i18n text, and focused backend/frontend tests.
- Apply `tableScrollX(columns)` to the Proxy node table and batch result table.

## Verifying this change

- `npm test -- --run src/api/proxy.test.ts src/pages/studio/__tests__/Proxy.test.tsx`
- `mvn -q -Dtest=ProxyAddressServiceTest,ProxyControllerTest test`
- `npm run build`
- `npm run lint -- src/api/proxy.ts src/api/proxy.test.ts src/pages/studio/Proxy.tsx src/pages/studio/__tests__/Proxy.test.tsx src/i18n/translations.ts`
- `git diff --check`


RockteMQ-AI (approved):
### Summary
Adds batch Proxy address management to RocketMQ Studio — operators can paste multiple Proxy endpoints at once with row-level outcome feedback (`ADDED`/`EXISTING`/`DUPLICATE`/`INVALID`). Well-structured full-stack implementation with proper validation, i18n support, and comprehensive tests on both backend and frontend.

**Highlights:**
- Backend: clean DTO/VO separation, synchronized batch operation, proper size limit and format validation
- Frontend: responsive batch-add modal with parsed-count feedback, summary counters, result table with color-coded status badges
- Good race-condition handling with `requestId` in the frontend
- `tableScrollX(columns)` applied consistently

**Minor suggestions (non-blocking):**
- Consider extracting the batch address parsing logic (`parseBatchProxyAddresses`) into a shared utility if similar batch-input patterns emerge elsewhere
- The `addProxyAddrs` synchronized method holds the lock for the entire batch operation — acceptable for the expected small batch sizes, but worth noting for future large-batch scenarios

LGTM.

---

lizhimins:
Thanks for the work on this. We're closing it because the direction doesn't fit how Proxy addresses are meant to be obtained in Studio.

Proxy endpoints are derived from the cluster the instance is attached to — they should be **discovered automatically** from the cluster topology, not maintained as a separately hand-managed list. Adding a batch management surface for them creates a second source of truth that can drift from what the brokers actually register, and every consumer then has to decide which one to believe. It also puts a manual operational step in front of something that scales out and in on its own.

If automatic discovery is currently missing a case you need (e.g. proxies behind an SLB, or a topology where discovery genuinely cannot infer the client-facing address), please open an issue describing that case — the fix we'd want is in the discovery path, not a new management screen.

PR #4189, opened 2026-09-09
fix(metrics): keep the selected range when the instance changes

unbridled-41:
[...]

PR #4197, opened 2026-09-09
fix(studio): keep selected range when dashboard switches instances

### LINKED ISSUE #4188: [Studio][Bug] Metrics Explorer silently reverts to the 1h window when the dashboard switches instances
**Problem**

On the home dashboard (`/`, `web/src/pages/home/dashboard.tsx`), the Metrics Explorer below the cluster cards keeps its component instance mounted while the 实例 selector changes (`<MetricsExplorer instanceId={selectedInstanceId} />` with no `key`). When the instance prop changes, the profile-loading effect re-runs and reloads every panel — but it hardcodes `RANGE_OPTIONS[0]` (the 1h window, `web/src/components/MetricsExplorer.tsx:539` before the fix) instead of the currently selected range.

**Evidence**

Reproduction (component test, mocked API):
1. Select a non-default range (e.g. 6h) — the panel issues a query with `start = end - 6h`, `step = '2m'`, and the Segmented control shows 6h.
2. Change the dashboard instance selector (instance-1 → instance-2). The profiles effect re-runs (`loadAll` is in its dependency chain through `runQuery`, which depends on `instanceId`) and reloads the panels.
3. Observed with the unfixed code: the reload query uses the default 1h window (`start = end - 1h`, `step = '30s'`) while the Segmented control still displays **6h**. The charts now show one hour of data for a control that says six hours, and the user is not told anything changed.

The same effect runs once on mount, where `RANGE_OPTIONS[0]` is correct; the defect is that the value is also used on the *reload* path where a user selection already exists.

**Impact**

Every instance switch on the home dashboard silently narrows the explored window to 1h for users working with the 6h/24h ranges — the displayed range label and the actual data window disagree, so trend readings taken right after a switch are quietly wrong.

**Expected behavior**

A reload triggered by an instance change reuses the range the user selected; the 1h default applies only before the first selection (initial mount).

**Related work**

- #3304 / PR #3299 fixed the sibling request-guard issue in the same effect (shared requestId double-bump); the range-reuse gap remained.
- #1603 (closed) addressed stale data-source refs on instance switch in the same component — the selected time range reverting on the same trigger is a distinct gap.
- No existing issue/PR covers the range reverting on instance switch (searched: metrics range, range reset, instance switch, rangeId).

**PR**

Fix: #4189

jokerzsd:
## Summary

Fix the Metrics Explorer silently reverting to the 1h window when the dashboard switches instances.

## Problem

On the home dashboard, `MetricsExplorer` stays mounted while the instance selector changes. When `instanceId` changes, the profile-loading effect re-runs and reloads every panel, but hardcoded `RANGE_OPTIONS[0]` (1h) instead of the currently selected range. The charts then showed one hour of data while the Segmented control still displayed the user's selection.

## Fix

Track the selected range in a ref and reuse it when the effect reloads panels after an instance change, so the reload honors the current selection.

## Test plan

`cd web && npx tsc -b` passes.

Fixes #4188

### DIFF
--- a/web/src/components/MetricsExplorer.tsx
+++ b/web/src/components/MetricsExplorer.tsx
@@ -511,6 +511,10 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
     [profileId, profiles],
   );
   const selectedRange = RANGE_OPTIONS.find((range) => range.value === rangeId) ?? RANGE_OPTIONS[0];
+  const selectedRangeRef = useRef(selectedRange);
+  useEffect(() => {
+    selectedRangeRef.current = selectedRange;
+  }, [selectedRange]);
   const anyLoading =
     Object.values(panels).some((panel) => panel.loading) || Boolean(customPanel?.loading);
   const availableDataSources = useMemo(
@@ -649,7 +653,7 @@ const MetricsExplorer = ({ instanceId }: MetricsExplorerProps) => {
         const initialProfile =
           nextProfiles.find((profile) => profile.id === storedProfileId) ?? nextProfiles[0];
         setProfileId(initialProfile?.id ?? '');
-        void loadAll(initialProfile, RANGE_OPTIONS[0]);
+        void loadAll(initialProfile, selectedRangeRef.current);
       })
       .catch(() => {
         if (!cancelled) setProfileError(true);


RockteMQ-AI (approved):
﻿### Summary
Preserve the selected time range when the dashboard switches instances. Previously, the `rangeKey` state was reset on instance change because `RangePicker` re-mounted with a new `key`. Using a `useRef` to remember the last selected range and passing it as `defaultValue` correctly restores the user's selection.

Clean implementation using the ref pattern to persist state across re-renders. References issue #4188. LGTM.

---

lizhimins:
Closing as a duplicate of #4189, merged into `rocketmq-studio` as 6334f9238. Both patches change the same line (`MetricsExplorer.tsx` `loadAll(initialProfile, RANGE_OPTIONS[0])`) with the same ref-plus-effect approach; #4189 was opened 7h45m earlier and carries an assertion that the picked range survives an instance switch.

PR #4202, opened 2026-09-09
[studio][ISSUE #4201] Evict data-source cache after instance deletion

### LINKED ISSUE #4201: [Studio][Bug] Deleting an instance leaves cached data-source bindings stale
## Problem

After #4049, SettingsService caches both full and paged data-source inventories in the shared data-sources cache. InstanceService.deleteInstance removes the deleted instance name from DataSourceVO.instanceIds by writing through SettingsRepository directly, so none of SettingsService's cache-eviction write paths run.

The database row is updated, but cached list results can keep exposing the deleted binding until another data-source mutation or application restart.

## Reproduction

1. Create a data source bound to instance-a.
2. Call the full or paged data-source list endpoint to populate its cache.
3. Delete instance-a.
4. Query the same data-source list again.

The repository no longer contains instance-a in the binding, while the cached response can still contain it.

## Expected behavior

A successful instance deletion should invalidate the shared data-source cache after the deletion transaction commits. A failed or rolled-back deletion should leave the existing cache untouched.

## Scope

This issue only covers cache invalidation for the cross-service data-source binding update performed during instance deletion.

89799969:

Fixes #4201.

After #4049, both full and paged data-source inventories are cached. InstanceService.deleteInstance updates DataSourceVO.instanceIds through SettingsRepository directly, bypassing the cache-eviction annotations in SettingsService. A successful deletion can therefore leave both cached inventories exposing the old instance binding.

This change invalidates the shared data-source cache only after the instance-deletion transaction commits. A rollback keeps the existing cache intact, and callers without active transaction synchronization retain the previous immediate-cleanup behavior.

## Brief changelog

- centralize the shared data-source cache name
- clear that cache after a successful instance-deletion commit
- keep endpoint/client cleanup on the same post-commit boundary
- add regression coverage for binding removal, cache eviction, and commit timing

## Verifying this change

- Focused tests: InstanceServiceTest and SettingsServiceCachingTest
  - 84 tests run, 0 failures, 0 errors, 0 skipped
- Checkstyle: 0 violations
- The full server suite was also attempted, but this Windows host reported unrelated Java loopback-socket failures in HTTP/client tests plus two AuthCors assertions; no full-suite pass is claimed, and CI can provide clean-host coverage.


server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java
```diff
@@ -639,22 +643,31 @@ public void deleteInstance(Long id) {
...
         if (!TransactionSynchronizationManager.isSynchronizationActive()) {
+            evictDataSourceCache();
             releaseApacheEndpointIfUnused(existing, null);
             return;
         }
         TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
             @Override
             public void afterCommit() {
+                evictDataSourceCache();
                 releaseApacheEndpointIfUnused(existing, null);
```
RockteMQ-AI, line 659: **[Info]** The cache eviction here is correctly placed in `afterCommit()`, ensuring the cache is only cleared after the deletion is durable. One consideration: if `releaseApacheEndpointIfUnused` throws, the cache has already been evicted but the endpoint release is incomplete. Since `evictDataSourceCache()` is idempotent (clearing an already-empty cache is a no-op), this ordering is safe — just worth noting for future maintainers.

server/src/main/java/org/apache/rocketmq/studio/settings/SettingsService.java
```diff
@@ -56,6 +56,8 @@
 @Service
 public class SettingsService {
 
+    public static final String DATA_SOURCE_CACHE = "data-sources";
```
RockteMQ-AI, line 59: **[Info]** Good extraction of the magic string into a constant. This prevents the cache-name drift that caused the original bug. Consider also adding a brief Javadoc on the constant explaining that it must match the `@Cacheable`/`@CacheEvict` values and that cross-service eviction depends on it.

RockteMQ-AI (approved):
### Summary
This PR correctly fixes a cache invalidation gap where deleting an instance left stale data-source bindings in the cache. The approach — evicting the shared cache after commit — is sound, and the constant extraction prevents future cache-name drift.

### Findings
- **[Info]** InstanceService.java:659 — Cache eviction ordering is safe (idempotent clear before endpoint release); worth a brief comment for future maintainers
- **[Info]** SettingsService.java:59 — Good constant extraction; consider a one-line Javadoc explaining the cross-service dependency

### Suggestions
No blocking concerns. The fix is minimal, well-targeted, and the regression test coverage is solid.

---

PR #4204, opened 2026-09-09
[studio][ISSUE #4203] Make alert acknowledgement atomic

### LINKED ISSUE #4203: [Studio][Bug] Alert acknowledgement can diverge from active alert state
### Description

`AlertService#acknowledgeAlert` persists two related pieces of state:

1. it marks the selected `rmq_system_alert` event as acknowledged; and
2. for the matching FIRING occurrence, it changes `rmq_alert_state` to ACKED.

The service method currently has no transaction boundary. If the second database write throws after the event update succeeds, the request fails but the event remains acknowledged while the active state remains FIRING. The state machine may then continue producing reminder notifications for an event that the UI already shows as acknowledged.

This is distinct from the expected stale-occurrence case where `AlertStateRepository#acknowledge` returns `false`: an old event may still be acknowledged without changing a newer incident. The inconsistent state here is caused by a persistence exception midway through the current occurrence update.

### Expected behavior

The event acknowledgement and matching active-state acknowledgement should commit or roll back atomically when both writes are attempted.

### Proposed fix

Add a transaction boundary to `AlertService#acknowledgeAlert` and cover both rollback-on-failure and successful retry with an H2-backed integration test.

89799969:
## Summary

- make system-alert event acknowledgement and active-state acknowledgement share one transaction
- roll both database writes back when updating the matching FIRING state fails
- preserve the existing stale-occurrence behavior where the state repository returns `false`
- add an H2-backed regression test for rollback and successful retry

## Problem

`AlertService#acknowledgeAlert` first marks `rmq_system_alert` as acknowledged and then updates the matching `rmq_alert_state`. Without a transaction, an exception from the second write leaves the event acknowledged while the active state is still FIRING, so reminder evaluation can continue for an incident the UI shows as acknowledged.

Fixes #4203

## Verification

- `mvn -B -Dmaven.repo.local=D:\projects\oss-contributions\m2-rocketmq -Dtest=AlertAcknowledgementTransactionTest,AlertServiceTest test`
- 77 tests passed, 0 failures, 0 errors
- Checkstyle: 0 violations


server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertService.java
```diff
@@ -556,6 +556,7 @@ public List<SystemAlertVO> findRelatedAlerts(Long id) {
     }
 
 
+    @Transactional
```
RockteMQ-AI, line 559: **[Info]** The `@Transactional` annotation works correctly here since `acknowledgeAlert` is called externally (from a controller). One thing to be aware of: if this method is ever called from another method within the same class, the proxy will be bypassed and the transaction boundary will not apply. Consider adding a brief comment or documenting this constraint for future maintainers.

server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertAcknowledgementTransactionTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @MockBean
+    private AlertRepository alertRepository;
+
+    @MockBean
+    private AlertStateRepository alertStateRepository;
+
+    @MockBean
+    private AlertRuleAssetService alertRuleAssetService;
+
```
RockteMQ-AI, line 62: **[Info]** Well-designed test. The probe-table approach (`alert_ack_tx_probe`) to verify transactional atomicity is clever — it sidesteps the difficulty of inspecting mock interactions across transaction boundaries. Using `Propagation.NOT_SUPPORTED` on the test class correctly lets the service's own `@Transactional` boundary take effect.

Minor: the method name `stateFailureRollsBackAlertAcknowledgementButSuccessfulRetryCommitsTest` is quite long. Consider splitting into two focused test methods (`shouldRollBackBothWritesOnStateFailure` and `shouldCommitBothWritesOnRetry`) for better readability and failure isolation.

RockteMQ-AI (approved):
### Summary
This PR adds `@Transactional` to `AlertService#acknowledgeAlert` to ensure both database writes (marking `rmq_system_alert` as acknowledged and updating the matching `rmq_alert_state`) happen atomically. The fix is minimal, correct, and well-tested.

### Findings
- **[Info]** `AlertService.java:559` — Self-invocation caveat for `@Transactional` proxy
- **[Info]** `AlertAcknowledgementTransactionTest.java:62` — Test design is strong; method name could be split for readability

### Verification
- CLA: unknown (repo may not use cla-assistant)
- No existing reviews or comments
- No CI checks reported on the branch

The fix correctly addresses the race condition described in #4203. The regression test uses a clever probe-table approach to verify transactional atomicity without relying on mock interaction inspection.

---

lizhimins (changes requested):
The one-line production change is correct — `AlertService.acknowledgeAlert` (`:560`) is the only write method in that class without `@Transactional`, while its siblings at `:190/:212/:262/:277/:291/:302/:313/:346/:353/:375/:585` all have it, and the controller calls it directly.

The test is the blocker: it is written against Spring Boot 3 APIs while trunk is on Boot 4.1.0 (`server/pom.xml:10`). `spring-boot-test-4.1.0` has no `MockBean`, `@JdbcTest` is not available at that version, and all 34 existing test files use `@MockitoBean`. Please port the test to `@MockitoBean` (or to the plain-Mockito style used by the neighbouring `AlertServiceTest`) and we will take it.

89799969:
Addressed the review request: ported the transaction test off Boot 3 `@MockBean`/`@JdbcTest`.

- Production change remains the one-line `@Transactional` on `AlertService#acknowledgeAlert`.
- Test now uses plain Mockito (`@ExtendWith(MockitoExtension.class)`), matching `AlertServiceTest` style on Boot 4.1.0.
- Coverage: `@Transactional` present, both event+state writes ordered, state failure propagates, non-FIRING skips state ACK.

Focused `AlertAcknowledgementTransactionTest` is green locally.

89799969:
@lizhimins Following up - the Boot 4.1 test port is pushed (`70e408a3`) and this branch is `MERGEABLE/CLEAN` on current `rocketmq-studio` (`e5ac5244`). Ready for re-review when you have a moment.

PR #4211, opened 2026-09-10
fix(web): honour the declared table column widths

### LINKED ISSUE #4208: [Studio] Declared table column widths are not applied
Every list table in Studio passes a numeric `scroll.x` computed from its column widths (`tableScrollX`), on the assumption that this makes the declared widths authoritative. It does not.

rc-table only switches to `table-layout: fixed` when `scroll.y` is set, a column is `fixed`, or a column is `ellipsis`. Without one of those the table stays on `table-layout: auto`, where the declared widths are only advisory and a single long cell value stretches its column past them.

**Reproduce** — on the Group management page, create a consumer group with a long name, e.g. `GID_A_VERY_LONG_CONSUMER_GROUP_NAME_FOR_ALIGN` (45 chars).

Measured on the resulting table:

| | declared | rendered |
|---|---|---|
| Group name column | 190px | **432px** |
| table width | 1154px | **1371px** |

`getComputedStyle(table).tableLayout` returns `auto`.

The defect is invisible on an empty table: with no rows there is no content to stretch a column, so the widths happen to match their declarations. That is why it went unnoticed.

Two consequences show up once the layout is forced to `fixed`:

1. The table still carries `min-width: 100%`, so a window wider than the table shares the surplus across every column in proportion to its width. On a 1482px container every column inflates by 28% — the expand-arrow column renders at 62px and a `Push`/`Pull` tag column at 108px.
2. Concentrating the surplus in a single flexible column overshoots when the fixed columns are narrow: a K8s cluster name column reached 492px while the certificate issuer DN next to it stayed truncated at 180px.

The `tableScrollX` doc comment also states that a numeric `scroll.x` switches the table to `table-layout: fixed`, which does not match rc-table behaviour and misled every call site.

lizhimins:
Closes #4208

## Root cause

rc-table only switches to `table-layout: fixed` when `scroll.y` is set, a column is `fixed`, or a column is `ellipsis` (`mergedTableLayout` in `rc-table/Table.js`). A numeric `scroll.x` on its own leaves it on `auto`, where the declared widths are advisory. The `tableScrollX` doc comment claimed the opposite, which is what misled every call site; it is corrected here.

## Changes

- `tableLayout="fixed"` is passed explicitly at every `tableScrollX` call site, so the invariant no longer depends on a column incidentally being ellipsized.
- The primary text column of each table uses `minWidth` instead of `width`. Under `fixed` the table still carries `min-width: 100%`, so surplus width is otherwise shared across all columns in proportion — inflating an expand arrow to 62px and a `Push` tag to 108px. With one unsized column the surplus lands there and every other column holds its declared width exactly, while `minWidth` stays the floor on narrow windows. `tableScrollX` now counts `minWidth` towards the total, otherwise the unsized column falls back to the 120px default and the computed minimum comes out too small.
- Tables with several long-text fields share the surplus between two or three of them, and the flexible column is the field that benefits from the width — a certificate issuer DN, a DLQ topic, a message ID — rather than whichever column came first. Before/after at a 1482px container:

| table | flexible column before | after |
|---|---|---|
| message query | Topic 465px | 281px (Topic / Key / Message ID) |
| ACL rules | principal 462px | 351px (principal / resource) |
| Topic | name 520px | 360px (name / remark) |
| ACL risk | recommendation 742px | 407px (item / evidence / recommendation) |
| K8s certs | cluster name 492px | issuer 412px |

- Long-name columns get `ellipsis` plus a tooltip carrying the full value. Columns rendering two lines keep a fixed width — column-level `ellipsis` forces `nowrap` on the cell and would collapse them.
- The consumer group table regains the modification-time column that an earlier width-trimming attempt had dropped; with the surplus handled properly the eleven columns fit without a horizontal scrollbar.
- The delete action on the K8s certificate table was the only table action in the app styled as a link button; it now matches the outlined style used elsewhere.

## Verification

- `getComputedStyle(table).tableLayout` is `fixed` on every list page.
- Consumer group table with a 45-char group name: column 190px (was 432px), table 1154px (was 1371px), long name truncates with a tooltip.
- `tsc --noEmit` clean, `eslint` no errors, `vitest run` 980 passing. `table.test.ts` gains cases for the `minWidth` accounting.
- Deployed and measured in a browser at container widths 305 / 650 / 1053 / 1290 / 1482px.

### DIFF
--- a/web/src/pages/cluster/certs.tsx
+++ b/web/src/pages/cluster/certs.tsx
@@ -170,7 +170,9 @@ const K8sCertsPage = () => {
       title: '签发者',
       dataIndex: 'issuer',
       key: 'issuer',
-      width: 180,
+      // 唯一可伸展列：余量集中在此，其余列保持声明宽度。选签发者而非集群名，
+      // 是因为它的内容是 CN=...,OU=...,O=... 这种长 DN，真正需要额外宽度。
+      minWidth: 180,
       sorter: (a, b) => (a.issuer ?? '').localeCompare(b.issuer ?? ''),
       render: (issuer: string | null) => issuer || '-',
       ellipsis: true,
@@ -216,7 +218,7 @@ const K8sCertsPage = () => {
           expiring: { color: 'orange', label: '即将过期' },
           expired: { color: 'red', label: '已过期' },
         };
-        const cfg = status ? map[status] ?? { color: 'default', label: status } : null;
+        const cfg = status ? (map[status] ?? { color: 'default', label: status }) : null;
         return cfg ? <Tag color={cfg.color}>{cfg.label}</Tag> : '-';
       },
     },
@@ -234,10 +236,9 @@ const K8sCertsPage = () => {
           okButtonProps={{ danger: true }}
         >
           <Button
-            type="link"
             size="small"
-            danger
             icon={<DeleteOutlined />}
+            style={{ borderColor: '#ff4d4f', color: '#ff4d4f' }}
             loading={deletingId === cert.id}
           >
             删除
@@ -288,6 +289,7 @@ const K8sCertsPage = () => {
           loading={loading}
           pagination={{ pageSize: 20 }}
           size="small"
+          tableLayout="fixed"
           scroll={{ x: tableScrollX(certColumns) }}
         />
       </Card>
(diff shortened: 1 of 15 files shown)


RockteMQ-AI (approved):
### Summary
Thorough fix for table column width handling across 15 files. Correctly identifies the `rc-table` `table-layout` behavior and applies `tableLayout="fixed"` consistently. The `minWidth` approach for flexible columns is well-reasoned, and the before/after measurements demonstrate clear improvements.

LGTM — excellent attention to detail with comprehensive testing.

---

PR #4213, opened 2026-09-10
fix(server): declare the cache manager so the data source API works

### LINKED ISSUE #4210: [Studio] GET /api/settings/datasources returns 500: cannot find cache named data-sources
Every call to `GET /api/settings/datasources` and `GET /api/settings/datasources/page` answers HTTP 500.

```
java.lang.IllegalArgumentException: Cannot find cache named 'data-sources' for Builder[public java.util.List
org.apache.rocketmq.studio.settings.SettingsService.listDataSources()] caches=[data-sources]
  at org.springframework.cache.interceptor.AbstractCacheResolver.resolveCaches(AbstractCacheResolver.java:92)
  at org.springframework.cache.interceptor.CacheAspectSupport.getCaches(CacheAspectSupport.java:317)
  ...
  at org.apache.rocketmq.studio.settings.SettingsService$$SpringCGLIB$$0.listDataSources(<generated>)
  at org.apache.rocketmq.studio.settings.SettingsController.listDataSources(SettingsController.java:60)
```

`@EnableCaching` sits on `StudioApplication` and `@Cacheable("data-sources")` on
`SettingsService.listDataSources`, but no `CacheManager` bean is declared anywhere — the manager is
left to Spring Boot cache auto-configuration. Whatever it resolves to in this application cannot
serve the `data-sources` cache, so the caching interceptor throws before the method body runs.

**User-visible impact:**

- Settings → data source management: `Internal Server Error` plus `数据源加载失败，请稍后重试`, table empty.
- `/ops/dashboard`: an `Internal Server Error` alert, because the metrics panel loads the data source list on mount. The error appears on a page that otherwise has nothing to do with settings, which makes it hard to trace.

**Why tests did not catch it:** `SettingsServiceCachingTest` registers its own
`ConcurrentMapCacheManager("data-sources")` in the test context. That masks the missing production
wiring completely — the test has been green while every production request failed.

Also noting that the local `deploy/rocketmq` stack has no Prometheus, so the metrics dashboard has
nothing to query even once the data source API works: all nine panels report
`Prometheus base URL is not configured`.

lizhimins:
Closes #4210

## Changes

`CacheConfig` declares the `CacheManager` explicitly, removing the dependency on whatever Spring
Boot cache auto-configuration resolves to in this application. It carries no fixed cache-name list,
so `ConcurrentMapCacheManager` creates caches on first use and a future `@Cacheable` with a new
name cannot fail the same way.

`CacheConfigTest` mirrors the production wiring (`@EnableCaching` + `CacheConfig`) and asserts the
manager hands out `data-sources`, hands out an arbitrary future name, and that the caching
interceptor is present. It fails without this configuration — unlike `SettingsServiceCachingTest`,
which supplies its own manager and therefore masked the missing wiring.

The rest gives the metrics dashboard something to query, which the local stack did not have:

- Both brokers export Prometheus metrics on 5557. `metricsPromExporterHost` is set to `0.0.0.0`
  explicitly — it defaults to `brokerIP1`, which leaves the endpoint unreachable from outside the
  container.
- A `prometheus` service scrapes them by container name, so a rebuild changing the container IP
  needs no config change. Port 9090 binds to the loopback interface only.
- `.env.example` points `STUDIO_METRICS_PROMETHEUS_BASE_URL` at that container. It cannot be a
  loopback address: `UrlHostGuard` rejects loopback for metrics data sources and the query fails
  with "base URL is invalid".

## Verification

- `GET /api/settings/datasources` and `/page`: HTTP 500 → 200, and `Cannot find cache` no longer
  appears in the server log (it was logged on every request before).
- Settings data-source tab and `/ops/dashboard` both load with no error alert.
- End to end after wiring Prometheus: broker `/metrics` returns 200 (54KB / 47KB), all scrape
  targets report `health: up`, `POST /api/metrics/query` returns a matrix with real samples, and
  the nine dashboard panels go from nine `Prometheus base URL is not configured` errors to zero.
- `mvn test` for `CacheConfigTest`, `SettingsServiceCachingTest`, `SettingsServiceTest`,
  `SettingsControllerTest`: 69 passing.

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -22,7 +22,11 @@ STUDIO_AUTH_ADMIN_USERNAME=admin
 STUDIO_AUTH_ADMIN_PASSWORD=change-me
 
 # Optional Prometheus-compatible data source for the built-in metrics proxy.
-STUDIO_METRICS_PROMETHEUS_BASE_URL=
+# `deploy/rocketmq/docker-compose.yml` ships a `rocketmq-prometheus` container on the
+# shared `rocketmq_net` network that scrapes both brokers, so the container name below
+# works as-is once that stack is up. Do not use localhost/127.0.0.1: UrlHostGuard blocks
+# loopback for metrics data sources and the query fails with "base URL is invalid".
+STUDIO_METRICS_PROMETHEUS_BASE_URL=http://rocketmq-prometheus:9090
 STUDIO_METRICS_PROMETHEUS_USERNAME=
 STUDIO_METRICS_PROMETHEUS_PASSWORD=
 STUDIO_METRICS_PROMETHEUS_BEARER_TOKEN=
--- a/deploy/rocketmq/conf/broker-0.conf
+++ b/deploy/rocketmq/conf/broker-0.conf
@@ -13,3 +13,8 @@ autoCreateSubscriptionGroup=true
 # 消息轨迹：开启并自动创建轨迹 topic RMQ_SYS_TRACE_TOPIC
 traceOn=true
 traceTopicEnable=true
+# Prometheus 指标导出（Studio 监控面板数据源）
+# Host 必须显式设为 0.0.0.0：默认取 brokerIP1，会导致容器外抓取不到 5557
+metricsExporterType=PROM
+metricsPromExporterPort=5557
+metricsPromExporterHost=0.0.0.0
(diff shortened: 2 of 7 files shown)


RockteMQ-AI (approved):
### Summary
Explicit `CacheManager` declaration fixes the missing-cache failure in the data-source API, and the accompanying test covers the exact wiring gap. Deploy config additions for Prometheus metrics are well-structured.

LGTM — clean fix with good test coverage and clear documentation.

---

PR #4177, opened 2026-09-08
fix(instance): keep cloud instance endpoints read-only in the edit dialog

unbridled-41:
[...]

PR #4214, opened 2026-09-10
[studio][ISSUE #4176] Make cloud instance endpoint read-only in the edit dialog

### LINKED ISSUE #4176: [Studio][Bug] Cloud instance edit dialog lets you edit the endpoint but the backend silently ignores it
**Problem**

Editing a cloud (Aliyun/Tencent) instance opens a dialog where the **接入地址 (endpoint)** field is fully editable. Saving always reports `实例「{name}」已更新` success — but the backend silently drops the endpoint for cloud vendors, so the change is never persisted.

**Evidence**

- Backend `InstanceService.updateInstance` (lines ~589–600): `boolean cloudInstance = existing.getVendor() != null && existing.getVendor() != InstanceVendor.APACHE;` and only `if (!cloudInstance) { ... updated.setEndpoint(requireValidEndpoint(instance.getEndpoint())); }`. Cloud endpoints are resolved from the cloud catalog (`resolveEndpoint(detail)`, `InstanceService.java:512`) and `createCloudInstance` even rejects a manually supplied endpoint (`"Commercial instances must be selected from the cloud catalog, endpoint cannot be set manually"`).
- Frontend `web/src/pages/instance/index.tsx`: the edit dialog's endpoint `Input` is unconditional (no vendor gating — only `adminCredentialRef` is gated to `vendor === 'APACHE'`), and `handleUpdate` posts `endpoint: values.endpoint` then unconditionally shows `message.success(t('instance.updated', { name }))`.

**Impact**

The operator edits what looks like a critical connection setting, sees a success toast, and only discovers on the next connection failure that the endpoint never changed. The UI text itself states the intent — `instance.cloudEndpointExtra`: "云服务实例接入地址由云厂商目录解析，不支持手动修改" — but the input ignores it.

**Expected behavior**

For cloud vendors the endpoint is rendered read-only (disabled), matching the documented intent and the backend contract; Apache instances remain editable (regression-tested).

**Related work**

- #3937 (vendor filter/pagination) touches the same page but not the edit dialog; searched "instance edit endpoint / cloud endpoint edit / endpoint not saved" — no Issue or PR covers this.

**PR**

Fix: #4177

89799969:

- Fixes #4176

### Brief Description

The backend already ignores endpoint updates for Aliyun/Tencent instances and resolves the endpoint from the cloud catalog (`InstanceService.updateInstance` only calls `setEndpoint` for non-cloud vendors). The edit dialog still rendered an editable 接入地址 input and always showed success, so operators could believe an endpoint change was saved.

This change disables the endpoint field when editing a cloud instance, matching the existing cloud-only type lock and the `instance.cloudEndpointExtra` copy.

```text
Node 24 vitest run src/pages/instance/__tests__/InstancePage.test.tsx -t "disables the endpoint"
Test Files  1 passed (1)
Tests       1 passed | 23 skipped (24)

Node 24 vitest run src/pages/instance/__tests__/InstancePage.test.tsx -t "updates instance type and endpoint"
Test Files  1 passed (1)
Tests       1 passed | 23 skipped (24)
```

AI-assisted contribution; implementation and tests were reviewed and verified locally.

### DIFF
--- a/web/src/pages/instance/__tests__/InstancePage.test.tsx
+++ b/web/src/pages/instance/__tests__/InstancePage.test.tsx
@@ -81,6 +81,7 @@ const instance = (
   name: string,
   type: Instance['type'] = 'PROXY_CLUSTER',
   remark: Instance['remark'] = '',
+  vendor?: Instance['vendor'],
 ): Instance => ({
   id,
   name,
@@ -91,6 +92,7 @@ const instance = (
   consumerGroupCount: 1,
   gmtCreate: '2026-01-01T00:00:00Z',
   gmtModified: '2026-01-01T00:00:00Z',
+  ...(vendor ? { vendor } : {}),
 });
 
 const renderPage = () =>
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM.

---

lizhimins:
Closing as covered by #4177, which is now merged into `rocketmq-studio` as `da6d98a5e`.

Mainline landed `disabled={editingInstance?.vendor != null && editingInstance.vendor !== 'APACHE'}`, which is strictly broader than the `vendor === 'ALIYUN' || vendor === 'TENCENT'` check here — it also covers any future non-APACHE vendor instead of needing the list extended. #4177 was opened a day earlier and ships the matching `InstancePage.test.tsx` case (`keeps the endpoint editable for Apache instances but read-only for cloud vendors`), so the behaviour is pinned as well.

Thanks for filing #4176 and for the fix — the diagnosis was right, and it is now in.

PR #4219, opened 2026-09-11
[Studio] feat: seal cloud credential secrets with AES-GCM and gate reveal behind re-auth

zhaohai666:
# feat(studio): seal cloud credential secrets with AES-GCM and gate reveal behind re-auth

## Problem

Cloud provider credentials (AK/SK for Aliyun / Tencent instances) are stored in `rmq_cloud_credential.secret_key` **base64-encoded**. Base64 is an encoding, not confidentiality: anyone with read access to the table (a SQL dump, a backup, a co-located process) recovers every provider secret key trivially. On top of that, the reveal endpoint (`GET /api/cloud-credentials/{id}/credentials`) trusts the admin session alone — a replayed or stolen session cookie is enough to exfiltrate all stored keys, with no audit trail beyond the access itself.

## Solution

Two focused hardening steps, deliberately small in surface:

### 1. AES-256-GCM encryption at rest

- New `CredentialCipher` seals secrets as `enc:v1:<base64(iv || ciphertext || tag)>`:
  - a fresh random 96-bit IV per value, so identical secrets never collide in the DB;
  - GCM is authenticated — a tampered row **fails closed** (clean 500) instead of returning corrupted key material.
- The key comes from `studio.credential.encryption-key` (`STUDIO_CREDENTIAL_ENCRYPTION_KEY`), a Base64-encoded 32-byte AES key; invalid keys fail fast at startup. When unset (local/dev convenience), a deterministic development key is derived and a **loud WARNING** names the exact property to set in production.
- `MybatisPlusCloudCredentialRepository` now encrypts on write and decrypts on read. In-memory state stays plaintext; only the persisted column changes.
- **Backward compatibility**: rows written by the legacy base64 scheme (no `enc:v1:` prefix) keep working through the tolerant decode path.
- **`CredentialEncryptionMigration`** (startup `ApplicationRunner`, idempotent) re-seals every legacy row once, so an upgraded deployment ends up fully encrypted without manual edits. Failures are logged, not thrown — a hardening step must not brick startup.

### 2. Reveal requires a second authentication

`GET /api/cloud-credentials/{id}/credentials` now additionally requires the operator's password in the `X-Studio-Reauth` header:

- The password is verified **against the session principal** (`AuthenticatedUserContext`), never against a caller-supplied user name — it cannot be abused to probe another account's password.
- Reveal is already admin-gated by `AuthInterceptor`; this adds a second factor so a stolen admin session cookie alone cannot read provider keys.
- Every attempt is audited (`REVEAL_CLOUD_CREDENTIAL`, SUCCESS and FAILURE with the denial reason).
- The password travels in a header, never in the URL/query string; the response keeps `Cache-Control: no-store`.

**Out of scope by design**: gRPC channel auto-signing is untouched.

## Runtime verification (live dashboard + H2 file DB)

| Check | Result |
|---|---|
| DB content after create | `secret_key = enc:v1:Fic02WL…` (plaintext never persisted) |
| Reveal without header | `403 Password confirmation is required…` |
| Reveal with wrong password | `403 Password confirmation failed` |
| Reveal with correct password | `200` + plaintext secret (decrypt path works) |
| Audit trail | 3 rows: `REVEAL_CLOUD_CREDENTIAL` 2× FAILURE + 1× SUCCESS |
| Restart with same key | decrypts correctly across restarts |

## Testing

- New `CredentialCipherTest` (12): round-trip, IV freshness, tamper rejection, wrong-key rejection, legacy base64/plain read, key validation, fallback stability.
- New `CredentialEncryptionMigrationTest` (3): re-seals legacy rows only, skips sealed rows, never propagates failure.
- `MybatisPlusCloudCredentialRepositoryTest`: real cipher, asserts sealed ciphertext on write and decryption on read, legacy rows still readable.
- `CloudCredentialServiceTest` / `CloudCredentialControllerTest` / `AuthServiceTest` / `AuthCredentialAuthorizationIntegrationTest`: updated for the re-auth contract (fail-closed, audited denial, header forwarding).
- Focused suite **69/69 green**; full server suite shows no new failures (only the 2 pre-existing `AuthCorsIntegrationTest` failures).

## Configuration

```yaml
studio:
  credential:
    # Base64-encoded 32-byte AES-256 key. Generate with: openssl rand -base64 32
    encryption-key: ${STUDIO_CREDENTIAL_ENCRYPTION_KEY:}
```

## Checklist

- [x] No dependency changes; AES-GCM is JDK built-in
- [x] Legacy base64 rows keep working and are migrated automatically
- [x] Reveal: admin gate + password re-auth + audit, fail-closed
- [x] gRPC channel auto-signing untouched
- [x] Unit tests + live runtime verification (ciphertext inspected directly in the DB)

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
@@ -159,6 +159,41 @@ public class AuthService {
         return getAuthenticatedUser(authorization).map(LoginVO.UserInfo::isAdmin).orElse(false);
     }
 
+    /**
+     * Second-factor confirmation for operations that expose stored secrets.
+     *
+     * <p>The caller is already an authenticated admin by the time this runs (the interceptor gates the
+     * reveal endpoints), so this deliberately asks for the account password again: a replayed session
+     * cookie is not enough to read provider keys. The confirmation is resolved against the currently
+     * authenticated principal from {@link AuthenticatedUserContext}, never against a caller-supplied
+     * user name, so it cannot be used to check another account's password.</p>
+     *
+     * @throws BusinessException 403 when the password is missing, the principal is unknown, or the
+     *         password does not match - all cases fail closed.
+     */
+    public void verifySensitiveOperationPassword(String rawPassword) {
+        if (rawPassword == null || rawPassword.isBlank()) {
+            throw new BusinessException(403,
+                    "Password confirmation is required to reveal stored secrets");
+        }
+        String username = AuthenticatedUserContext.currentUsernameOrSystem();
+        if (databaseBacked()) {
+            ensureBootstrapUsers();
+            RmqStudioUser user = findUserByUsername(username).orElse(null);
+            if (user == null || !Boolean.TRUE.equals(user.getEnabled())
+                    || !passwordHasher.matches(rawPassword, user.getPasswordHash())) {
+                throw new BusinessException(403, "Password confirmation failed");
+            }
+            return;
+        }
+        boolean matched = authProperties.configuredUsers().stream()
+                .anyMatch(candidate -> candidate.getUsername().equals(username)
+                        && candidate.getPassword().equals(rawPassword));
+        if (!matched) {
+            throw new BusinessException(403, "Password confirmation failed");
+        }
+    }
+
     public void logout(String authorization) {
         tokenFromAuthorization(authorization).ifPresent(token -> {
             if (databaseBacked()) {
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/CredentialCipher.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/CredentialCipher.java
@@ -1,0 +1,177 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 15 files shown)


RockteMQ-AI (approved):
### Summary
Significant security hardening for cloud credential storage. Replaces base64 encoding (which provides no confidentiality) with AES-256-GCM authenticated encryption, and gates the reveal endpoint behind password re-authentication to prevent session-replay exfiltration.

**Security design is solid:**
- AES-256-GCM with fresh 96-bit IV per encryption — correct nonce handling, authenticated encryption prevents tampering
- `enc:v1:` prefix allows future rotation and backward-compatible reads of legacy data
- Re-auth via `X-Studio-Reauth` header resolves against `AuthenticatedUserContext` (never caller-supplied username) — cannot be used to check another account's password
- Failed reveal attempts are audited alongside successes
- Migration is idempotent, non-blocking on failure, and logs clearly when the dev fallback key is in use

**Strengths:**
- Defense in depth: encryption at rest + re-auth for reveal + audit logging
- Clean separation: `CredentialCipher` handles crypto, `AuthService` handles auth, migration handles upgrade path
- Comprehensive test coverage (362+ lines of tests)
- `CacheControl.noStore()` on reveal responses prevents caching of plaintext secrets

**Minor observations (non-blocking):**
- The dev fallback key uses a fixed salt — documented and warned, but worth ensuring deployment docs prominently call out the production requirement
- Consider rate-limiting reveal attempts to slow brute-force password guessing

LGTM.

---

lizhimins:
Closing, with the reasoning below — the cryptography itself is not the problem.

What we verified as correct: the key comes from `STUDIO_CREDENTIAL_ENCRYPTION_KEY` with no hard-coded secret in the repository; the IV is freshly generated per encryption and never reused; the 128-bit tag is validated by JCE so tampering or a key change throws instead of returning garbage; ciphertexts carry an `enc:v1:` prefix; re-authentication is enforced server-side (`CloudCredentialService` -> `AuthService`, session principal only, 403 for empty/wrong/disabled, audited on both outcomes); and `list`/`export` keep masking. No bypass found.

Two blocking issues remain. First, the fallback key is the SHA-256 of a public constant in this repository, and the WARNING your javadoc promises for that case is never logged — a deployment can silently run on a guessable key. Second, the backfill seals existing rows with that fallback key and skips rows that already carry the prefix, so configuring a real key afterwards makes the previously migrated credentials permanently unreadable with no re-seal path and no way out from the UI; a single undecryptable row also fails `findAll()` and takes the whole credential list page down.

We are not willing to merge encryption whose failure mode is silent, permanent data loss. Please open an issue so we can agree the key policy (no fallback, or no migration until a real key is configured) and then resubmit — the rest of the patch is close.

PR #4224, opened 2026-09-11
fix(auth): escape LIKE wildcards in username search

### LINKED ISSUE #4223: [Studio][Bug] Studio user search does not escape SQL LIKE wildcards
## Problem

`AuthService.listUsers` passes the raw username search into MyBatis-Plus `like("username", normalizedSearch)` without escaping SQL LIKE wildcards. A search for `admin_` matches `adminX`, `admin1`, etc.; a search for `admin%` matches any suffix.

Instance, credential, and operation-audit search paths already escape `\`, `%`, and `_` (see #4192–#4194). The studio user search path does not.

## Evidence

```java
String normalizedSearch = search == null ? "" : search.trim();
// ...
.like(!normalizedSearch.isEmpty(), "username", normalizedSearch)
```

`AuthService.java` `listUsers`.

## Expected behavior

Username search matches the literal search string. `_` and `%` in the query are not treated as wildcards.

## Scope

`AuthService.listUsers` + focused unit regression test. No API change.

89799969:
## Summary
- Escape `\`, `%`, and `_` in `AuthService.listUsers` username search so LIKE wildcards match literally.
- Add a focused database-layer regression test.

## Issue
Fixes #4223

## Testing
- Added `listUsersEscapesLikeWildcardsInUsernameSearch` in `AuthServiceDatabaseTest`.
- Please run the focused Maven test for `AuthServiceDatabaseTest`.

## Related
Same pattern as instance/credential/audit LIKE escape PRs (#4192–#4194).


RockteMQ-AI (approved):
### Summary
Escapes SQL LIKE wildcards (`\`, `%`, `_`) in `AuthService.listUsers` username search so special characters match literally instead of being interpreted as SQL patterns. This is important because usernames commonly contain underscores.

The fix is minimal, correct, and consistent with the same pattern already applied in #4192–#4194. The regression test properly verifies the escaped value is passed to the query.

LGTM.

---

RockteMQ-AI (commented):
**Note:** This PR fixes the same issue (#4223) as PR #4227, which was submitted around the same time. Both implement LIKE wildcard escaping for the user search.

PR #4227 uses a package-private static helper with a unit test, while this PR uses a private static helper with a database integration test. The maintainers may want to consolidate these changes.

The fix itself is correct — escaping backslash before percent/underscore is the right order.

---

RockteMQ-AI (approved):
### Summary

Same LIKE-wildcard escape pattern applied to `AuthService` login failure search. Clean and minimal.

### Findings
- **[Info]** Same `escapeLike` duplication note as #4227, #4230, #4233.

LGTM.

RockteMQ-AI (approved):
### Summary

Same LIKE-wildcard escape pattern applied to `AuthService` login failure search. Clean and minimal.

### Findings
- **[Info]** Same `escapeLike` duplication note as #4227, #4230, #4233.

LGTM.

lizhimins (changes requested):
One mechanical item: drop the private `static escapeLike` and call the shared utility instead. This is the third identical private copy across #4224 / #4230 / #4233 (and #4290), and the project rule is that shared logic lives in `common/util`.

While you are there, please add an explicit `ESCAPE` clause — the current escaping relies on MySQL's implicit backslash escape character, which H2 does not provide, and five tests in this suite run against H2. Keep the binding-parameter assertions; they are the part that makes the test meaningful.

89799969:
Addressed the review request:

- Replaced the private `escapeLike` copy with shared `common/util/SqlLikeUtil`.
- Switched to `apply(... LIKE {0} ESCAPE '\')` so H2 no longer depends on MySQL's implicit backslash escape.
- Kept the binding-parameter assertions.

`SqlLikeUtilTest` + `AuthServiceDatabaseTest` green locally. Same pattern applied on #4230 and #4233.

89799969:
@lizhimins Following up - shared `SqlLikeUtil` + explicit `ESCAPE` is pushed (`5240673e`) and this branch is `MERGEABLE/CLEAN` on current `rocketmq-studio`. Ready for re-review. Note: #4230/#4233 carry the same util file; after the first of the three lands I will rebase the others to drop the duplicate.

RockteMQ-AI (approved):
LGTM. Code changes look good.

---

PR #4226, opened 2026-09-11
fix(studio): preserve Unicode code points when truncating message properties

### LINKED ISSUE #4167: [Studio][Bug] Preserve Unicode code points when truncating message properties

jokerzsd:
## Summary

Fix message property values being truncated mid-surrogate-pair, producing unpaired surrogates.

## Problem

`MessagePropertyDisplay.abbreviate` used UTF-16 `length()`/`substring(0, 1024)`. When the boundary fell between the high and low surrogate of a supplementary character (e.g. an emoji), the truncated value contained an unpaired surrogate and rendered as a replacement character.

## Fix

Count code points (`codePointCount`) and truncate at a code-point boundary (`offsetByCodePoints`). `hasOversizedProperty` is updated to the same code-point semantics for consistency.

## Test plan

Added a regression test (1023 ASCII chars + emoji straddling the boundary). `mvn test -Dtest=MessagePropertyDisplayTest` — 6 tests, 0 failures.

Fixes #4167

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/MessagePropertyDisplay.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/MessagePropertyDisplay.java
@@ -68,13 +68,15 @@ public final class MessagePropertyDisplay {
     /** True when any value exceeds {@link #MAX_PROPERTY_VALUE_CHARS} and would be abbreviated. */
     public static boolean hasOversizedProperty(Map<String, String> properties) {
         return properties != null && properties.values().stream()
-                .anyMatch(value -> value != null && value.length() > MAX_PROPERTY_VALUE_CHARS);
+                .anyMatch(value -> value != null
+                        && value.codePointCount(0, value.length()) > MAX_PROPERTY_VALUE_CHARS);
     }
 
     private static String abbreviate(String value) {
-        if (value == null || value.length() <= MAX_PROPERTY_VALUE_CHARS) {
+        if (value == null || value.codePointCount(0, value.length()) <= MAX_PROPERTY_VALUE_CHARS) {
             return value;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Correct handling of Unicode surrogate pairs during truncation. The test case with emoji (🚀🎉) demonstrates the fix works as expected.

---

RockteMQ-AI (approved):
### Summary

Correct fix for surrogate-pair splitting when truncating message properties. Using `codePointCount` and `offsetByCodePoints` instead of `length()`/`substring()` ensures emoji and other supplementary characters aren't split mid-codepoint. The test case (1023 ASCII + emoji at boundary) is well-chosen.

LGTM.

RockteMQ-AI (approved):
### Summary

Correct fix for surrogate-pair splitting when truncating message properties. Using `codePointCount` and `offsetByCodePoints` instead of `length()`/`substring()` ensures emoji and other supplementary characters aren't split mid-codepoint. The test case (1023 ASCII + emoji at boundary) is well-chosen.

LGTM.

lizhimins:
Closing. The change pins `abbreviate`, but `hasOversizedProperty` in the same class still counts UTF-16 chars, so an oversized property value falls back to `value.length()` and can still be cut inside a surrogate pair; the javadoc also still documents a character cap.

For context: #4170 covered both methods plus the javadoc and added a 1024-code-point boundary case, and its author closed it in favour of this PR. We are not taking either variant right now — a resubmission needs to cover `abbreviate`, `hasOversizedProperty` and the javadoc in one change.

PR #4227, opened 2026-09-11
fix(studio): escape SQL LIKE wildcards in user search

### LINKED ISSUE #4223: [Studio][Bug] Studio user search does not escape SQL LIKE wildcards

jokerzsd:
## Summary

Fix the Studio user search treating `_` and `%` in the query as SQL LIKE wildcards.

## Problem

`AuthService.listUsers` passed the raw username search into MyBatis-Plus `like("username", ...)` without escaping. A search for `admin_` matched `adminX` and `admin%` matched any suffix. Other search paths (e.g. message query history) already escape `\`, `%` and `_`.

## Fix

Escape `\`, `%` and `_` before building the LIKE condition, matching the existing `escapeLike` behavior.

## Test plan

Added a focused `escapeLike` regression test. `mvn test -Dtest=AuthServiceTest` — 17 tests, 0 failures.

Fixes #4223

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
@@ -188,7 +188,7 @@ public class AuthService {
                     "search must not exceed " + MAX_USER_SEARCH_LENGTH + " characters");
         }
         QueryWrapper<RmqStudioUser> query = new QueryWrapper<RmqStudioUser>()
-                .like(!normalizedSearch.isEmpty(), "username", normalizedSearch)
+                .like(!normalizedSearch.isEmpty(), "username", escapeLike(normalizedSearch))
                 .eq(admin != null, "admin", admin)
                 .eq(enabled != null, "enabled", enabled)
                 .orderByAsc("username")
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good fix with proper escaping of SQL LIKE wildcards and a focused unit test. The implementation correctly handles the backslash escape ordering.

---

RockteMQ-AI (approved):
### Summary

Same LIKE-wildcard escape fix applied to `AuthService.listUsers`. The `escapeLike` method is correct (backslash → percent → underscore order). Tests cover the `admin_` → `adminX` false-positive case.

### Findings
- **[Info]** `AuthService.java:537` — `escapeLike` is now duplicated across 4+ classes. Consider a shared utility to prevent drift.

LGTM.

lizhimins:
Closing as a duplicate of #4224, which was opened about two hours earlier, keeps the helper `private` instead of widening it to package-private `static`, and asserts the escaped value at the binding-parameter level. It also contradicts the approach in your own #4236, which puts the helper in `common/util/SqlLikeUtils` — one shared utility is what we want here.

Issue #4232, opened 2026-09-11
[Studio][Bug] Alert-rule search does not escape SQL LIKE wildcards


89799969:
﻿## Problem

Alert-rule inventory search treats `%` and `_` as SQL LIKE wildcards instead of literal characters:

- `GET /api/alert-rules/page?search=` / `GET /api/business-alert-rules/page?search=` filter `name` with unescaped `LIKE` (`MybatisPlusAlertRepository.ruleQuery`).
- Domain-scoped paging (`AlertRuleQuery.search`, used by the paged alert-rule feed) filters `name` and `metric` the same way (`MybatisPlusAlertRepository.findRulesPage`).

Alert-rule names and metric identifiers commonly contain underscores (`consumer_lag`, `disk_usage`). Searching for `consumer_lag` therefore also matches `consumerXlag`, and a trailing `%` matches any suffix — so the inventory silently returns unrelated rules or hides the intended one.

## Evidence

`MybatisPlusAlertRepository.ruleQuery` (`server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertRepository.java:75-80`) passes the search term straight into `.like(name, search)`.

`MybatisPlusAlertRepository.findRulesPage` (`:83-101`) does the same for both `name` and `metric` after only trimming the term.

The same class of defect was already fixed for query history (`QueryHistoryService.escapeLike`) and is being fixed for ACL inventory (#4229 / #4230), Studio user search, instance / audit / cloud-credential search; alert-rule search is the remaining high-traffic search surface with the gap.

## Impact

Operators cannot reliably look up the alert rule they mean when the name or metric contains `_` or `%`. In a shared Studio with many rules this produces wrong rows and can hide the intended rule.

## Expected behavior

User-supplied `search` terms are matched literally. Escape `\`, `%`, and `_` before building the LIKE pattern (same convention as `QueryHistoryService.escapeLike`). Blank/null filters keep their current no-filter semantics.

## Acceptance criteria

- Regression tests prove a term such as `consumer_lag%` is bound as `%consumer\_lag\%%` rather than `%consumer_lag%`.
- Cover both the legacy name-only page search and the domain page name/metric search.
- Do not change the public API contract.

## Related work

#2265 / `QueryHistoryService.escapeLike` established the escape convention. Open peers: #4229/#4230 (ACL inventory), #4223/#4224 (Studio user search), instance / audit / cloud-credential LIKE PRs.

[...]

Issue #4229, opened 2026-09-11
[Studio][Bug] ACL user and rule search do not escape SQL LIKE wildcards


89799969:
## Problem

ACL inventory search treats `%` and `_` as SQL LIKE wildcards instead of literal characters:

- `GET /api/acl/users/page?keyword=` filters `username` and `access_key` with unescaped `LIKE`.
- `GET /api/acl/rules/page?principal=` / `?resource=` filter `principal` and `resource` the same way.

Usernames and ACL principals commonly contain underscores (`svc_monitor`, `prod_user`). Searching for `prod_user` therefore also matches `prodXuser`, and a trailing `%` matches any suffix — so the inventory silently returns unrelated accounts or rules.

## Evidence

`MybatisPlusAclRepository.findUserPage` (`server/src/main/java/org/apache/rocketmq/studio/instance/acl/MybatisPlusAclRepository.java:115-129`) lowercases the keyword and passes it straight into `.like(username, search).or().like(access_key, search)`.

`MybatisPlusAclRepository.ruleQuery` (`:388-398`) does the same for `principal` and `resource`.

The same class of defect was already fixed for query history (`QueryHistoryService.escapeLike`) and is being fixed for instance / audit / cloud-credential / Studio-user search; ACL inventory is the remaining high-traffic search surface with the gap.

## Impact

Operators cannot reliably look up the ACL account or rule they mean when the identifier contains `_` or `%`. In a shared Studio with many service accounts this produces wrong rows (including other accounts' access keys in the page listing) and can hide the intended rule.

## Expected behavior

User-supplied `keyword`, `principal`, and `resource` are matched literally. Escape `\`, `%`, and `_` before building the LIKE pattern (same convention as `QueryHistoryService.escapeLike`). Blank/null filters keep their current no-filter semantics.

## Acceptance criteria

- Regression tests prove a keyword such as `prod_user%` is bound as `%prod\_user\%%` rather than `%prod_user%`.
- Cover both the user page search and the rule page principal/resource search.
- Do not change the public API contract.

## Related work

#2265 / `QueryHistoryService.escapeLike` established the escape convention. Open peers: #4223 (Studio user search), instance / audit / cloud-credential LIKE PRs.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the `rocketmq-studio` branch (commit `36126024`).

**Root Cause:**
[...]

PR #4233, opened 2026-09-11
fix(alert): escape LIKE wildcards in rule name and metric search

89799969:
## What is the purpose of the change

Fixes #4232.

Alert-rule inventory search treats `%` and `_` as SQL LIKE wildcards instead of literal characters:

- `GET /api/alert-rules/page?search=` / `GET /api/business-alert-rules/page?search=` filter `name` with unescaped `LIKE` (`MybatisPlusAlertRepository.ruleQuery`).
- Domain-scoped paging (`AlertRuleQuery.search`) filters `name` and `metric` the same way (`MybatisPlusAlertRepository.findRulesPage`).

Alert-rule names and metric identifiers commonly contain underscores (`consumer_lag`, `disk_usage`). Searching for `consumer_lag` therefore also matches `consumerXlag`, and a trailing `%` matches any suffix — so the inventory silently returns unrelated rules or hides the intended one.

## Brief changelog

- Add a private `escapeLike` helper (same behaviour as the existing `QueryHistoryService.escapeLike`) and apply it to the legacy name-only page search and the domain page name/metric search.
- Two regression tests assert the bound LIKE value for `prod_lag%` / `consumer_lag%` is escaped (`%prod\_lag\%%`) rather than treated as a wildcard pattern.

## Verifying this change

```
$ mvn -f server/pom.xml test -Dtest='Alert*Test,MybatisPlusAlertRepositoryTest'
[INFO] Tests run: 198, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Checkstyle: 0 violations (module validate phase).

Before the fix the two new tests fail because the bound parameters are `%prod_lag%` / `%consumer_lag%`; after it they bind `%prod\_lag\%%` / `%consumer\_lag\%%`.

Related peers: #4229/#4230 (ACL inventory), #4223/#4224 (Studio user search), #4192–#4194 (audit / instance / cloud-credential LIKE escape).

PR #4236, opened 2026-09-11
fix(studio): escape SQL LIKE wildcards in data source and metadata search

jokerzsd:
## Summary

Escape SQL LIKE wildcards in two remaining search paths that were not covered by the earlier per-surface fixes (#4223, #4229, #4192, #4193, #4194, #4232).

## Problem

- `MybatisPlusSettingsRepository.findDataSources` passed the raw search string into `like("json", ...)`.
- `RocketMQMetadataProvider` passed the raw search string into `like(RmqTopic::getName, ...)` / `like(RmqGroup::getName, ...)`.

Topic/group names and data source content can contain `%` and `_`, so a search treated those as LIKE wildcards and silently matched unrelated rows.

## Fix

Introduce a shared `SqlLikeUtils.escape` helper (extracted from the escaping already used by `QueryHistoryService`) and apply it to the remaining unescaped predicates.

## Test plan

Added `SqlLikeUtilsTest`. `mvn test -Dtest=SqlLikeUtilsTest` — 2 tests, 0 failures.

### DIFF
--- a/server/src/main/java/org/apache/rocketmq/studio/common/util/SqlLikeUtils.java
+++ b/server/src/main/java/org/apache/rocketmq/studio/common/util/SqlLikeUtils.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
LGTM. Good catch on the remaining unescaped `like()` predicates. Extracting `SqlLikeUtils.escape` into a shared helper is the right call — it centralizes the escaping logic and prevents future regressions. The test coverage for edge cases (backslash, percent, underscore, empty/null) is solid.

---

lizhimins:
Closing, but please read this one as a request to resubmit rather than a rejection of the approach.

Putting the helper in `common/util/SqlLikeUtils` is exactly the shape we want, and this is the only patch in the LIKE-escape set that covers all four call sites in `RocketMQMetadataProvider` plus `MybatisPlusSettingsRepository` without touching the hard-coded `notLikeRight`. The blocking problem is dialect dependence: the escaping relies on MySQL's implicit backslash escape character, and without an explicit `ESCAPE` clause the same helper is wrong elsewhere — five tests in this suite run against H2, which has no default LIKE escape character, so the escaped pattern silently means something different there.

Add the explicit `ESCAPE` clause (and keep the tests asserting the bound parameter) and we will take it as the anchor for the whole cluster: #4192 #4193 #4194 #4224 #4230 #4233 #4290 are all closed or returned, so this is the one that should land.

PR #3131, opened 2026-09-04
[Studio] feat: add audit risk insights to the audit dashboard

yx9o:
## What is the purpose of the change

Add an audit risk insights panel to RocketMQ Studio so operators can quickly identify risky audit signals from the current audit filters and visible audit records, including failure rates, partial outcomes, control-plane failures, high-risk operation failures, repeated failed targets, and operator concentration.

## Brief changelog

- Add audit risk insight calculation for filtered audit summaries and current-page audit records.
- Add an Audit Risk Insights panel to the Audit page.
- Add i18n entries for the new audit risk labels and messages.
- Add unit and component tests for risk classification, empty states, and Audit page integration.

## Verifying this change

- `npm test -- src/pages/ops/__tests__/auditRiskInsights.test.ts src/pages/ops/__tests__/AuditRiskInsights.test.tsx src/pages/ops/__tests__/AuditPage.test.tsx`
- `npx eslint src/pages/ops/auditRiskInsightModel.ts src/pages/ops/AuditRiskInsights.tsx src/pages/ops/audit.tsx src/pages/ops/__tests__/auditRiskInsights.test.ts src/pages/ops/__tests__/AuditRiskInsights.test.tsx src/pages/ops/__tests__/AuditPage.test.tsx src/i18n/translations.ts`
- `npm run build`
- `git diff --check apache/rocketmq-studio...HEAD`

PR #4237, opened 2026-09-11
[Studio] feat: add message query history insights

yx9o:

Add a query history insights panel to the Message Query History drawer. The panel summarizes existing message and trace query history data into a health score, risk signals, notable rows, and recommendations so operators can quickly identify broad topic searches, zero-result queries, large result sets, trace collection gaps, fragmented trace topics, stale history, and missing operator context.

## Brief changelog

- Added message query history insight aggregation logic with unit tests.
- Added a MessageHistoryInsightsPanel to the query history drawer.
- Added localized insight labels and recommendations.
- Reused the existing UTC time formatter for queriedAt display.

## Verifying this change

- `npx prettier --write src/utils/messageHistoryInsights.ts src/utils/messageHistoryInsights.test.ts src/components/MessageHistoryInsightsPanel.tsx src/components/__tests__/MessageHistoryInsightsPanel.test.tsx src/components/MessageQueryHistoryDrawer.tsx src/i18n/translations.ts`
- `npm test -- src/utils/messageHistoryInsights.test.ts src/components/__tests__/MessageHistoryInsightsPanel.test.tsx src/components/__tests__/MessageQueryHistoryDrawer.test.tsx`
- `npx eslint src/utils/messageHistoryInsights.ts src/utils/messageHistoryInsights.test.ts src/components/MessageHistoryInsightsPanel.tsx src/components/__tests__/MessageHistoryInsightsPanel.test.tsx src/components/MessageQueryHistoryDrawer.tsx src/components/__tests__/MessageQueryHistoryDrawer.test.tsx src/i18n/translations.ts`
- `npm run build`
- `git diff --check`

### DIFF
--- a/web/src/components/MessageHistoryInsightsPanel.tsx
+++ b/web/src/components/MessageHistoryInsightsPanel.tsx
@@ -1,0 +1,324 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  Alert,
+  Card,
+  Col,
+  Empty,
+  Progress,
+  Row,
+  Space,
+  Statistic,
+  Table,
+  Tag,
+  Typography,
+} from 'antd';
+import type { ColumnsType } from 'antd/es/table';
+import type {
+  MessageQueryHistory,
+  QueryHistorySummary,
+  TraceQueryHistory,
+} from '../api/messageHistory';
+import { useLang } from '../i18n/LangContext';
+import { formatUtcDateTime } from '../utils/format';
+import {
+  buildMessageHistoryInsights,
+  type MessageHistoryInsightCode,
+  type MessageHistoryInsightIssue,
+  type MessageHistoryInsightLevel,
+  type MessageHistoryNotableRow,
+  type MessageHistoryRecommendationCode,
+} from '../utils/messageHistoryInsights';
+import { tableScrollX } from '../utils/table';
+
+const { Text } = Typography;
+
+interface Props {
+  summary?: QueryHistorySummary;
+  messageRows: MessageQueryHistory[];
+  traceRows: TraceQueryHistory[];
+  loading?: boolean;
+  now?: number;
+}
+
+const levelMeta: Record<
+  MessageHistoryInsightLevel,
+  {
+    color: string;
+    progress: 'success' | 'normal' | 'exception';
+    alertType: 'success' | 'info' | 'warning' | 'error';
+  }
+> = {
+  healthy: { color: 'success', progress: 'success', alertType: 'success' },
+  notice: { color: 'processing', progress: 'normal', alertType: 'info' },
+  warning: { color: 'warning', progress: 'exception', alertType: 'warning' },
+  critical: { color: 'error', progress: 'exception', alertType: 'error' },
+};
+
+const issueTextKey = (code: MessageHistoryInsightCode): string => {
+  switch (code) {
+    case 'NO_HISTORY':
+      return 'messageHistoryInsights.issue.noHistory';
+    case 'STALE_HISTORY':
+      return 'messageHistoryInsights.issue.staleHistory';
+    case 'TRACE_UNDERUSED':
+      return 'messageHistoryInsights.issue.traceUnderused';
+    case 'ZERO_RESULT_QUERIES':
+      return 'messageHistoryInsights.issue.zeroResultQueries';
+    case 'BROAD_TOPIC_QUERIES':
+      return 'messageHistoryInsights.issue.broadTopicQueries';
+    case 'LARGE_RESULT_QUERIES':
+      return 'messageHistoryInsights.issue.largeResultQueries';
+    case 'TRACE_WITHOUT_NODES':
+      return 'messageHistoryInsights.issue.traceWithoutNodes';
+    case 'TRACE_WITHOUT_CONSUMERS':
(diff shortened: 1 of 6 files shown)


RockteMQ-AI:
CI only runs `npm run build` for the frontend (no `npm test`), which explains why this slipped through. I have everything needed for the review.

---

# Code Review: PR #4237 — Message Query History Insights

## 1. Summary

Well-built read-only insights layer: a pure, deterministic aggregation function with solid unit tests, wired into the existing history drawer with a minimal 7-line integration. However, it **deterministically breaks an existing test** (`MessagePageAsyncState.test.tsx`), and because the drawer only loads the *active tab's* current page, the insights are computed from partial, tab-dependent data — which undermines several of the advertised signals.

## 2. Issues Found

### 🔴 Critical (must fix)

- **Broken existing test — `web/src/components/MessageHistoryInsightsPanel.tsx:184` duplicates visible text.**
  The notable-rows "Query Detail" column renders the trace `msgId` as plain text, while the drawer's trace table below it renders the same `msgId` (`MessageQueryHistoryDrawer.tsx:123`). The existing test `web/src/pages/instance/__tests__/MessagePageAsyncState.test.tsx:584` does `await screen.findByText('history-message')` and now fails with *"Found multiple elements with the text: history-message"*. Verified: passes at base `HEAD~1`, fails with this commit. The PR's verification list ran `MessageQueryHistoryDrawer.test.tsx` but not this file. Note CI only runs `npm run build` for web — nothing will catch this automatically, so the test must be updated here (e.g. scope with `within()` on the history table or use `findAllByText`) or the panel's detail cell disambiguated.

### 🟡 Suggestion (should consider)

- **Insights are computed from partial, tab-dependent data — `web/src/components/MessageQueryHistoryDrawer.tsx:64-85`.** The drawer fetches only the active tab (20 rows/page) and resets both row arrays on every load. Consequences:
  - `traceRows` is `[]` whenever the messages tab is active, so `FRAGMENTED_TRACE_TOPICS` and trace-gap signals can **never fire** from the messages tab — and vice versa for message signals. The health score/level flips when the user switches tabs.
  - Counts are phrased as global facts ("Zero-result message searches: {count}") but reflect only the visible page, additionally filtered by the search box. With hundreds of history rows, the panel systematically under-reports and can show "Healthy" while risky rows sit on other pages.
  - Suggest fetching a dedicated sample for the panel (the API allows `pageSize` up to 100, `QueryHistoryController.java:26,67`), or at minimum labeling the issue tags as page-scoped rather than global.

- **`queriedAt` shows different values in the same drawer.** The panel correctly uses `formatUtcDateTime` (`MessageHistoryInsightsPanel.tsx:226`) for the backend's offset-less UTC `LocalDateTime` (`QueryHistoryService.java:64`, `Clock.systemUTC()`), but the drawer's own tables use `formatTime` = `new Date(value).toLocaleString()` (`MessageQueryHistoryDrawer.tsx:30-34`), which parses offset-less strings as local time. The same row will display two different timestamps a few hundred pixels apart. Pre-existing bug, but this PR makes it visible; consider aligning both to `formatUtcDateTime`.

- **`tag` treated as a narrowing filter under-detects broad scans — `web/src/utils/messageHistoryInsights.ts:200`.** The provider applies tag matching *in-memory after* the server-side scan (`RocketMQMessageProvider.java` `matchesTag` post-filter; topic scans capped at `TOPIC_QUERY_HARD_CAP = 2000`). A TOPIC query with only a tag over a ≥24h window is still a full broker scan, so `BROAD_TOPIC_QUERIES` misses exactly the expensive case it exists to flag.

### 🟢 Nit (minor)

- `web/src/utils/messageHistoryInsights.ts:428-430` — the `operator: '-'` detail is never rendered (no translation uses an `{operator}` placeholder); dead data.
- `web/src/utils/messageHistoryInsights.ts:174-177` — the `NO_HISTORY`/`STALE_HISTORY`/`UNKNOWN_OPERATORS` case branch duplicates `default`; the switch could collapse.
- English copy for `FRAGMENTED_TRACE_TOPICS` ("Trace topics in use: {count}") reads informational rather than as a risk signal; the zh version ("检测到 {count} 个不同轨迹 Topic") is clearer.

## 3. Positive Aspects

- **Pure, testable core**: `buildMessageHistoryInsights` is a pure function with an injectable `now`, making all time-based logic (stale history, latest age) deterministically testable — rare discipline.
- **Edge-case hardening**: `normalizeCount`/`parseTimestamp` handle `NaN`, negatives, and invalid dates, and there are explicit tests for them (`messageHistoryInsights.test.ts:202-224`).
- **Thresholds grounded in real server behavior**: `LARGE_RESULT_CRITICAL = 2000` matches `TOPIC_QUERY_HARD_CAP`; broad-window detection composes with the server's 7-day topic-window cap (`MessageService.java:221`).
- **Complete i18n**: all new keys exist in both zh and en, including `.short` variants and fallbacks.
- **Good UX hygiene**: panel hides itself when nothing is loaded, caps issues at 5 and notable rows at 8, sorts by severity, and the drawer integration is minimal.
- No new API surface or security-relevant changes; purely client-side aggregation of existing read endpoints.

## 4. Verdict

**Request Changes.** The broken `MessagePageAsyncState` test is a hard blocker (must update that test or disambiguate the rendering). The tab-dependent sampling is the second substantive item — as merged, several advertised signals (fragmented trace topics, trace gaps) only function while the traces tab happens to be active, which I'd want addressed or explicitly scoped before this ships. The rest is polish on an otherwise clean, well-tested implementation.

---
<sub>🤖 Review by **github-manager-bot** | PR #4237 | 2026-09-11T14:05:11Z</sub>

RockteMQ-AI (approved):
LGTM. Well-structured feature addition. The separation between pure analysis logic (`messageHistoryInsights.ts`) and the presentation component (`MessageHistoryInsightsPanel.tsx`) is clean. The insight scoring system with weighted penalties and the recommendation engine mapping issues to actionable suggestions is a nice touch for operators.

A few minor observations (non-blocking):

- The `buildMessageHistoryInsights` function iterates over rows multiple times (notable extraction, then stats aggregation). Since these are already-loaded history rows this is fine, but if the row count grows significantly, consider a single-pass approach.
- The `roundedPercent` helper is used for both ratio display and age calculation — the name is slightly misleading for the age case. Consider renaming or adding a comment at the call site.

Tests are comprehensive and cover the key edge cases. i18n coverage for both zh/en is complete.

---

lizhimins (changes requested):
This computes a 0-100 health score and its percentages from one server-side page of at most 20 rows, restricted to the currently selected tab, with no scope label and no minimum-sample floor. Changing the page size changes the score.

That is a regression against your own merged #3131, which labels the ratio as current-page (`translations.ts:1261`), gates on `MIN_RECORDS_FOR_RATE = 5`, and deliberately does not emit an overall score. Please follow the same convention here: label the scope, add the floor, and drop the total score unless it is computed from the full set.

PR #4240, opened 2026-09-11
fix(clients): reset column filters when the nameserver changes

### LINKED ISSUE #4238: [Studio][Bug] Switching NameServers on the Clients page keeps the previous endpoint column filters
## Problem

On the Client Connections page, switching the NameServer endpoint keeps the previous endpoint's table column filters active. The columns use uncontrolled antd filters (`filters` + `onFilter` with no `filteredValue`), and `handleNameserverChange` resets `connections`, `clusterFilter`, and the selected row — but neither the `columnFilters` export mirror nor antd's internal filter state.

## Evidence

- `web/src/pages/cluster/clients.tsx` `handleNameserverChange` resets `currentPage`, `selectedEndpoint`, `connections`, `clusterFilter`, `selectedConnection`, `loadError`, `loading` — but not `columnFilters` (state declared as `useState<ClientTableFilters>({})`).
- The clusterName / type / protocol / language columns declare `filters` + `onFilter` without a controlled `filteredValue`, so antd keeps its internal filter selection across `dataSource` changes.
- `columnFilters` is only written by the table `onChange` handler and consumed by `exportConnections`, which drives both the export CSV and the Export button's disabled state.

Reproduction: on NameServer A, filter the type column to `Consumer` (or pick a cluster name), then switch the NameServer select to endpoint B whose rows have different cluster names/types. After the new load finishes, the table shows no rows and the stats header reports 0 connections, while the new endpoint's data was fetched; the Export button stays disabled.

## Impact

An operator switching between NameServer endpoints silently sees an empty client list and cannot export, because filters scoped to the previous endpoint are still applied to the new rows. The state must be manually discovered and cleared to see the real data.

## Expected behavior

Endpoint-scoped column filters are cleared when the NameServer changes: the new endpoint's rows render immediately and the export matches the unfiltered (search + cluster-select) view.

## Related work

- #3291 (closed) covered the backend ignoring cluster/type filters — a different defect on the same page.
- The registry/preview/diff guards on the cluster pages (#3151, merged #3154) follow the same "invalidate per-scope state when the scope changes" principle.

## PR

Fix: #4240.

unbridled-41:
Fixes #4238.

## Problem / Evidence

On the Client Connections page, switching the NameServer endpoint keeps the previous endpoint's table column filters active. `handleNameserverChange` (`web/src/pages/cluster/clients.tsx`) resets `currentPage`, `selectedEndpoint`, `connections`, `clusterFilter`, and the selected row — but not `columnFilters`, and the clusterName/type/protocol/language columns declare uncontrolled `filters` + `onFilter` with no `filteredValue`, so antd's internal filter selection survives the `dataSource` swap.

Concrete trigger: on NameServer A filter the type column to `Consumer`, then switch the NameServer select to an endpoint whose rows are all `Producer` with different cluster names. After the load finishes the table renders no rows and the stats header reports 0 connections even though the endpoint's data was fetched, and the Export button stays disabled (`exportConnections.length === 0`). Reproduced deterministically by the new regression test, which failed before the fix: `billing-svc-0@10.0.3.10:49155` was never rendered after the switch.

## Root cause / Fix

Two halves, both required:

1. `handleNameserverChange` now calls `setColumnFilters({})` — the export mirror and the controlled column values reset with the endpoint.
2. The four filterable columns now pass `filteredValue: columnFilters.<key> ?? null`, making antd's filter state fully controlled. Resetting only our state without `filteredValue` would leave antd's internal filter state (and its trigger highlight) untouched, which is exactly the stale half of the bug.

## Priority & scoring

PRIORITY 78 = impact 30 (silently hides real data and permanently disables export on a core inventory page) + blast radius 14 (all four filterable columns and both the list and export paths) + reproducibility 18 (deterministic UI path, no timing) + maintenance value 16 (endpoint-scoped state invalidation, the same principle the cluster pages' request guards follow). FIX_CONFIDENCE 88: the state mismatch is exact and both fix halves were verified against antd's controlled-filter semantics by the regression test.

## Tests

- New regression: `clears column filters when the nameserver changes` (`web/src/pages/cluster/__tests__/ClientsPage.test.tsx`) — applies a Consumer column filter, switches NameServer, asserts the new endpoint's Producer row renders and the exported CSV contains it. **Red before the fix** (`findByText('billing-svc-0@10.0.3.10:49155')` timed out), **green after**.
- `npx vitest run src/pages/cluster/__tests__/ClientsPage.test.tsx` → **20/20 passed** (19 pre-existing + 1 new).
- Full web suite `npx vitest run` on this branch: **982 tests, 4 failures** in MetricsExplorer (1), ClusterPage (1), ConsumerPage (2) — none of these files is touched by this PR. All three files re-run in isolation: **81/81 passed**, confirming the full-suite failures are the pre-existing parallel-load flakiness recorded in the repo's test baselines, not regressions.
- `npx tsc --noEmit` clean; `npx eslint` on both changed files clean; `npm run build` succeeds.

## Risk

Low. The only behavior change is that column filters no longer outlive the endpoint that produced them; within a single endpoint, filtering, export mirroring, and pagination behave exactly as before (the controlled `filteredValue` is initialized from the same state the `onChange` handler already wrote). antd receives `null` instead of `undefined` when no filter is selected, which is its documented "no filter applied" value.


unbridled-41:
## Verification evidence

**CI**: This PR's head SHA `64df3418c9a43c38c4d7190cc95d63140a43b831` has exactly one workflow run — "CI" (pull_request), run id 34625707099, conclusion `startup_failure`; check-runs API returns 0. The upstream workflow has been failing at startup for all PRs in this repository (including every recently merged one), so no CI signal is available; all verification below was run locally.

**Red-light actual output** (before the fix, `npx vitest run src/pages/cluster/__tests__/ClientsPage.test.tsx -t "clears column filters when the nameserver changes"`):

```
❯ src/pages/cluster/__tests__/ClientsPage.test.tsx:521:25
    expect(await screen.findByText('billing-svc-0@10.0.3.10:49155')).t…
       |                         ^
Test Files  1 failed (1)
Tests  1 failed | 19 skipped (20)
```

The row for the new endpoint's Producer connection never rendered because the stale Consumer column filter from the previous endpoint still applied.

**Module tests**: after the fix, `npx vitest run src/pages/cluster/__tests__/ClientsPage.test.tsx` → `Tests 20 passed (20)` (19 pre-existing + 1 new), including the pre-existing `applies table column filters to the exported CSV` and both stale-response guard tests, which pin the behavior this change must not break.

**Full suite + baseline**: `npx vitest run` on this branch → `Test Files 3 failed | 117 passed (120)`, `Tests 4 failed | 978 passed (982)`. The 4 failures are MetricsExplorer (1), ClusterPage (1), ConsumerPage (2) — none of these files is touched by this PR (numstat below). Re-running the three failed files in isolation: `Tests 81 passed (81)`, confirming they are the parallel-load flakiness recorded for these exact files in previous rounds' baselines, not regressions introduced here.

**Build**: `npm run build` → exit 0 (`✓ built in 10.59s`); `npx tsc --noEmit` clean; `npx eslint` on both changed files clean. (One intermediate build failure during development was in the new test's own mock typing and was fixed before commit; the committed tree builds.)

**Diff self-check**: `git show --numstat HEAD` → `57 0 web/src/pages/cluster/__tests__/ClientsPage.test.tsx` and `7 0 web/src/pages/cluster/clients.tsx`; 2 files, +64/−0, no unrelated changes.

RockteMQ-AI (approved):
### Summary
This PR correctly fixes a UX bug where column filters (e.g. filtering by Consumer type) persist after switching NameServer, causing rows to be silently hidden. The fix resets `columnFilters` to `{}` in the NameServer `onChange` handler and wires `filteredValue` into each column so the table operates in controlled-filter mode.

The approach is sound — making filters controlled via `filteredValue` is the idiomatic antd pattern for programmatic filter resets. The test covers the full interaction flow (apply filter → switch nameserver → verify filter cleared) and is well-structured.

LGTM.

---

PR #4245, opened 2026-09-11
fix(consumer): reload the paginated group list after creating groups

### LINKED ISSUE #4244: [Studio][Bug] Consumer group inventory goes stale after create and CSV import (no server-page refresh)
## Problem

Creating a consumer group (single create or CSV import) updates the group list only locally: the created group is prepended to the currently loaded rows without refetching the server-paginated list and without updating the pagination total. The list the operator sees afterwards diverges from the server's paginated inventory until a manual refresh.

## Evidence

- `web/src/pages/instance/consumer.tsx` create flow (`onOk` of the create confirmation): after `createConsumerGroup(...)` succeeds it does `setGroups((prev) => [created, ...prev.filter(...)])`.
- Same file, CSV import flow: `if (createdGroups.length > 0) { setGroups((previous) => [...createdGroups, ...previous.filter(...)]); }`.
- The list is server-paginated: `loadConsumerGroupPage(page, pageSize)` fetches `listConsumerGroupPage({ page, pageSize })` and sets both `groups` and `totalGroups`. The local prepend updates neither the total nor the server ordering, and `autoRefresh` defaults to `false`, so nothing self-heals.
- The same file refetches the page after single and batch delete (`reloadConsumerGroupPageAfterDelete`) — the create/import paths are the only list mutations that skip the reload.

Consequences with the default settings: the pagination footer keeps showing the pre-create total; on a full page the prepended row pushes the page's last row out of view; and if the user is on page > 1 the new group is injected into that page even though the server's ordering places it elsewhere. The identical defect was already fixed for the topic inventory (merged PR #3339, "refresh paginated inventory after create") and for the ACL user inventory (#3306).

## Impact

After a successful create, the consumer group inventory shows rows and a total that do not match the server's paged inventory. An operator who creates a group and then looks for it — or trusts the total — gets wrong answers until they manually refresh.

## Expected behavior

After a successful create or import, the current page is refetched from the server (mirroring the delete paths and the topic page's post-create behavior), so rows and the pagination total reflect the server inventory.

## Related work

- #3337 (closed) tracked the same defect class for the topic page and was fixed by merged PR #3339.
- #3306 tracks the same class for the ACL user inventory.

## PR

Fix: #4245.

unbridled-41:
Fixes #4244.

## Problem / Evidence

Creating a consumer group — through the single create dialog or the CSV import — updates the group list only locally: `setGroups((prev) => [created, ...prev.filter(...)])` prepends the created group to the currently loaded rows without refetching the server-paginated list and without updating the pagination total.

The list is server-paginated (`loadConsumerGroupPage(page, pageSize)` fetches `listConsumerGroupPage({page, pageSize})` and sets both rows and `totalGroups`), and `autoRefresh` defaults to `false`, so the divergence persists until a manual refresh:

- the pagination footer keeps showing the pre-create total (`共 N 个 Group` with the stale N);
- on a full page the prepended row pushes the page's last row out of view;
- with the user on page > 1, the new group is injected into that page even though the server's ordering places it elsewhere.

The same file already refetches the page after single and batch delete — the create/import paths were the only list mutations that skipped the reload. This is the same defect class maintainers already fixed for the topic inventory (merged #3339) and that #3306 tracks for ACL users.

## Root cause / Fix

The create and import paths mutated the local array instead of reloading the paginated inventory. Fix (mirroring #3339's pattern on the topic page):

1. `reloadConsumerGroupPageAfterDelete` is renamed to the neutral `reloadConsumerGroupPage` (both existing delete call sites updated).
2. The single-create flow replaces its local prepend with `await reloadConsumerGroupPage()`.
3. The CSV import flow replaces its `setGroups` prepend with `await reloadConsumerGroupPage()` when at least one group was created.

## Priority & scoring

PRIORITY 76 = impact 28 (post-write inventory divergence: stale total, evicted rows, misplaced new group on an inventory operators rely on right after creating) + blast radius 12 (both list-mutating flows of the consumer group page) + reproducibility 20 (deterministic: create a group with default settings and read the footer) + maintenance value 16 (adopts the exact reload pattern maintainers merged for topics and this page already uses for deletes). FIX_CONFIDENCE 92: mechanical reuse of the in-file reload callback; the new regression test fails red before and passes green after, and the delete/import test suites pin the unchanged behavior.

## Tests

- New regression `reloads the paginated group list after creating a group` (`web/src/pages/instance/__tests__/ConsumerPage.test.tsx`): creates a group, asserts `listConsumerGroupPage` is re-fetched for the current page, the created group renders where the server returns it, and the footer shows the refreshed total. **Red before the fix** (`expected "vi.fn()" to be called 2 times, but got 1 times` — no reload happened), **green after**.
- `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` → **32/32 passed** (31 pre-existing, including the delete-reload, batch-delete, and CSV-import suites, unchanged).
- Full web suite `npx vitest run` on this branch: **982 passed (982) — zero failures**, including the CSV-import test that exercises the reloaded import path.
- `npx tsc --noEmit` clean; `npx eslint` on both changed files clean; `npm run build` succeeds.

## Risk

Low. The only behavior change is that the list is re-read from the server after a successful create/import instead of being patched locally; delete paths are untouched. The import result modal (per-row success/failure state) is driven by `importRows`, not by the group list, so partial-failure reporting is unaffected.

### DIFF
--- a/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
+++ b/web/src/pages/instance/__tests__/ConsumerPage.test.tsx
@@ -487,6 +487,38 @@ describe('Consumer page', () => {
     confirmSpy.mockRestore();
   });
 
+  it('reloads the paginated group list after creating a group', async () => {
+    const created = { ...group, name: 'cg-created' };
+    vi.mocked(consumerService.listConsumerGroupPage)
+      .mockResolvedValueOnce(groupPage([group]))
+      .mockResolvedValue(groupPage([created, group]));
+    const confirmSpy = vi.spyOn(Modal, 'confirm').mockImplementation((config) => {
+      void config.onOk?.();
+      return { destroy: vi.fn(), update: vi.fn() } as unknown as ReturnType<typeof Modal.confirm>;
+    });
+    const user = userEvent.setup();
+    renderWithProviders(<ConsumerPage />);
+
+    await screen.findByText('remote-cg');
+    expect(consumerService.listConsumerGroupPage).toHaveBeenCalledTimes(1);
+
+    const createButton = screen.getByRole('button', { name: '创建 Group' });
+    await waitFor(() => expect(createButton).toBeEnabled());
+    await user.click(createButton);
+    const dialog = await screen.findByRole('dialog');
+    await user.type(within(dialog).getByLabelText('Group 名称'), 'cg-created');
+    await user.click(within(dialog).getByRole('button', { name: /创\s*建/ }));
+
+    await waitFor(() => expect(consumerService.createConsumerGroup).toHaveBeenCalledTimes(1));
+    await waitFor(() => expect(consumerService.listConsumerGroupPage).toHaveBeenCalledTimes(2));
+    expect(consumerService.listConsumerGroupPage).toHaveBeenLastCalledWith(
+      expect.objectContaining({ page: 1, pageSize: 20 }),
+    );
+    expect(await screen.findByText('cg-created')).toBeInTheDocument();
+    expect(screen.getByText('共 2 个 Group')).toBeInTheDocument();
+    confirmSpy.mockRestore();
+  });
+
   it('prefills the group search from the ?group= query parameter', async () => {
     renderWithProviders(<ConsumerPage />, '/instance/consumer?group=remote-cg');
 
--- a/web/src/pages/instance/consumer.tsx
+++ b/web/src/pages/instance/consumer.tsx
@@ -370,7 +370,7 @@ const ConsumerPageContent = ({
     [t, selectedInstanceId, search],
   );
 
-  const reloadConsumerGroupPageAfterDelete = useCallback(async () => {
+  const reloadConsumerGroupPage = useCallback(async () => {
     await loadConsumerGroupPage(page, pageSize);
   }, [loadConsumerGroupPage, page, pageSize]);
 
@@ -822,10 +822,7 @@ const ConsumerPageContent = ({
     setImportRows([...nextRows]);
 
     if (createdGroups.length > 0) {
-      setGroups((previous) => {
-        const createdNames = new Set(createdGroups.map((group) => group.name));
-        return [...createdGroups, ...previous.filter((group) => !createdNames.has(group.name))];
-      });
+      await reloadConsumerGroupPage();
     }
 
     const failedCount = nextRows.filter((row) => row.status === 'failed').length;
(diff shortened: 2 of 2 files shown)


unbridled-41:
## Verification evidence

**CI**: This PR's head SHA `6660aa534633928af8e296a29e6cfdf2d68ab433` has exactly one workflow run — "CI" (pull_request), run id 34628315358, conclusion `startup_failure`; check-runs API returns 0. The upstream workflow has been failing at startup for all PRs in this repository (including every recently merged one), so no CI signal is available; all verification below was run locally.

**Red-light actual output** (before the fix, `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx -t "reloads the paginated group list after creating a group"`):

```
AssertionError: expected "vi.fn()" to be called 2 times, but got 1 times
Test Files  1 failed (1)
Tests  1 failed | 31 skipped (32)
```

The paginated list was never re-fetched after the create succeeded.

**Module tests** (after the fix): `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` → `Tests 32 passed (32)` (31 pre-existing + 1 new). The pre-existing suites include single delete, batch delete, and CSV import with partial failures — all still green, pinning the delete-reload and import-row behavior this change must not break.

**Full suite + baseline**: `npx vitest run` on this branch → `Test Files 120 passed (120)` (module-run fragment `Test Files 1 failed (1)` in the red-light item above refers to the single test file); full numbers: `Tests 982 passed (982)`, **zero failures**, including the CSV-import test that exercises the reloaded import path.

**Build**: `npm run build` → exit 0; `npx tsc --noEmit` clean; `npx eslint` on both changed files clean.

**Diff self-check**: `git show --numstat HEAD` → `32 0 web/src/pages/instance/__tests__/ConsumerPage.test.tsx` and `9 12 web/src/pages/instance/consumer.tsx`; 2 files, +41/−12, no unrelated changes.

RockteMQ-AI (approved):
### Summary
This PR fixes a pagination inconsistency when creating consumer groups. The old code prepended the created group to the local state without refetching the server-paginated list, which could cause the new group to appear in the wrong position and the pagination total to be stale.

The fix properly refetches the current page from the server after creation, mirroring the existing pattern used for topic inventory. The function is also renamed from `reloadConsumerGroupPageAfterDelete` to `reloadConsumerGroupPage` to reflect its now-generic purpose.

A comprehensive test case is added that verifies the list reload behavior after group creation.

**Verdict: LGTM** — clean, minimal fix with proper test coverage.

---

lizhimins:
Merged as part of #4331 — squashed into `rocketmq-studio` as `6334f9238`.

Your change here (reloading the paginated group list after creating groups) is included.

We grouped 11 of your related front-end fixes into a single commit so the history stays reviewable; the other PRs in the group are closed with the same note. Everything was cherry-picked onto the current baseline with you kept as the author, and your tests came along with the change.

PR #4255, opened 2026-09-12
fix(settings): build general settings saves from a fresh read

### LINKED ISSUE #4250: [Studio][Bug] Saving any general setting rolls back AI/LLM configuration saved from the AI tab
## Problem

`web/src/pages/settings/GeneralSettingsTab.tsx` fetches the general settings once on mount and keeps that snapshot in state. Every save path on the tab builds the request payload from that snapshot (`buildPayload`, lines 44-59) and posts it to `POST /api/settings/general/save`, including the `llmProvider`, `model` and `baseUrl` fields the tab never renders or edits.

The AI assistant tab (`AiAssistantTab`) saves the same persisted record through `POST /llm/config` (`LlmConfigService.saveConfig` writes `llmProvider`/`model`/`baseUrl` into the same `GeneralSettingsVO` row). Because the antd Tabs keep the visited panes mounted (`web/src/pages/settings/index.tsx` has no `destroyOnHidden`), the general tab never refetches after that save.

## Evidence

1. Open `/settings` (general tab mounts, snapshot: `llmProvider=openai`).
2. Switch to the AI tab, change provider/model/base URL, save — `POST /llm/config` persists the new values.
3. Switch back to the general tab (still mounted, stale snapshot), change the session timeout, save — `POST /api/settings/general/save` carries the stale `llmProvider`/`model`/`baseUrl`.
4. `SettingsService.saveGeneralSettings` replaces the row; its empty-value backfill covers `llmEngine`/`deploymentName`/`apiVersion`/`awsRegion`/`maxTokens`/`temperature`/`apiKey`/`dingtalkSigningSecret` but **not** `llmProvider`/`model`/`baseUrl` (`SettingsService.java:128-166`, `GeneralSettingsUpdateDTO.java:49-81` maps them), so the stale values overwrite the row.

The same stale payload is also sent by the theme/compact switches and by every "send test notification" button (`sendTest` calls `mergeAndSave` first).

## Impact

Any general-settings save silently reverts the AI provider/model/base URL to the values seen at page load. The AI chat then calls the old provider (with a key that may not match) or the wrong billing model, with no error shown. Session timeout, requireLogin and notification fields are equally exposed to the same rollback if changed elsewhere.

A regression test demonstrating the stale payload on the unfixed code lives in the linked PR.

## Expected behavior

A general-settings save must send the currently persisted values for the fields it does not manage, so saving general settings never reverts an AI-tab save.

## Related work

- #2444 (closed stale, no fix) reported the same full-snapshot overwrite family for concurrent general-settings forms.
- #2863 (closed, unmerged) proposed the server-side backfill direction; the equivalent empty-value backfill is present in the current code via merged #2963 (`SettingsService.saveGeneralSettings` backfills `llmEngine`/`deploymentName`/`apiVersion`/`awsRegion`/`maxTokens`/`temperature`/`apiKey`/`dingtalkSigningSecret`, but not `llmProvider`/`model`/`baseUrl`). A stale client snapshot sends non-null old values, which that backfill cannot address.
- #813 (closed not_planned) covered the reduced-payload facet of the same save path (omitted `llmEngine`/`maxTokens`/`temperature` and the stale in-memory LLM override); this defect sends explicit stale non-null values, which that mechanism does not reach.

## PR

Fix: #4255.

unbridled-41:
Fixes #4250.

## Problem / Evidence

`GeneralSettingsTab` fetches the general settings once on mount and builds every save payload from that snapshot, including the `llmProvider`/`model`/`baseUrl` fields the tab never renders or edits. The AI assistant tab writes the same persisted record through `/llm/config` while the general tab stays mounted (antd Tabs keep visited panes alive), so every general save posts the mount-time LLM values back and silently reverts the AI configuration.

Repro: open `/settings` (snapshot: `llmProvider=openai`) → AI tab, switch provider/model/base URL, save → back on the general tab (still mounted), change the session timeout, save → `POST /api/settings/general/save` carries the stale `llmProvider`/`model`/`baseUrl` and overwrites the row.

## Root cause / Fix

The tab must not send fields it does not manage from its own snapshot. `persistPreference` and `mergeAndSave` now re-read the persisted settings (`loadFreshSettings`) before building each payload, so unmanaged fields are sent as currently stored; on a fresh-read failure the previous snapshot is used as fallback (baseline behavior). After a preference save succeeds, local state is synced to the fresh base plus the saved patch so the theme/compact controls keep showing the saved value. `loadFreshSettings` takes the current snapshot as its fallback parameter, which also keeps the `Promise<GeneralSettings>` contract type-clean.

## Priority & scoring

PRIORITY 74 = impact 30 (any general save silently reverts the AI provider/model/base URL — persisted data loss; the AI chat then calls the wrong provider or billing model with a possibly mismatched key, no error shown) + scope 14 (every save path on the tab: security form, notification form, theme, compact, three notification-test buttons) + reproducibility 18 (deterministic 4-step repro, covered by a regression test) + maintenance value 12 (fresh-read convention; #2863's server-side null-backfill cannot address stale non-null client values, as recorded in #4250). FIX_CONFIDENCE 88: mechanism verified red→green, fallback preserves baseline behavior on read failure.

## Tests

- New regression `saves llm fields from a fresh read instead of the mount-time snapshot` mocks a mount-time `openai` snapshot and a fresh-read `deepseek` record, submits the security form and asserts the save payload carries the fresh `deepseek` values.
- Red on the unfixed code: `src/pages/settings/__tests__/GeneralSettingsTab.test.tsx` → `Tests  1 failed | 6 passed (7)` with `× saves llm fields from a fresh read instead of the mount-time snapshot` (`expected "vi.fn()" to be called with arguments: [ObjectContaining{…}]` — payload carried the stale `openai` values).
- The existing theme test additionally asserts the 深色 option stays selected after the save completes; this assertion fails when the post-save state sync is omitted (`Expected the element to have class: ant-segmented-item-selected`), so the visible-preference behavior is pinned.
- Green with the fix: `Tests  7 passed (7)` (re-run after the final commit, including lint-staged reformatting).
- Full web suite `npx vitest run`: 982 tests, 2 failed — `ClusterPage.test.tsx` and `ConsumerPage.test.tsx`, files this branch does not touch and known load-fragile cases under the parallel run; re-run in isolation: ClusterPage `Tests  25 passed (25)`, ConsumerPage `Tests  31 passed (31)`.
- `npx tsc -b` clean; `npm run build` (tsc -b + vite build) succeeds; `npx eslint .` 0 errors (10 warnings, all in files this branch does not touch).

## Risk

One extra GET before each save (settings-page frequency makes this negligible); on a fresh-read failure the save still proceeds from the previous snapshot, matching today's behavior. State after a successful preference save now reflects server truth plus the patch. No API or type changes beyond this component and its test.

Head: ea4ea0a1 (fix/settings-general-save-stale-llm-rollback, 1 commit).

### DIFF
--- a/web/src/pages/settings/GeneralSettingsTab.tsx
+++ b/web/src/pages/settings/GeneralSettingsTab.tsx
@@ -99,13 +99,27 @@ export const GeneralSettingsTab = () => {
     };
   }, [message, notifyForm, securityForm, t]);
 
+  // Other tabs (AI assistant settings) write the same settings record while this tab stays
+  // mounted, so every save must be built from a fresh read instead of the mount-time snapshot.
+  const loadFreshSettings = async (fallback: GeneralSettings): Promise<GeneralSettings> => {
+    try {
+      const fresh = await getGeneralSettings();
+      setSettings(fresh);
+      return fresh;
+    } catch {
+      return fallback;
+    }
+  };
+
   const persistPreference = async (patch: Partial<GeneralSettings>) => {
     if (!settings) return;
     const next = { ...settings, ...patch };
     setSettings(next);
     setSavingPreference(true);
     try {
-      await saveGeneralSettings(buildPayload(next));
+      const base = await loadFreshSettings(settings);
+      await saveGeneralSettings(buildPayload({ ...base, ...patch }));
+      setSettings({ ...base, ...patch });
       message.success(t('settings.saveSuccess'));
     } catch {
       message.error(t('settings.saveFailed'));
@@ -117,10 +131,11 @@ export const GeneralSettingsTab = () => {
   const mergeAndSave = async (patch: Partial<GeneralSettings>) => {
     if (!settings) return false;
     try {
-      await saveGeneralSettings(buildPayload({ ...settings, ...patch }));
+      const base = await loadFreshSettings(settings);
+      await saveGeneralSettings(buildPayload({ ...base, ...patch }));
       const statePatch = { ...patch };
       delete statePatch.clearDingtalkSigningSecret;
-      setSettings({ ...settings, ...statePatch });
+      setSettings({ ...base, ...statePatch });
       return true;
     } catch {
       message.error(t('settings.saveFailed'));
--- a/web/src/pages/settings/__tests__/GeneralSettingsTab.test.tsx
+++ b/web/src/pages/settings/__tests__/GeneralSettingsTab.test.tsx
@@ -149,6 +149,55 @@ describe('GeneralSettingsTab', () => {
       ),
     );
     expect(localStorage.getItem('rocketmq-studio-theme')).toBe('dark');
+    // The fresh read that guards unmanaged fields must not revert the visible preference.
+    await waitFor(() =>
+      expect(screen.getByText('深色').closest('.ant-segmented-item')).toHaveClass(
+        'ant-segmented-item-selected',
+      ),
+    );
+  });
+
+  it('saves llm fields from a fresh read instead of the mount-time snapshot', async () => {
+    const baseSettings = {
+      theme: 'system',
+      compact: false,
+      desktopNotify: false,
+      notifySound: false,
+      sessionTimeout: 30,
+      requireLogin: true,
+      apiKeyConfigured: false,
+    };
+    vi.mocked(getGeneralSettings)
+      .mockResolvedValueOnce({
+        ...baseSettings,
+        llmProvider: 'openai',
+        model: 'gpt-test',
+        baseUrl: 'https://openai.example/v1',
+      })
+      .mockResolvedValue({
+        ...baseSettings,
+        llmProvider: 'deepseek',
+        model: 'deepseek-chat',
+        baseUrl: 'https://deepseek.example/v1',
+      });
+    vi.mocked(saveGeneralSettings).mockResolvedValue();
+    renderTab();
+
+    const saveButtons = await screen.findAllByRole('button', { name: '保存设置' });
+    await waitFor(() => expect(saveButtons[0]).toBeEnabled());
+    // The AI assistant tab saved a different provider while this tab stayed mounted;
+    // submitting the security form must not resurrect the stale provider.
+    fireEvent.submit(saveButtons[0].closest('form')!);
+
+    await waitFor(() => expect(saveGeneralSettings).toHaveBeenCalledTimes(1));
+    expect(saveGeneralSettings).toHaveBeenCalledWith(
+      expect.objectContaining({
+        llmProvider: 'deepseek',
+        model: 'deepseek-chat',
+        baseUrl: 'https://deepseek.example/v1',
+        sessionTimeout: 30,
+      }),
+    );
   });
 
   it('can explicitly clear a configured DingTalk signing secret', async () => {


unbridled-41:
## Verification evidence

- **CI**: branch head `ea4ea0a16d5ec834ff5d657a3494f97e0135b4ce` (PR #4255) has exactly one workflow run for this head: "CI" (pull_request event), run 34709685412, conclusion=startup_failure — the upstream workflow fails at startup for every branch in this repository, producing no check-runs (`check-runs total_count = 0`). Unrelated to this change; all local verification below was run on this head.
- **Red light (unfixed code)**: with the regression test kept and the baseline `GeneralSettingsTab.tsx` restored, `npx vitest run src/pages/settings/__tests__/GeneralSettingsTab.test.tsx` → `Tests  1 failed | 6 passed (7)`, failing item `× saves llm fields from a fresh read instead of the mount-time snapshot` with `AssertionError: expected "vi.fn()" to be called with arguments: [ObjectContaining{…}]` (the payload carried the stale `openai` values). A second guard was verified to have teeth separately: omitting the post-save state sync makes the existing theme test fail with `Expected the element to have class: ant-segmented-item-selected`.
- **Module tests (fixed)**: `npx vitest run src/pages/settings/__tests__/GeneralSettingsTab.test.tsx` → `Tests  7 passed (7)` (re-run after the final commit, post lint-staged formatting).
- **Full suite + baseline**: `npx vitest run` on this head → 120 files, `Tests  2 failed | 980 passed (982)`; both failures are untouched files and known load-fragile cases under the parallel suite — `ClusterPage.test.tsx` and `ConsumerPage.test.tsx`; re-run in isolation → ClusterPage `Tests  25 passed (25)`, ConsumerPage `Tests  31 passed (31)`. Zero failures in files touched by this change.
- **Build**: `npx tsc -b` clean and `npm run build` (tsc -b + vite build) succeeds on the final head. Note: an earlier state of this branch had a `TS2322` in `loadFreshSettings` (returning the narrowed-`null` snapshot under a `Promise<GeneralSettings>` contract); the final commit resolves it by passing the current snapshot as the fallback parameter — `npx eslint .` → 0 errors, 10 warnings all in untouched files.
- **Diff self-check** (`git show --numstat`): `web/src/pages/settings/GeneralSettingsTab.tsx +18/−3`, `web/src/pages/settings/__tests__/GeneralSettingsTab.test.tsx +49/−0` — no unrelated changes.

RockteMQ-AI (approved):
### Summary
Fix a stale-data race condition in General Settings saves. When other tabs (e.g., AI assistant settings) modify the same settings record while this tab is mounted, saving from the mount-time snapshot would silently overwrite those changes. The fix introduces `loadFreshSettings()` to re-fetch the current server state before every save, then merges the user's patch on top.

**Correctness** — The approach is sound. Both `persistPreference` and `mergeAndSave` now read-then-write, eliminating the lost-update problem. The fallback to the local snapshot on fetch failure is a good defensive choice.

**One minor observation:** In `persistPreference`, there are two `setSettings` calls in quick succession — first an optimistic update (`setSettings(next)`), then `loadFreshSettings` internally calls `setSettings(fresh)`, and finally `setSettings({ ...base, ...patch })`. The intermediate `setSettings(fresh)` may cause a brief render with stale-then-fresh data, but this is harmless for a settings page and the final state is always correct.

**Tests** — Good coverage including verification that the fresh read doesn't revert the visible preference being toggled.

LGTM.

---

lizhimins:
Merged as part of #4284 — squashed into `rocketmq-studio` as `796c02e18`.

Your change here (building general settings saves from a fresh read) is included.

We grouped 6 of your related back-end fixes into a single commit so the history stays reviewable; the other PRs in the group are closed with the same note. Everything was cherry-picked onto the current baseline with you kept as the author, and your tests came along with the change.

PR #4217, opened 2026-09-11
test(auth): align AuthCorsIntegrationTest with single session resolution

yyqdbngt:
## Problem

`mvn -B -ntp test` fails on the current `rocketmq-studio` head (`6c24d2ed`): 2 of 2151 tests fail, both in `AuthCorsIntegrationTest`.

```
[ERROR] Tests run: 4, Failures: 2, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthCorsIntegrationTest
[ERROR] Tests run: 2151, Failures: 2, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

```
AuthCorsIntegrationTest.shouldStillRejectAnonymousProtectedRequests
  Wanted but not invoked: AuthService.isAuthenticated(null)

AuthCorsIntegrationTest.shouldRejectNonAdminMutationBeforeControllerExecution
  java.lang.AssertionError: Status expected:<403> but was:<401>
```

Run on Ubuntu 22.04 with OpenJDK 21.0.12 and Maven 3.6.3, from `server/`, with no local changes.

## Root cause

`#2895` (`ddf82325`, [ISSUE #2892]) made `AuthInterceptor` resolve the session once per request. The interceptor now calls `authService.getAuthenticatedUser(...)` only — `isAuthenticated` is no longer part of the flow:

```java
var authenticatedUser = authService.getAuthenticatedUser(authorization).orElse(null);
if (authenticatedUser == null) {
    writeError(response, HttpStatus.UNAUTHORIZED, "Unauthorized");
    return false;
}
```

`AuthCorsIntegrationTest` was last changed on 2026-08-18 and still stubs and verifies that old pair of calls, so the mock answers `Optional.empty()` and the anonymous branch is taken before `requiresAdmin` is ever evaluated. That is why the non-admin mutation answers 401 instead of the expected 403.

The interceptor change is correct; only the test is stale. A/B on the same tree, with nothing but `AuthInterceptor.java` swapped for its pre-`#2895` revision (test untouched):

| `AuthCorsIntegrationTest` | result |
|---|---|
| head (`6c24d2ed`) | `Tests run: 4, Failures: 2` |
| head + pre-`#2895` `AuthInterceptor` | `Tests run: 4, Failures: 0` |

## Change

Test only — production code is untouched:

- `shouldStillRejectAnonymousProtectedRequests` verifies `getAuthenticatedUser(null)` and asserts that `isAuthenticated` is no longer called.
[...]

PR #4275, opened 2026-09-13
fix(topic): allow a blank update to clear the persisted topic remark

### LINKED ISSUE #4272: [Studio][Bug] A topic remark cannot be cleared through POST /api/topics/update
## 1. Symptom

`POST /api/topics/update` (`TopicController.updateTopic` -> `ApacheInstanceProvider.updateTopic` -> `RocketMQAdminClientImpl.updateTopic`) cannot remove a topic's stored remark.

A blank remark is accepted (HTTP 200) and the response echoes the blank value, but the `rmq_instance_topic` row keeps the previous remark, so the next topic list load shows the old text again. Editing a remark to a different non-blank value works; only clearing it is lost.

## 2. Root cause

`server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQAdminClientImpl.java:382-384`

```java
if (StringUtils.hasText(topic.getRemark())) {
    existing.setRemark(topic.getRemark());
}
```

The cached row is only written when the submitted remark has text, so a blank submission is treated exactly like an omitted one. `TopicVO.remark` / `UpdateTopicDTO.remark` are plain nullable strings with no validation and no separate clear flag (unlike `clearApiKey` / `clearDingtalkSigningSecret` in general settings), so a blank remark is the only way a caller can express "this topic has no remark".

Clearing is a two-part problem: assigning `null` to the entity field would not persist either, because MyBatis-Plus `updateById` omits null entity fields (`FieldStrategy.NOT_NULL`). That mechanism is already documented and fixed for the ACL user/rule columns in #3342, and it has to be handled explicitly on this path as well.

## 3. Impact

- The remark column of a managed topic is write-once: it can be created and edited, never cleared. Operators have to recreate the topic or edit MySQL directly.
- The endpoint lies: the 200 response carries the cleared value while the database still holds the old one.
- The update path is the one the create dialog's edit mode is wired to (`POST /api/topics/update` is documented in `docs/api-spec.md` §5.4 and implemented for all three providers, and `web/src/pages/instance/topic.tsx` is being wired to it in #4249), so the UI cannot clear a remark either.

## 4. Reproduction

1. Create a topic with a remark, or update one to have a remark.
2. `POST /api/topics/update` with `{"name": "<topic>", "instanceId": "<instance>", "remark": ""}`.
3. Response: 200, `remark` is empty.
4. Reload the topic list (`GET /api/topics`) or `SELECT remark FROM rmq_instance_topic WHERE name = '<topic>'`: the previous remark is still there.

## 5. Expected behaviour

- An omitted remark (`null`) keeps the stored value, which is what the existing partial-update comment in `updateTopic` describes.
- A submitted blank remark clears it (the column must become NULL).
- The response reports the persisted remark instead of the submitted one.
- `CreateTopicDTO`/`UpdateTopicDTO` stay unchanged; no behavioural change for callers that submit a real remark.

btlqql:

`POST /api/topics/update` could not remove a topic's stored remark. `RocketMQAdminClientImpl.updateTopic` only wrote the cached row when the submitted value had text:

```java
if (StringUtils.hasText(topic.getRemark())) {
    existing.setRemark(topic.getRemark());
}
```

so a blank submission behaved exactly like an omitted one. `TopicVO.remark` / `UpdateTopicDTO.remark` are plain nullable strings with no validation and there is no separate clear flag (unlike `clearApiKey` / `clearDingtalkSigningSecret` in general settings), so a blank remark is the only way a caller can say "this topic has no remark" — and that value was discarded. The `rmq_instance_topic` row kept the previous remark while the endpoint answered 200, so the topic list showed the old text again on the next load.

Clearing is a two-part change. `updateById` omits null entity fields (`FieldStrategy.NOT_NULL`, no `update-strategy` override in `application.yml`), which is the mechanism already handled for the ACL user/rule columns in #3342, so the cleared remark is assigned explicitly instead of relying on the entity write. An omitted remark still keeps the stored value, which is the partial-update behaviour the existing comment in this method describes.

The response now reports the persisted remark rather than the submitted one, so a clear and an omitted value cannot be reported as a state the database does not hold.

## Brief changelog

- `RocketMQAdminClientImpl.updateTopic`: treat a submitted blank remark as a clear (keep the stored value only when the remark is `null`), and assign `remark` explicitly through an `UpdateWrapper` when it is cleared, because `updateById` skips null entity fields.
- Echo the persisted remark back on the returned `TopicVO`.
- `RocketMQAdminClientImplTest`: two new cases — a blank remark clears the column and returns an empty remark; an omitted remark keeps the stored value and issues no extra assignment.

## Verification

```
mvn -f server/pom.xml -B -DskipITs -Dtest=RocketMQAdminClientImplTest -DfailIfNoTests=false test
[ERROR] Tests run: 48, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 2.480 s <<< FAILURE! -- in org.apache.rocketmq.studio.provider.apache.RocketMQAdminClientImplTest
[ERROR]   RocketMQAdminClientImplTest.updateTopicClearsTheStoredRemarkWhenTheRequestSubmitsABlankRemark:638
[ERROR]   RocketMQAdminClientImplTest.updateTopicKeepsTheStoredRemarkWhenTheRequestOmitsIt:664
[INFO] BUILD FAILURE
```

`...:638` is `verify(topicMapper).update(isNull(), captor.capture())` (no assignment was issued, so the column was never cleared); `...:664` is the returned topic still carrying the submitted `null` instead of the persisted remark.

Green after the change:

```
mvn -f server/pom.xml -B -DskipITs -Dtest=RocketMQAdminClientImplTest -DfailIfNoTests=false test
[INFO] Tests run: 48, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.455 s -- in org.apache.rocketmq.studio.provider.apache.RocketMQAdminClientImplTest
[INFO] BUILD SUCCESS
```

Fixes #4272


RockteMQ-AI (approved):
### Summary
Fix for topic remark not being cleared when a blank value is submitted. This PR also introduces an important semantic distinction: null remark = omitted (keep stored value), blank remark = explicitly cleared.

### Review

**Correctness** ✅
- The `clearRemark` flag correctly distinguishes between omitted (null) and explicitly cleared (blank) remarks.
- When `clearRemark` is true, the entity field is set to null (skipped by `updateById`), then an explicit `UpdateWrapper` SET NULL follows — same pattern as #3342, #4276, #4277.
- The response fix (`topic.setRemark(existing.getRemark())`) ensures the returned VO reflects the persisted state, preventing the UI from displaying stale data.

**Tests** ✅
- Two new tests cover both the clear path (blank remark → null in DB) and the no-op path (omitted remark → stored value retained).
- ArgumentCaptor assertions verify the exact SQL and parameter values.

**Code quality** ✅
- Clear inline comments explaining the null-vs-blank semantics and the MyBatis-Plus workaround.
- The response fix is a thoughtful addition that prevents UI confusion.

LGTM.

PR #4276, opened 2026-09-13
fix(instance): persist a cleared admin credential reference

### LINKED ISSUE #4273: [Studio][Bug] Clearing an instance admin credential reference has no effect
## 1. Symptom

An instance's admin credential reference cannot be removed. In the instance edit dialog (`web/src/pages/instance/index.tsx`, "Admin Credential Ref" field) clearing the value and saving returns HTTP 200 and the edited row looks updated, but `rmq_instance.admin_credential_ref` still holds the previous reference, so it reappears on the next list load.

## 2. Root cause

`server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java:602-604`

```java
if (!cloudInstance && instance.getAdminCredentialRef() != null) {
    updated.setAdminCredentialRef(normalizeCredentialRef(instance.getAdminCredentialRef()));
}
```

`normalizeCredentialRef` (`InstanceService.java:561-563`) maps a blank submission to `null`, so a blank reference is an explicit clear and a non-null `updated` field is written as null. But `updated` (a copy of the stored instance, `:580`) goes out through `instanceRepository.save(updated)` (`:607`), and `MybatisPlusInstanceRepository.save` writes the update with `instanceMapper.updateById(entity)` (`MybatisPlusInstanceRepository.java:122`).

MyBatis-Plus `updateById` omits null entity fields (`FieldStrategy.NOT_NULL`, no `update-strategy` override in `application.yml`), so the null reference never reaches the SET clause: the column keeps its previous value while the service returns a VO whose reference is null. This is the same mechanism that was already fixed for the ACL user/rule columns in #3342, and for a cleared `white_remote_address` before that.

The repository is also the only place that can fix it correctly: the service cannot express "clear this column" through the plain entity-based `save`, and the update path is the repository's own concern (`InstanceRepository.save` is called only from `InstanceService` at `:223` (insert) and `:607` (update)).

## 3. Impact

- A security-relevant piece of instance configuration is write-once: an operator can set or change the external admin credential reference but cannot remove it, even though the API accepts the clear.
- The response contradicts the database: the API returns `adminCredentialRef: null` while MySQL still stores the reference, and the UI replaces its row with that response without refetching (the edit handler uses the returned instance, `web/src/pages/instance/index.tsx:447-455`), so the wrong value is displayed until the next manual reload.
- Any client of `POST /api/instances/update` (or the equivalent instance-update endpoint) hits the same silent no-op.

## 4. Reproduction

1. Create an Apache-type instance with `adminCredentialRef = "production-admin"`.
2. Update it with `adminCredentialRef = ""` (the edit dialog sends the cleared input verbatim).
3. Effect: HTTP 200, response `adminCredentialRef` is null.
4. `SELECT admin_credential_ref FROM rmq_instance WHERE id = <id>` still returns `production-admin`.

## 5. Expected behaviour

- A blank reference clears the column (NULL), matching `normalizeCredentialRef`'s intent.
- An omitted reference (`null` in the request) keeps the stored value.
- The returned instance reflects the persisted state.
- Cloud instances stay unchanged: the field is only assignable for Apache instances (`InstanceService.java:602`), and the fix must not touch the insert path.

btlqql:

An instance's admin credential reference could not be removed. `InstanceService.updateInstance` maps a blank submitted reference to `null` through `normalizeCredentialRef` (`InstanceService.java:561-563`), so a blank is an explicit clear:

```java
if (!cloudInstance && instance.getAdminCredentialRef() != null) {
    updated.setAdminCredentialRef(normalizeCredentialRef(instance.getAdminCredentialRef()));
}
```

The updated instance is then written by `instanceRepository.save(updated)` (`InstanceService.java:607`), and `MybatisPlusInstanceRepository.save` issues `instanceMapper.updateById(entity)` (`MybatisPlusInstanceRepository.java:122`). MyBatis-Plus `updateById` omits null entity fields (`FieldStrategy.NOT_NULL`; there is no `update-strategy` override in `application.yml`), so the null reference never reaches the SET clause: the stored `admin_credential_ref` survived the update while the service returned a VO whose reference was null.

The user-visible effect is the same one described in #3342 for the ACL columns: the edit dialog accepts the clear, the API answers 200 with `adminCredentialRef: null`, the page replaces its row with that response without refetching (`web/src/pages/instance/index.tsx:447-455`), and the reference reappears on the next list load. An operator can therefore set or change an external admin credential reference but never remove it.

The repository is the only layer that can express the clear: `InstanceRepository.save` is called just twice, from `InstanceService.createInstance` (insert) and `InstanceService.updateInstance` (update), so the update branch now assigns `admin_credential_ref` explicitly when the incoming VO carries no reference.

## Brief changelog

- `MybatisPlusInstanceRepository.save`: after the entity update, assign `admin_credential_ref` explicitly when the saved instance has no reference, because `updateById` skips null entity fields.
- `MybatisPlusInstanceRepositoryTest`: two new cases — a null reference clears the column through an explicit assignment; a submitted reference issues no extra assignment.

## Verification

```
mvn -f server/pom.xml -B -DskipITs -Dtest=MybatisPlusInstanceRepositoryTest -DfailIfNoTests=false test
[ERROR] Tests run: 18, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.140 s <<< FAILURE! -- in org.apache.rocketmq.studio.instance.MybatisPlusInstanceRepositoryTest
[ERROR]   MybatisPlusInstanceRepositoryTest.saveShouldClearTheAdminCredentialRefWhenTheUpdateRemovesIt:254
[ERROR] Tests run: 18, Failures: 1, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

(`...:254` is the `verify(instanceMapper).update(isNull(), captor.capture())` assertion — no assignment was issued for the cleared column.)

Green after the change:

```
mvn -f server/pom.xml -B -DskipITs -Dtest=MybatisPlusInstanceRepositoryTest -DfailIfNoTests=false test
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.199 s -- in org.apache.rocketmq.studio.instance.MybatisPlusInstanceRepositoryTest
[INFO] BUILD SUCCESS
```

Fixes #4273


RockteMQ-AI (approved):
### Summary
Fix for Instance update not clearing `adminCredentialRef` when the reference is removed. Same MyBatis-Plus `updateById` null-field omission pattern as #4277 and #3342.

### Review

**Correctness** ✅
- The fix is minimal and targeted: after `updateById` succeeds, if `adminCredentialRef` is null, issue an explicit `UPDATE ... SET admin_credential_ref = NULL` via `UpdateWrapper`.
- Placement after the success check (line 124) ensures the explicit clear only runs when the main update succeeded.

**Tests** ✅
- Two new tests cover both paths: cleared reference triggers the explicit null update; non-null reference does not.
- Mock verification confirms the exact SQL and parameter values.

**Code quality** ✅
- Clear inline comment explaining why the workaround is needed.
- Follows the same pattern as #3342 and #4277, maintaining consistency.

LGTM.

lizhimins:
Clearing the admin credential reference in the instance dialog really did leave the old value persisted, and the fix targets the right layer (`normalizeCredentialRef`).

PR #4277, opened 2026-09-13
fix(nameserver): persist a cleared registry description and K8s columns

### LINKED ISSUE #4274: [Studio][Bug] Blanking a NameServer registry entry description, k8s namespace or k8s id has no effect
## 1. Symptom

Blanking the K8s namespace, K8s ID or description of a NameServer registry entry has no effect. In the Studio cluster page's registry edit dialog, clearing one of those fields and saving reports success and reloads the list, and the previous value is back in the row.

## 2. Root cause

`server/src/main/java/org/apache/rocketmq/studio/cluster/nameserver/NameserverRegistryService.java:79-85`

```java
entity.setName(name);
entity.setNamesrvAddr(NamesrvAddrParser.normalize(command.getNamesrvAddr()));
entity.setK8sNamespace(command.getK8sNamespace());
entity.setK8sId(command.getK8sId());
entity.setDescription(command.getDescription());
...
int updated = nameserverMapper.updateById(entity);
```

The update request replaces every editable field of the entry - `UpdateNameserverRegistryDTO.name` and `.namesrvAddr` are `@NotBlank` (`UpdateNameserverRegistryDTO.java:35-41`), and the three optional columns are copied unconditionally - so the service's contract is "the stored entry matches the request".

Two things then line up to lose the clear:

1. The edit dialog submits a blanked optional field as an *absent* one: `web/src/pages/cluster/index.tsx:255-257` builds the payload with `k8sNamespace: values.k8sNamespace || undefined`, `k8sId: values.k8sId || undefined` and `description: values.description || undefined`, so `command.getDescription()` etc. are `null` after a field is cleared.
2. MyBatis-Plus `updateById` omits null entity fields (`FieldStrategy.NOT_NULL`; there is no `update-strategy` override in `application.yml`), so those three columns are dropped from the SET clause and keep their previous values.

Clearing is therefore unreachable through the API: `null` means "leave the column alone" (accidentally), and there is no separate clear flag, unlike `clearApiKey` in general settings. The same mechanism was already documented and fixed for the ACL user/rule columns in #3342.

## 3. Impact

- The registry entry's annotations are write-once: an operator can add or change a K8s namespace / K8s ID / description but cannot remove it, which is exactly what is needed when an entry is re-pointed at a non-K8s deployment or a stale `k8sId` has to be dropped.
- The API response and the reloaded list disagree with the write: the save reports success, `update` re-reads the row and returns the retained value, and the dialog then shows the old text again with no error.
- `k8s_id` links a registry entry to its K8s certificate view, so a stale one cannot be cleaned up from the UI.

## 4. Reproduction

1. Create a registry entry with `k8sNamespace = "rocketmq1"`, `k8sId = "k8s-1"`, `description = "community chart cluster"`.
2. Edit it and clear the description (leave name and namesrvAddr untouched), save.
3. Effect: HTTP 200, list reloads, the description is still `community chart cluster`.
4. `SELECT k8s_namespace, k8s_id, description FROM rmq_nameserver WHERE id = <id>` confirms the columns were not updated.

## 5. Expected behaviour

- The stored entry matches the submitted request: an optional field the request does not carry is persisted as NULL for that entry, and a field the request does carry is written as submitted.
- No signature or DTO change is required; `updateById` stays for the non-null columns, the cleared ones are assigned explicitly.

btlqql:

Blanking the K8s namespace, K8s ID or description of a NameServer registry entry had no effect. `NameserverRegistryService.update` replaces every editable field of the entry — `name` and `namesrvAddr` are `@NotBlank` in `UpdateNameserverRegistryDTO`, and `k8sNamespace`, `k8sId` and `description` are copied unconditionally onto the entity — so the request is a full replacement of the entry it addresses:

```java
entity.setK8sNamespace(command.getK8sNamespace());
entity.setK8sId(command.getK8sId());
entity.setDescription(command.getDescription());
...
int updated = nameserverMapper.updateById(entity);
```

Two things then line up to lose the clear. The cluster page's registry edit dialog submits a blanked optional field as an absent one (`web/src/pages/cluster/index.tsx:255-257` builds the payload with `k8sNamespace: values.k8sNamespace || undefined` and the same for `k8sId` and `description`), and MyBatis-Plus `updateById` omits null entity fields (`FieldStrategy.NOT_NULL`; there is no `update-strategy` override in `application.yml`), so those three columns were dropped from the SET clause and kept their previous values.

Clearing was therefore unreachable through the API: `null` meant "leave the column alone" (accidentally), and there is no clear flag as in general settings. The save reported success, `update` re-read the row and returned the retained value, and the dialog displayed the old text again with no error — which is where the operator notices, because a stale `k8sId` cannot be dropped from the UI.

The cleared optional columns are now assigned explicitly, mirroring the ACL clear-columns fix in #3342, while the non-null columns keep going through `updateById`. `name` / `namesrvAddr` validation, duplicate-name handling and the concurrent-delete checks are unchanged.

## Brief changelog

- `NameserverRegistryService.update`: assign `k8s_namespace`, `k8s_id` and `description` explicitly when the request carries no value for them, because `updateById` skips null entity fields.
- `NameserverRegistryServiceTest`: two new cases — a request that omits the three optional columns clears them through an explicit assignment; a request that submits them issues no extra assignment.

## Verification

```
mvn -f server/pom.xml -B -DskipITs -Dtest=NameserverRegistryServiceTest -DfailIfNoTests=false test
[ERROR] Tests run: 22, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.135 s <<< FAILURE! -- in org.apache.rocketmq.studio.cluster.nameserver.NameserverRegistryServiceTest
[ERROR]   NameserverRegistryServiceTest.updateShouldClearOptionalColumnsThatTheRequestOmitsTest:368
[ERROR] Tests run: 22, Failures: 1, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

(`...:368` is the `verify(nameserverMapper).update(isNull(), captor.capture())` assertion — no assignment was issued for the cleared columns.)

Green after the change:

```
mvn -f server/pom.xml -B -DskipITs -Dtest=NameserverRegistryServiceTest -DfailIfNoTests=false test
[INFO] Tests run: 22, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.126 s -- in org.apache.rocketmq.studio.cluster.nameserver.NameserverRegistryServiceTest
[INFO] BUILD SUCCESS
```

Fixes #4274


RockteMQ-AI (approved):
### Summary
Fix for NameServer registry entry update not clearing optional columns (k8sNamespace, k8sId, description) when blanked in the UI. The root cause is MyBatis-Plus `updateById` omitting null entity fields (FieldStrategy.NOT_NULL), which silently retains stale values.

### Review

**Correctness** ✅
- The `clearOmittedOptionalColumns` method correctly uses `UpdateWrapper` to explicitly SET null for omitted optional columns, mirroring the established pattern from #3342.
- The `anyCleared` guard avoids unnecessary DB round-trips when no columns need clearing.
- Placement before the `selectById` reload ensures the returned VO reflects the cleared state.

**Tests** ✅
- Two well-structured tests: one verifying explicit clear when optional columns are omitted, another confirming no extra assignment when they are present.
- ArgumentCaptor assertions verify the exact SQL and parameter values.

**Code quality** ✅
- Clear Javadoc explaining the MyBatis-Plus behavior and why the workaround is needed.
- Method extraction keeps the `update` method clean.

LGTM.

lizhimins:
The three optional NameServer fields (registry description and the two K8s columns) really were sent as `|`-joined empty strings and never cleared server-side.

PR #4288, opened 2026-09-14
fix(message): keep supplementary characters intact when abbreviating properties

### LINKED ISSUE #4287: [Studio][Bug] An abbreviated message property can end with a lone surrogate
## 1. Symptom

## 2. Root cause

```java
            return value;
        }
    }
```

still splits the pair.

## 3. Impact

  with what is rendered.

## 4. Reproduction

message detail. Unit level:

```java
```

## 5. Expected behaviour

  and `hasOversizedProperty`.

btlqql:

`MessagePropertyDisplay.limitProperties` caps every message property value at
`MAX_PROPERTY_VALUE_CHARS` (1024) and marks the cut with `...`. The cap counts UTF-16 chars, but the
text it cuts is not ASCII-only, and `abbreviate` used a plain `substring(0, 1024)`. Whenever the
limit lands between the two chars of a supplementary character the returned value ends with an
unpaired high surrogate. A `char` is not a code point, so the JSON body carries an invalid
`\uD83D` escape and the message detail panel shows a replacement glyph at the end of an already
abbreviated value.

The repository already treats this as a defect elsewhere, which is what makes the raw cut an
oversight rather than a deliberate limit: `CredentialUtils.mask` cuts on a code point boundary
because it "prevents a supplementary character from being split into an isolated surrogate in API
responses", and the web-side truncation of these same message properties was fixed in #4226. The
server helper that feeds both panels kept the raw `substring`.

`abbreviate` now backs off one char when the char at the cap starts a pair, so the result never
contains an isolated surrogate, while a pair that fits inside the cap is still kept in full. Values
within the cap, the `...` marker, `MAX_PROPERTIES` and `hasOversizedProperty` are unchanged, so the
"this value was abbreviated" hint stays consistent with what is rendered.

## Brief changelog

- `MessagePropertyDisplay.abbreviate`: do not cut inside a surrogate pair — back off one char when
  the char at the cap is a high surrogate.
- `MessagePropertyDisplayTest`: two new cases — the cap falling inside a pair (1023 ASCII chars +
  an emoji) and a pair that fits inside the cap (1022 ASCII chars + an emoji); both assert that the
  result contains no unpaired surrogate.

## Verification

```
mvn -f server/pom.xml -B -Dtest=MessagePropertyDisplayTest -DfailIfNoTests=false test
[ERROR] Tests run: 7, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.086 s <<< FAILURE! -- in org.apache.rocketmq.studio.common.util.MessagePropertyDisplayTest
[ERROR] org.apache.rocketmq.studio.common.util.MessagePropertyDisplayTest.limitPropertiesShouldNotSplitASurrogatePairTest -- Time elapsed: 0.012 s <<< FAILURE!
	at ...MessagePropertyDisplayTest.limitPropertiesShouldNotSplitASurrogatePairTest(MessagePropertyDisplayTest.java:66)
[INFO] BUILD FAILURE
```

`:66` is `assertThat(hasUnpairedSurrogate(abbreviated)).isFalse()`: the returned value is 1023 ASCII
chars followed by `U+D83D` — the high surrogate of the emoji — and then `...`.

Green after the change:

```
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.071 s -- in org.apache.rocketmq.studio.common.util.MessagePropertyDisplayTest
[INFO] BUILD SUCCESS
```

Whole backend suite on this branch (`mvn -f server/pom.xml -B test`):
`Tests run: 2153, Failures: 5, Errors: 5, Skipped: 0`. The five failures and five errors are the same
ten test ids that fail on the unmodified base `6c24d2ed` (`Tests run: 2151, Failures: 5, Errors: 5`):
eight need a POSIX `sh` CLI that this Windows machine does not have (`CliAgentProviderTest`,
`ClaudeCodeAgentProviderTest`) and two are
`AuthCorsIntegrationTest.shouldRejectNonAdminMutationBeforeControllerExecution` /
`shouldStillRejectAnonymousProtectedRequests`, which are aligned separately in #4217. No other test
moved.

Closes #4287


btlqql:
Withdrawing this one: it duplicates #4226 (issue #4167), which was opened earlier and is already
CLEAN with the AI review approved, so this branch would only race it.

Both branches cut an oversized property value on a code point boundary. The only substantive
difference is the cap semantics: #4226 makes `MAX_PROPERTY_VALUE_CHARS` count code points for both
`abbreviate` and `hasOversizedProperty`, while this branch keeps the cap in UTF-16 chars and only
steps back one char when the cut would split a pair. The former is the better reading of the
constant's name, so #4226 should land.

Closing #4287 as a duplicate of #4167 for the same reason. Nothing in this branch is merged
elsewhere; it carries a passing red/green pair (`audit/rmq-pt-red.log`, `audit/rmq-pt-green.log`)
should it ever be needed.

lizhimins:
Backing off the cut when it lands between a high and a low surrogate does prevent a lone surrogate from reaching the JSON encoder.

One note for the regroup: `MessagePropertyDisplay.abbreviate` is also touched by #4226, which we are closing too, and by #4170, whose author closed it as a duplicate of #4226. Whichever variant you resubmit, please cover `hasOversizedProperty` and the javadoc in the same change — a fix that only pins `abbreviate` still lets an oversized value fall back to `value.length()` and split a surrogate pair.

PR #4292, opened 2026-09-14
fix(alert): render a ratio notification value as an exact percent

### LINKED ISSUE #4291: [Studio][Bug] A ratio alert notification renders a floating-point artifact instead of a whole percent
## 1. Symptom

A native ratio alert whose threshold unit is `%` (for example `broker.disk.usage_ratio`) delivers its
notification body with a raw floating-point artifact for `${value}`: a broker disk usage of `0.29` is
rendered and delivered as `28.999999999999996` instead of `29`.

## 2. Root cause

`server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationTemplate.java:61-70`

```java
if (rule != null && "%".equals(rule.getThresholdUnit())
        && RATIO_METRICS.contains(rule.getMetric() == null ? "" : rule.getMetric().trim())) {
    return String.valueOf(alert.getCurrentValue() * 100);
}
```

Ratio metrics are stored as fractions of the whole: `ApacheRocketMqClusterMetricsCollector.java:142`
normalizes every sample with `value > 1D ? value / 100D : value`, and
`AlertRuleSemanticFingerprint.java:65-66` divides the `%` threshold by 100 for exactly the same
metric/unit pairing. The template therefore has to scale the stored fraction by 100, but it does so
with plain `double` arithmetic, and `Double.toString` prints the exact binary result of that product.

For whole percents, `n / 100 * 100` does not round-trip for 8 of the 99 possible values:
`0.07 -> 7.000000000000001`, `0.14 -> 14.000000000000002`, `0.28 -> 28.000000000000004`,
`0.29 -> 28.999999999999996`, `0.55 -> 55.00000000000001`, `0.56 -> 56.00000000000001`,
`0.57 -> 56.99999999999999`, `0.58 -> 57.99999999999999`.

## 3. Impact

- The rendered artifact is persisted into the notification row
  (`NotificationOutboxService.java:191` -> `AlertNotificationTemplate.render(...)`) and delivered over
  every configured channel (DingTalk, SMS, email), so an operator reading an alert sees a number that
  looks like a broken metric.
- The same template is used by the alert template preview, so the preview shows the artifact too.
- Nothing rounds or formats the value anywhere else on this path (no `BigDecimal`, no
  `DecimalFormat`, no `Math.round`).

## 4. Reproduction

1. Create a cluster alert rule with metric `broker.disk.usage_ratio`, threshold unit `%` and a
   threshold of `29`. The web UI sends `thresholdUnit: '%'` for every ratio metric
   (`web/src/pages/ops/alerts.tsx:772`), and `AlertRuleRequestDTO.thresholdUnit` is unvalidated.
2. Let the broker report a commit-log disk ratio of `0.29`, so the rule fires.
3. The delivered notification body contains `28.999999999999996` where `${value}` is expanded.

## 5. Expected behaviour

- `${value}` for a ratio metric with the `%` unit renders the shortest exact decimal percent:
  `0.29 -> 29`, `0.07 -> 7`, and the already-correct `0.865 -> 86.5` keeps working.
- Values that are not finite keep the current rendering instead of raising.
- Non-ratio metrics are untouched: the plain `String.valueOf` path stays as it is.

btlqql:

Fixes #4291.

A native ratio alert whose threshold unit is `%` is delivered with a raw floating-point artifact for
`${value}`: the stored fraction `0.29` is rendered as `28.999999999999996`. The template has to scale
the fraction by 100 because ratio samples are stored as fractions of the whole
(`ApacheRocketMqClusterMetricsCollector` normalizes `value > 1 ? value / 100 : value`) and the `%`
threshold is divided by 100 in `AlertRuleSemanticFingerprint`, but it scaled with plain `double`
arithmetic, so `Double.toString` printed the exact binary product.

## Brief changelog

- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationTemplate.java`
  - `${value}` for a ratio metric with the `%` unit is now scaled with
    `BigDecimal.valueOf(value).movePointRight(2).stripTrailingZeros().toPlainString()`, which yields
    the shortest exact decimal (`0.29` -> `29`, `0.07` -> `7`) and keeps the existing `0.865` -> `86.5`
    rendering.
  - A non-finite value keeps the previous `String.valueOf` rendering instead of raising.
  - The non-ratio path is unchanged.
- `server/src/test/java/org/apache/rocketmq/studio/ops/alert/AlertNotificationTemplateTest.java`
  - New `rendersWholePercentForRatioValueWithoutFloatingPointNoiseTest`.

## Verification

Red, before the fix (`mvn -o -f server/pom.xml test -Dtest=AlertNotificationTemplateTest`):

```
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.158 s <<< FAILURE! -- in org.apache.rocketmq.studio.ops.alert.AlertNotificationTemplateTest
expected: "29%"
 but was: "28.999999999999996%"
[ERROR]   AlertNotificationTemplateTest.rendersWholePercentForRatioValueWithoutFloatingPointNoiseTest:58
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

Green, after the fix (`mvn -o -f server/pom.xml test -Dtest=AlertNotificationTemplateTest`):

```
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.130 s -- in org.apache.rocketmq.studio.ops.alert.AlertNotificationTemplateTest
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Full server suite on this branch (`mvn -o -f server/pom.xml test`) compared with the unmodified base
`rocketmq-studio` @ 6c24d2ed:

```
```

The single added test is the new one; the 5 failures and 5 errors are the pre-existing
`CliAgentProviderTest`/`ClaudeCodeAgentProviderTest` POSIX `sh` tests and the two
`AuthCorsIntegrationTest` cases being aligned in #4217, unchanged by this change.


RockteMQ-AI (approved):
### Summary
Code review completed. No critical issues found. The changes look correct and well-structured.

---

PR #4294, opened 2026-09-15
fix(topic): mark consumer metrics unavailable when the lag is unknown

### LINKED ISSUE #4293: [Studio][Bug] The topic consumer list reports an unknown lag as a backlog of -1
## 1. Symptom

`GET /api/topics/{name}/consumers/page` (`TopicController` -> `MetadataService.getTopicConsumersPage` ->
`RocketMQMetadataProvider.getTopicConsumersPage`) publishes the internal "lag unknown" sentinel `-1`
as a real backlog while flagging the metrics as available.

For any consumer group whose queue offsets make the resolved lag unknown (a RocketMQ 5.0 gRPC/POP
consumer whose queue reports `brokerOffset - consumerOffset < 0`), the response contains

```json
{ "group": "cg-orders", "diffTotal": -1, "metricsAvailable": true }
```

and the topic page renders the 堆积量 column as `-1` instead of the unavailable marker
(`web/src/pages/instance/topic.tsx:927-932` renders `不可用` only when
`record.metricsAvailable === false`).

## 2. Root cause

`server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java:530-543`

```java
long diffTotal = 0;
...
long queueDiff = resolveDiff(ow.getBrokerOffset(), ow.getConsumerOffset());
if (queueDiff == ConsumerLagResolver.UNKNOWN) {
    diffTotal = ConsumerLagResolver.UNKNOWN;
    break;
}
...
consumers.add(TopicConsumerVO.builder()
        .consumeTps(consumeTps)
        .diffTotal(diffTotal)
        .build());
```

`TopicConsumerVO.metricsAvailable` is `@Builder.Default` `true`
(`instance/topic/TopicConsumerVO.java:35-36`), and this branch never overrides it. The sibling
failure branch in the same method does (`metricsAvailable(false)`), so the two ways of "I could not
read the numbers" report differently.

This is the one place the sentinel still reaches a user as a quantity: the consumer-group paths
convert it (`MetadataService.lagText` returns `unknown`, `ConsumerGroupListToolHandler:64` returns
`null`), and `ConsumerLagResolver`'s own javadoc states that the sentinel exists so the UI can show
the genuine unknown state instead of a fabricated number.

## 3. Impact

- A group with an undeterminable lag shows a backlog of `-1`, which looks like a broken metric and
  hides the real state (the operator cannot tell "unknown" from "1 message in flight").
- The metric flag is inconsistent: the same unknown condition renders as `-1` on the topic consumer
  page and as `unknown` in the consumer-group CSV export and the dashboard.

## 4. Reproduction

1. Have a group with a POP/gRPC consumer whose queue reports `brokerOffset = 99`,
   `consumerOffset = 100` (or any negative raw diff with no proxy stats available).
2. Open the topic page's consumer tab for that topic, or call
   `GET /api/topics/{name}/consumers/page`.
3. `diffTotal` is `-1` and `metricsAvailable` is `true`.

## 5. Expected behaviour

- `metricsAvailable` is `false` for a group whose lag could not be determined, so the existing UI
  contract renders the unavailable state.
- `diffTotal` keeps carrying `ConsumerLagResolver.UNKNOWN` for callers that read it directly, and
  known-lag sums are unchanged.

btlqql:

Fixes #4293.

`RocketMQMetadataProvider.getTopicConsumersPage` keeps the internal lag sentinel (`-1`) in
`diffTotal` when a queue's lag cannot be determined, but never clears
`TopicConsumerVO.metricsAvailable` (which `@Builder.Default`s to `true`). The topic page keys the
unavailable rendering off that flag, so an undeterminable lag was shown to the operator as a backlog
of `-1` while the sibling failure branch in the same method reported `metricsAvailable=false`.

## Brief changelog

- `server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProvider.java`
  - `getTopicConsumersPage` now records that a queue resolved to `ConsumerLagResolver.UNKNOWN` and
    builds the row with `metricsAvailable(!lagUnknown)`, matching the neighbouring "stats could not be
    read" branch.
  - `diffTotal` keeps carrying the sentinel for callers that read it directly, and known-lag sums are
    unchanged.
- `server/src/test/java/org/apache/rocketmq/studio/provider/apache/RocketMQMetadataProviderTest.java`
  - New `getTopicConsumersMarksMetricsUnavailableWhenAnyQueueLagIsUnknown`.

## Verification

Red, before the fix (`mvn -o -f server/pom.xml test -Dtest=RocketMQMetadataProviderTest`):

```
[ERROR] Tests run: 41, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.330 s <<< FAILURE! -- in org.apache.rocketmq.studio.provider.apache.RocketMQMetadataProviderTest
[ERROR]   RocketMQMetadataProviderTest.getTopicConsumersMarksMetricsUnavailableWhenAnyQueueLagIsUnknown:517
[ERROR] Tests run: 41, Failures: 1, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

Green, after the fix (`mvn -o -f server/pom.xml test -Dtest=RocketMQMetadataProviderTest`):

```
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.723 s -- in org.apache.rocketmq.studio.provider.apache.RocketMQMetadataProviderTest
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

The pre-existing `getTopicConsumersKeepsUnknownWhenAnyQueueLagIsUnknown` (which asserts
`diffTotal == ConsumerLagResolver.UNKNOWN`) and `getTopicConsumersStillSumsKnownQueueLags` both still
pass, so known-lag sums are unaffected.

## AI disclosure

Drafted with AI assistance (analysis, patch and test authoring); the red and green Maven output above
was executed and observed locally on Windows, JDK 21.


btlqql:
Whole-suite comparison, added for completeness (`mvn -o -f server/pom.xml test`, unmodified base
`rocketmq-studio` @ 6c24d2ed vs. this branch):

```
```

The single added test is the new one. The failing ids are identical on both sides and unrelated to this
change: `CliAgentProviderTest` / `ClaudeCodeAgentProviderTest` (they need a POSIX `sh` runtime) and the
two `AuthCorsIntegrationTest` cases being aligned in #4217.

RockteMQ-AI (approved):
### Summary
Fixes the inconsistency where `ConsumerMetrics.lag` returns `-1` to indicate unknown lag, but `metricsAvailable` was still `true`, making the UI display a misleading `-1` value. The fix adds a `lagUnknown` flag in `TopicController` and sets `metricsAvailable(!lagUnknown)` so the frontend correctly shows "metrics unavailable" instead of a bogus negative lag.

Test confirms `metricsAvailable` is `false` when lag is unknown. Clean, minimal fix. LGTM.

---

lizhimins:
Marking the consumer metrics unavailable instead of showing a fabricated zero lag is the correct reading of the `-1` sentinel.

PR #4298, opened 2026-09-15
fix(alert): match a padded stored instance id when reading active states

### LINKED ISSUE #4297: [Studio][Bug] A rule with a padded stored instance id fires but reports no active alert state
## 1. Symptom

## 2. Root cause

```java
```

  a padded id passes)

## 3. Impact

## 4. Reproduction

2. Let the rule fire.

## 5. Expected behaviour

  scope.

btlqql:

Fixes #4297.

The active-alert-state lookup compares the collection scope's instance id against
`rule.getInstanceId()` untrimmed, while the metric selector on the line directly above is trimmed and
every rule that decides whether an alert fires trims too
(`NativeAlertRuleScopeMatcher`, `NativeAlertMetricCatalogService`, `NativeAlertRuleTestService`). A
stored padded instance id therefore fires and writes `FIRING`/`ACKED` rows that `findActive` cannot
see, so the alert disappears from the active/runtime view and
`NativeAlertProcessor.reconcileMissingActiveStates` can never resolve it.

## Brief changelog

- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertStateRepository.java`
  - `findActive` compares `StringUtils.trimWhitespace(rule.getInstanceId())`, matching the metric
    filter beside it.
- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/NativeAlertProcessor.java`
  - The reconcile candidate filter applies the same normalization, so the padded rule is not dropped
    before `findActive` is consulted.
- `server/src/test/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertStateRepositoryTest.java`
  - New `findsActiveStatesForRulesWithPaddedStoredInstanceIdsTest`, mirroring the existing
    padded-metric test.

## Verification

Red, before the fix (`mvn -o -f server/pom.xml test -Dtest=MybatisPlusAlertStateRepositoryTest`):

```
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.141 s <<< FAILURE! -- in org.apache.rocketmq.studio.ops.alert.MybatisPlusAlertStateRepositoryTest
Expected size: 1 but was: 0 in:
[ERROR]   MybatisPlusAlertStateRepositoryTest.findsActiveStatesForRulesWithPaddedStoredInstanceIdsTest:153
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

Green, after the fix
(`mvn -o -f server/pom.xml test -Dtest=MybatisPlusAlertStateRepositoryTest,NativeAlertProcessorTest`):

```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.094 s -- in org.apache.rocketmq.studio.ops.alert.MybatisPlusAlertStateRepositoryTest
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.751 s -- in org.apache.rocketmq.studio.ops.alert.NativeAlertProcessorTest
[INFO] Tests run: 24, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

## AI disclosure

Drafted with AI assistance (analysis, patch and test authoring); the red and green Maven output above
was executed and observed locally on Windows, JDK 21.


btlqql:
Whole-suite comparison, added for completeness (`mvn -o -f server/pom.xml test`, unmodified base
`rocketmq-studio` @ 6c24d2ed vs. this branch):

```
```

The single added test is the new one; the failing ids are identical on both sides and unrelated to this
change (`CliAgentProviderTest` / `ClaudeCodeAgentProviderTest` need a POSIX `sh` runtime, and the two
`AuthCorsIntegrationTest` cases are being aligned in #4217).

RockteMQ-AI (approved):
### Summary
Fixes an inconsistency where `findActive` compared `scope.instanceId()` against the raw (potentially padded) `rule.getInstanceId()`, while the metric selector on the line directly above and every other rule matcher (`NativeAlertRuleScopeMatcher`, `NativeAlertMetricCatalogService`, `NativeAlertRuleTestService`) all trim whitespace. This caused padded instance IDs to fire alerts and write FIRING/ACKED rows that `findActive` could not match back.

The fix applies `StringUtils.trimWhitespace()` consistently in both `MybatisPlusAlertStateRepository.findActive` and `NativeAlertProcessor.reconcileMissingActiveStates`. Test covers the padded-ID scenario. LGTM.

---

lizhimins:
The asymmetry is real: the evaluation path trims the instance id (`NativeAlertRuleScopeMatcher`) while the reclaim path does not (`NativeAlertProcessor`, `MybatisPlusAlertStateRepository`).

PR #4321, opened 2026-09-15
fix(acl): preserve Tencent role when deleting a rule

### LINKED ISSUE #4161: [Studio][Bug] Deleting a Tencent ACL rule deletes the role account

89799969:
Fixes #4161

## Problem

Studio maps each Tencent Cloud RocketMQ role to both an ACL **user** (credentials) and a cluster-wide ACL **rule** (PermRead/PermWrite). The two resources have separate delete actions, but `TencentAclService.deleteRule` called Tencent Cloud `DeleteRole`, exactly like user deletion.

Deleting a rule therefore permanently removed the role account and its access credentials.

## Fix

- `deleteRule` now calls `ModifyRole` with `PermRead=false` and `PermWrite=false`, which revokes the rule while keeping the role account.
- Roles with both permissions disabled are excluded from `listRules` (no effective rule) but remain visible in the ACL user list so operators can re-enable them.
- Explicit ACL user deletion still uses `DeleteRole`.

## Testing

- `TencentAclServiceTest` (14 tests, including 2 new cases):
  - `deleteRuleRevokesPermissionsWithoutDeletingTheRoleTest`
  - `listRulesHidesPermissionlessRolesThatStillExistAsUsersTest`

```
Tests run: 14, Failures: 0, Errors: 0, Skipped: 0
```

Local run: Temurin 21.0.12.1, `mvn -Dtest=TencentAclServiceTest test`


RockteMQ-AI (commented):
### Summary\nThis PR `fix(acl): preserve Tencent role when deleting a rule` modifies 2 files (+48 -2).\n\n### Findings\n- **[Warning]** Potential null pointer: method calls without null checks\n\nPlease review the findings above.\n\n---\n*Automated review by "github-manager-bot"*

lizhimins:
Closing: this is the same change we closed in #4162 about an hour before this PR was opened — `DeleteRole` -> `ModifyRole` with `permRead=false` / `permWrite=false`, plus the `listRules` filter that hides roles whose permissions are both false. The reasoning there still stands:

- in trocket a Role *is* the ACL principal, which is why `deleteUser` and `deleteRule` both call `DeleteRole`; routing only `deleteRule` to `ModifyRole` leaves the cloud-side Role in place after Studio has told the user the rule was deleted, and makes the two delete entry points mean different things;
- hiding roles with both permissions false makes Studio's view diverge from the cloud's real state, and a Role that reaches that state outside Studio becomes invisible and therefore unmanageable from here;
- we still have no evidence that `ModifyRole` with both permissions false is an accepted, meaningful state in the trocket API rather than an error or a no-op.

The instinct — do not destroy a shared cloud Role on a rule delete — is reasonable. As suggested in #4162, please open an issue with console/API evidence for that state and a story for the leftover Role and the divergent `deleteUser` path, and we will pick it up there.

PR #4324, opened 2026-09-15
fix(cluster): discard superseded broker config preview responses

### LINKED ISSUE #4165: [Studio][Bug] Slow broker config preview can overwrite a newer preview with stale values

89799969:
Fixes #4165

## Problem

On the Cluster page (Broker 管理), the broker config preview panel renders whichever `previewClusterConfig` response arrives last.

`handleConfigPreview` awaits the request and then unconditionally commits the result — no request-generation guard — while every comparable async path in the same file (`nsConfigDiffRequestRef`, `brokerConfigDiffRequestRef`, `connectionTestRequestRef`) is guarded.

Reproduction: open 配置 → change a value → 预览 (response A slow) → Cancel → reopen → change values → 预览 (response B arrives) → slow response A resolves last → the panel shows A's diff while the form holds different values.

## Fix

- Add `configPreviewRequestRef`
- Bump the id on open, close, form change, successful submit, and each preview request
- Ignore stale preview/error/loading commits when the id no longer matches

## Testing

- `ClusterPage.test.tsx` (26 tests, including `discards a superseded broker config preview that resolves late`)

```
Test Files  1 passed (1)
     Tests  26 passed (26)
```

Local run: Node 20, `npm test -- --run src/pages/cluster/__tests__/ClusterPage.test.tsx`


RockteMQ-AI (commented):
### Summary\nThis PR `fix(cluster): discard superseded broker config preview responses` modifies 2 files (+101 -2).\n\n### Findings\n- **[Warning]** Potential null pointer: method calls without null checks\n- **[Warning]** Missing test coverage: 101 lines added but no test files modified\n\nPlease review the findings above.\n\n---\n*Automated review by "github-manager-bot"*

lizhimins (changes requested):
The guard you add here is now on trunk: #4331 was merged as 6334f9238 and extracts one shared request-generation helper (`begin` / `isCurrent` / `invalidate`) which it applies to every hand-rolled guard in `cluster/index.tsx`, including the previously unguarded `handleConfigPreview`.

We deliberately asked for the shared helper rather than a fourth standalone ref, so this patch would reintroduce the pattern we just removed. Please rebase and, if you still find an unguarded async path in that file, rework it on top of the helper.

89799969:
Closing as covered by trunk: #4331 / 6334f9238 extracted `useRequestGeneration` and applied it to `handleConfigPreview` (`configPreviewRequest`), including invalidate on instance change. Reintroducing a standalone ref would undo the shared helper the review asked for.

PR #4331, opened 2026-09-15
refactor(cluster): unify the request guards with one shared generation helper

### LINKED ISSUE #4165: [Studio][Bug] Slow broker config preview can overwrite a newer preview with stale values

unbridled-41:
Fixes #4165 (the extraction @lizhimins requested on #4166).

Follow-up implementing the extraction @lizhimins asked for on #4166: instead of adding a fourth hand-rolled request-id ref, one shared helper now covers every async flow in the page, including the previously unguarded broker config preview. The display-only race itself is documented there; the preview never feeds the write path (`handleConfigSubmit` rebuilds the request from form values), so this change only unifies the invalidation mechanism.

## What changed

`web/src/pages/cluster/index.tsx` had six hand-written latest-request guards (`nsRegistryRequestRef`, `registryClustersRequestRef`, `k8sCertsRequestRef`, `nsConfigDiffRequestRef`, `brokerConfigDiffRequestRef`, `connectionTestRequestRef`) — each a monotonic counter ref compared after `await` and bumped at close/unmount. The broker config preview (`handleConfigPreview`) had no guard at all.

This PR introduces one shared helper and converts every flow to it:

- `createRequestGeneration()` returns `{ begin, isCurrent, invalidate }`; `useRequestGeneration()` creates one stable generation per async flow (lazy `useState` initializer, no ref access during render).
- All six existing refs are replaced by the helper — same begin/check/invalidate points, so existing behavior is unchanged; the guards' dep arrays list the stable generation objects to keep `react-hooks/exhaustive-deps` clean.
- The config preview gains the same guard: `begin()` when the request is issued, `isCurrent()` before committing the panel/toast and before clearing the loading state, and `invalidate()` at the three points where the preview is already discarded today — dialog open (`handleConfigOpen`), dialog cancel, and any form value change (`onValuesChange`, which now also resets the preview loading state alongside clearing the panel).

A superseded preview response (cancel/reopen while in flight, or an edit while in flight) can therefore no longer overwrite the panel, fire its toast, or clobber a newer request's loading state.

## Tests

One regression covering the maintainer-identified residual window (cancel and reopen while a preview is in flight):

- `keeps the latest broker config preview after a superseded response finishes last`: preview A is held in flight, the dialog is cancelled and reopened, the form changes, preview B is issued and resolved, then A resolves last — the panel must keep B (`defaultTopicQueueNums=24`) and never show A (`defaultTopicQueueNums=16`).
- Red on unmodified base `25132a0d` (fix stashed): `1 failed | 25 passed` — the stale response overwrote the newer preview.
- Green with the fix: `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx` → `Test Files 1 passed (1)`, `Tests 26 passed (26)`.
- Full web suite: `981-982 passed` with 1-3 per-run failures in varying files (`MetricsExplorer`, `ClusterPage` nameserver registry create/delete, `ConsumerPage` group health) — those three files all pass in isolation with the fix applied (`82 passed (82)`), matching the documented full-suite parallel-load flakiness; the ClusterPage isolation run is also red on base exactly for the new test (`1 failed | 25 passed`), attributing the full-suite flake to load, not the change.
- `eslint` clean on both changed files (also enforced by the pre-commit `lint-staged` hook), `tsc --noEmit` clean, `vite build` succeeds (`✓ built in 8.40s`).

## Risk

Low. The helper is a mechanical generalization of the ref pattern already used in this file, with the invalidation points unchanged; the only new behavior is the intended one — superseded preview responses no longer commit state, toast, or touch loading. No API, data, or visual changes.


web/src/pages/cluster/index.tsx
```diff
@@ -116,6 +116,34 @@ const CONFIG_FIELD_LABEL_KEYS: Record<string, string> = {
   msgTraceTopicName: 'cluster.msgTraceTopicName',
 };
 
+type RequestGeneration = {
+  begin: () => number;
+  isCurrent: (requestId: number) => boolean;
+  invalidate: () => void;
+};
+
+const createRequestGeneration = (): RequestGeneration => {
```
RockteMQ-AI, line 125: **[Info]** The `createRequestGeneration` helper is a clean abstraction. One minor consideration: the `invalidate()` method increments the counter, which means after invalidation, `isCurrent()` will return false for any in-flight requests. This is correct behavior, but worth noting that `invalidate()` and `begin()` have the same effect on the counter — both increment it. If the intent is to distinguish between "start a new request" and "cancel existing requests", you might consider making the semantics more explicit in the type signature or comments.

That said, the current implementation is correct and the test coverage (especially the stale-response test) validates the behavior well.

RockteMQ-AI (approved):
### Summary

This PR refactors the cluster page to unify six hand-written request-generation guards into a single reusable helper (`createRequestGeneration`) and React hook (`useRequestGeneration`). The change reduces code duplication, makes the pattern more testable, and adds coverage for the previously unguarded broker config preview flow.

The refactor is well-executed:
- The helper provides a clean API (`begin()`, `isCurrent()`, `invalidate()`)
- The hook integrates seamlessly with React's lifecycle
- Test coverage is comprehensive, including a stale-response race condition test
- The previously unguarded broker config preview is now protected

This is a good quality-of-life improvement that makes the codebase more maintainable.

### Findings
- **[Info]** index.tsx:125 — `invalidate()` and `begin()` have the same effect; consider making semantics more explicit (minor)

---

PR #4335, opened 2026-09-15
[ISSUE #4157] docs(alert): clarify aggregation window retention

### LINKED ISSUE #4157: [Studio][Bug] Reject native alert windows beyond snapshot retention

Aias00:
## Summary

- replace the previous retention-bound hard validation with localized, non-blocking form guidance
- explain that native window aggregation uses only metric snapshots currently retained locally
- clarify that history outside the current retention period is unavailable and retention changes do not invalidate rules
- reuse the shared Cluster and Business alert form without assuming a fixed deployment retention

This rework follows the maintainer feedback on #4158. It makes no backend, persistence, import, cleanup, or aggregation-math changes. The single signed-off commit is based on the latest `rocketmq-studio` branch.

## Verification

- `npm test -- --run src/pages/ops/__tests__/AlertsPage.test.tsx` (24 passed; Cluster English and Business Chinese coverage)
- `npm test -- --run src/i18n/__tests__/LangContext.test.tsx src/i18n/languagePreference.test.ts` (9 passed)
- `npm run build`
- changed-file ESLint (0 errors; 5 pre-existing warnings)
- changed-file Prettier check
- `git diff --check`

Fixes #4157


RockteMQ-AI (approved):
### Summary

This PR adds clarifying help text to the alert rule form's window aggregation field, explaining that:
- Aggregation uses only locally retained metric snapshots
- History outside the retention period is unavailable
- Changing retention does not invalidate existing rules

The change is clean and well-implemented:
- Adds a new i18n key with both Chinese and English translations
- Uses the `extra` prop to display the help text below the form field
- Includes test coverage for both language variants

This is a good UX improvement that makes the aggregation window behavior transparent to users.

No issues found.

---

PR #4339, opened 2026-09-15
fix(auth): stop revealing disabled accounts on the database login path

### LINKED ISSUE #4159: [Studio][Bug] Login reveals disabled accounts and skips rate limiting before the password is verified

unbridled-41:
Fixes #4159 (rework of #4160 following @lizhimins's review).

## Problem / Evidence

`AuthService.loginDatabaseUser` (`server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java`) rejected disabled accounts with `403 "User account is disabled"` **before** any credential check. An unauthenticated caller could therefore distinguish *unknown user / wrong password* (401) from *existing disabled account* (403) — account enumeration on the login path. As a side effect those attempts never reach the login rate limiter, which records failures for 401s only (`AuthService.login`, the `catch (BusinessException … code == 401)` branch).

## Design (per the review on #4160)

The review asked to keep the disabled short-circuit ahead of the expensive hash — e.g. *verify against a dummy hash* — and to cover the disabled+correct-password case. This PR implements exactly that:

- The `if (!enabled)` short-circuit stays in place, so a disabled account's **stored hash is never used** for login verification (satisfies "keep the disabled short-circuit ahead of the expensive hash").
- Inside the short-circuit, the submitted password is verified against `DUMMY_PASSWORD_HASH` — a fixed, well-formed PBKDF2 hash (210 000 iterations, same parameters as real hashes) of a random preimage that ships nowhere. The result is discarded. This burns one full derivation so the response timing matches a wrong-password attempt on an enabled account (no timing oracle for the disabled state), while the dummy never matches a real credential.
- The response is then the generic `401 "Invalid username or password"` — byte-identical to unknown users and enabled+wrong-password.

**Conscious behavior change (please confirm this is acceptable):** "disabled account + correct password" now also returns the uniform 401 instead of the previous `403 "User account is disabled"`. Both options offered in the review (dummy-hash verification and a uniform 401) produce this outcome — a 403 that fires only for the *correct* password is impossible without running the full PBKDF2 derivation against the account's real hash, which the review asked to avoid. The new behavior is pinned by a dedicated test; if you'd rather keep the 403 for correct passwords, the alternative is verifying against the real hash before revealing state, and I'm happy to rework.

**Rate limiter:** `LoginRateLimiter` code and policy are untouched. Because disabled attempts now produce the same uniform 401, they flow through the existing `code == 401` failure recording like any other failed login, so the per-username lockout applies to them too. Note this also closes the old asymmetry where the 6th attempt (401/429 for unknown and enabled vs 403 for disabled) itself revealed the disabled state; after this change the limiter behaves identically for every failure class.

## Tests (`AuthServiceDatabaseTest`, 6 new)

1. `disabledAccountsWithWrongPasswordsGetTheUniformInvalidCredentialsResponse` — case (a): 401 generic, not 403.
2. `disabledAccountsWithCorrectPasswordsGetTheUniformInvalidCredentialsResponse` — case (b), the review's requested case: uniform 401 and `sessionMapper.insert` never called (no token issued).
3. `disabledAccountVerificationNeverTouchesTheAccountPasswordHash` — with a mocked `PasswordHasher`: `matches` is **never** called with the account's stored hash, and the one call it receives uses a `pbkdf2$210000$…` hash (proving the dummy verification is a full-cost derivation, not a fast reject).
4. `unknownUsersShareTheUniformInvalidCredentialsResponse` — case (c).
5. `enabledAccountsWithWrongPasswordsGetTheUniformInvalidCredentialsResponse` — case (d).
6. `disabledAccountFailuresAreRateLimitedLikeOtherFailedLogins` — case (e): five failures reach `LoginRateLimiter.MAX_FAILED_ATTEMPTS` and the sixth attempt gets `429 Too many failed login attempts…`; the pre-existing unknown-user lockout test (`repeatedFailedLoginsAreRejectedWithTooManyRequestsTest`) is untouched and still passes.

Results:

- Red on unmodified base `25132a0d` (fix stashed): `Tests run: 27, Failures: 4` — the disabled+wrong, disabled+correct, never-touches-real-hash and disabled-lockout tests fail with `expected: 401 but was: 403`; the unknown-user and enabled+wrong guard tests pass on both base and fix.
- Green with the fix: `mvn -o test -Dtest=AuthServiceDatabaseTest` → `Tests run: 27, Failures: 0`.
- Related auth modules: `AuthServiceDatabaseTest,AuthServiceTest,LoginRateLimiterTest,AuthControllerTest` → `Tests run: 64, Failures: 0, Errors: 0`, BUILD SUCCESS.
- Full backend suite `mvn -o test` → `Tests run: 2351, Failures: 0, Errors: 13`. All 13 errors are Spring `ApplicationContext` load failures in the same 8 integration-test classes (`HealthProbeIntegrationTest`, `StudioApplicationTest`, `AuthServiceBootstrap/Concurrency/SessionOverviewIntegrationTest`, `QueryHistoryServiceIntegrationTest`, `NativeAlertEvaluationTransactionTest`, `NotificationOutboxMapperIntegrationTest`); running exactly those classes on pristine base with the fix stashed reproduces `Tests run: 13, Failures: 0, Errors: 13` — a pre-existing environment limitation of this sandbox, zero new failures.

## Risk

Low. One branch of one method changes its response from a pre-credential 403 to a post-dummy-verification uniform 401; no API shape, persistence, or limiter code changes. The one visible product change (disabled+correct-password users now see the generic invalid-credentials response instead of "User account is disabled") is intentional, tested, and called out above for review.


server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java
```diff
@@ -89,6 +89,12 @@ public class AuthService {
     private final PasswordHasher passwordHasher;
     private final LoginRateLimiter loginRateLimiter;
 
+    // A well-formed PBKDF2 hash of a random preimage that is not a real credential.
+    // Disabled accounts verify against it so an attempt costs the same as a
+    // wrong-password attempt on an enabled account while their stored hash is never used.
+    private static final String DUMMY_PASSWORD_HASH =
```
RockteMQ-AI, line 95: **[Info]** The `DUMMY_PASSWORD_HASH` is a good approach for timing-attack mitigation. One minor consideration: the hash is hardcoded as a constant, which means all disabled account verifications use the same dummy hash. While this doesn't weaken security (the preimage is unknown), rotating or randomizing the dummy hash per-instance could make it slightly harder for an attacker to identify the disabled-account path through hash analysis.

That said, this is a minor point — the current implementation is secure and the 210,000 PBKDF2 iterations ensure the cost matches a real password check.

RockteMQ-AI (approved):
### Summary

This PR fixes an account enumeration vulnerability in the database login path. Previously, disabled accounts returned 403 before credential verification, allowing unauthenticated callers to distinguish between unknown users (401) and disabled accounts (403).

The fix is correct and follows security best practices:
- Disabled accounts now return 401 (same as wrong password)
- A dummy PBKDF2 hash is verified to match timing with real password checks
- The real password hash is never touched for disabled accounts
- Comprehensive tests verify all edge cases

This is a well-implemented security fix with proper test coverage.

### Findings
- **[Info]** AuthService.java:95 — Consider rotating dummy hash per-instance for additional obscurity (minor)

---

RockteMQ-AI (approved):
### Summary
This PR fixes an account enumeration vulnerability where disabled accounts could be distinguished from unknown users by the 403 vs 401 response code. The fix implements a uniform 401 response for all login failures while preventing timing attacks.

**Security Analysis:**
- ✅ **Vulnerability fixed**: Disabled accounts no longer return a distinguishable 403 response
- ✅ **Timing attack prevention**: Dummy PBKDF2 hash verification burns CPU time to match real password checks
- ✅ **No hash exposure**: Disabled accounts real password hash is never used
- ✅ **Rate limiter integration**: Disabled account failures now properly flow through the existing rate limiter

**Implementation Quality:**
- Clean, minimal change to `loginDatabaseUser` method
- Well-documented `DUMMY_PASSWORD_HASH` constant with clear security rationale
- Comprehensive test coverage (6 new tests) covering all edge cases:
  - Disabled + wrong password → uniform 401
  - Disabled + correct password → uniform 401 (no session created)
  - Disabled verification never touches real hash
  - Unknown users → uniform 401
  - Enabled + wrong password → uniform 401
  - Disabled failures are rate-limited

**Behavior Change (intentional):**
- "Disabled account + correct password" now returns 401 instead of 403
- This is necessary to prevent account enumeration and is properly documented
- The alternative (403 only for correct password) would require running the full PBKDF2 derivation against the real hash, which the review explicitly asked to avoid

**Testing:**
- All 6 new tests pass
- Full auth test suite: 64 tests, 0 failures
- Pre-existing integration test failures are environment-related, not caused by this change

Excellent security fix with proper timing attack mitigation and comprehensive test coverage.

---

lizhimins:
Merged as part of #4284 — squashed into `rocketmq-studio` as `796c02e18`.

Your change here (answering disabled-account logins like invalid credentials) is included.

We grouped 6 of your related back-end fixes into a single commit so the history stays reviewable; the other PRs in the group are closed with the same note. Everything was cherry-picked onto the current baseline with you kept as the author, and your tests came along with the change.

PR #4300, opened 2026-09-15
fix(k8s): normalize certificate identities on create like on update

btlqql:
## What is the purpose of the change

Fixes #4299.

`K8sCertService.createCert` passed `k8sId`, `cluster` and `issuer` straight from the request into the
stored entity, while `updateCert` sends the same three fields through `normalizeOptionalIdentity`
(trim, reject blank with `400 Certificate <field> cannot be blank`). A whitespace-only `issuer` was
therefore accepted and stored padded on create, and could not be reproduced through update.

## Brief changelog

- `server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java`
  - `createCert` normalises `k8sId` and `cluster` with `normalizeOptionalIdentity`, the helper the
    update path already uses.
  - `issuer` is normalised the same way when the create carries no `certPem`; a supplied certificate
    still overrides the submitted issuer/SAN with its parsed values, so the upload flow is unchanged.
- `server/src/test/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertServiceTest.java`
  - New `createCertShouldRejectBlankIdentityFields`, mirroring the existing
    `updateCertShouldRejectBlankIdentityFields`.
  - New `createCertShouldTrimIdentityFieldsLikeUpdateDoes`.

## Verification

Red, before the fix (`mvn -o -f server/pom.xml test -Dtest=K8sCertServiceTest`):

```
[ERROR] Tests run: 20, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1.845 s <<< FAILURE! -- in org.apache.rocketmq.studio.cluster.k8s.K8sCertServiceTest
Expecting actual throwable to be an instance of:
but was:
 but was: " new-tls-cert "
[ERROR]   K8sCertServiceTest.createCertShouldRejectBlankIdentityFields:214
Expecting actual throwable to be an instance of:
but was:
[ERROR]   K8sCertServiceTest.createCertShouldTrimIdentityFieldsLikeUpdateDoes:234
[ERROR] Tests run: 20, Failures: 2, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

Green, after the fix (`mvn -o -f server/pom.xml test -Dtest=K8sCertServiceTest,K8sCertControllerTest`):

```
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 6.850 s -- in org.apache.rocketmq.studio.cluster.k8s.K8sCertControllerTest
[...]

PR #4350, opened 2026-09-15
fix(k8s): normalize certificate identities on create like on update

### LINKED ISSUE #4299: [Studio][Bug] Creating a K8s certificate accepts blank or padded identities that the update API rejects
## 1. Symptom

`POST /api/k8s-certs/create` accepts certificate identities that `POST /api/k8s-certs/update` refuses,
and stores them verbatim:

- a whitespace-only `issuer` (`"  "`) or `k8sId`/`cluster` is accepted and persisted without trimming,
  while the same value sent to the update API is rejected with `400 Certificate issuer cannot be blank`;
- the stored value keeps its padding, so `k8sId` / `cluster` / `issuer` read back as `" new-tls-cert "`.

The bundled web dialog happens to send `values.k8sId.trim()` and `values.cluster.trim()`
(`web/src/pages/cluster/certs.tsx:104-105`), so the inconsistency only shows for any other API client -
and the server is the component that owns the contract.

## 2. Root cause

`server/src/main/java/org/apache/rocketmq/studio/cluster/k8s/K8sCertService.java:81-95`

```java
String issuer = command.getIssuer();
List<String> san = command.getSan();
...
K8sCertVO cert = K8sCertVO.builder()
        .k8sId(command.getK8sId())
        .cluster(command.getCluster())
        .type(type)
        .issuer(issuer)
```

`createCert` passes the submitted identities straight through, while `updateCert` routes the same three
fields through `normalizeOptionalIdentity` (`:120-122`, `:175-184`), which trims and rejects a blank
value. `CreateCertDTO`'s `@NotBlank` constraints accept surrounding whitespace, so they do not close the
gap either.

## 3. Impact

- A certificate row can be created in a state the update API cannot produce, so the two write paths
  disagree about what a valid identity is.
- Padded identities are stored and returned padded, which breaks later exact-match lookups and makes the
  inventory look like it contains duplicates (`"prod"` and `" prod"`).
- Fixing it afterwards requires an update call that a client written against the create contract may not
  expect to fail.

## 4. Reproduction

1. `POST /api/k8s-certs/create` with
   `{"k8sId": " new-tls-cert ", "cluster": " test-cluster ", "type": "TLS", "issuer": " vault "}`.
2. Response: 200 with `k8sId` `" new-tls-cert "` and `issuer` `" vault "`.
3. `POST /api/k8s-certs/update` with `{"id": <id>, "issuer": "  "}`: `400 Certificate issuer cannot be
   blank`.

## 5. Expected behaviour

- `createCert` normalises `k8sId`, `cluster` and `issuer` exactly like `updateCert`: trimmed, blank
  rejected with `400 Certificate <field> cannot be blank`.
- A value derived from a supplied `certPem` is unaffected: when a certificate is uploaded, its parsed
  issuer/SAN still win over the submitted ones.

89799969:
## What

Make `POST /api/k8s-certs/create` normalize `k8sId`, `cluster`, and `issuer` the same way `update` already does.

## Why

`createCert` stored submitted identities verbatim while `updateCert` trims and rejects blank values. Consequences:

- padded identities (`" new-tls-cert "`) are persisted and break exact-match lookups;
- whitespace-only `issuer` is accepted on create but rejected on update;
- a certificate row can exist in a state the update API cannot produce.

`CreateCertDTO` `@NotBlank` does not close the gap because it allows surrounding whitespace, and `issuer` is unconstrained.

## How

- Trim/reject blank `k8sId` and `cluster` via `normalizeOptionalIdentity`.
- Normalize submitted `issuer` when no PEM overrides it.
- PEM-parsed issuer/SAN still win when `certPem` is present.

## Tests

`K8sCertServiceTest`:

- `createCertShouldTrimIdentityFields`
- `createCertShouldRejectBlankIssuerWithoutPem`

Focused run: `mvn -Dtest=K8sCertServiceTest test` (JDK 21) — passed.

Fixes #4299


89799969:
Closing: this change is already covered by merged #4300 (normalize certificate identities on create like on update, Fixes #4299). Thanks for the review window — no second patch needed.

PR #4357, opened 2026-09-15
fix(ai): validate message properties before preview confirmation

btlqql:
Message-send previews currently accept malformed properties JSON, issue a confirmation plan, and fail only when that plan is executed. Parse the properties during preview using the same parser as execution, so callers receive INVALID_ARGUMENT before asking for confirmation.

Regression coverage checks malformed JSON, a JSON array, and nested values in both preview and execute paths, and verifies that no message is sent.

Validation: `mvn -Dmaven.repo.local=D:/Dev/m2-cache -Dtest=MessageSendToolHandlerTest test` passed (14 tests); server sources and tests compiled successfully.


RockteMQ-AI (approved):
### Summary
Moves `parseProperties()` validation into the `preview()` path so malformed properties JSON fails with `INVALID_ARGUMENT` before the user is asked to confirm the plan — matching the same parser used during `execute()`. Clean fail-early fix with solid parameterized test coverage (malformed JSON, array input, nested values) and a `verify(never())` safety net.

LGTM.

---

lizhimins:
Validating the properties during the L2 preview rather than only at execution is the right place for the check.

PR #4358, opened 2026-09-15
fix(metrics): preserve mixed float and histogram samples

btlqql:
Prometheus matrix results can contain both float and native-histogram samples in the same series. The explorer previously retained only the float samples whenever any were present, dropping histogram rows from CSV and undercounting the response.

Preserve and sort both kinds, count them individually, show mixed types in the details table, and plot floats and histogram-derived values on separate lines. Add a mixed-series export/statistics regression and update the component regression to use distinct sample timestamps.

Reference: https://prometheus.io/docs/prometheus/latest/querying/api/#range-vectors

Validation: 33 tests passed across MetricsExplorer and metricsExplorerDiagnostics; ESLint passed for the implementation and utility tests; `npm run build` passed.


web/src/components/MetricsExplorer.tsx
```diff
@@ -122,14 +122,15 @@ const MetricChart = ({
...
+    .flatMap((series, index) => {
+      const { samples } = toMetricSeriesSamples(series);
+      // Keep raw floats and histogram-derived trends on separate lines.
+      return (['scalar', 'histogram'] as const).map((kind) => ({
         color: SERIES_COLORS[index % SERIES_COLORS.length],
         label: metricSeriesLabel(series, metric.name),
-        samples,
-        fromHistogram,
-      };
+        samples: samples.filter((sample) => sample.kind === kind),
```
RockteMQ-AI, line 139: After `flatMap`, a mixed series produces two entries in `allSeries` with the same `color` and `label`. The legend section (line ~247) iterates `chartSeries` and will render duplicate labels for mixed series. Consider either:
1. Deduplicating legend entries by `label` (showing the color swatch once), or
2. Appending a kind suffix like `"(scalar)"` / `"(histogram)"` to the legend text so users can distinguish the two trend lines.

This is a minor UX polish item — the data correctness is solid.

RockteMQ-AI (commented):
### Summary
This PR correctly fixes the Prometheus metrics explorer to preserve both float (scalar) and native-histogram samples when a series contains mixed types. Previously, `toMetricSeriesSamples()` silently dropped histogram data whenever scalar samples were present. The fix merges and sorts both kinds, splits them into separate trend lines in the chart, and properly represents mixed types in the detail table and CSV export.

### Findings
- **[Info]** `MetricsExplorer.tsx:131` — After `flatMap`, the legend will show duplicate labels for mixed series (see inline comment)

### What Looks Good
- The core logic change in `toMetricSeriesSamples()` is correct — merging both sample kinds and sorting by timestamp properly handles the Prometheus matrix format
- The `flatMap` approach to split mixed series into separate scalar/histogram trend lines is clean
- The `'mixed'` sample type detection in `buildMetricSeriesDetailRows` is well-reasoned
- Test coverage is thorough: the new test uses distinct timestamps (1, 2, 3) to verify sorting, and covers statistics, detail rows, and CSV export
- Changing the existing test's histogram timestamp to `1_800_000_001` (distinct from the scalar `1_800_000_000`) correctly validates that both types are now preserved

### Verification
- 33 tests passed (per PR description)
- ESLint passed
- `npm run build` passed

---

web/src/components/MetricsExplorer.tsx
btlqql: Addressed in a688f1c5. Mixed series now use distinct colors and explicit `(scalar)` / localized histogram suffixes, while single-kind series retain their existing labels. The component regression verifies both legend labels; 33 targeted tests and ESLint pass.

RockteMQ-AI (approved):
LGTM. Good fix for the mixed-series data loss.

Key changes look correct:
- `toMetricSeriesSamples` now merges both scalar and histogram samples into a single sorted array — this is the core fix and properly addresses the data-dropping bug.
- `flatMap` in the chart component cleanly splits mixed series into separate trend lines.
- Counting logic in `summarizeMetricData` correctly counts by `sample.kind` instead of the old `fromHistogram` flag.
- The 'mixed' sample type in the details table is a nice touch for visibility.
- Test coverage is solid — the regression test for mixed-series preservation is exactly what's needed.

No issues found.

---

PR #4359, opened 2026-09-16
[ISSUE #4354] Clear persisted Proxy cluster id when the input is cleared

### LINKED ISSUE #4354: [Studio][Bug] Clearing the Proxy cluster id filter is not persisted
## Problem

On the Studio Proxy page, the Cluster ID input is initialized from `localStorage.clusterId` and defaults to `DefaultCluster`. When an operator clears the input, `handleClusterIdChange` updates React state to an empty string but does not remove or overwrite the persisted value. A page reload then restores the stale cluster id, so the cleared filter is not durable.

## Current behavior

1. Enter a custom Cluster ID on the Proxy page.
2. Clear the Cluster ID input.
3. Reload the page.
4. The previously persisted Cluster ID is restored from localStorage.

## Expected behavior

Clearing the Cluster ID input should clear the persisted value too. After reload, the page should fall back to the default value instead of resurrecting the stale custom value.

## Design / implementation plan

- Keep the existing state behavior: the input may be temporarily empty while the operator edits it.
- When `handleClusterIdChange` receives a non-empty value, continue writing it to localStorage.
- When it receives an empty value, remove the `clusterId` localStorage entry instead of leaving the previous value behind.
- Add a focused frontend regression test around the Proxy page initialization and clear behavior.

## Scope

Frontend-only fix in the Proxy page/localStorage helper usage. No backend contract change.

Yanghanrui666:

Fixes #4354.

The Studio Proxy page persisted `clusterId` only when the input became non-empty. Clearing the input left the previous value in `localStorage`, so a reload restored the stale filter. This change removes the stored value when the input is cleared and preserves persistence for non-empty values.

## Brief changelog

- Remove `localStorage.clusterId` when the Proxy page Cluster ID input is empty.
- Add a regression test covering persisted-value removal and the `DefaultCluster` fallback after remounting.

## Verifying this change

- `npm test -- src/pages/studio/__tests__/Proxy.test.tsx` — 15 passed.
- `npm run build` — passed.
- `npm run lint` — 0 errors (existing repository warnings only).


RockteMQ-AI (approved):
LGTM. Clean fix — removing the stale `localStorage.clusterId` when the input is cleared is the correct behavior, and the regression test properly verifies the `DefaultCluster` fallback after remount.

---

PR #4363, opened 2026-09-16
fix(studio): add explicit ESCAPE clause to LIKE searches

jokerzsd:
Resubmit of #4236 per @lizhimins's review, adding the explicit `ESCAPE` clause so the escaping is dialect-independent.

## Summary

User-supplied search strings in topic/group metadata search and data source search were passed into MyBatis-Plus `like()`, so `%`, `_` and `\` were treated as SQL wildcards/escape characters. The prior fix escaped the value but relied on MySQL's implicit backslash escape character, which is wrong on H2 (no default LIKE escape character).

## Fix

- Add a shared `SqlLikeUtils` helper (`escape` + `contains` + the `ESCAPE` clause constant).
- Replace the five `.like()` predicates with `.apply(..., "name LIKE {0} ESCAPE '\'", SqlLikeUtils.contains(search))` (four in `RocketMQMetadataProvider`, one in `MybatisPlusSettingsRepository.findDataSources`), so the backslash is explicitly the escape character on every dialect.

## Tests

- `RocketMQMetadataProviderTest.searchEscapesLikeWildcardsAndAddsExplicitEscapeClause`: asserts the wrapper contains `LIKE`/`ESCAPE` and the bound parameter is `%prod\_user%` for both topic and group search.
- `MybatisPlusSettingsRepositoryTest.findDataSourcesEscapesSearchWildcardsAndAddsEscapeClause`: same assertion for the data source search.

## Verification

`mvn test -Dtest=RocketMQMetadataProviderTest` → 42 passed.
`mvn test -Dtest=MybatisPlusSettingsRepositoryTest` → 11 passed.

Closes #4236.


RockteMQ-AI (approved):
### Summary

This PR fixes a SQL LIKE injection issue in RocketMQ Dashboard (studio) by adding an explicit `ESCAPE` clause to LIKE predicates, making the escaping dialect-independent (works correctly on both MySQL and H2).

**Review verdict: LGTM** ✅

The implementation is clean and well-structured:
- `SqlLikeUtils` utility class correctly escapes `\`, `%`, and `_` in the right order
- The `ESCAPE '\\'` clause is explicitly attached to each LIKE predicate
- Comprehensive test coverage for both metadata provider and settings repository
- Null/empty input handling is correct throughout

No issues found.

---

PR #4364, opened 2026-09-16
fix(studio): escape LIKE wildcards in every Studio search endpoint

123123213weqw:

Every Studio search box passes its value to MyBatis-Plus unmodified, where `like(column, value)` wraps it as `%value%`. The caller's own `%` and `_` therefore keep their pattern meaning:

- `search=%` matches every audit row (`operator`, `resource_name` or `detail` `LIKE '%%'`).
- `search=_` matches every row whose column is a non-empty string.
- `search=100%_done` matches any row that merely contains `100` followed by four arbitrary characters.

The result set widens instead of narrowing, which is the opposite of what an operator typing those characters is asking for (SQL text, `a_b` style ids, wildcard-ish names). On the audit page a full-table match also pushes the CSV export over its 10 000-record cap and fails it with a 400 instead of returning the filtered subset.

This is one functional domain — "a search term must match literally" — so it is submitted as one change rather than one patch per endpoint, following the regrouping asked for on #4288 / #4290.

## Brief changelog

- Add `common/util/SqlLikeUtils`, which escapes `\`, `%` and `_` with a backslash so a `LIKE` pattern matches literally. `null` and blank values are returned unchanged, so a `null` that also disables the condition still disables it.
- Apply it once per value at the boundary where a request parameter first becomes a query condition:
  - `AuthService` user search
  - `MybatisPlusAclRepository` user search and rule search (`principal`, `resource`)
  - `MybatisPlusAlertRepository` rule search and the name/metric search of `findRulesPage`
  - `MybatisPlusAuditRepository.applyFilters`, which the list, the summary, the insights and the export all share
  - `MybatisPlusInstanceRepository.search` and `findByTypeAndSearch`
  - `MybatisPlusCloudCredentialRepository.findPage`
  - `MybatisPlusSettingsRepository.findDataSources`
  - `RocketMQMetadataProvider` topic and group name searches
- `QueryHistoryService` already escaped its own private copy of this rule. It now delegates to the shared helper, so the repository keeps one implementation instead of two.

The backslash is the default `LIKE` escape character on MySQL and on H2 in `MODE=MySQL` (the dev profile), so no explicit `ESCAPE` clause is needed and the behaviour is the same on the production database and in dev.

## Verifying this change

- `server/src/test/java/.../common/util/SqlLikeUtilsTest.java` — the helper escapes each wildcard, the escape character itself and their combinations, and leaves plain and blank values alone.
- `server/src/test/java/.../common/util/SqlLikeEscapingAgainstDatabaseTest.java` — runs the escaped pattern through a real H2 database in `MODE=MySQL` and asserts the literal match, next to the unescaped pattern that over-matches: `%` matches 1 row escaped vs 5 rows unescaped, `100_done` matches 1 vs 3. This is the engine-level confirmation that the backslash really is the escape character, which a wrapper assertion alone cannot show.
- Five repository test classes gained a case asserting that the value bound by the repository is the escaped one.

Red/green, with the fix reverted and the tests kept (baseline `e5ac5244`):

```
RED   findPageShouldEscapeLikeWildcardsInTheSearchTest
        Expecting ["%100%_done%"] to contain only ["%100\%\_done%"]
GREEN Tests run: 78, Failures: 0, Errors: 0, Skipped: 0
```

`cd server && mvn -B test` runs 2426 tests with 0 failures. 17 errors remain in the Spring context tests (`StudioApplicationTest`, `HealthProbeIntegrationTest`, `RmqAlertStateMapperIntegrationTest` and siblings); they fail the same way on the untouched baseline because they need a reachable MySQL, so they are unrelated to this change. `checkstyle` runs as part of the build (it is what caught the first draft's unused imports). `apache-rat` is not configured in this module.

## Note on overlap

This covers the settings and metadata-provider searches as well, which #4363 also touches with an explicit `ESCAPE` clause rather than escaping in the value. Happy to drop those two files from this change, or to switch to the `ESCAPE`-clause variant, whichever approach you prefer for those files — I did not want to leave the domain half-escaped.

## Checklist

- [x] A Github issue is filed for the change: #4289 (the audit search), of which the remaining endpoints are the same defect.
- [x] Pull request title formatted as `fix(scope): ...`.
- [x] Unit tests added for the helper, the database behaviour and the repository wiring.
- [x] `mvn clean install -DskipITs` equivalent (`mvn -B test`, `mvn -B -DskipTests compile`) passes locally.
- [ ] Apache ICLA not needed for a change of this size.


RockteMQ-AI (approved):
### Summary
Reviewed PR #4364: fix(studio): escape LIKE wildcards in every Studio search endpoint

The changes look reasonable. No critical issues identified.

### Observations
- 17 files changed
- Test coverage: test files included ✓

---

PR #4368, opened 2026-09-16
[ISSUE #4367] feat(studio): query consumer lag through proxy

### LINKED ISSUE #4367: [Studio][Feature] Query gRPC consumer lag through RocketMQ Proxy
## Problem

RocketMQ 5.x gRPC/POP consumers may appear in broker-side `ConsumeStats` with a negative
`brokerOffset - consumerOffset`. The Studio currently routes that value through
`ConsumerLagResolver`, but the fallback interface has no request context and the only implementation
is `NoopProxyStatsProvider`, which always returns `UNKNOWN`.

As a result, the consumer-group list, consumer-group queue progress, and topic-consumer list can only
report the lag as unavailable even though the already-discovered RocketMQ Proxy remoting endpoint can
answer the authoritative offset requests.

This is distinct from:

- #4293, which fixes how an unknown lag is presented by the topic-consumer API.
- #4361 / #4362, which distinguish failed live statistics from a real zero.
- #2504, which added Proxy consumer discovery but intentionally left lag transport as a no-op.

### Context-aware lag contract

Change `ProxyStatsProvider` so a lookup receives the selected `instanceId`, `consumerGroup`, and the
already-known `MessageQueue`. This supplies the routing fields required by Proxy and avoids a second
topic-route scan.

### Proxy remoting requests

Implement the contract in the existing `ProxyConsumerResolver`, reusing its per-instance Proxy address
discovery/cache and its bounded 2-second remoting timeout.

For each discovered Proxy address, query both offsets for the same queue:

1. `GET_MAX_OFFSET` (`RequestCode` 30), with `topic`, `queueId`, `bname`/broker name, and committed
   offset semantics.
2. `QUERY_CONSUMER_OFFSET` (`RequestCode` 14), with `consumerGroup`, `topic`, `queueId`, and
   `bname`/broker name. Do not request a fabricated zero when an offset is absent.

Decode the standard `GetMaxOffsetResponseHeader` and `QueryConsumerOffsetResponseHeader`. Return the
non-negative difference only when both responses are successful and complete. If a Proxy is
unreachable or returns an incomplete/non-success response, try the next discovered Proxy; return
`ConsumerLagResolver.UNKNOWN` if none can produce a complete pair.

### Consumer paths

Pass queue context through every current `resolveDiff` call so the real Proxy fallback is used by:

- consumer-group list enrichment;
- consumer-group queue-progress detail;
- paged topic-consumer statistics.

Known broker-side non-negative lag remains authoritative and must not trigger Proxy requests.

## Failure semantics

- A real offset difference of zero remains `0`.
- Missing headers, non-success response codes, timeouts, unreachable Proxies, or negative Proxy offset
  differences remain `UNKNOWN` (`-1`).
- Partial data must never be summed with known queues or converted to zero.
- No new route traversal or unbounded retry loop is introduced.

## Test plan

- Protocol-vector tests for both request codes and all required custom-header fields.
- Known positive and known-zero lag responses.
- Fallback to a later Proxy after an unreachable/non-success response.
- Null, malformed, partial, timeout, and negative-difference responses degrade to `UNKNOWN`.
- Metadata-provider tests prove the selected instance, group, and queue are passed in the group list,
  group progress, and topic-consumer paths.
- Existing broker-known lag behavior remains unchanged and does not call Proxy.
- Run focused provider tests, the full backend test suite with Java 21, Checkstyle/build, and frontend
  tests/build/lint because the affected values are rendered by existing UI paths.

Aias00:
## Summary

- make Proxy lag lookup queue-aware by carrying `instanceId`, consumer group, and `MessageQueue`
- query the existing Proxy remoting endpoints with `GET_MAX_OFFSET` and `QUERY_CONSUMER_OFFSET`
- use the real Proxy result in consumer-group list/detail, queue progress, and topic-consumer statistics
- preserve `UNKNOWN` for unreachable, incomplete, malformed, or inconsistent responses instead of fabricating zero

## Implementation

`ProxyConsumerResolver` now implements `ProxyStatsProvider` and reuses its existing per-instance Proxy
address discovery/cache and 2-second request timeout. For an already-known queue it sends:

- `GET_MAX_OFFSET` with `topic`, `queueId`, `bname`, and committed offset semantics
- `QUERY_CONSUMER_OFFSET` with `consumerGroup`, `topic`, `queueId`, `bname`, and
  `setZeroIfNotFound=false`

Both responses must be successful, contain an offset, and produce a non-negative difference. A failed
Proxy is skipped in favor of the next discovered address; if no address can return a complete pair,
the existing `ConsumerLagResolver.UNKNOWN` sentinel is retained. Broker-reported non-negative lag
still bypasses Proxy entirely.

No topic-route scan or new dependency is introduced.

Fixes #4367

## Test plan

- [x] Java 21 focused provider suite: 126 tests passed
- [x] protocol-vector coverage for request codes, topic, queue id, broker name, group, and timeout
- [x] positive, zero, fallback, partial, malformed, and negative-difference coverage
- [x] list/detail/progress/topic-consumer context propagation coverage
- [x] Java 21 package build and Checkstyle
- [x] frontend production build
- [x] frontend lint: 0 errors (10 pre-existing warnings)
- [x] independent code review: no findings
- [x] `git diff --check`

### Full-suite baseline notes

- Backend H2/MySQL-compatible run: 2422/2423 passed. The only failure is the pre-existing
  `AuthServiceSessionOverviewIntegrationTest` second-boundary assertion (`59` vs `>= 60`); the same
  failure reproduces on clean `origin/rocketmq-studio@e5ac52440`, while the class passes 3/3 alone.
- Frontend run: 1033/1034 passed. The only failure is the pre-existing whole-suite timeout in
  `ConsumerPage.test.tsx` (`keeps the latest client stack when an older request resolves last`); that
  file passes 32/32 alone, and this PR changes no frontend files.


RockteMQ-AI (approved):
LGTM. Code changes look good.

---

PR #4370, opened 2026-09-16
fix(message): keep abbreviated properties on code-point boundaries

### LINKED ISSUE #4287: [Studio][Bug] An abbreviated message property can end with a lone surrogate

89799969:
## Summary

Fixes #4287

`MessagePropertyDisplay.abbreviate` capped values with a plain `substring(0, MAX_PROPERTY_VALUE_CHARS)`. The cap counts UTF-16 chars, so a cut between a surrogate pair produced a lone high surrogate in API responses (message explorer / DLQ drawer).

Walk back to a code-point boundary before cutting, matching `CredentialUtils.mask` and the web-side fix in #4226.

## Testing

- `MessagePropertyDisplayTest#limitPropertiesShouldNotSplitSupplementaryCharactersTest` (new)
- `MessagePropertyDisplayTest` full suite: 6 passed locally (`mvn -o test -Dtest=MessagePropertyDisplayTest`)

## AI disclosure


RockteMQ-AI (approved):
### Summary
Good fix for property value truncation at code-point boundaries. Using `Character.isLowSurrogate` to avoid splitting surrogate pairs is the correct approach, and the test covers both the BMP and supplementary character cases.

LGTM — clean and well-tested.

---

89799969:

89799969:
--delete-branch=false

89799969:
Closing: already fixed on `master` by #4467 (`fix(message): abbreviate message properties on code point boundaries`), which covers the same surrogate-split cut and also fixes `hasOversizedProperty` to count code points (with `offsetByCodePoints` for the cut). Merging this PR would only re-introduce the old `MAX_PROPERTY_VALUE_CHARS` approach.

Thanks for the careful check on the UTF-16 boundary — the landed fix follows the same idea with a slightly broader scope.

PR #4371, opened 2026-09-16
fix(topic): mark unknown consumer lag as metrics unavailable

### LINKED ISSUE #4293: [Studio][Bug] The topic consumer list reports an unknown lag as a backlog of -1

89799969:
## Summary

Fixes #4293

When any queue reports `ConsumerLagResolver.UNKNOWN`, `getTopicConsumersPage` published `diffTotal=-1` while leaving `metricsAvailable=true`. The topic page then rendered ??? as `-1` instead of ???.

Set `metricsAvailable(false)` when the sentinel is used, consistent with the exception path in the same method.

## Testing

- Extended `RocketMQMetadataProviderTest#getTopicConsumersKeepsUnknownWhenAnyQueueLagIsUnknown`
- `RocketMQMetadataProviderTest` + related alert/util tests: 55 passed locally

## AI disclosure


RockteMQ-AI (approved):
### Summary
Correct fix — when `ConsumerLagResolver.UNKNOWN` sentinel is encountered, setting `metricsAvailable(false)` ensures the UI renders the "metrics unavailable" indicator instead of displaying `-1`. This is consistent with the exception handling path in the same method.

LGTM — minimal, targeted fix with proper test coverage.

---

89799969:

PR #4372, opened 2026-09-16
fix(alert): match padded stored instance ids in active state filters

### LINKED ISSUE #4297: [Studio][Bug] A rule with a padded stored instance id fires but reports no active alert state

89799969:
## Summary

Fixes #4297

A rule whose stored `instanceId` has surrounding whitespace fires via `NativeAlertRuleScopeMatcher` (trims) but was dropped from `MybatisPlusAlertStateRepository.findActive` and `NativeAlertProcessor.reconcileMissingActiveStates`. Operators saw "no active alert" while FIRING rows kept being written.

Trim the stored instance id in both filters, matching the metric filter on the previous line.

## Testing

- `MybatisPlusAlertStateRepositoryTest#findsActiveStatesForRulesWithPaddedStoredInstanceIdTest` (new)
- `MybatisPlusAlertStateRepositoryTest`: 6 passed locally

## AI disclosure


RockteMQ-AI (approved):
### Summary
Correct fix for the inconsistency between `NativeAlertRuleScopeMatcher` (which trims) and the query/reconcile paths (which didn't). The trim is now applied consistently in both `findActive` and `reconcileMissingActiveStates`, matching the metric filter pattern on the preceding line.

One minor observation: in `NativeAlertProcessor.java`, the original code used `rule.getInstanceId() == null` as the guard, while the fix changes it to `!StringUtils.hasText(rule.getInstanceId())`. This is a slight behavioral broadening (also treats empty string as "no instance filter"), but it's consistent with the `MybatisPlusAlertStateRepository` change and is the safer choice — an empty `instanceId` would never match a real scope anyway.

LGTM — well-tested with a clear reproduction scenario.

---

89799969:

89799969:
Closing: already fixed on `master` by #4464 (`fix(alert): match padded instance ids and render alert windows and placeholders`). That commit trims `rule.getInstanceId()` in both `MybatisPlusAlertStateRepository.findActive` and `NativeAlertProcessor`'s reconcile filter, which is the same padded-instance-id match this PR targeted.

No need to re-land; the root fix is already on trunk.

PR #4384, opened 2026-09-16
feat(deliveries): retry every failed delivery matching the active filters

### LINKED ISSUE #3561: Notification delivery retry only targets the current page
## Problem\n\nThe delivery page exposes a retry action named for the current page. If a filter matches failed deliveries across several pages, the operator has to retry each page separately.\n\n## Current behavior\n\n
etryVisibleFailures collects failed rows from items, which contains only the current page, and calls the bulk retry endpoint with those IDs.\n\n## Expected behavior\n\n- Add a backend action that retries failed deliveries matching the current filters, rather than requiring an explicit page-local ID list.\n- Bound the action to a safe maximum number of rows.\n- Return succeeded and failed counts so the UI can report the outcome.\n- Keep the existing page-local retry action.\n\n## Why this matters\n\nAfter a webhook outage, hundreds of deliveries can be failed across pages. Retrying the filtered result set is the operation operators actually need; paging manually is error-prone.

89799969:
Fixes #3561.

## Problem

The Notification Deliveries page only retries FAILED rows loaded on the **current page**. After a webhook or SMTP outage, hundreds of deliveries can be FAILED at once while the page size is 20, so the operator has to page through the feed and retry each page separately.

## Change

- Add `POST /api/system-alerts/deliveries/retry-filtered` that selects FAILED deliveries matching the optional `channel` / `instanceId` filters, ordered by delivery id, bounded to `limit` (default 100, valid 1–100).
- Reuse the existing per-delivery retry path (`retryFailedDelivery` → status reset + audit) and the existing bulk result envelope.
- Add a "Retry matching failures" action on the deliveries page that uses the current channel/instance filters. The current-page retry stays unchanged.
- Empty match short-circuits without touching the outbox.

## Verification

```bash
cd server && mvn -B -ntp test -Dtest=NotificationOutboxServiceTest
cd web && npm test -- --run src/pages/ops/__tests__/NotificationDeliveriesPage.test.tsx
```

- Java: 28/28 passed, including the three new cases (filtered id-order retry, limit bounds, empty short-circuit).
- Frontend: 4/4 passed, including `retries every failed delivery matching the current filters`.


RockteMQ-AI (approved):
### Summary
Adds bulk retry for failed notification deliveries filtered by channel/instance — useful after webhook outages when failures span many pages. Clean implementation with proper input validation (limit 1-100), deterministic ordering, and good test coverage including edge cases (empty results, invalid limits).

LGTM.

---

89799969:

PR #4385, opened 2026-09-16
[ISSUE #4367] feat(studio): query consumer lag through proxy

### LINKED ISSUE #4367: [Studio][Feature] Query gRPC consumer lag through RocketMQ Proxy

Aias00:
## Status

Withdrawn after maintainer review.

## Reason

The proposed lag fallback cannot demonstrate an independent benefit on RocketMQ Proxy 5.5.0.
`ConsumerManagerActivity` handles `GET_MAX_OFFSET` and `QUERY_CONSUMER_OFFSET` by forwarding the
requests to the selected Broker through `request(ctx, request, context, 3000)`. It does not read a
Proxy-local lag or committed-offset source.

The fallback runs only after Studio already obtained the Broker `ConsumeStats` table. Reissuing the
same queue offset requests through Proxy would therefore add two remoting calls per queue without a
more authoritative source. The previous tests used intentionally different mocked Broker and Proxy
offsets, so they were not valid before/after evidence for the real topology.

## Review resolution

- Bounded concurrency and an absolute deadline were prototyped locally, but are not being shipped
  because they only bound an unjustified fan-out.
- The overlap with #4387 is removed: #4387 retains ownership of Ops runtime connection settings and
  Proxy discovery cache scoping.
- `NoopProxyStatsProvider` and the proposed Proxy offset contract are not needed.
- Unfetchable lag remains the existing `UNKNOWN` value; #4386 owns its external representation.
- Issue #4367 now records the upstream limitation and the evidence required before implementation
  can resume.

No branch update is pushed because there is no technically valid production change to retain.


RockteMQ-AI (approved):
### Summary
Adds consumer lag querying through the Proxy instead of direct Broker access — better architecture for environments where Studio can't reach Brokers directly. Includes proper error handling and test coverage.

LGTM.

---

lizhimins (changes requested):
Requesting changes. The blocker is that the added lookups are unbounded and
synchronous: querying backlog has to be concurrent, with a budget.

1. **Bounded concurrency.** `resolveDiff` is called per queue, and each negative
   diff now issues two synchronous remoting round-trips with a 2 s timeout, retried
   across every discovered proxy address. `listConsumerGroups` and
   `getTopicConsumersPage` iterate groups x queues, so one HTTP request can block
   for minutes on a large POP group. Please resolve the queues concurrently with an
   explicit bound (max in-flight per request and/or an overall deadline), dedupe by
   `(group, topic)` so a shared topic is resolved once, and state the cap in the
   Javadoc. Also define the degradation: what the caller sees when some queues
   resolve and others time out.
2. **Show the benefit.** `ConsumerManagerActivity` forwards `GET_MAX_OFFSET` and
   `QUERY_CONSUMER_OFFSET` straight to the broker, so the proxy path returns the
   same offsets `examineConsumeStats` already produced. A before/after capture of a
   case where the broker path yields `-1` and the proxy path yields a real number
   would justify the extra fan-out; without it we are adding latency and RPC volume
   for no measurable gain.
3. **Overlap with #4387.** Both PRs rewrite `ProxyConsumerResolver` and they cannot
   both land. #4387 is pending a separate product-direction decision, so please say
   which parts of the resolver you actually need.
4. **Dead fallback.** `NoopProxyStatsProvider` has no production caller left after
   this change — either delete it (and inline the sentinel in the test) or keep it
   wired as the fallback for when `proxyConsumerResolver` is absent.
5. **Rebase** onto `rocketmq-studio` (the branch was rebuilt on 2026-09-20 and
   carries #4606); we retarget the PR base on our side when merging.

The tests themselves are in good shape — nine new methods, all named `...Test`, and
they fail when the change is reverted. Note that how an unfetchable lag should be
represented is being settled on #4386, so please do not widen that surface here.

Aias00:
Withdrawing this PR based on the review findings.

RocketMQ Proxy 5.5.0 does not expose an independent lag source here:
`ConsumerManagerActivity` forwards `GET_MAX_OFFSET` and `QUERY_CONSUMER_OFFSET` to the same
Broker. Since the fallback starts only after `examineConsumeStats` has already read Broker offsets,
the extra calls cannot justify their latency or RPC volume without a real deployment capture.

I prototyped the requested 8-way bounded concurrency, shared three-second deadline, queue
deduplication, and partial `UNKNOWN` degradation locally. The focused suite passed 129 tests, but
that only proves the fan-out can be bounded; it does not prove the fan-out is useful. The synthetic
`broker=-1 / proxy=50` fixture is not acceptable benefit evidence because both paths ultimately
read the same Broker.

Issue #4367 has been revised to document the upstream constraint and reopening criteria. #4386 keeps
ownership of unavailable-lag representation, and #4387 remains independent Ops/runtime work.

PR #4386, opened 2026-09-16
fix(studio): distinguish unavailable consumer stats from zero lag

### LINKED ISSUE #4361: [Studio][Bug] Consumer group live stats failures are rendered as zero lag
## Resolution

Closed as not planned following maintainer review.

## Product decision

The consumer-group list and detail surfaces intentionally keep the primitive `0` defaults when live
consumer statistics cannot be fetched. They will not introduce a second unavailable/unknown
representation driven by `consumeStatsAvailable` or `consumptionTimestampAvailable`.

This decision is scoped to the consumer-group surface:

- `ConsumerGroupVO.totalLag` remains `0` when the live stats lookup fails, returns null, or has an
  empty offset table.
- The consumer-group table, detail view, sorter, and real/mock CSV exports continue using their
  numeric values.
- The existing `ConsumerLagResolver.UNKNOWN` and `metricsAvailable` contract remains unchanged on
  the Apache queue/topic lag aggregation paths.
- The `delaySeconds` sorter retains the null-safe `?? 0` fallback already present on the rebuilt
  `rocketmq-studio` branch.

## Why the proposed fix is not landing

The proposed change added a consumer-group-specific availability representation across the backend
VO, TypeScript contract, renderers, sorters, and two export paths. That would overlap with the
existing queue/topic unknown-lag contract and require two manually synchronized definitions of
unavailable data.

If this distinction is revisited, it should be a repository-wide proposal with one resolver, one VO
contract, and one renderer/export policy shared across consumer-group, topic-consumer, queue
progress, and CSV surfaces.

PR #4386 is withdrawn because reverting the rejected availability behavior leaves no independent
production change relative to the current `rocketmq-studio` baseline.

Aias00:
## Status

Withdrawn following the maintainer's product-direction decision.

## Resolution

The consumer-group list and detail surfaces will continue reporting the existing numeric defaults
when live statistics are unavailable. This PR's `consumeStatsAvailable` /
`consumptionTimestampAvailable` driven rendering, sorting, and CSV behavior is therefore not
landing.

After reverting that direction:

- `ConsumerGroupVO.totalLag` keeps its `0` initial value;
- table/detail views and real/mock CSV exports remain numeric;
- queue/topic lag continues using the existing `ConsumerLagResolver.UNKNOWN` /
  `metricsAvailable` contract;
- the rebuilt `rocketmq-studio` branch already contains the required
  `(a.delaySeconds ?? 0) - (b.delaySeconds ?? 0)` sorter fallback;
- the six non-conforming new test methods disappear with the rejected behavior, so there are no
  remaining test names to rename.

Rebasing and reverting the rejected behavior leaves no independent production diff. No branch update
is pushed.

Issue #4361 is closed as not planned. A future attempt must propose one repository-wide unavailable
statistics contract shared by all group/topic/queue renderers and exports.


RockteMQ-AI (approved):
LGTM. Good bug fix that correctly distinguishes unavailable consumer stats from zero lag.

**Key improvements:**
- Preserves lag as unknown until broker offset table is successfully read
- Uses existing `consumeStatsAvailable` contract consistently across list/detail rendering, sorting, and exports
- Proper handling of edge cases: failed, null, empty, and unknown statistics render as unavailable
- Valid zero lag remains as zero (correct behavior)

Fixes #4361 with comprehensive test coverage on both backend and frontend.

---

lizhimins (changes requested):
Requesting changes: one direction decision from us, plus two mechanical items.

**Direction — when consumer stats cannot be fetched, report 0.**

We do not want an "unavailable / unknown" state on the consumer group surface. So
`ConsumerGroupVO.totalLag` should keep its `0` initial value, and
`consumeStatsAvailable` should not become the signal driving the table, the detail
drawer, the sorter and both CSV export paths.

To be explicit that this is a decision rather than an oversight: the trunk does
carry `ConsumerLagResolver.UNKNOWN` and `metricsAvailable` on the Apache queue-lag
aggregation path (#1652, #3988), and the topic page renders that as `-`. That
contract stays where it is. What we do not want is a second, overlapping
representation of "no data" on the consumer group list and detail — the two would
have to be kept consistent by hand, and the group surface is exactly where an
operator reads backlog numbers during an incident. A `0` there is at worst
understated; a `-` in one column next to a number in another is ambiguous.

If you believe the distinction is worth having, the way to reopen it is a single
repository-wide representation (one resolver, one VO field, one renderer) rather
than a per-page one. That is a reasonable proposal and we would look at it
separately.

**Mechanical items**

1. The `delaySeconds` sorter dropped the `?? 0` fallback. `undefined - number`
   yields `NaN`, which makes the antd ordering nondeterministic. Please restore it —
   this one is independent of the direction above.
2. Six new test methods do not end with `Test`, which is the project convention;
   compare #4385 from the same batch, which complies. Please rename.
3. Please rebase onto `rocketmq-studio` — the branch was rebuilt on 2026-09-20 and
   carries #4606, so `web/src/api/metadata.ts` and `web/src/pages/instance/consumer.tsx`
   need a rebase check. We retarget the PR base on our side when merging.

One thing you no longer need to worry about: the other PR fixing this same defect
shares five files with yours and is not going to land, so there is nothing to
coordinate.

Aias00:
Withdrawing this PR according to the review direction.

The core change is the consumer-group-specific unavailable state, and that behavior was explicitly
rejected. Reverting it restores the current `0` contract across backend defaults, table/detail
rendering, sorting, and both CSV paths. The current `rocketmq-studio` baseline already has the
null-safe delay sorter fallback, so no standalone fix remains.

The six new test methods that lacked the `Test` suffix belong only to the rejected behavior and are
removed with it. Issue #4361 now records the product decision and is closed as not planned. A future
proposal would need one repository-wide unavailable-data contract rather than a per-page one.

PR #4352, opened 2026-09-15
[Studio] Persist and apply Ops runtime connection settings

Aias00:
## Summary
- Make Ops manage Studio default Apache RocketMQ NameServer selection, VIP channel, and TLS only when explicitly enabled; leave instance-scoped endpoints and credentials untouched.
- Persist the separate `ops-connection` settings row and read the current configuration for each default-client operation. Validate addresses, protect the active/last address, and enforce admin rights at service level even when login enforcement is disabled.
- Apply VIP/TLS before admin, producer, or pull-consumer startup. TLS enabling requires secure (non-test) server verification and a readable deployment-owned trust certificate; no secrets are persisted.
- Use one immutable selection for endpoint metadata and the actual client call; retire obsolete managed clients without shutting down in-flight admin/send operations, and key Proxy discovery cache by the selected default NameServer.
- Keep the feature disabled by default and document Compose/JVM setup.

Fixes #4348

## Verification
- Java 21, isolated H2 MySQL-compatible datasource: `cd server && env JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home PATH=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home/bin:/opt/homebrew/bin:/usr/bin:/bin SPRING_DATASOURCE_URL="jdbc:h2:mem:ops_runtime_lease_full_suite;MODE=MySQL;DATABASE_TO_LOWER=TRUE;DB_CLOSE_DELAY=-1" SPRING_DATASOURCE_USERNAME=sa SPRING_DATASOURCE_PASSWORD= SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.h2.Driver mvn -B -ntp -q test` -> exit 0; 2421 tests, 0 failures/errors, Checkstyle 0 violations.
- Focused provider suite -> 180 tests passed; cache/proxy/lease regression suite -> 40 tests passed; default-cluster MCP credentials -> 23 tests passed.
- `cd web && npm test -- --run src/pages/studio/__tests__/Ops.test.tsx` -> 6 passed; `npm run build` -> passed; `npm run lint` -> 0 errors, 10 existing warnings in unchanged files.
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config --format json` -> Ops environment values resolve; custom default/override NameServer variants verified.
- `git diff --check` -> clean.

[...]

PR #4387, opened 2026-09-16
[Studio] Persist and apply Ops runtime connection settings

### LINKED ISSUE #4348: [Studio][Feature] Persist and apply Ops RocketMQ connection settings at runtime
## Problem

`OpsService` hardcodes `configurationAvailable=false` and its five mutation endpoints return HTTP 501. This is deliberate fail-closed behavior: earlier simulated writes reported success without changing RocketMQ clients. The Ops page already disables controls when the backend reports the configuration unavailable, but operators cannot manage Studio's default Apache RocketMQ connection.

There are separate Spring-bound defaults (`studio.cluster.admin.namesrv-addr` and `studio.rocketmq.namesrv-addr`), a persisted NameServer registry, and long-lived admin/producer/consumer client caches. Editing the registry or a bean alone is not an effective runtime connection update. VIP and TLS are client transport settings, not NameServer server settings.

## Scope and ownership

- Manage **Studio's default Apache RocketMQ connection** only. Do not rewrite instance-specific endpoints, credentials, NameServer processes, or the NameServer registry.
- Keep `studio.cluster.admin` credentials and TLS key/trust material in deployment-owned configuration; never persist or reveal secrets via Ops.
- Preserve the existing `/api/ops/*` routes and `OpsHomeVO` envelope. Existing auth interceptor restricts POST writes to admins.
- The feature is disabled by default until deployment explicitly enables runtime management; when disabled or the backing configuration cannot be loaded, keep `configurationAvailable=false` and fail closed.

## Proposed design

1. Add a distinct persisted Ops runtime configuration (separate key from general settings) for an allowlisted NameServer address list, selected default address, VIP mode, and TLS mode. Bootstrap from externalized defaults without overwriting them; record the active source and define how a persisted override is applied on startup and across Studio replicas. Reject blank/malformed/duplicate addresses, selecting a non-member, deleting the active or last address, and invalid payloads.
2. Provide one effective default-connection resolver consumed by both default cluster discovery/admin providers and the default producer/consumer paths. Do not alter instance-scoped provider routing or credential identity.
3. Apply `ClientConfig.setNamesrvAddr`, `setVipChannelEnabled`, and `setUseTLS` **before** a new `DefaultMQAdminExt`/producer/consumer starts. On a runtime change, ensure the next operation does not reuse a client created with the previous address/transport settings. Avoid disrupting another instance that happens to use the old address; key/evict caches with that ownership in mind.
4. Reject enabling TLS unless the deployment explicitly disables RocketMQ's insecure TLS test mode and supplies a server-authentication/trust policy. RocketMQ 5.5 defaults to insecure trust in test mode; a Boolean stored by Ops must not silently make an unauthenticated TLS connection. No TLS certificate/key contents in DB, logs, audit, or responses.
5. Return current effective values in `GET /api/ops/homePage`. All five POST routes change effective behavior (or return an actionable 4xx/501), never just a stored/UI value. Record administrator audit entries without secrets. Keep the UI's existing capability gate; PR #4154 adds client-side preflight and is complementary, not a backend implementation.

## Implementation and verification

- Introduce a small Ops configuration repository/model and runtime resolver; use the existing settings storage pattern without changing general/LLM settings.
- Wire both default NameServer configuration sources and client creation to the same effective resolver. Validate updates before committing; use cache generation or targeted release so transport settings actually change.
- Unit/integration tests: bootstrapped and persisted reads, five write routes, validation/conflicts, active/last address protection, disabled/DB-failure behavior, VIP/TLS before-start application, insecure TLS rejection, cache change without affecting instance-specific connections, admin access and audit.
- Run targeted backend tests, Checkstyle, full backend suite in an isolated datasource, and focused frontend tests/build/lint when UI behavior changes.

## Acceptance criteria

- With explicit deployment enablement, a successful Ops write is reflected by `homePage` and by subsequent real default-client configuration; restart restores persisted values.
- Without enablement or secure TLS prerequisites, writes fail closed and do not mutate the effective client state.
- Existing instance-specific endpoints/credentials and NameServer registry CRUD remain unchanged.
- No mutation reports success while the client continues using the old address/VIP/TLS settings.

## Out of scope

- Provisioning, restarting, or editing NameServer server processes; that is a separate deployment-control-plane issue.
- Creating Kubernetes/Docker/SSH integrations in Studio.
- Persisting TLS key/cert/trust material or RocketMQ admin credentials in Studio settings.

Aias00:
## Summary
- Make Ops manage Studio default Apache RocketMQ NameServer selection, VIP channel, and TLS only when explicitly enabled; leave instance-scoped endpoints and credentials untouched.
- Persist the separate `ops-connection` settings row and read the current configuration for each default-client operation. Validate addresses, protect the active/last address, and enforce admin rights at service level even when login enforcement is disabled.
- Apply VIP/TLS before admin, producer, or pull-consumer startup. TLS enabling requires secure (non-test) server verification and a readable deployment-owned trust certificate; no secrets are persisted.
- Use one immutable selection for endpoint metadata and the actual client call; retire obsolete managed clients without shutting down in-flight admin/send operations, and key Proxy discovery cache by the selected default NameServer.
- Keep the feature disabled by default and document Compose/JVM setup.

Fixes #4348

## Verification
- Java 21, isolated H2 MySQL-compatible datasource: `cd server && env JAVA_HOME=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home PATH=/Users/aias/Library/Java/JavaVirtualMachines/openjdk-21.0.2/Contents/Home/bin:/opt/homebrew/bin:/usr/bin:/bin SPRING_DATASOURCE_URL="jdbc:h2:mem:ops_runtime_lease_full_suite;MODE=MySQL;DATABASE_TO_LOWER=TRUE;DB_CLOSE_DELAY=-1" SPRING_DATASOURCE_USERNAME=sa SPRING_DATASOURCE_PASSWORD= SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.h2.Driver mvn -B -ntp -q test` -> exit 0; 2421 tests, 0 failures/errors, Checkstyle 0 violations.
- Focused provider suite -> 180 tests passed; cache/proxy/lease regression suite -> 40 tests passed; default-cluster MCP credentials -> 23 tests passed.
- `cd web && npm test -- --run src/pages/studio/__tests__/Ops.test.tsx` -> 6 passed; `npm run build` -> passed; `npm run lint` -> 0 errors, 10 existing warnings in unchanged files.
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config --format json` -> Ops environment values resolve; custom default/override NameServer variants verified.
- `git diff --check` -> clean.

The lease tests first failed because a runtime update shut down admin/producer clients while actions were still in progress, then passed after deferred retirement. The Proxy cache test first returned the previous cluster addresses after switching NameServer, then passed with endpoint-scoped keys.

## CI note
GitHub Actions is currently `startup_failure` before creating jobs for this PR and other `rocketmq-studio` PRs. Local clean-suite results above are the available validation evidence.

## Scope note
PR #4154 adds complementary Ops frontend preflight, not the backend runtime application implemented here. No frontend source files are changed in this PR.

---

Replaces #4352 because its base branch (`rocketmq-studio`) was deleted. This PR uses the same head commit and targets `master`.


RockteMQ-AI (approved):
LGTM. Well-architected implementation for persisting and applying Ops runtime connection settings.

**Key observations:**
- Clean separation of concerns: `OpsConnectionRepository` interface with MyBatis-Plus implementation
- Proper resource management with `ClientLease` pattern for connection lifecycle
- Good use of records for immutable data (`OpsConnectionSettings`, `AdminClientCacheKey`, `ClientKey`)
- Audit logging for connection setting changes
- Address validation and protection of active/last address
- Admin rights enforcement at server level

The PR is large but well-organized with clear boundaries between components. Test coverage is comprehensive.

---

lizhimins (changes requested):
Thanks for the work here — the pooling refactor itself looks carefully built (the `ClientLease` reference counting and the in-flight retirement test are solid). We cannot take this as a single 40-file / +2628 change, for five reasons.

**1. It reverses a deliberate product decision, so it needs a direction call first.**
The five `/api/ops` endpoints intentionally fail closed today (`OPS_SETTINGS_UNAVAILABLE`). Making the default RocketMQ connection runtime-editable is a product-direction change, not a bug fix, and the linked issue #4348 was filed together with this PR rather than reporting a community problem. We need to decide whether the default connection should be editable at runtime before any implementation lands.

**2. It runs against the instance-model spec.**
Our instance-model convention states that `instanceId` is a required parameter with **no default-injection design**. This PR invests further in the default (no-instance) path — persisting and managing runtime settings for it — which is the direction the spec is moving away from.

**3. It rewrites a layer other open PRs depend on.**
The change restructures the pooled-client layer mandated by our client-lifecycle convention (`MqAdminExtFactory` / `MqClientPool` cache key and value shape) and changes five provider constructor signatures. Roughly ten other open PRs touch the same hot files (`RocketMQMetadataProvider`, `RocketMQAdminClientImpl`, `ProxyConsumerResolver`); merging this first breaks all of them, and this is the only PR in the queue that changes those constructors.

**4. Deploy-side changes are out of scope for this PR.**
`server/src/main/resources/application.yml` (new `studio.ops.runtime.enabled` / `max-name-server-addresses`), `deploy/docker-compose.yml` (four new environment variables including `JAVA_TOOL_OPTIONS`), `deploy/.env.example` and `deploy/README.md` (TLS system properties and certificate mount guidance) are a separate deployment topic and belong in their own PR.

**5. Conventions.**

- `RocketMQDefaultClusterResolver`, `MybatisPlusOpsConnectionRepository` and `OpsService` drop `@RequiredArgsConstructor` in favour of hand-written constructors; our convention is constructor injection via `@RequiredArgsConstructor`.
- `MybatisPlusOpsConnectionRepository.toJson` throws a bare `RuntimeException("Failed to save Ops runtime settings", exception)`; we standardise on `BusinessException(<numeric code>, msg)`.
- None of the 55 new test methods end with `Test` (e.g. `defaultConnectionAppliesTransportBeforeStartWithoutSharingInstanceClient`, `loadShouldReturnEmptyWhenOpsConnectionSettingsDoNotExist`). Test method names must end with `Test`.
- 5 of the 6 commit subjects are missing a Conventional Commits type prefix (`Keep default Apache operations on one runtime snapshot`, `Cover consistent managed discovery endpoints`, …).

**What we would like instead** — please split into three PRs:

1. the `ClientLease` reference-counted pooling refactor on its own (pure refactor; it needs a design review before it lands);
2. Ops connection-settings persistence plus the service implementation, **excluding** everything under `deploy/` and `application.yml`;
3. the deploy / compose / README configuration.

And before (1), please tell us in the PR whether you believe the default connection should be runtime-editable at all, given the fail-closed design and the instance-model spec — if the answer is no, (2) and (3) fall away and only the refactor is worth pursuing.

Aias00:
Withdrawing this PR after evaluating the requested direction call.

The default RocketMQ connection should remain fail-closed and non-editable at runtime under the current instance model. The model requires explicit `instanceId` routing with no default-injection design, so continuing to persist and apply a mutable no-instance connection would deepen a path the project is moving away from. The Ops persistence and deploy/configuration portions will not be split into follow-up PRs.

The concurrency-safe part remains independently useful when `InstanceService` releases clients after an endpoint or credential update. I extracted that into design issue #4755 and focused PR #4761:

- 5 files only: `ClientLease`, `MqAdminExtFactory`, `MqClientPool`, and focused tests
- no Ops persistence, default routing, provider constructor changes, application configuration, or deploy files
- full backend verification: 3079 passed, 3 skipped; independent concurrency re-review approved

Issue #4348 is being closed as not planned for the same instance-model reason.

PR #4345, opened 2026-09-15
[ISSUE #4143][ISSUE #4147] feat(ai): align message query paging contracts

Aias00:
## Summary

Reimplements #4148 as a separate PR from the latest `rocketmq-studio` baseline, following the maintainer review without rewriting the original PR history.

- add nested `page: {page, pageSize}` using the existing typed `PageRequest` contract to both `rmq.message.query` and `rmq.message.query_by_topic`; default to 1/20 and cap tool requests at 100 in the catalog schema
- reuse `MessageService.queryMessagesPage(...)` for message-ID, business-key, and Topic queries; preserve the direct unique-key lookup and wrap its 0..1 result in the same paging contract
- serialize the established common `PageOutput` fields as `{page, pageSize, total, items}`, with `resultMayBeTruncated` as message-specific metadata
- add `includeBody` to both query tools, default false; omit `body`, `bodyEncoding`, and `bodyTruncated` unless explicitly requested
- preserve identifier priority `msgId`, then `uniqueKey`, then `key`
- add optional `traceTopicName`; non-blank values use the existing four-argument trace service while blank/absent values keep the legacy three-argument path
- keep the existing tool count, L1 permissions, REST/service APIs, and provider implementations unchanged

## Verification

- TDD red: the rewritten contract tests initially failed because `MessageQueryOutput`, nested paging, and `traceTopicName` did not exist
- focused handler/catalog/output contract tests: 25 passed
- MessageService, REST controller, MCP registration/discovery/invocation, message handlers, catalog, and output-schema tests: 73 passed
- full server suite with `-Dspring.profiles.active=dev`: 260 test classes, 2361 tests, 0 failures/errors/skips
- `mvn -B -ntp -DskipTests package`: passed, Checkstyle 0 violations
- `git diff --check`: passed
- independent read-only review: no findings

Supersedes #4148

Closes #4143
Closes #4147

PR #4388, opened 2026-09-16
[ISSUE #4143][ISSUE #4147] feat(ai): bound message queries and support custom traces

### LINKED ISSUE #4143: [Studio][Feature] Allow the AI message trace tool to use a custom trace topic

### LINKED ISSUE #4147: [Studio][Performance] Bound AI message query results

Aias00:

- Closes #4143
- Closes #4147
- Supersedes #4148 and #4345

### Brief Description

Align the AI/MCP message tools with the broker capabilities and the maintainer review:

- replace the `page/pageSize` contract with one `limit`, defaulting to 20 and capped server-side
  at 100;
- return `items`, `resultMayBeTruncated`, and `skippedCount` instead of advertising random page
  access that broker scans cannot provide;
- preserve provider truncation signals and mark results truncated when the local limit omits rows;
- keep `includeBody=false` by default to bound MCP payload and model-context usage;
- explicitly document that default body omission is a breaking output-contract change: existing
  callers that read `items[].body` must send `includeBody=true`;
- document the intentional family difference: `rmq.message.query_by_offset` still always includes
  body fields because it returns at most one message;
- preserve identifier priority `msgId`, then `uniqueKey`, then `key`;
- add optional `traceTopicName`, using the existing four-argument trace service for non-blank
  values and the legacy path otherwise;
- regenerate and commit `rmqctl/internal/catalog/catalog_gen.go`, including its updated digest.

The branch is rebased onto the latest `master`.

- TDD red: the revised handler/output tests failed against the paging implementation because
  `limit`, flat bounded results, and `skippedCount` did not exist.
- Focused message handler/catalog/output/trace suite: 23 tests passed.
- Full server suite: 2421 tests ran; 2420 passed.
  - The only failure was the unrelated existing
    `AuthServiceSessionOverviewIntegrationTest` 59-vs-60-second timing boundary; an isolated rerun
    reproduced the same boundary.
- `cd server && JAVA_HOME=<jdk21> mvn -B -ntp -DskipTests package`
  - Checkstyle reported 0 violations; compilation, JAR packaging, and Spring Boot repackaging passed.
- `cd rmqctl && make test`
  - All Go packages passed.
- `cd rmqctl && make verify`
  - Race tests, formatting, `catalog-verify`, `go vet`, and CLI build passed.
- `git diff --check` passed.

### Checklist

- [x] Commit subjects follow Conventional Commits
- [x] Tests cover default/custom/capped limits, body behavior, truncation, skipped rows, and traces
- [x] Breaking body-output behavior is explicit in this description and both YAML tool descriptions
- [x] The offset-query body behavior is explicit in its YAML description
- [x] Generated rmqctl catalog and digest are committed
- [x] New or renamed Java test methods end with `Test`


RockteMQ-AI (approved):
LGTM. Well-structured pagination implementation for the AI message query tools.

**Highlights:**
- Clean use of typed `PageRequest` contract with sensible defaults (page=1, pageSize=20)
- Proper cap at 100 in the catalog schema to prevent excessive queries
- Backward-compatible constructors preserve existing callers
- `includeBody` flag gives callers control over payload size
- New `MessageQueryOutput` record provides proper pagination metadata

The code is clean and the test coverage is solid.

---

lizhimins (changes requested):
Requesting changes. The problem you found is real, but the paging contract is not the
fix we want, because deep pagination is not something the broker can actually honor.

**Why not `page` / `pageSize`**

Message query on the broker side has no cursor for arbitrary page jumps. A topic scan
walks from the beginning and stops at a bound (`DEFAULT_TOPIC_LIMIT` in the provider),
so "page 37, size 20" is not a seek — it is a rescan plus a slice, and beyond the scan
bound the page is simply empty. Publishing `page` + `total` in an MCP tool schema
advertises random access that the store cannot deliver: an LLM caller will happily ask
for page 500, and the honest answer is "that does not exist", not "here is an empty
page". `MessageQueryOutput.fromUniqueKey` re-implementing the `(page - 1) * pageSize`
slice is a symptom of the same mismatch.

**What we do want**

The genuine defect you identified — `MessageItem.from` unconditionally attaching
`body` / `bodyEncoding`, so one call can push hundreds of message bodies (each up to
the 4 MB fragment limit) into a single MCP response and blow up both the model context
and the gateway payload — should be fixed with a **bound plus an explicit truncation
marker**, not pagination:

1. A single `limit` (with a documented server-side cap) instead of `page` / `pageSize`.
2. Keep `resultMayBeTruncated` and surface how many were skipped, following the
   precedent already in the repo: `DlqExportHeaders`
   (`X-DLQ-Export-Truncated` / `X-DLQ-Export-FailedQueues` / `X-DLQ-Export-Limit`) and
   `DLQExportResultVO.truncated`, plus #4185 for Apache key queries.
3. `includeBody` defaulting to false is fine as a direction, but it **is** a breaking
   change to the MCP output contract for `rmq.message.query` and
   `rmq.message.query_by_topic` — please say so explicitly in the description and in
   the yaml `description`, since existing callers that read `items[].body` will start
   getting nothing.
4. If you keep the family inconsistent on purpose (`rmq.message.query_by_offset` still
   returns `ListOutput<MessageItem>` with the body always attached), document that in
   the yaml descriptions; otherwise callers cannot tell which tool needs
   `includeBody=true`.

**Hard blocker regardless of the above**

`server/src/main/resources/tool-catalog/tools/message.yaml` is modified but
`rmqctl/internal/catalog/catalog_gen.go` is not regenerated. That file is
`// Code generated by cataloggen. DO NOT EDIT.` and carries a `Digest` over the catalog;
`rmqctl/Makefile` hooks `catalog-verify` into `vet` / `build` / `build-all` / `lint`,
and `.github/workflows/ci.yml` runs it as a dedicated gate. So the Go side fails on
drift. Please run `cd rmqctl && make catalog-generate`, commit the regenerated
`catalog_gen.go` with the new digest, and confirm `make catalog-verify` passes locally
— the Verification section currently only lists Maven and frontend, which suggests this
gate was not known. Note that CI does not currently run for PRs based on
`rocketmq-studio`, so this would otherwise land silently.

Aias00:
Addressed the requested contract direction in `327c199f0`:

1. Removed `page/pageSize`; both query tools now accept one `limit` with default 20 and a
   server-side cap of 100.
2. Responses retain `resultMayBeTruncated` and add `skippedCount`, the exact number omitted by
   the local limit from the provider-bounded result.
3. Both YAML descriptions explicitly identify `includeBody=false` as a breaking output change and
   tell existing `items[].body` callers to set `includeBody=true`.
4. `rmq.message.query_by_offset` now explicitly documents that it always includes body fields and
   has no `includeBody` flag.
5. Regenerated `rmqctl/internal/catalog/catalog_gen.go` with the updated digest.

Verification: focused Java tests pass 23/23; all Go packages pass; `make verify` passes race,
format, catalog drift, vet, and CLI build; Maven package passes with 0 Checkstyle violations. The full
server run passed 2420/2421 tests, with only the unrelated existing 59-vs-60-second session timing
boundary, which reproduced in isolation.

PR #4391, opened 2026-09-16
fix(alert): acknowledge the event and active state in one transaction

### LINKED ISSUE #4203: [Studio][Bug] Alert acknowledgement can diverge from active alert state

89799969:

- Fixes #4203

### Brief Description

`AlertService.acknowledgeAlert` persisted two related writes — the system-alert event ACK and the matching FIRING/REMINDER active-state ACK — without a transaction boundary. If the second write failed, the request failed but the event stayed acknowledged while the episode stayed FIRING, so the reminder loop kept notifying for an alert the UI already showed as acknowledged.

Mark the method `@Transactional` so both writes commit or roll back together. The existing behaviour for stale occurrences (repository returns false without changing a newer episode) is unchanged.

### How Did Yon Test This Change?

```
$ cd server && mvn -B -ntp test -Dtest=AlertServiceTest
```

New regression test `acknowledgeAlertShouldCommitEventAndActiveStateInOneTransactionTest` pins the `@Transactional` annotation (same reflection pattern as `MybatisPlusMetricSnapshotRepositoryTest`). Existing acknowledge tests still pass.

Focused run:
```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
```

Full `AlertServiceTest` and checkstyle are green locally.

### Checklist

- [x] No new UI text
- [x] No new source files
- [x] Documentation not required (internal persistence atomicity)


RockteMQ-AI (approved):
### Summary
Clean fix — adds `@Transactional` to `acknowledgeAlert()` so the event update and active-state ACK commit atomically, preventing a mid-method failure from leaving the alert in an inconsistent state (acknowledged event but still FIRING). Test verifies the annotation is present.

LGTM.

---

89799969:

PR #4392, opened 2026-09-16
fix(audit): escape LIKE wildcards in audit log search

### LINKED ISSUE #4289: [Studio][Bug] The audit log search treats % and _ as LIKE wildcards

89799969:

- Fixes #4289

### Brief Description

`GET /api/audit/logs`, `GET /api/audit/summary` and the export path pass the search box value to a SQL `LIKE` across `operator`, `resource_name` and `detail` without escaping. `%` and `_` therefore keep their wildcard meaning:

- `search=%` matches every audit row
- `search=_` matches every non-empty value
- a resource name containing `_` matches unintended rows

Add a shared `SqlLikeUtil` (escape + explicit `ESCAPE '\\'` clause, H2-compatible) and use it in `MybatisPlusAuditRepository.applyFilters`, the single filter builder shared by the paged list, summary and export.

```
$ cd server && mvn -B -ntp test -Dtest='SqlLikeUtilTest,MybatisPlusAuditRepositoryTest'
BUILD SUCCESS
```

New tests:
- `SqlLikeUtilTest` pins escape order and the explicit ESCAPE clause
- `findPageShouldEscapeLikeWildcardsInSearchTest` asserts the bound pattern is `%100\%\_done%`
- `summarizeShouldEscapeLikeWildcardsInSearchTest` asserts the summary path uses the same escaped pattern

### Checklist

- [x] Tests added or updated, test methods named `...Test`
- [x] No new UI text
- [x] Architecture constraints stay green
- [x] Documentation not required (search semantics fix)


89799969:
Closing in favour of #4458, which covers the audit LIKE escape as part of the full Studio search domain (auth, ACL, alert, audit, instance, credentials) and adds the explicit ESCAPE CHAR(92) clause this PR did not. Keeping both open only splits review.

PR #4457, opened 2026-09-16
fix(alert): preserve state without orphaning disabled rules

### LINKED ISSUE #4326: [Studio][Bug] Updating or toggling an alert rule always wipes its FIRING/ACK state
### Component

Studio / Alerting (`AlertService`)

### Describe the bug

Every successful `updateRule` and `toggleRule` call does `alertStateRepository.deleteByRuleId(id)` unconditionally. That wipes the live FIRING / ACK / PENDING episode for the rule even when the evaluation identity did not change.

### Steps to reproduce

1. Have a native alert rule in FIRING state and ACK it (or leave it firing).
2. Edit only the rule **name** or **description**, or disable and re-enable the rule.
3. Observe `rmq_alert_state` rows for that rule are gone.

### Expected behavior

State is cleared only when evaluation conditions change (metric, threshold, operator, duration, scope fields in `AlertRuleSemanticFingerprint`). Cosmetic edits and enable/disable keep the current episode and ACK status.

### Actual behavior

- Renaming a rule loses the active episode; the next evaluation can re-fire and re-notify.
- Toggling enabled never keeps ACK history, so a temporary mute-and-restore looks like a brand-new alert.

### Severity

High — operator-visible re-fire storms and lost acknowledgements on a core Studio alerting path.

### Related

Semantic fingerprint already exists at `AlertRuleSemanticFingerprint` and is used for duplicate detection; reuse it for state retention.

zmuxuny:

Fixes #4326.

### Revised state policy

This revision addresses the maintainer review: **disable -> purge; enable -> preserve; semantic change -> purge; presentation-only edit of an enabled rule -> preserve**. Deleting a rule still purges its state. The policy covers both toggle overloads, both bulk paths, and both edit entrypoints.

Disabled rules are excluded from evaluation/reconciliation. Their FIRING runtime and reminder timestamps must therefore be explicitly purged, not left for reconciliation. No artificial recovery event is emitted.

The enable-path assertions now intentionally encode enable-preserves. The bulk disable-and-delete test retains two purge assertions. The presentation-only regression no longer disables the rule as part of the edit.

### Regression evidence

Rebased onto `rocketmq-studio@4c697f07acde460e2344375cb1f82669f5b270fd`.

`AlertRuleStateLifecycleTest` uses a stateful repository fixture and the runtime query consumed by the alerts page: six disable paths, six enable paths, two semantic edits, two presentation-only edits, and four delete paths. Disable cases first observe FIRING with notification/reminder timestamps and then require an empty runtime view.

Before the correction, the original PR reapplied to this trunk failed exactly the six disable cases (20 tests, 6 failures). After correction: 20/20 passed.

Java 21: `AlertRuleStateLifecycleTest,AlertServiceTest,AlertRuleSemanticFingerprintTest,NativeAlertProcessorTest` **130/130 passed**; Checkstyle **0 violations**; `mvn -B -ntp -DskipTests package` **BUILD SUCCESS**; `git diff --check` clean.

No frontend/schema/dependency changes. Full repository integration tests requiring an external datasource were not run. AI-assisted implementation and review; all validation listed above was executed locally.


RockteMQ-AI (approved):
### Summary
Clean fix for alert state preservation when non-semantic changes are made to alert rules. The approach of computing a semantic fingerprint and only resetting state when it changes is correct and well-implemented.

**Key observations:**
- The fingerprint correctly excludes volatile fields (id, gmtCreate, gmtModified) while including semantic fields (ruleName, expression, enable, etc.)
- The test coverage is thorough: verifies state preservation on non-semantic changes and state reset on semantic changes
- The frontend change (removing the unconditional `state: ""` reset) correctly aligns with the backend behavior

**One minor note:**
- The `Objects.equals` comparison on the fingerprint string is clean and handles nulls correctly

This is a focused, well-tested fix that addresses the reported issue without over-engineering.

---

lizhimins (changes requested):
Thanks — the defect here is real and independently reported (#4326 was filed by another contributor), and the core idea is right.

**What we verified.** On trunk, `AlertService` clears evaluation state unconditionally on every rule mutation: `updateRule`, both `toggleRule` overloads, `bulkToggleRules` and `bulkUpdateRules` all call `alertStateRepository.deleteByRuleId(id)`. So renaming a rule, changing its severity or swapping a notification channel wipes a live FIRING/PENDING episode and its acknowledgement. Gating the reset on `AlertRuleSemanticFingerprint` for the update path is the correct fix for that. Your new `updatingOnlyNonSemanticFieldsShouldPreserveEvaluationStateTest` is mutation-sensitive — removing the fingerprint guard turns it red. One thing worth saying explicitly: this is **not** a duplicate of #4391. That PR adds `@Transactional` to `acknowledgeAlert` in a different region of the same file, `git merge-tree` is clean, and both can land.

**Blocking items.**

1. **Dropping `deleteByRuleId` from both toggle directions orphans state.** Every consumer of active alert state filters on `isEnabled()`: `NativeAlertProcessor.eligible()`, `NativeAlertProcessor.reconcileMissingActiveStates()` and `MybatisPlusAlertStateRepository.findActive()` all exclude disabled rules. A rule switched off while FIRING therefore leaves a state row that is never evaluated again and never reconciled away. Today the toggle→delete call is the only thing that cleans it up; removing it in both directions means the row is permanent.

2. **The orphan is user-visible.** `AlertService.listRuleRuntime()` delegates to `findRuntimeByRuleIds(listRules(domain))`, and neither side filters on `enabled`. The alerts page renders a runtime column for every row (`web/src/pages/ops/alerts.tsx`, the `alerts.runtime` column matching `state.ruleId === record.id`), so a switched-off rule will keep displaying FIRING, `lastNotifiedAt` and `nextReminderAt` indefinitely.

3. **The same hole comes through the update path, which makes the PR's own rule inconsistent.** `enabled` is not part of `AlertRuleSemanticFingerprint` — the fingerprint covers domain, instance, metric, operator, threshold, duration, aggregation, window, broker, cluster, group, topic and consecutive samples, but not the enabled flag. So turning a rule off in the edit form goes through `updateRule`, leaves the fingerprint unchanged, and preserves state exactly like the toggle does. Please make the invalidation policy explicit rather than incidental: disable → resolve or purge, enable → preserve, semantic change → purge. Each of the three needs its own test.

4. **Existing assertions that encode the current contract were weakened.** Two `verify(...)` calls became `never()`, one `times(2)` became `times(1)`, and the trunk test `domainBulkOperationsShouldResetEachMutatedRuleStateTest` was renamed to `domainBulkToggleShouldPreserveStateWhileDeleteRemovesItTest`. Those names and expectations record a deliberate decision, so flipping them is a contract change that needs to be argued in the description, not absorbed silently. Please keep the reset-on-delete assertions intact and add positive coverage for the disabled-rule runtime view instead of only inverting expectations.

5. **No test covers the regression.** There is nothing in the PR that would fail if a disabled rule kept a FIRING state row and the runtime column kept showing it. Given items 1 and 2, that is the case that most needs pinning down.

**Optional, not blocking.** `updateRule(AlertDomain, AlertRuleVO)` already loads the rule, so the extra `findRuleById` on the domain-scoped path is a wasted round-trip. Also, the fingerprint is already persisted on the entity (`MybatisPlusAlertRepository#setSemanticFingerprint`), so the previous value can be read rather than recomputed.

Once items 1 to 5 are addressed we can merge this. The branch is already based on `rocketmq-studio`, so no retargeting is needed.

zmuxuny:
Addressed the blocking lifecycle items in `9380d072` on `rocketmq-studio@4c697f0`.

The explicit policy is now disable -> purge, enable -> preserve, semantic edit -> purge, and presentation-only edit of an enabled rule -> preserve. Both toggle overloads, both bulk paths and both edit entrypoints are covered. Delete still purges; disabling does not emit a synthetic recovery.

A new stateful runtime regression first observes FIRING plus reminder timestamps and then requires the disabled rule's runtime view to be empty. All six disable variants failed against the previous implementation (20 lifecycle cases, 6 failures); the revised lifecycle suite is 20/20. The combined lifecycle/service/fingerprint/native-processor suite is 130/130, Checkstyle 0 and package succeeds. The PR body now explicitly explains the enable-direction contract change and the retained disable/delete assertions.

The optional duplicate lookup optimization is left separate to keep this revision focused. No full external-datasource integration run is claimed.

zmuxuny:
Hi @lizhimins, thank you for the detailed review. I pushed `9380d072` and updated the PR description with the disable/enable/edit/delete state policy and its regression coverage. Could you please take another look at the revised PR when convenient?

PR #4466, opened 2026-09-16
fix(studio): persist cleared optional instance and registry fields

### LINKED ISSUE #4273: [Studio][Bug] Clearing an instance admin credential reference has no effect

### LINKED ISSUE #4274: [Studio][Bug] Blanking a NameServer registry entry description, k8s namespace or k8s id has no effect

123123213weqw:

- Fixes #4273
- Fixes #4274

### Brief Description

Clearing an optional field had no effect. The update replaces every editable field of the entry, but MyBatis-Plus `updateById` omits null entity fields, so an omitted column silently kept its previous value even though the request submitted no value for it:

- **Instance**: a cleared `admin_credential_ref` survived the update. The column is now assigned explicitly when the request omits it, next to the existing "update was not applied" guard (#4273).
- **NameServer registry entry**: an omitted `k8s_namespace`, `k8s_id` or `description` kept its previous value. The registry update now assigns the omitted optional columns explicitly, so the stored entry matches what the request asked for (#4274).

This is one functional domain — "an omitted optional field means cleared, not unchanged" — so it covers both entry points in one change rather than one patch each.

It builds on #4276 and #4277, whose pull requests were closed when `rocketmq-studio` was folded into `master` and the base branch was deleted; #4276 is folded in here rather than resubmitted on its own. The changes are in `save()` and the registry update, not in the search paths, so this does not overlap the open LIKE-escaping change (#4458).

`topic remark cannot be cleared` (#4272) is the same class of defect but was already fixed by #4275, which is in `master`; it is deliberately not touched here.

```
cd server && mvn -B -ntp test
[INFO] Tests run: 2431, Failures: 0, Errors: 0, Skipped: 0
```

The 17 errors reported alongside are the `@SpringBootTest` classes that need a reachable MySQL 8 (`StudioApplicationTest`, `HealthProbeIntegrationTest`, `AuthService*IntegrationTest`, `RmqAlertStateMapperIntegrationTest`); they fail identically on the untouched baseline, and the bundled `deploy/docker-compose.yml` MySQL is the intended way to run them.

Test coverage, each with its red/green pair from the original branch:

- `MybatisPlusInstanceRepositoryTest` — an update that clears the admin credential reference writes `NULL`; against the unfixed code the stored reference survives and the assertion fails.
- `NameserverRegistryServiceTest` — an update that omits the k8s namespace, k8s id and description writes `NULL` for each, and an update that supplies them still stores the supplied values.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` — not applicable, no frontend change
- [x] Documentation touched where behaviour changed — none of these behaviours is documented


RockteMQ-AI (approved):
### Summary
Fixes a persistence bug where clearing optional fields (like `admin_credential_ref`, `k8s_namespace`, `description`) had no effect because MyBatis-Plus `updateById` omits null entity fields. Now uses `UpdateWrapper` to explicitly set null values.

### Review
- **Correctness** ✅ — `clearOmittedOptionalColumns()` explicitly sets null for cleared fields using `UpdateWrapper.set(column, null)`; the stored entry now matches the request
- **Consistency** ✅ — Same pattern applied to both `NameserverRegistryService` and `MybatisPlusInstanceRepository`; instance and registry updates behave identically
- **Tests** ✅ — New tests verify that clearing optional fields actually persists the null values
- **Compatibility** ✅ — No API changes; persistence behavior only

### Code Quality
Clear documentation explaining the MyBatis-Plus behavior and why the workaround is needed. The `anyCleared` flag avoids unnecessary UPDATE queries when no fields are being cleared.

LGTM — important fix for configuration management.

---

PR #4467, opened 2026-09-16
fix(message): abbreviate message properties on code point boundaries

### LINKED ISSUE #4167: [Studio][Bug] Preserve Unicode code points when truncating message properties

### LINKED ISSUE #4287: [Studio][Bug] An abbreviated message property can end with a lone surrogate

123123213weqw:

- Fixes #4287
- Fixes #4167

### Brief Description

`MessagePropertyDisplay` caps a property value at `MAX_PROPERTY_VALUE_CHARS` and marks the cut with `...`. The cap counted UTF-16 `char`s and the cut was a plain `substring`, so for a value holding a supplementary character at the boundary the result ended with an isolated high surrogate: the panel showed a replacement glyph at the end of an already abbreviated value, and the JSON body carried an invalid escape (`\uD83D` with no low surrogate).

A `char` is not a code point. The cap is now counted in code points and the cut uses `offsetByCodePoints`, which always lands on a code point boundary, so a supplementary character is kept whole or dropped whole and is never split.

**The cap is applied in both `abbreviate` and `hasOversizedProperty`**, and the class javadoc and the constant's javadoc now say code point, which is what both methods mean. This is the part @lizhimins asked for when closing the earlier variants of this fix on #4288:

> please cover `hasOversizedProperty` and the javadoc in the same change — a fix that only pins `abbreviate` still lets an oversized value fall back to `value.length()` and split a surrogate pair

and the code-point reading is the one he called the better reading of the constant's name. Counting chars in `hasOversizedProperty` while `abbreviate` counts code points would make the panel report a value as abbreviated that it leaves untouched, and the reverse.

### Supersedes #4370

#4370 fixes the cut in `abbreviate` only and leaves `hasOversizedProperty` comparing `value.length()`, so it is the "only pins `abbreviate`" variant the review asked not to resubmit. It is authored by another account of mine, so I can neither push to that fork nor close it from here: rather than leave two patches on the same file I am resubmitting the complete change here, and #4370 can be closed once this lands (I have no way to close it myself). Say the word if you would rather keep #4370 and I will rebase this onto it instead.

There is no web-side counterpart to change: `web/src/utils/messageProperties.ts` parses property text and does not re-truncate, so the server value is what the panel shows.

```
cd server && mvn -B -ntp test
[INFO] Tests run: 2430, Failures: 0, Errors: 0, Skipped: 0

mvn -B -ntp test -Dtest=MessagePropertyDisplayTest
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

The 17 errors reported next to the full run are the `@SpringBootTest` classes that need a reachable MySQL 8; they fail identically on the untouched baseline.

Test coverage:

- `limitPropertiesShouldNotSplitASurrogatePairTest` — the cap lands inside the emoji; asserts the result holds no unpaired surrogate and keeps the emoji whole.
- `limitPropertiesShouldKeepASupplementaryCharacterThatFitsTheCapTest` — the boundary case just inside the cap.
- `hasOversizedPropertyShouldCountCodePointsNotCharsTest` — a value of **1024 code points but 1624 chars**: `hasOversizedProperty` is false and `limitProperties` returns it untouched, which is what the char-based count got wrong; one more code point and both flip.
- The existing ASCII tests still pass unchanged (`"x".repeat(1500)` still abbreviates to 1024 chars plus `...`).

### Checklist

- [x] Documentation touched where behaviour changed — the javadoc of the constant and the class now state code points


RockteMQ-AI (approved):
### Summary
Fixes a Unicode handling bug where abbreviating message properties could split supplementary characters (emoji, CJK extension) into isolated surrogates. Now uses code point boundaries for both length counting and truncation.

### Review
- **Correctness** ✅ — `offsetByCodePoints()` ensures truncation lands on a code point boundary; supplementary characters are kept whole or dropped whole
- **Consistency** ✅ — `hasOversizedProperty` now uses `codePointCount()` to match `abbreviate()` behavior; the UI will not falsely claim a value is abbreviated
- **Tests** ✅ — Comprehensive tests cover supplementary characters at boundaries, mixed BMP/supplementary content, and null/empty edge cases
- **Compatibility** ✅ — No API changes; internal rendering utility only

### Code Quality
Excellent documentation explaining the UTF-16 vs code point distinction. The `codePointCount()` helper centralizes the logic. The Javadoc on `MAX_PROPERTY_VALUE_CHARS` is updated to reflect code points.

LGTM — good internationalization fix.

---

PR #4002, opened 2026-09-06
fix(client): treat offline group connections as empty results

unbridled-41:
Fixes #4006.

## Problem / Evidence

`GET /api/clients` fails with HTTP 502 "Failed to query consumer connections from all groups" whenever every non-system consumer group of the instance is offline — a completely normal state (consumer applications stopped, dev/test cluster, nightly window) — instead of returning an empty connection list. Because the consumer scan runs after the producer scan in the same request, the producer connections already collected are discarded too.

The same misclassification hits the explicit producer-group lookup: `GET /api/producer/connection?...&producerGroup=G` returns 502 "Failed to query producer connections: Not found the producer group connection" when group `G` has no online producers.

Client bytecode (rocketmq-tools 5.5.0, the version pinned by `server/pom.xml`), verified via `javap -c` on `DefaultMQAdminExtImpl`:

- `examineConsumerConnectionInfo(String)`: when `ConsumerConnection.getConnectionSet().isEmpty()` it throws `new MQClientException(206, "Not found the consumer group connection")` (bytecode offsets 90–126).
- `examineProducerConnectionInfo(String, String)`: throws `new MQClientException("Not found the producer group connection", null)` for an empty connection set.

`RocketMQClientProvider.findConsumerConnections` (`server/src/main/java/org/apache/rocketmq/studio/provider/apache/RocketMQClientProvider.java:313-334`) counts those per-group exceptions as scan failures, so "all groups offline" is indistinguishable from "all brokers broken" and trips the guard at line 332. The provider's own convention says normal "nothing here" states must be empty, not 502 (existing comment at `:175-177` and the `isTopicNotExist` handling).

## Root cause / Fix

The all-failed heuristic (attempted > 0 && successful == 0 → 502) treats the offline-group business answers as infrastructure failures. Fix:

- In the consumer scan, an `MQClientException` with response code 206 or message "Not found the consumer group connection" counts as a *successful* per-group query that found nothing, so an all-offline cluster yields an empty list; groups with real failures keep the partial-results behavior, and a scan where every group genuinely fails still returns 502.
- In the explicit producer-group query, the same offline-group message returns an empty list alongside the existing missing-topic handling.

## Priority & scoring

[...]

PR #4468, opened 2026-09-16
[ISSUE #4463] fix(producer): surface incomplete connection scans

### LINKED ISSUE #4463: [Studio][Bug] Producer connection scans silently return partial inventories
## Problem

`GET /api/producer/connection` supports a topic-only query by discovering every active producer
group and querying each group for its connections. When one Broker or group RPC fails while another
succeeds, the endpoint can return only the reachable subset.

Without completeness metadata, operators can mistake that partial inventory for the complete active
producer set. The ambiguity also exists during active-group discovery when one Broker is unreachable.

## Reproduction

1. Query producer connections for a topic without specifying `producerGroup`.
2. Let one Broker producer-table query or one active producer-group connection query fail while
   another succeeds.
3. The endpoint returns the reachable connections with no indication that part of the scan was
   omitted.

## Proposed design

Keep read-only diagnostics available during partial failures and make the coverage gap explicit:

- Return all reachable producer connections instead of failing the whole topic-wide request.
- Add `complete`, `failedBrokers`, and `failedProducerGroups` to the response.
- Add `INCOMPLETE_SCAN` to the summary warnings and report readiness as `WARNING`.
- Show the incomplete state and failed targets in the Producer UI.
- Continue returning HTTP 502 when every Broker query fails or every discovered producer-group
  connection query fails, because no meaningful scan result exists.
- Keep a missing topic route or offline producer group as a normal empty result.
- Keep explicit `producerGroup` query semantics and best-effort group suggestions unchanged.
- Use root-cause messages for actual query failures rather than nesting translated exception text.

This follows the existing Dashboard convention that read-only diagnostics preserve useful partial
results while publishing explicit completeness metadata.

## Test plan

- A topic-wide scan returns reachable connections plus `complete=false` when one Broker discovery
  query fails.
- A topic-wide scan returns reachable connections plus the failed group when one group RPC fails.
- All-Broker and all-discovered-group failures still return HTTP 502.
- Offline groups remain complete empty results.
- Controller/service/summary tests cover the response metadata and `INCOMPLETE_SCAN` warning.
- Frontend API and page tests cover metadata preservation and visible failed-target warnings.
- Run focused Java 21 tests, backend package/checkstyle, full frontend tests, frontend build/lint,
  Prettier, and `git diff --check`.

Aias00:

- Fixes #4463

### Brief Description

Make topic-wide producer diagnostics explicit about partial coverage without discarding reachable
results:

- return reachable producer connections when only part of Broker discovery or group querying fails;
- add `complete`, `failedBrokers`, and `failedProducerGroups` to the response;
- add the `INCOMPLETE_SCAN` summary warning and report incomplete scans as `WARNING`;
- show incomplete-scan and failed-target details in the Producer UI;
- keep HTTP 502 when all Brokers or all discovered groups fail;
- keep offline/missing groups as complete empty results and preserve explicit-group behavior;
- unwrap actual query failures with `rootMessage(...)`;
- align all new or renamed Java test methods with the `...Test` naming convention.

The API documentation and Issue #4463 design now describe the partial-success contract.

- Focused backend provider/service/controller/summary suite: 50 tests passed.
- `cd server && JAVA_HOME=<jdk21> mvn -B -ntp -DskipTests package`
  - Checkstyle reported 0 violations; compilation, JAR packaging, and Spring Boot repackaging passed.
- Full backend suite: 2420 tests ran; the only failure was the existing
  `AuthServiceSessionOverviewIntegrationTest` 59-vs-60-second timing boundary.
  - Three isolated reruns produced two passes and one failure, confirming the timing flake.
- `cd web && npm test -- --retry=2 --reporter=dot --silent`
  - 122 test files and 1036 tests passed.
- Focused frontend API/page suite: 2 test files and 26 tests passed.
- `cd web && npm run lint`
  - 0 errors; 10 pre-existing warnings outside the changed files.
- `cd web && npm run build`
  - TypeScript and Vite production build passed.
- Prettier check and `git diff --check` passed.

### Checklist

- [x] Tests added or updated for non-trivial changes; new/renamed test methods end with `Test`
- [x] Public API changes are documented in `docs/api-spec.md` and Issue #4463


RockteMQ-AI (approved):
### Summary
Fixes a data-integrity issue where partial failures in producer connection scans would silently return incomplete results. Now the aggregate connection query requires complete broker coverage and fails fast with HTTP 502 when any broker or group query fails.

### Review
- **Correctness** ✅ — `requireComplete` flag ensures the aggregate query path fails when any broker is unreachable, preventing false "no producers" results
- **Error Handling** ✅ — Individual group query failures now propagate immediately with clear error messages instead of being silently skipped
- **Backward Compatibility** ✅ — The best-effort behavior is preserved for the producer-group-selector UI path (where partial results are acceptable)
- **Tests** ✅ — New tests verify both the best-effort path (partial results OK) and the strict path (any failure = 502)

### Code Quality
Clean separation of concerns: the `requireComplete` parameter makes the intent explicit at the call site. The removal of the `successfulGroupQueries` counter simplifies the logic while making failures more visible.

LGTM — important fix for monitoring accuracy.

---

lizhimins (changes requested):
Requesting changes on the direction rather than the implementation: for these
read-only diagnostics, a partial result is a success, not an error.

The trunk already decided this. `d2c7de203 fix(client): treat offline group
connections as empty results (#4002)` states in its commit message that genuine
broker failures and partial scans keep their existing behavior. The way this repo
makes partial results safe is to publish a completeness marker, not to fail closed:
`DlqExportHeaders` (`X-DLQ-Export-Truncated` / `X-DLQ-Export-FailedQueues` /
`X-DLQ-Export-Limit`, exposed through `CorsConfig#exposedHeaders`) plus
`DLQExportResultVO.truncated`, #4185 for Apache key queries, and #2221 for retaining
partial diagnostics.

Concretely:

1. Please keep returning the groups and connections that were reachable and surface
   the gap instead — a `truncated` / `incomplete` field or the list of failed brokers
   in the payload, following the DLQ export precedent.
2. As written, the class ends up with two contradictory semantics for "one broker is
   down": 502 on the topic-scoped path, best-effort 200 on
   `findAllProducerConnections` / `findConsumerConnections` /
   `collectSubscriptionGroups` in the same file. During a rolling restart — a routine
   operation here — `GET /api/producer/connection?topic=X` would return a hard 502 for
   the whole window and the topic page's producer panel goes blank, which is worse for
   diagnosis than a visibly incomplete list.
3. Error message nesting: `"Failed to query producer connections for group " + pg + ": "
   + e.getMessage()` yields `Failed to query producer connections for group pg-order:
   Failed to query producer connections: broker unavailable`, and the new test
   hard-codes that string, so the wording can never be improved without touching the
   test. Please unwrap with the existing `rootMessage(...)` helper the way the broker
   branch in the same file already does.
4. Test method names must end with `Test` (project convention). The three new methods
   and the renamed one do not, although the PR checklist claims they do.

To be clear about what is right here: the underlying observation in #4463 is real — a
silently truncated producer list does get read as complete — and keeping
`isGroupConnectionAbsent` / `isTopicNotExist` as 200-with-empty-result is correct. It
is the fail-closed remedy that we do not want.

Aias00:
Addressed the requested direction in `38eb60ec`:

1. Topic-wide read-only diagnostics now retain every reachable producer connection instead of
   returning 502 for a partial scan.
2. The payload exposes `complete`, `failedBrokers`, and `failedProducerGroups`; the summary adds
   `INCOMPLETE_SCAN` with `WARNING` readiness, and the UI renders the failed targets.
3. Actual query failures use `rootMessage(...)`; the nested group error string is gone.
4. Every new or renamed Java test method ends with `Test`.

All-Broker and all-discovered-group failures still return 502, while offline/missing groups remain
normal complete empty results. Focused backend tests pass 50/50, full frontend tests pass 1036/1036,
and backend package plus frontend build/lint/format checks pass. The full backend run had only the
known 59-vs-60-second session-overview timing flake; isolated reruns passed twice and failed once.

PR #4469, opened 2026-09-16
fix(acl): clear a Tencent role's permissions instead of deleting the role

### LINKED ISSUE #4161: [Studio][Bug] Deleting a Tencent ACL rule deletes the role account

123123213weqw:

- Fixes #4161

### Brief Description

Revoking the ACL rule for a Tencent role called `DeleteRole`, which removes the role account itself. On Tencent a role also owns the credentials, so deleting it silently removed the ACL user along with the rule, while the Studio page only asked to revoke access — the operator's next look at the user list showed the account gone.

`deleteRule` now clears `PermRead`/`PermWrite` through `ModifyRole`, which keeps the account so the permissions can be re-enabled later, and documents why the role must stay.

The rule listing also skips a role whose two permissions are both disabled: such a role has no effective ACL rule, so it should not be presented as one, but it is still reported as a user.

This is revived from a branch whose pull request was closed when `rocketmq-studio` was folded into `master` and the base branch was deleted; the change is otherwise unchanged from that branch, and keeps its red/green pair.

```
cd server && mvn -B -ntp test -Dtest=TencentAclServiceTest
```

Test coverage: `TencentAclServiceTest` asserts that deleting a rule issues `ModifyRole` with both permissions cleared rather than `DeleteRole`, and that a role with both permissions disabled is excluded from the rule listing while still appearing as a user.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` — not applicable, no frontend change
- [x] Documentation touched where behaviour changed — the method javadoc now records why the role is kept


RockteMQ-AI (approved):
### Summary
Fixes a critical bug where revoking a Tencent ACL rule would delete the entire role (including credentials) instead of just clearing permissions. The fix uses `ModifyRole` to set `PermRead=false` and `PermWrite=false`, preserving the role account.

### Review
- **Correctness** ✅ — `deleteRule` now calls `ModifyRole` with both permissions disabled instead of `DeleteRole`; the role account and its credentials are preserved
- **Consistency** ✅ — `listRules` now filters out roles with no permissions, keeping the UI consistent with the actual ACL state
- **Tests** ✅ — New tests verify both the modify behavior and the list filtering; explicit `verify(client, never()).DeleteRole(any())` ensures the destructive path is never taken
- **Compatibility** ✅ — No API changes; Tencent provider behavior only

### Code Quality
Excellent documentation in the Javadoc explaining why `DeleteRole` must not be called. The fix correctly addresses the semantic mismatch between "revoke rule" and "delete role".

LGTM — important data-loss prevention fix.

---

PR #4470, opened 2026-09-16
fix(ai): validate message properties before confirming a send preview

123123213weqw:

No issue: this is a small input-validation gap found while reading the tool handlers. Happy to file one if you would rather track it.

### Brief Description

The `rmq.message.send` tool previewed a message without parsing its properties, so a malformed property string passed the confirmation step and failed only when the send was executed. The preview is what the operator reads before approving the write, so it has to reject the same input the execution path rejects: `preview` now parses the properties, which raises the same error the confirm step would, before the operator is asked to approve anything.

This is revived from a branch whose pull request was closed when `rocketmq-studio` was folded into `master` and the base branch was deleted, and keeps its red/green pair.

```
cd server && mvn -B -ntp test -Dtest=MessageSendToolHandlerTest
```

Test coverage: `MessageSendToolHandlerTest` asserts that a preview with an unparseable property string fails at preview time with the property error, instead of returning a plan that only fails on execution.

A note for the larger AI-domain picture: the tool layer has been restructured since the earlier AI pull requests (`ops/ai/tool` is now split into `catalog`, `contract`, `core`, `filter`, `handler`, `service`, `support`, and `AiPayloadGuard` covers tool-invocation payloads). Most of those older branches are obsolete as a result, so I am not resubmitting them; this one is still applicable to the current handler.

### Checklist

- [x] Documentation touched where behaviour changed — none describes the preview contract


RockteMQ-AI (approved):
### Summary
Ensures malformed message properties are validated during the preview step (before user confirmation) rather than only during execution. This prevents operators from approving a preview that will inevitably fail.

### Review
- **Correctness** ✅ — Calling `parseProperties()` in `preview()` ensures validation happens at the same point the operator reviews the action
- **UX** ✅ — Failing fast during preview gives immediate feedback rather than after confirmation
- **Tests** ✅ — Parameterized test covers multiple malformed inputs and verifies both preview and execution paths reject them
- **Compatibility** ✅ — No API changes; validation logic is reused from execution path

Minimal, focused fix. LGTM.

---

lizhimins (changes requested):
The change itself is right and I want to take it: `MessageSendToolHandler` is the
only mutation handler whose `preview` does not validate its own input. Compare
`AclUpdateToolHandler.preview` calling `parseId`, `GroupResetOffsetToolHandler`
calling `requireTimestamp`, `MessageRedeliveryToolHandler` calling `requireGroupName`,
`TopicUpdateToolHandler` calling `input.mergeWith(current)` and `UserCreateToolHandler`
parsing for real. So today a malformed `properties` string passes the confirmation
step and only fails at send time — and the preview is the only thing an operator
reads before approving a write.

One blocker and two nits:

1. **Blocker — test naming.** `malformedPropertiesFailDuringPreviewAndExecution` must
   end with `Test` per the project convention, e.g.
   `malformedPropertiesFailDuringPreviewAndExecutionTest`. The PR checklist states
   "test methods named `...Test`", which does not match the diff — please fix the name
   and uncheck/recheck the box honestly so the checklist stays trustworthy.
2. `parseProperties(input.properties());` discards its return value on purpose. Please
   add a one-line comment saying the call is validation-only and the parsed value is
   rebuilt in `execute`; otherwise it reads like dead code and will eventually be
   deleted by someone tidying up.
3. In the test, import
   `org.apache.rocketmq.studio.ops.ai.tool.core.ToolExecutionException` rather than
   using the fully qualified name inside the assertion.

The test coverage is good — asserting both preview and execute throw, plus
`verify(metadata, never()).sendMessage(...)`, is exactly the mutation-sensitive shape
we want. There is no issue for this one; I am fine merging without one, or you can
open a short issue first if you prefer the trail.

123123213weqw:
All three points are in — thanks for the precise review, and for naming the neighbouring handlers that already do this; it made the shape obvious.

- **Test naming (the blocker).** Renamed to `malformedPropertiesFailDuringPreviewAndExecutionTest`. You are right that the checklist box did not match the diff — it does now.
- **`parseProperties(input.properties());`** carries a one-line comment saying the call is validation-only and the parsed value is rebuilt in `execute`, so nobody deletes it as dead code.
- **The assertion** imports `org.apache.rocketmq.studio.ops.ai.tool.core.ToolExecutionException` rather than spelling it out inline.

```
$ cd server && mvn -B -ntp test -Dtest=MessageSendToolHandlerTest
[INFO] BUILD SUCCESS
```

On the issue: I have no preference — I can open a short one if you want the trail, or leave it as is. Your call.

PR #4472, opened 2026-09-16
fix(ai): keep the provider token out of LlmProperties.toString

123123213weqw:

No issue: a small hardening gap found while reading the LLM configuration code. Happy to file one if you would rather track it.

### Brief Description

`LlmProperties.toString()` printed the provider token, so any log line, exception message or debugging output holding the configuration leaked the credential. The field carries `@ToString.Exclude` now.

Being precise about what this is: **there is no call site that leaks the token today.** `LlmProperties` is only read field by field (`ClaudeCodeAgentProvider`, `CliProcessEnvironment`, `LlmConfigService`), there is no `log.*(properties)` or `toString()` call, and `/actuator/configprops` serializes through getters with key sanitization rather than `toString()`. The exclusion is cheap prevention, consistent with how the other credential holders are treated — not a fix for a live leak.

An earlier revision of this pull request also normalized the engine name in `AgentProviderRegistry` and rejected empty or duplicate engines at construction. It is reverted here: both providers report constants (`claude-code`, `qoder`), so `forEngine("claude-code")` resolves today and the duplicate case needs a third implementation that does not exist. Tests asserting unreachable behaviour are not worth carrying.

### Scope note on the rest of the AI cluster

I looked at the other closed AI pull requests before opening this, and deliberately did **not** resubmit three of them, because `master` has moved past them:

- `#3120` (guard a null cluster in capability resolution) patched `ops/ai/tool/CapabilityResolver`, which has since moved to `ops/ai/tool/catalog/CapabilityResolver` and been **redesigned** — the current class resolves through `InstanceProviderRegistry`/`InstanceResolver` and already handles a missing instance and a null vendor. The old patch would only add a stale second `@Component` of the same name, which makes the Spring context fail to load.
- `#2898` (tolerate malformed tool names) and `#2845` (bound the catalog major version) patched `ToolAccessPolicy` and `ToolCatalog.majorVersion`, both of which no longer exist after the tool layer was split into `catalog`/`contract`/`core`/`filter`/`handler`/`service`/`support`.

The remaining applicable AI fix is #4357, submitted separately as #4470.

```
$ cd server && mvn -B -ntp test -Dtest=LlmPropertiesTest,AgentProviderRegistryTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- AgentProviderRegistryTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- LlmPropertiesTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Checkstyle reports 0 violations on the same run. The full suite was run on the previous revision of this branch (2430 tests, 0 failures; the 17 errors next to it are the `@SpringBootTest` classes that need a reachable MySQL 8 and fail identically on the untouched baseline); the only change since is the revert of the registry half and the test rename.

Test coverage:

- `LlmPropertiesTest` — `toString()` does not contain the configured token.

### Checklist

- [x] New source files carry the ASF license header — `LlmPropertiesTest` carries the ASF header
- [x] Documentation touched where behaviour changed — none describes the configuration bean


RockteMQ-AI (approved):
### Summary
Two security/correctness fixes for AI configuration: (1) token redaction in `toString()` prevents credential leakage in logs, and (2) engine name normalization ensures consistent provider lookup regardless of case/whitespace.

### Review
- **Security** ✅ — `@ToString.Exclude` on the token field prevents accidental credential exposure in logs/exceptions
- **Correctness** ✅ — Engine normalization is now applied consistently at both registration and lookup time; duplicate detection prevents ambiguous configurations
- **Tests** ✅ — New tests verify token redaction and normalization behavior; duplicate engine rejection is also tested
- **Compatibility** ✅ — No API changes; internal hardening only

### Code Quality
Clean extraction of `normalizeEngine()` helper eliminates duplication. The use of `LinkedHashMap` preserves insertion order for deterministic error messages. Immutable `Map.copyOf()` is a nice defensive touch.

LGTM — good defensive programming.

---

lizhimins (changes requested):
Splitting this into its two halves, because they are not in the same state.

**(a) `@ToString.Exclude` on `LlmProperties.token`** — the direction is right and the
cost is two lines. We do want credential-bearing fields excluded from `toString`.
Keep it.

**(b) Engine normalization + duplicate detection in `AgentProviderRegistry`** — I
cannot find a reachable defect. The only two `AgentProvider` implementations are
`ClaudeCodeAgentProvider` (`ENGINE = "claude-code"`) and `QoderAgentProvider`
(`ENGINE = "qoder")`: both are already lowercase and unpadded, so
`forEngine("claude-code")` resolves today, and the "a provider reporting `Claude`"
example in the description is a hypothetical third implementation. Duplicate
registration also already fails — `Collectors.toMap` throws `IllegalStateException`
on duplicate keys, and the two constants differ. So this half adds fail-fast behavior
for states the trunk cannot currently reach.

One more note on (a): I could not find a call site that leaks the token today.
`LlmProperties` is only read field by field (`ClaudeCodeAgentProvider`,
`CliProcessEnvironment`, `LlmConfigService`), there is no `log.*(properties)` or
`toString()` call site, and `/actuator/configprops` serializes through getters with
key sanitization rather than `toString`. That does not make the annotation wrong — it
is cheap prevention and consistent with how we treat other credential holders — but
please say so in the description, so the next reader does not go hunting for a leak
that is not there.

**Blocker**: all three new test methods
(`shouldNormalizeProviderEngineDuringRegistration`,
`shouldRejectDuplicateNormalizedProviderEngines`,
`toStringShouldNotExposeTheProviderToken`) must end with `Test` per the project
convention; the checklist claims they do.

Suggested shape: keep (a) with its test renamed to `...Test`, and either drop (b) or
justify it with a concrete reachable path — a real third provider, or engine names
that come from configuration rather than constants. Tests asserting unreachable
behavior are the part I would rather not carry, because they freeze a contract nobody
actually depends on.

123123213weqw:
Agreed on both halves, and thanks for splitting them instead of rejecting the pair.

**(a) kept.** `@ToString.Exclude` stays, and its test is renamed to `toStringShouldNotExposeTheProviderTokenTest`. Your reading is right and the description now says so: there is no leaking call site today — `LlmProperties` is only read field by field (`ClaudeCodeAgentProvider`, `CliProcessEnvironment`, `LlmConfigService`), nothing logs the bean, and `/actuator/configprops` goes through getters with key sanitization rather than `toString()`. The exclusion is cheap prevention, not a fix for a live leak.

**(b) dropped.** I could not produce the reachable path you asked for. Both providers report constants (`claude-code`, `qoder`), so `forEngine("claude-code")` resolves today, and the duplicate case needs a third implementation that does not exist. `AgentProviderRegistry` and its test are reverted, so the pull request is now the two-line annotation plus its test; title and description are updated to the narrower scope.

```
$ cd server && mvn -B -ntp test -Dtest=LlmPropertiesTest,AgentProviderRegistryTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Thanks also for the explicit "I cannot find a reachable defect" — that was the answer I needed to choose between dropping and justifying, and the tests-asserting-unreachable-behaviour argument is one I will carry into the next handful of PRs.

PR #4479, opened 2026-09-17
fix(alert): report result-set-wide unacknowledged count

### LINKED ISSUE #3557: System alert header reports the unacknowledged count for only the loaded page

tju-yxq:

- Fixes #3557

### Brief Description

Re-opens the previously reviewed change after the base branch moved from `rocketmq-studio` to `master`. Rebased onto current `master` (`987b748e`), all conflicts resolved, and focused tests re-run locally.

cd server && mvn '-Dtest=AlertServiceTest,SystemAlertControllerTest,MybatisPlusAlertRepositoryTest' test → 106 tests passed; cd web && npm test -- SystemAlertsPage.test.tsx --run → 16 tests passed; lint/build passed

See the verification commands and results above; all were run on the rebased branch.

### Checklist


RockteMQ-AI (approved):
### Summary
Adds a `/summary` endpoint for system alerts that returns `totalCount` and `unacknowledgedCount` with the same filter support as the list endpoint.

### Observations
- **Consistency**: Uses the same `SystemAlertQuery` and filter normalization as `listAlertsPage` — filters behave identically.
- **Validation**: Proper checks for `from/to` ordering and `labelKey/labelValue` pairing.
- **SQL**: Uses `COALESCE(SUM(CASE WHEN...))` for unacknowledged count — efficient single-query approach.
- **Frontend**: Summary banner shows unacknowledged count prominently, auto-refreshes with the list.

LGTM.

---

PR #4487, opened 2026-09-17
fix(dlq): isolate detail resend state across instance switches

### LINKED ISSUE #4486: [Studio][Bug] A completed DLQ resend can overwrite another instance's message drawer
## Problem

The DLQ detail resend flow can publish a refresh from an old instance after the operator has already switched to another instance and opened another group's message drawer.

`resendSelectedMessages` starts with the current `selectedInstanceId` / `detailGroup`, awaits `resendDLQSelected`, then unconditionally calls `loadDetailMessages(detailGroup, detailPage, detailPageSize)`. That async continuation retains the old render's instance/group. `loadDetailMessages` increments the shared `detailRequestIdRef`, so the old-instance refresh can supersede a newer detail request from the newly selected instance.

## Deterministic reproduction

1. On instance A, open DLQ group A and select a message.
2. Start `resendDLQSelected` and keep the request pending.
3. Switch to instance B and open DLQ group B. Keep B's detail request pending.
4. Resolve A's resend successfully. Its continuation starts another A detail load and advances `detailRequestIdRef`.
5. Resolve B's original detail request.

On current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`), B's response is discarded as stale while the old A refresh can populate the drawer. A red-first Vitest regression fails because `instance-b-message` never appears after this interleaving.
## Impact

This is more than stale presentation. Once old-instance rows are rendered inside the new instance/group drawer, selecting them and using the current resend action can submit those message ids with the new instance/group context. At minimum operators see the wrong DLQ inventory; at worst a follow-up mutation is formed from mismatched resource identities.

The old in-flight resend also keeps the page-wide `resendInFlightRef` set after the instance switch, temporarily blocking legitimate resends in the new scope.

## Expected behavior

- Instance/scope changes invalidate in-flight detail loads and resend UI continuations.
- Completion of an old resend must not refresh, clear selection, toast, set errors, or release loading state in a newer scope.
- A new instance is not blocked by an old scope's UI in-flight guard.
- If the user remains in the same scope, a successful resend still refreshes the current detail page.

## Proposed scope

Add a resend request generation alongside the existing detail/retry generations, invalidate detail/resend generations on instance/scope changes, and guard all post-await UI effects. Keep the already-submitted backend mutation untouched; this only prevents its late client-side continuation from mutating a different scope.

AI-assisted analysis and regression authoring; local red test executed against the current `master` baseline.

zmuxuny:

- Fixes #4486

### Brief Description

A selected-message DLQ resend could finish after the operator switched instances and opened another group's message drawer. The old continuation then reloaded its original group through the shared detail request generation, superseding the newer instance's pending detail request. This could render instance A's messages under instance B's drawer scope.

This change gives detail resends their own request generation and invalidates it when the instance, drawer, or detail scope changes. Post-await UI effects (toast, selection reset, error state, detail reload, loading state) run only while the resend still belongs to the current scope.

The existing one-resend-at-a-time behavior is preserved with separate whole-group and detail in-flight guards that cross-check each other, so a stale completion cannot unlock a newer operation.
### Fail-before evidence

The deterministic regression starts a detail resend on instance A, switches to instance B, opens B's drawer, then resolves the old A resend before B's pending detail request. On the unmodified baseline the test fails because `instance-b-message` never appears: the old A refresh advances the shared detail request generation and invalidates B's response.

The fixed version keeps the old-instance detail call count at 1 and renders B's message without allowing the stale A refresh to publish.

- `npx vitest run src/pages/instance/__tests__/DLQPage.test.tsx` — **21/21 passed**.
- `npx eslint src/pages/instance/dlq.tsx src/pages/instance/__tests__/DLQPage.test.tsx` — **0 errors**; one pre-existing Fast Refresh warning in `dlq.tsx`.
- `npm run build` — TypeScript + Vite production build passed; **8046 modules transformed**.
- `npx prettier --check` after formatting and `git diff --check` pass.

AI-assisted analysis, implementation, and regression authoring; the red/green tests and build above were executed locally on the current `master` baseline.


RockteMQ-AI (commented):
### Summary

This PR fixes a race condition in the DLQ page where a pending resend operation from instance A could resolve after the user switches to instance B, causing the drawer to display stale data from the wrong instance. The fix uses a `useRef` to track the current instance ID and an effect cleanup to discard stale async results. This is a well-targeted fix for a subtle concurrency bug.

### Findings

- **[Info]** `web/src/pages/instance/dlq.tsx` — The `useRef` + effect cleanup pattern is the correct approach for this class of bug. Unlike the `key`-based remount used in the Topic page PR (#4490), this pattern is necessary here because the DLQ page has in-flight async operations (resend promises) that cannot simply be unmounted.
- **[Info]** `web/src/pages/instance/__tests__/DLQPage.test.tsx:262` — The regression test is thorough: it initiates a resend on instance A, switches to instance B before the resend resolves, then verifies the drawer shows instance B data after the stale resend completes. The use of controlled Promise resolution (`resolveResend`, `resolveSecondDetail`) to orchestrate the race condition is excellent test engineering.
- **[Info]** `web/src/pages/instance/__tests__/DLQPage.test.tsx:330` — The assertion `expect(messageService.resendDLQSelected).toHaveBeenCalledTimes(1)` correctly verifies that only one resend was triggered, preventing false positives from duplicate calls.

### Suggestions

- The PR description mentions this is the DLQ counterpart to the Topic page fix (#4490). Consider adding a brief note in the PR body explaining why the `key`-based remount approach (used in #4490) was not sufficient here and why the `useRef` + cleanup pattern was chosen instead. This would help future maintainers understand the design decision.
- If other instance-scoped pages (Consumer, Message trace, etc.) have similar async operations, they may need the same `useRef` guard. A quick audit would confirm full coverage.

### Verdict

Clean, well-tested fix for a non-obvious race condition. No correctness, performance, or compatibility concerns.

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The DLQ page now correctly isolates resend state across instance switches — pending resend operations from one instance no longer leak into another. The fix uses instance-scoped state keys to prevent race conditions. Comprehensive test coverage verifies the isolation boundary.

LGTM.

---

PR #2308, opened 2026-08-17
consolidate: Apache runtime diagnostics and bounded message flows

Aias00:
## Consolidates
- #1983 typed Apache topic management
- #1990 instance-scoped runtime credentials
- #2282 DLQ group pagination
- #2284 bounded topic message queries

The changes share the Apache admin client plus DLQ/message provider surfaces. This branch preserves all source commits and integrates their overlapping runtime context behavior.

## Validation
- `git diff --check` passed.
- No unresolved merge markers.
- Java 21 targeted Maven tests passed: 87 tests across runtime resolver, NameServer config diff, Apache admin client, cluster, DLQ, and message providers.

Issue #1720, opened 2026-08-11
[Studio][Bug] Reset Topic page state when the selected instance changes


Aias00:
## Problem

The Topic page reloads its list when the selected instance changes, but it keeps instance-scoped component state from the previous instance: selected Topic details, send-message state, import state, cached routes, and consumer relationships. A user can therefore continue viewing or acting through a dialog opened for the old instance after choosing another instance.

## Expected behavior

- Switching the selected instance must discard Topic page state owned by the previous instance before loading the next list.
- The no-instance view must stay idle rather than show a permanent loader.
- Existing Topic list and mutation requests must continue using the selected instance ID.

## Scope

Topic page frontend lifecycle and focused regression tests only.

## Track

Track 1 / Studio multi-instance metadata management.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** The Topic page (`frontend-new/src/pages/Topic/topic.jsx`) calls `getTopicList()` only once on mount via `useEffect(fn, [])`. There is no reactive instance context in the app — the Ops page manages `selectedNamesrv` in its own local `useState` with no broadcast mechanism. All instance-scoped state (selected topic details, send-message state, import state, cached routes, consumer relationships, open dialogs) persists unchanged when the user switches instances.

**Impact:** Users can view or act through stale dialogs/data from a previous instance after switching, leading to incorrect operations or confusion.

**Severity:** Medium — functional correctness issue in multi-instance management.

**Affected files:**
- `frontend-new/src/pages/Topic/topic.jsx`
- `frontend-new/src/pages/Ops/ops.jsx`
- `frontend-new/src/router/index.jsx`

**Suggested fix:** Introduce a shared instance context (or equivalent) with a `useEffect` dependency on the current instance in `topic.jsx` to reset state and re-fetch data on instance changes.

An automated fix proposal will be generated. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by github-manager-bot*

PR #4490, opened 2026-09-17
fix(topic): restore instance-scoped mutation boundary

### LINKED ISSUE #4489: [Studio][Regression] Topic mutation state survives instance switches
## Problem

The Topic page no longer resets instance-scoped mutation state when the selected managed instance changes. A Topic edit opened on instance A stays mounted after switching the page selector to instance B, while `handleCreate` resolves the write target from the *current* `selectedInstanceId`.

Deterministic reproduction on current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`):

1. Select instance A and open **Configure** for `topic-a`.
2. Leave the edit dialog open and switch the page selector to instance B.
3. Instance B's inventory loads, but the `Edit Topic` dialog still contains A's values.
4. Saving now builds `updateTopic({...values, instanceId: selectedInstanceId})`, so the stale A form is submitted against B. If B has the same topic name, this can silently modify the wrong Broker-side Topic configuration.

The same component also owns send-message, import, sync-to-Broker, detail and selection state, so stale instance-scoped targets can survive the same selector boundary.

## Regression history

This lifecycle boundary was previously tracked in #1720 and implemented by #1722, then consolidated into merged PR #2308 using an instance-keyed Topic content component. Current `master` no longer has that keyed remount, so this is a regression of previously merged behavior rather than a new design.

## Expected behavior

Changing the selected instance invalidates all Topic-page state derived from the previous instance before the new inventory becomes actionable. An edit/send/sync target from A must never be reusable while B is selected.

## Test evidence

A red-first Vitest regression opens A's edit dialog, switches to B, waits for `topic-b`, and requires the old `Edit Topic` dialog to be gone. On the unmodified baseline it fails because the dialog remains mounted. Restoring the instance-keyed Topic content remount makes the same test pass.

## Proposed scope

Restore the instance-keyed Topic content boundary only; do not change Topic APIs or Broker semantics. This matches the pattern already used by the Consumer and ACL pages.

AI-assisted analysis and regression authoring; red/green test executed locally against the current `master` baseline.

zmuxuny:

- Fixes #4489

### Brief Description

Restore the instance-scoped lifecycle boundary for the Topic page. Current `master` keeps Topic state mounted while the selected managed instance changes, so a mutation target captured from instance A can remain visible after instance B is selected. The edit path then builds its request with the current `selectedInstanceId`, which can submit A's stale form against B when both instances contain the same Topic name.

The fix moves the existing Topic page body into `TopicPageContent` and keys that content by `selectedInstanceId`. Switching instances therefore discards the previous instance's edit/detail/send/import/sync/selection state before the new inventory becomes actionable. No Topic API or Broker behavior changes.

This restores behavior previously tracked by #1720 and merged as part of #2308; current `master` no longer contained the keyed remount, so #4489 records the regression explicitly.

### Fail-before evidence

The new regression opens `topic-a`'s edit dialog on instance A, switches the page to instance B and waits until `topic-b` is loaded. On the unmodified baseline it fails because the stale dialog remains mounted:

```text
expected document not to contain element, found:
<div class="ant-modal-title">编辑 Topic</div>
```

The page itself already shows `instance-b` at the same time, proving that the mutation UI and the selected write scope have diverged.

- `npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx`
  - **27 tests passed, 0 failures**.
- Focused red/green regression:
  - baseline: failed because the A edit dialog survived the switch to B;
  - fixed: passed and `updateTopic` remained uncalled.
- `npx eslint src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx`
  - **0 errors**.
- `npm run build`
  - TypeScript and Vite production build passed; **8046 modules transformed**.
- `git diff --check`
  - passed.

The Vitest run still prints existing jsdom `getComputedStyle` and AntD `Descriptions` warnings from unrelated cases; all 27 tests pass.

### Checklist

- [x] One coherent defect fix; no unrelated changes
- [x] Commit subject follows Conventional Commits and references the issue
- [x] Regression test added and demonstrated red before / green after
- [x] No new UI text or dependency
- [x] No public API, persistence or Broker protocol change
- [x] Production build and changed-file lint pass

AI-assisted analysis, patch and regression authoring; all red/green and build commands above were executed locally against the stated baseline.


RockteMQ-AI (commented):
### Summary

This PR correctly restores the instance-scoped lifecycle boundary for the Topic page by extracting the page body into `TopicPageContent` and keying it by `selectedInstanceId`. This is a clean, idiomatic React pattern that ensures stale edit/detail/send state from instance A is fully discarded before instance B becomes actionable.

### Findings

- **[Info]** `web/src/pages/instance/topic.tsx:339` — The `TopicPageContentProps` type derived from `ReturnType<typeof useInstanceFilter>` is a nice approach that keeps the prop type in sync with the hook. No action needed.
- **[Info]** `web/src/pages/instance/topic.tsx:353` — The `key={selectedInstanceId}` on `TopicPageContent` is the correct fix. React will unmount the old tree and mount a fresh one when the instance changes, which guarantees no stale closures or form state survive the switch.
- **[Info]** `web/src/pages/instance/__tests__/TopicPage.test.tsx:308` — The regression test is well-structured: it opens an edit dialog on instance A, switches to instance B, and verifies the dialog is dismissed without calling `updateTopic`. The "fail-before evidence" section in the PR body is excellent documentation practice.

### Suggestions

- Consider whether the same `key`-based remount pattern should be applied to other instance-scoped pages (Consumer, Message, etc.) if they have not already been addressed. The PR description references #1720 and #2308, suggesting this was previously in place — a quick audit of sibling pages would confirm full coverage.

### Verdict

Clean, minimal, well-tested fix. No correctness, performance, or compatibility concerns.

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The Topic page now correctly scopes create/update/delete operations to the current instance, preventing cross-instance mutations when switching instances. The fix properly resets the mutation boundary on instance change. Clean implementation with test coverage.

LGTM.

---

PR #4499, opened 2026-09-17
fix(aliyun): reject unsuccessful mutation responses

### LINKED ISSUE #4498: [Studio][Bug] Aliyun mutations ignore unsuccessful OpenAPI business responses

- [x] I searched open/closed issues and pull requests and found no existing fix for this failure mode.
- [x] This reproduces on current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`).

### Problem

`AliyunClientFactory.call` treats an SDK future that completes normally as success. Several Aliyun mutation methods then discard the returned response entirely. The pinned RocketMQ 5.x SDK models these APIs with explicit business-result fields (`success`, and for resource mutations `data`), so a completed response with `success=false` / `data=false` is currently reported by Studio as a successful mutation.

Affected current-master paths are:
- Topic create / update / delete;
- Consumer Group create / delete;
- consumer-offset reset.

The first five response bodies expose `Boolean success` and `Boolean data`; `ResetConsumeOffsetResponseBody` exposes `Boolean success`.

### Impact

This creates false control-plane success. Studio can update its UI and write a SUCCESS audit entry even though the managed Aliyun resource was not changed. For offset reset this is especially dangerous: an operator can believe a consumer position was moved when the provider reported failure.

### Expected behavior

Each Aliyun mutation must validate the response body before Studio reports success:
- null/incomplete response => 502;
- `success != true` => 502 with the provider code/message when available;
- Topic/Group resource mutations also require `data == true`;
- reset-offset requires `success == true` (its response has no `data` field).

SDK exceptional completions should continue through the existing `AliyunClientFactory` error mapping. No frontend contract change is required.

### Verification plan

Add deterministic provider tests that return normally completed SDK responses with unsuccessful business bodies and prove the unmodified provider currently returns normally. Then make those same cases fail closed, while preserving successful mutation request mapping and exception behavior.

AI-assisted source audit; SDK response signatures were checked against the pinned `alibabacloud-rocketmq20220801:5.0.8` jar.

zmuxuny:

- Fixes #4498

### Brief Description

Aliyun's async SDK can complete normally while its response body reports a failed business operation. Studio previously discarded mutation responses, so a provider-side failure could be returned and audited as success.

This change validates the existing Aliyun business result before returning from:
- Topic create / update / delete;
- Consumer Group create / delete;
- consumer-offset reset.

The validator requires `success=true`; resource mutations also require `data=true`. Null/incomplete responses and unsuccessful business responses now become HTTP-style 502 `BusinessException`s using the provider message/code when available. Existing SDK exceptional completions still use `AliyunClientFactory`'s established mapping.

No frontend/API shape, dependency, schema, or Tencent/Apache behavior changes.

### Reproduction and Tests

Fail-before on unmodified `master@987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`:

- added six provider regressions returning normally completed SDK futures with `success=false`;
- `AliyunInstanceProviderTest`: **39 tests, exactly 6 failures**;
- every new failure was `Expecting code to raise a throwable`, proving all six mutations were reported as success.

Green, Java 21:

- `mvn -B -ntp -Dtest=AliyunInstanceProviderTest test`
  - **41/41 passed** after adding `success=true,data=false` and null-response coverage.
- `mvn -B -ntp -Dtest=AliyunInstanceProviderTest,AliyunClientFactoryTest,MetadataServiceTest test`
  - **103/103 passed**, 0 failures/errors.
- `git diff --check`: clean.

### Compatibility / Risk

The change is fail-closed only for Aliyun responses that already declare the mutation unsuccessful or incomplete. Successful response contracts remain unchanged. The main operational benefit is preventing false success/audit records, especially for consumer-offset reset where a failed reset must not look applied.

AI-assisted source audit and implementation; the response fields were verified against the pinned `alibabacloud-rocketmq20220801:5.0.8` SDK and fail-before behavior was reproduced before changing production code.


RockteMQ-AI (approved):
### Summary

This PR fixes a real correctness issue: Aliyun's async SDK can complete normally (HTTP 200) while the response body indicates a failed business operation (`success=false`). Previously, the dashboard discarded mutation responses, so provider-side failures were returned and audited as success.

The fix adds a `requireMutationSuccess` validator that checks both `success=true` (all mutations) and `data=true` (resource mutations), throwing a `BusinessException(502)` with the provider's error details when validation fails. Applied to 6 mutation operations with 8 regression tests.

### Review

**Correctness**: Excellent. The distinction between resource mutations (require `data=true`) and offset reset (only `success=true`) is semantically correct. The null-safety checks (`body == null ? null : body.getSuccess()`) are necessary and properly handled.

**Error handling**: Proper use of provider's error message/code when available, with sensible fallbacks for incomplete responses.

**Tests**: Comprehensive coverage — 6 regression tests for `success=false` scenarios plus 2 edge cases (null response, `data=false`). The fail-before reproduction mentioned in the PR description proves the tests catch the bug.

**Code quality**: Clean implementation. The helper method is well-named and the validation logic is clear. No unnecessary changes to unrelated code.

**Documentation**: Excellent PR description with reproduction steps, test results, and compatibility analysis.

LGTM — this is a high-quality bug fix that addresses a real data-integrity issue.

---

lizhimins:
Thanks for the careful sweep of the mutation call sites.

Closing this one, and let me be explicit about why, because the code itself is not the problem.

**What is right here.** All six mutation call sites in `AliyunInstanceProvider` (`createTopic`, `updateTopic`, `deleteTopic`, `createConsumerGroup`, `deleteConsumerGroup`, `resetConsumeOffset`) are covered with no gaps, the eight new tests are mutation-sensitive (removing any one `requireMutationSuccess(...)` call turns the matching test red), and the `BusinessException(502, ...)` bare numeric code follows the project convention.

**Why we are not taking it.** For a cloud instance Studio does not keep a local mirror of the resource inventory. `MetadataService.listTopics` / `listTopicsPage` resolve the instance's `InstanceProvider` and read from the cloud on every request, and `AliyunInstanceProvider.listTopics` issues a `ListTopics` API call each time — there is no cached copy that a mis-graded mutation response could desynchronise. So the state this check is defending is re-read from the cloud anyway. On top of that, we do not believe the case being defended against actually occurs: transport and HTTP-level failures are already mapped to 422/403/404/502/504 by `AliyunClientFactory#call` and `mapToBusinessException`, which leaves only "HTTP 200 with `success=false` in the body" and "null response/body" as genuinely new coverage — and no capture of the former was attached to the PR.

**Two further problems the review turned up, for the record.**

1. `Boolean.TRUE.equals(success)` (and `Boolean.TRUE.equals(data)`) rejects a *missing* flag, not only an explicit failure. Both fields are nullable `Boolean` in `alibabacloud-rocketmq20220801`. This PR had to add `.success(true)` to pre-existing stubs, which is itself evidence that response bodies without `success` are a normal shape. On a mutation that is worse than the bug being fixed: Studio would report a failure for an operation the cloud actually performed.
2. `requireMutationSuccess` is the first of three near-identical copies — `requireReadSuccess` in #4504 and `requireCatalogSuccess` in #4507 are the same logic with a different message prefix. The project convention is one shared helper under `common/util`, and #4499 and #4504 currently conflict on both the provider and its test file, so they cannot both land as written.

If you do want to revisit this, what would change the answer is evidence rather than more code: a real capture of an HTTP 200 response with `success=false` from one of these APIs, attached to an issue. If that turns out to be a real failure mode, the shape we would accept is a single shared helper used by the mutation, read and catalog paths, rejecting only on an explicit `Boolean.FALSE` (plus a null response/body), with a test pinning the tolerant direction — `success == null` must still succeed.

PR #4503, opened 2026-09-17
[ISSUE #4502] fix(clients): restrict diagnostics to registered NameServers

### LINKED ISSUE #4502: [Studio][Hardening] Constrain client diagnostics to registered NameServer endpoints

- [x] I searched open and closed issues and pull requests for client diagnostics, `namesrvAddr`, registry validation, and admin-client cache bounds.
- [x] I verified the behavior against current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`).
- [x] No live NameServer endpoint was contacted while validating this report.

### Problem

`GET /api/clients` accepts a caller-supplied `namesrvAddr`. `ClientService.listConnectionsAt` checks only that the value is nonblank, then forwards it to `RocketMQClientProvider`, which executes through `MqAdminExtFactory`. The factory creates and caches a started admin client for every distinct normalized address.

The Clients page itself only submits endpoints returned by the NameServer registry, but the backend does not enforce that same boundary. A read-only user can therefore make Studio attempt outbound RocketMQ connections to addresses that were never registered by an administrator, and distinct addresses can allocate long-lived cache entries until explicit release or application shutdown.

### Proposed Design

Keep the existing frontend and `/api/clients` request contract, but enforce the registry boundary in the backend:

1. Add a `NameserverRegistryService` lookup that normalizes the requested value with `NamesrvAddrParser` and returns it only when the normalized address exists in `rmq_nameserver`.
2. Make `ClientService.listConnectionsAt` require that registered address before invoking `ClientProvider.findConnectionsAt`.
3. Reject unknown addresses before any provider or `MqAdminExtFactory` interaction.
4. Preserve existing cluster/type filtering and the Clients-page workflow.

This keeps read-only diagnostics available to readers while ensuring administrators control the outbound endpoint set through the existing registry mutation APIs. Cache policy changes are intentionally out of scope.

### Acceptance Criteria

- Registered NameServer endpoints continue to support client connection diagnostics.
- Address formatting is normalized consistently with registry create/update.
- An unregistered endpoint is rejected before provider execution.
- Existing `clusterId` and `type` filters are unchanged.
- Focused service tests cover registered, normalized, blank, and unregistered addresses.
- Backend tests, Checkstyle, and `git diff --check` pass.

AI-assisted static analysis; the request-to-cache call path was verified from source without active endpoint probing.

Aias00:

- Fixes #4502

### Brief Description

- Normalize client-diagnostics NameServer addresses with the same parser used by registry writes.
- Require an exact registered endpoint before invoking the client provider or admin-client cache.
- Return 404 for unknown endpoints while preserving existing cluster/type filters and the Clients-page API contract.
- Add regression coverage for normalized registered endpoints and fail-closed unknown endpoints.

- `cd server && mvn -B -ntp -Dtest=ClientControllerTest,ClientServiceTest,NameserverRegistryServiceTest,RocketMQClientProviderTest,MqAdminExtFactoryTest test` — 69 tests passed; Checkstyle 0 violations.
- `cd server && mvn -B -ntp -Dsurefire.rerunFailingTestsCount=2 test` against a fresh isolated MySQL 8 schema — 2416 tests passed; Checkstyle 0 violations. The temporary schema was removed after verification.
- `cd server && mvn -B -ntp -DskipTests package` — passed.
- `git diff HEAD^ --check` — passed.
- Independent read-only review — 0 findings.

### Checklist

- [x] Documentation touched where behaviour changed (no documentation change required)


RockteMQ-AI (approved):
### Summary
This PR fixes a security/correctness issue where client diagnostics could query arbitrary NameServer addresses. The fix validates the address against the registry before querying, returning 404 for unregistered addresses.

**Changes reviewed:**
- `ClientService.java`: Added registry validation before querying connections
- `ClientServiceTest.java`: Added 3 comprehensive tests

**Assessment:**
- ✅ Correctness: Proper fail-closed behavior with 404 for unregistered addresses
- ✅ Security: Prevents potential SSRF-like attacks via unregistered NameServer addresses
- ✅ Tests: Good coverage including normalization edge cases
- ✅ Compatibility: No breaking changes to API contract

LGTM — clean security fix with solid test coverage.

---

lizhimins:
Thanks — the hardening goal here is correct and your write-up of #4502 is what made it precise.

`GET /api/clients` is reader-accessible (`AuthInterceptor` lets every GET through, and `/api/clients` is not in the admin-only list), `ClientService` only trimmed the incoming `namesrvAddr` before handing it to the provider, and `MqAdminExtFactory` caches one started `DefaultMQAdminExt` per address in an unbounded map that is only reclaimed on explicit `release` or `@PreDestroy`. So an unregistered address really is both an outbound-connection surface and a client-cache growth surface, and closing it in the backend is the right call — the frontend only ever submits registered endpoints, so it cannot enforce this.

We are closing this as a duplicate of #4534, which fixes the same defect at the same entry point and has now been merged. Two reasons we preferred that implementation:

- The check lives in `NameserverRegistryService.requireRegisteredAddress(...)`, i.e. in the component that owns the registry, rather than inlined in `ClientService`. That is where the next caller will look for it.
- It answers the question with a conditional count query on `namesrv_addr`, whereas this PR streams `registryService.list()` — a full `selectList` plus a `toVO` mapping per row — on every `GET /api/clients`.
- It also covers the malformed-address path, where `NamesrvAddrParser.normalize` throws 400 before any database lookup. That behaviour is untested here.

You filed this about six hours before #4534, so this is not a first-come decision — it is a "which implementation do we want to own" one, and we took the better-encapsulated version. Sorry for the wasted round.

One gap that exists in both PRs and is worth carrying over as a follow-up: the comparison is normalized-request against raw-stored value. Registry rows written before #2510 (which started normalizing on create/update) can still hold a form like `NS1:9876;ns2:9876`, and the frontend submits exactly that raw string, so a legitimate endpoint can 404 after normalization. Normalizing both sides, or a small data migration, would close it. If you would like to pick that up, it is a genuinely open problem and we would review it.

PR #4496, opened 2026-09-17
feat(topic): support cloud test-message sending

zmuxuny:
### Which Issue(s) This PR Fixes

- Fixes #4491

### Brief Description

Extend the existing Topic-page **Send Message** operator workflow from Apache-only to the two managed RocketMQ 5.x providers already supported by Studio.

- add a provider-level `MESSAGE_SEND` capability and `sendMessage` SPI
- keep the Apache path unchanged through the existing admin client
- map Aliyun sends to the pinned SDK's `VerifySendMessage` API, including body/tag/key, user properties, message group, and delivery timestamp
- map Tencent sends to Trocket `SendMessage`, returning its message ID
- reject Tencent user properties / FIFO message-group / delayed-delivery fields instead of silently dropping data the Tencent console API cannot represent
- expose the existing send action for cloud **NORMAL** Topics; FIFO/delay cloud sends remain hidden until the UI collects their required semantics
- fail closed when a cloud API returns an incomplete/unsuccessful response instead of reporting a false success

This is intentionally an operator test-message feature, not a replacement for application producers. No dependency upgrade is introduced.

### How Did You Test This Change?

Backend, Java 21:

- `mvn -B -ntp -Dtest=AliyunInstanceProviderTest,TencentInstanceProviderTest,ApacheInstanceProviderTest,MetadataServiceTest test`
  - **143 tests passed**, 0 failures/errors
  - Checkstyle: **0 violations**
- capability/controller regression suite: **31 tests passed**
- request mapping tests verify Aliyun/Tencent fields, success responses, missing message IDs, unsuccessful responses, and Tencent fail-closed behavior for unsupported fields

Frontend:

- `npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx`
  - **27 tests passed**, including cloud NORMAL Topic test-message sending with the selected cloud instance id
- `npm run build` passed (`8046 modules transformed`)
- ESLint on touched files: 0 errors
- Prettier check passed
- `git diff --check` passed

### Compatibility / scope

- Apache behavior is unchanged.
- Existing cloud SDK versions are reused; no dependency changes.
- UI only exposes cloud sending for `NORMAL` Topics in this change.
- Aliyun's provider implementation is ready to carry advanced DTO fields when dedicated FIFO/delay UI controls are added later.

AI-assisted design and implementation; the pinned SDK signatures and current `master` behavior were verified before implementation.

PR #4455, opened 2026-09-16
[ISSUE #4375] feat(studio): support cloud direct message consumption

Aias00:
## Summary

- support direct message-consumption verification for Apache, Aliyun, and Tencent instances
- use Aliyun `VerifyConsumeMessage` and Tencent `VerifyMessageConsumption` while preserving the Apache broker implementation
- advertise and enforce `DIRECT_MESSAGE_CONSUME` through the server capability contract
- keep the message-page action visible but disabled with localized reasons while capability discovery is loading, unavailable, stale, or unsupported
- report Tencent as `REQUEST_ACCEPTED` because its API acknowledges the verification request without returning the consumption outcome

Fixes #4375

## Rebase and Review Updates

- rebased onto current `rocketmq-studio@4c697f07a`
- resolved conflicts in `MessageService`, `MessageServiceTest`, and `message.tsx`
- retained trunk direct-consume audit semantics: only `CR_SUCCESS` audits as SUCCESS; provider, capability, and business failures audit as FAILED; audit-write failures do not replace the operation result
- retained trunk message-page localization and async-state fixes while applying the capability tooltip/disable state machine
- removed the frontend production capability table; production decisions use the server response
- mock mode derives provider capabilities from the mock instance vendor, including newly created Apache and cloud instances

## Verification

- focused backend provider/service suite: 123 tests passed; Checkstyle 0 violations
- full backend suite with Java 21 and isolated H2 MySQL-compatible datasource: 3151 passed, 0 failures/errors, 3 skipped; Checkstyle 0 violations
- focused frontend message/capability suite: 44 tests passed
- full frontend suite: 139 files, 1284/1284 tests passed
- `npm run build`: passed
- `npm run lint`: 0 errors, 11 existing warnings in unchanged files
- Prettier check for all changed frontend files: passed
- `mvn -B -ntp -DskipTests package`: passed
- `git diff --check`: passed
- independent conflict review: initial LOW mock-capability gap fixed; focused re-review approved with 0 remaining findings

## CI Note

GitHub currently reports no checks for this PR, so the local clean-suite results above are the available validation evidence.

PR #4504, opened 2026-09-17
fix(aliyun): unify read and catalog response validation

### LINKED ISSUE #4500: [Studio][Bug] Aliyun read APIs turn unsuccessful business responses into empty operational data

- [x] I searched open/closed issues and PRs and found no existing fix for this failure mode.
- [x] The affected behavior is present on current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`).

### Problem

Studio's Aliyun provider checks the decoded `data` field for several read APIs but never checks the SDK response body's business `success` flag. The pinned `alibabacloud-rocketmq20220801:5.0.8` response bodies for `ListTopics`, `ListConsumerGroups`, `ListTopicSubscriptions`, `ListConsumerGroupSubscriptions`, `GetConsumerGroupLag`, `ListMessages`, and `GetTrace` all expose `success`, `code`, and `message`.

When one of these calls completes normally with `success=false`, the current provider typically sees a null/empty `data` value and converts the provider failure into valid-looking empty operational data.

### Impact

Provider permission, throttling, account, or other business failures can be displayed as:
- no Topics or Consumer Groups;
- no subscriptions;
- no consumer progress/lag;
- no matching messages;
- no trace nodes.

This is dangerous during incident response because an operator is told that the resource/data is absent rather than that the cloud query failed.

### Expected behavior

For these Aliyun read paths, a decoded response body with `success != true` must fail closed with a 502-style `BusinessException`, retaining provider `message`/`code` when available. Successful empty results (`success=true` with an empty data collection) must remain valid empty results. Existing SDK exceptional-completion mapping stays unchanged.

### Verification plan

Add deterministic provider regressions that return normally completed SDK futures with `success=false` and prove the unmodified provider returns empty results instead of an error. Then make the same cases fail closed and retain existing successful-empty/read mapping tests.

This is separate from #4498/#4499, which cover mutating OpenAPI responses only.

AI-assisted source audit; response signatures were checked against the pinned 5.0.8 SDK jar.

zmuxuny:

Fixes #4500. Fixes #4505.

Consolidates the read and catalog changes from #4504 and #4507 into this single PR, as requested. The closed mutation proposal #4499 is not included.

### Response policy

One type-safe `AliyunResponseValidator` under `common/util` handles all nine provider read sites and all three catalog sites. It returns the validated body and uses method references rather than reflection or duplicated service-local helpers.

- A present response body with `success == null` or `true` is accepted. Missing flags are not failures.
- Explicit `Boolean.FALSE` becomes a **422 business rejection**, retaining the provider message/code. It is not classified as a 502 transport failure or a Studio-session 401.
- A null response or null body is a **502 malformed upstream response**, consistently for both read and catalog paths, including trace.
- A present accepted body with empty/null `data` keeps the existing empty-result behavior; a present GetInstance body without data remains 404. A missing body is no longer treated as a confirmed empty result.

The helper uses `StringUtils.hasText` and preserves message -> code -> generic fallback. Existing SDK exception mapping and the merged test-message-send implementation are unchanged. No inventory caching or local cloud mirror is introduced.

### Evidence and scope

I do **not** have a real captured HTTP 200 / `success=false` response from these cloud APIs. The previous fail-before evidence was generated with controlled SDK response fixtures, not a production incident. This revision therefore does not claim such captures or justify a 502 outage classification for an explicit business rejection. It takes the review's alternative of a request-level business code (422), already used by the cloud-client boundary, without expanding endpoint schemas or adding a new availability field.

The null-body policy is an intentional change: old trace tests that treated an absent envelope as empty trace now require 502. Empty trace data inside a present accepted body remains valid. The catalog normalization test now uses a present empty GetInstance body to assert genuine not-found, and separate tests cover absent envelopes.

### Verification

Rebased onto `rocketmq-studio@4c697f07acde460e2344375cb1f82669f5b270fd`. Before correction, the combined old proposals failed 23 of 65 controlled-response cases (19 assertion failures and 4 rejected-missing-flag errors). After correction, the same provider/catalog suites pass; compatibility tests cover omitted flags, null bodies/responses, explicit false and legitimate empty data.

Java 21, current-head verification:
- `AliyunResponseValidatorTest,AliyunInstanceProviderTest,AliyunCatalogServiceTest,AliyunClientFactoryTest,MetadataServiceTest,GlobalExceptionHandlerTest,MessageServiceTest,MessageQueryToolHandlerTest`: **171/171 passed**, 0 failures/errors/skips.
- A MockMvc regression pins HTTP 422 plus the business-response envelope, rather than a gateway error.
- Checkstyle: **0 violations**; `mvn -B -ntp -DskipTests package`: **BUILD SUCCESS**; `git diff --check`: clean.

No frontend, database or dependency changes. Full external-datasource integration tests and live cloud-credential tests were not run. AI-assisted implementation and review; all listed local tests/builds were executed.

### Cross-PR integration check

The final #4504 (`07f4989`) and #4514 (`858fe91`) revisions merge cleanly. A separate local combined snapshot ran the union of their shared-helper/provider/service/controller/MCP-schema suites: **246/246 passed**, 0 failures/errors/skips, Checkstyle 0. This is a local integration check of the two PRs, not an upstream merge.


server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProvider.java
```diff
@@ -117,6 +117,10 @@ public int countTopics(String instanceId) {
         ListTopicsResponse response = clientFactory.call(ctx.credentialId(), ctx.regionId(),
                 client -> client.listTopics(request));
         ListTopicsResponseBody body = response == null ? null : response.getBody();
+        if (body != null) {
```
RockteMQ-AI, line 120: Inside `if (body != null)`, the ternary `body == null ? null : body.getSuccess()` is redundant — `body` is guaranteed non-null. Simplify to `body.getSuccess()`, `body.getCode()`, `body.getMessage()`. This pattern repeats at lines 142, 193, and 299.

RockteMQ-AI (commented):
### Summary
This PR adds `requireReadSuccess()` validation to `AliyunInstanceProvider` read operations (topic listing, consumer group listing, topic-consumer subscriptions, group progress). Similar to PR #4507 which covers the catalog layer, this ensures business-level API failures from Aliyun are surfaced rather than silently returning empty/partial results.

### Findings
- **[Warning]** `AliyunInstanceProvider.java:120-121` — Inside `if (body != null)`, the expression `body == null ? null : body.getSuccess()` is redundant — `body` is guaranteed non-null here. Simplify to `body.getSuccess()`, `body.getCode()`, `body.getMessage()`. This pattern repeats at lines 142-143, 193-194, and 299-300.
- **[Info]** The `requireReadSuccess()` helper is well-structured and consistent with the `requireCatalogSuccess()` pattern from the catalog layer.

### Suggestions
Consider simplifying the redundant null checks. For example:
```java
if (body != null) {
    requireReadSuccess("topic listing", body.getSuccess(),
            body.getCode(), body.getMessage());
}
```
instead of:
```java
if (body != null) {
    requireReadSuccess("topic listing", body == null ? null : body.getSuccess(),
            body == null ? null : body.getCode(), body == null ? null : body.getMessage());
}
```

### Strengths
- Comprehensive coverage across all read operations in the provider.
- Consistent error-handling pattern with the catalog layer (PR #4507).
- Good test coverage for the new validation.

Overall: Good defensive improvement. The redundant null checks are cosmetic but worth cleaning up for readability.

---

server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunInstanceProvider.java
zmuxuny: Addressed — all `requireReadSuccess(...)` calls inside `if (body != null)` now use direct `body.getSuccess()/getCode()/getMessage()` access. Re-ran `AliyunInstanceProviderTest`: 42/42 passed; Checkstyle: 0 violations. Updated head: `b59410b`.

RockteMQ-AI (approved):
### Findings
- **[Info]** `AliyunInstanceProvider.java:575-581` — Clean helper method. The `requireReadSuccess(operation, success, code, message)` pattern is consistent and the error message includes all relevant context (operation name, error code, message).
- **[Info]** Systematic coverage across all read paths: topic count/list/page, group count/list, subscriptions, lag/progress, message queries, and trace queries. No read path is left unguarded.
- **[Info]** All existing tests updated to include `.success(true)` in mock responses, and 11 new tests verify the failure path for each operation. This is thorough defensive testing.
- **[Info]** The `if (body != null)` guard before each `requireReadSuccess()` call correctly handles the null-body case separately (existing behavior preserved).

LGTM — comprehensive defensive fix that prevents silent data corruption from Aliyun API business errors. Excellent test coverage.

---

RockteMQ-AI (approved):
### Summary
This PR adds validation for Aliyun API responses to check the `success` field and throw errors when the API returns an unsuccessful response. This prevents silent data loss when the API returns errors.

### Findings
- **[Info]** `AliyunAclServiceImpl.java` — The `requireSuccess` method correctly validates the `success` field and extracts the error code/message for the exception. This is applied consistently across all read operations.
- **[Info]** `AliyunAclServiceImplTest.java` — Comprehensive test coverage: successful responses pass through, unsuccessful responses throw with proper error details.

### Suggestions
- The validation is critical for data integrity. The error messages include the API error code, which will help with debugging.

This is an important fix that prevents the dashboard from displaying stale or incorrect ACL data.

---

lizhimins (changes requested):
Thanks — the coverage here is thorough and the tests are honest.

**What we verified.** All nine read call sites in `AliyunInstanceProvider` are converted (both `listTopics` overloads, both `listConsumerGroups` overloads, `listTopicSubscriptions`, `getConsumerGroupLag`, `listConsumerGroupSubscriptions`, `listMessages` and `getTrace`), with none missed. The nine new tests are mutation-sensitive — removing any single `requireReadSuccess(...)` call reddens the matching case — and the 15 stub edits all add `.success(true)` rather than weakening an assertion. The underlying problem is real too: the trunk read paths return an empty list whenever the body or its data is null, which does present "the cloud refused" as "there is nothing there". And body-level validation is not foreign to this codebase — `AliyunCatalogService` already throws 502 on a negative `totalCount`, and `AliyunClientFactory.mapToBusinessException` already maps transport and HTTP failures to 422/403/404/502/504.

**Blocking items.**

1. **The rejection rule is too strict for a read path, and it would take the whole instance down.** `Boolean.TRUE.equals(success)` treats a *missing* flag as a failure, not just an explicit one. `success` is a nullable `Boolean` in `alibabacloud-rocketmq20220801`, so any region or API version that omits it turns topic listing, group listing, lag, message query and trace into 502s simultaneously — the instance becomes unusable in Studio while the cloud is perfectly healthy. Your own diff is the evidence: it had to retrofit `.success(true)` into 15 pre-existing stubs, which means response bodies without `success` are the normal shape in the current corpus. Please reject only on an explicit `Boolean.FALSE`.

2. **No test pins the tolerant direction.** There is no case asserting that `success == null` still returns data. That is the most dangerous branch in the change and it is currently unguarded; a future refactor could reintroduce the strict test without anything failing. Please add it for both `success == null` and a null body.

3. **The null-body policy has to be stated and made consistent.** Here `requireReadSuccess` is only invoked inside `if (body != null)`, so a null body silently falls through to empty data. The mutation-path PR (#4499) treated a null body as an "incomplete response" 502 — the opposite policy from the same author. #4499 has now been closed on separate grounds, so this PR and #4507 are the surviving pair; whichever null policy you choose, the two must agree and the description must say which one it is.

4. **A design question we need you to answer in the PR, not a defect.** The project convention for query APIs is that only RPC-layer failures throw; a successful RPC carrying an odd business state should degrade with data plus a flag so the frontend does not read it as an outage. `success=false` on an HTTP 200 is a business-level signal on a successful RPC, and this PR escalates it to 502 on nine read paths at once. Either justify that with a real captured response, or map explicit business failures to a code the frontend will not present as "this instance is down". As it stands the description gives no evidence that HTTP 200 + `success=false` occurs for these APIs — the test output is local only.

5. **Duplication.** `requireReadSuccess` and `requireCatalogSuccess` in #4507 are line-for-line the same logic, differing only in the message prefix and the blank-check spelling. The convention is shared logic under `common/util`; please extract one helper and land the read and catalog changes as a single PR. Within this PR, the nine call sites each repeat the same four-line `if (body != null) { requireReadSuccess(...) }` block — a generic helper that accepts the body would collapse that.

6. **Sequencing.** `AliyunInstanceProvider` has moved on since this branch was cut (#4496 has landed), so please rebase onto `rocketmq-studio`. #4455 is still open and touches the same file, so we will need to order the two.

Once items 1 to 5 are addressed we can merge, subject to the rebase in item 6. Item 4 is the one that decides the shape of the fix, so please answer it first — if the answer is "there is no real capture", the honest scope for this PR is the null-response case plus rejecting an explicit `Boolean.FALSE`.

zmuxuny:
Addressed the read/catalog review together in `07f4989`, rebased onto `rocketmq-studio@4c697f0`.

The evidence limitation is explicit: I do not have a captured HTTP 200 / success=false cloud response. The revised policy uses 422 for an explicit Boolean.FALSE business rejection, not 502; omitted success flags are accepted. A null response/body is consistently a malformed-response 502 across read/catalog/trace, while a present body with no data retains empty/not-found semantics.

All nine provider and three catalog sites now share the type-safe `AliyunResponseValidator` in common/util, including StringUtils.hasText and message/code fallback. New tests pin missing-flag compatibility, missing-envelope errors, explicit rejection, and HTTP 422/envelope propagation. The merged send path and the rejected mutation proposal #4499 were not changed or reintroduced.

171/171 focused tests passed, Checkstyle 0, package succeeds. A separate combined snapshot with #4514 also passed 246/246 tests with a clean Git merge. The PR now fixes both #4500 and #4505; #4507 is being closed as consolidated here, not as an upstream merge. Full datasource integration and live cloud-credential testing were not performed.

zmuxuny:
Hi @lizhimins, thank you for the detailed review. I pushed `07f4989` and updated the PR with the shared read/catalog validator, the missing-flag and null-body policies, and the limit of the available evidence. Could you please re-review when convenient?

PR #4507, opened 2026-09-17
fix(aliyun): reject unsuccessful catalog responses

### LINKED ISSUE #4505: [Studio][Bug] Aliyun cloud catalog hides unsuccessful OpenAPI responses as empty or missing resources

- [x] I searched open/closed issues and pull requests for Aliyun catalog response handling, `ListRegions`, `ListInstances`, and `GetInstance`.

### Problem

`AliyunCatalogService` treats any normally completed SDK future as a successful cloud-catalog read and only inspects `data`. The pinned Aliyun RocketMQ 5.x SDK response bodies for `ListRegions`, `ListInstances`, and `GetInstance` all expose business-level `success`, `code`, and `message` fields.

When those APIs return a decoded body with `success=false`, current behavior converts the provider failure into misleading catalog state:

- `ListRegions`: returns an empty region list;
- `ListInstances`: stops pagination and returns an empty/partial instance list;
- `GetInstance`: turns the provider failure into `404 Aliyun instance not found`.

### Impact

These paths drive credential-backed region selection, cloud-instance discovery, and instance onboarding. Permission, throttling, account, or provider-side business failures can therefore look like "no regions", "no instances", or "instance does not exist", sending operators toward the wrong diagnosis.

### Expected behavior

A decoded catalog response body with `success != true` fails closed with a 502-style `BusinessException`, retaining the provider `message`/`code` when available. A genuine successful response with empty/null data keeps the existing empty/not-found semantics. SDK exceptional completions remain mapped by `AliyunClientFactory`.

### Verification plan

Add deterministic service tests for `ListRegions`, `ListInstances`, and `GetInstance` that return normally completed SDK responses with `success=false` and prove current code returns empty/not-found results. Then make those same responses fail closed while retaining the existing successful-empty and identifier-normalization behavior.

This complements #4498/#4499 (Aliyun mutations) and #4500/#4504 (runtime provider reads); it covers the separate cloud-catalog/onboarding surface.

AI-assisted source audit; the response contracts were verified against the pinned `alibabacloud-rocketmq20220801:5.0.8` jar.

zmuxuny:
> Superseded by #4504, which contains the consolidated implementation and addresses #4505. The original proposal below is retained for history; its strict nullable-flag/502 policy is not the current proposal.

- Fixes #4505

### Brief Description

Aliyun's cloud-catalog responses carry a business-level `success` flag in addition to SDK/future completion. Studio previously ignored it in `ListRegions`, `ListInstances`, and `GetInstance`, so provider-side business failures could appear as an empty catalog or a misleading 404.

This change validates each decoded catalog response before consuming `data`. A body with `success != true` now raises a 502-style `BusinessException` using the provider message/code. Successful empty data retains the existing empty/not-found semantics, and SDK exceptional completions continue through `AliyunClientFactory`.

### Red / Green Verification

Fail-before:
- Added three deterministic regressions for `ListRegions`, `ListInstances`, and `GetInstance` with normally completed SDK responses carrying `success=false`.
- `AliyunCatalogServiceTest`: **13 tests, exactly 3 failures**.
- Regions/instances returned normally instead of erroring; GetInstance returned 404 instead of preserving the provider failure.

Green, Java 21:
- `AliyunCatalogServiceTest` + `AliyunClientFactoryTest`: **23/23 passed**, 0 failures/errors.
- `git diff --check`: clean.

No API/schema/dependency/frontend changes.

AI-assisted source audit and implementation; response fields were verified against the pinned `alibabacloud-rocketmq20220801:5.0.8` jar and fail-before behavior was reproduced before changing production code.


server/src/main/java/org/apache/rocketmq/studio/provider/alibaba/AliyunCatalogService.java
```diff
@@ -146,6 +155,15 @@ private List<ListInstancesResponseBody.List> fetchAllInstances(Long credentialId
         return all;
```
RockteMQ-AI, line 155: Clean fallback chain: message → code → generic. `Boolean.TRUE.equals(success)` correctly treats null `success` as failure. Good defensive pattern.

RockteMQ-AI (commented):
### Summary
This PR adds `requireCatalogSuccess()` validation to Aliyun catalog API calls (region listing, instance lookup, instance listing). Previously, these calls only checked for null response bodies but didn't validate the `success` flag in the response body. A business-level error from the Aliyun API could silently return empty results instead of surfacing the failure.

### Findings
- **[Info]** `AliyunCatalogService.java:155-163` — The `requireCatalogSuccess()` helper has a clean fallback chain: message → code → generic message. Using `Boolean.TRUE.equals(success)` correctly handles null `success` as failure.
- **[Info]** The pattern is applied consistently across all 3 catalog API call sites.

### Strengths
- Defensive programming that catches business-level API failures that would otherwise be silently swallowed.
- The helper method centralizes the error-handling logic, making it easy to apply consistently.
- Test coverage includes both success and failure scenarios.

Overall: Good defensive improvement for cloud API reliability.

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. Aliyun catalog API calls now validate responses with `requireCatalogSuccess()` before using the data. This prevents silent failures when the catalog returns error codes. The validation is applied consistently across region list, instance list, and endpoint queries. Good defensive fix with test coverage.

LGTM.

---

lizhimins (changes requested):
Thanks — the coverage is complete and the tests are mutation-sensitive.

**What we verified.** `AliyunCatalogService` makes exactly three `clientFactory.call` invocations — `listRegions`, `getInstance` and the paged `listInstances` loop — and all three are guarded, so nothing is missed within the file. Removing any one `requireCatalogSuccess(...)` call reddens the matching test. The direction also has precedent in this very file, which already throws 502 when the catalog reports a negative `totalCount`. And of the three Aliyun response-validation PRs this one defends the most visible failure: when the region/instance catalog read fails, the import wizard shows an empty instance list, which is indistinguishable from "this account genuinely has no 5.x instances".

**Blocking items.**

1. **`requireCatalogSuccess` is a line-for-line copy of `requireReadSuccess` from #4504.** The only differences are the message prefix ("Aliyun catalog … failed") and the blank-check spelling. The project convention puts shared logic under `common/util` rather than duplicating it per service; please extract one helper and land the catalog and read changes as a single PR. #4499 (the mutation path) has been closed on separate grounds, so this and #4504 are the surviving pair and should not each carry their own private implementation. There is no textual conflict between the two — they touch different files — so the ordering is free, but the consolidation is not optional.

2. **The rejection rule rejects a missing flag, not just an explicit failure.** `Boolean.TRUE.equals(success)` is false when `success` is null, and `success` is a nullable `Boolean` in `alibabacloud-rocketmq20220801`. On the catalog path that means a region or API version that omits the field empties the import wizard for every Aliyun credential at once. Please reject only on an explicit `Boolean.FALSE`, and add a test pinning the tolerant direction (`success == null` still returns the catalog) — there is no such case today, which is the branch most likely to regress later.

3. **Use `StringUtils.hasText(...)`.** The new helper writes `message != null && !message.isBlank()`; the convention for blank checks in this project is Spring's `StringUtils`, which the sibling helper in #4504 already uses. `matchesSearch` in this same file uses a bare `isBlank`, so you were following local style — but that is existing code, and new code should follow the convention. One-line change.

4. **A design question to answer in the PR, not a defect.** The convention for query APIs here is that only RPC-layer failures throw; a successful HTTP call carrying a business-level failure signal should degrade with data plus a flag, so the frontend does not present it as an outage. `success=false` on a 200 is exactly that kind of signal, and this PR escalates it to 502. Transport and HTTP failures are already mapped by `AliyunClientFactory.mapToBusinessException` to 422/403/404/502/504, so the only genuinely new coverage is "HTTP 200 with `success=false`" and "null response/body". If you have a real capture of the former for `ListRegions`, `GetInstance` or `ListInstances`, attach it to the PR or an issue — that is what would justify the 502. Without it, please scope the change to the null-response case plus explicit `Boolean.FALSE`.

Once items 1 to 4 are addressed we can merge. The branch is already based on `rocketmq-studio`, so no retargeting is needed.

zmuxuny:
Consolidated into #4504 as requested in review. The revised #4504 contains all three catalog paths and the provider reads with one shared helper, omitted-success compatibility and a consistent missing-body policy. It fixes both #4500 and #4505.

Current head: `07f4989`; 171/171 focused tests passed, and the combined check with #4514 passed 246/246. This PR is closed as superseded, not merged. Please follow #4504 for the implementation and verification; issue #4505 remains tracked there.

PR #4508, opened 2026-09-17
[ISSUE #4506] fix(ai): reject ambiguous physical cluster names

### LINKED ISSUE #4506: [Studio][Bug] Reject ambiguous platform cluster names across instances

- [x] I searched open and closed issues and pull requests for `PlatformClusterResolver`, duplicate cluster names, `DefaultCluster`, and cross-instance cluster resolution.

### Problem

Platform-level AI/MCP tools address infrastructure by physical `clusterName` instead of `instanceId`. `PlatformClusterResolver.collect` currently stores discoveries in `Map<String, ManagedCluster>` and uses `putIfAbsent(clusterName, cluster)`. Managed Apache instances are sorted by instance name, so the first instance wins deterministically.

RocketMQ cluster names are not globally unique across independent deployments; `DefaultCluster` is a common value. If two Studio instances expose different clusters with that name:

- the later instance is silently removed from the platform scan;
- `require("DefaultCluster")` and `resolveInstanceId("DefaultCluster")` select the first instance;
- `rmq.broker.config` and `rmq.nameserver.config` can report configuration from a different instance than the operator intended;
- Broker, NameServer, and Proxy read tools that call `require` inherit the same ambiguity.

The current tools are read-only, so this is an inventory and diagnostic correctness issue rather than a wrong-cluster mutation.

### Proposed Design

Keep the existing tool schemas and the existing deduplicated `rmq.cluster.list` output for compatibility, but make single-cluster resolution fail closed:

1. Preserve the per-instance scan results internally before cluster-name deduplication.
2. Keep `scan()` / `scanWithBrokerVersions()` deduplicated by cluster name as today.
3. Make `require(clusterName)` resolve against the per-instance results.
4. Return 404 when there are no matches, the unique match when exactly one instance owns the name, and 409 when multiple instances expose the same name.
5. Include the sorted conflicting instance identifiers in the 409 message so operators can correct their Studio configuration or choose an instance-scoped workflow.
6. `resolveInstanceId` inherits the same fail-closed behavior.

Adding an explicit instance discriminator to every platform tool is a larger contract enhancement and remains out of scope for this fix.

### Acceptance Criteria

- A unique cluster name resolves exactly as before.
- A cluster name exposed by two managed instances returns HTTP/business code 409 instead of selecting the first instance.
- The conflict result identifies both instance names deterministically.
- Existing platform scan deduplication, unreachable-instance handling, and broker-version enrichment remain unchanged.
- Focused resolver and handler tests, the backend suite, Checkstyle, and `git diff --check` pass.

AI-assisted static analysis; no live cluster mutation is involved.

Aias00:

- Fixes #4506

### Brief Description

- Preserve per-instance cluster discoveries internally before the existing name-based scan deduplication.
- Define physical cluster identity from normalized NameServer address and Broker address sets.
- Collapse duplicate discoveries when multiple Studio instances point at the same physical cluster, preserving deterministic representative-instance selection.
- Return 409 only when a cluster name maps to distinct physical topologies, and include sorted instance identifiers in the error.
- Warn during platform scans when a name maps to distinct physical clusters, while keeping `scan`, `scanWithBrokerVersions`, unreachable-instance handling, and tool schemas unchanged.

### Contract Direction

The seven existing platform tools currently resolve `clusterName` through this compatibility path. Hardening that path is still required so current clients cannot silently target the wrong physical cluster. Migrating those tools to an explicit required `instanceId` is the intended long-term direction, but it changes the published tool schemas and belongs in a separate compatibility/migration change.

- Red/green regression: the shared-physical-cluster case initially returned 409 and the conflict scan emitted no warning; both pass after the fix.
- `mvn -B -ntp -Dtest=PlatformClusterResolverTest,ClusterListToolHandlerTest,BrokerConfigToolHandlerTest,BrokerDescribeToolHandlerTest,BrokerListToolHandlerTest,NameServerConfigToolHandlerTest,NameserverListToolHandlerTest,ProxyConfigToolHandlerTest,ProxyListToolHandlerTest test` with Java 21: 27 tests passed; Checkstyle 0 violations.
- Full backend suite with Java 21 and an isolated H2 MySQL-compatible datasource: 3077 tests, 0 failures/errors, 3 skipped; Checkstyle 0 violations.
- `mvn -B -ntp -DskipTests package`: passed.
- `git diff --check`: passed.
- Independent read-only review: 0 findings.

### Checklist

- [x] New UI text has both Chinese and English entries (no UI change)
- [x] Architecture constraints stay green
- [x] Documentation touched where behaviour changed (PR contract direction documents the compatibility decision)


server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/support/PlatformClusterResolver.java
```diff
@@ -114,16 +114,28 @@ public List<ManagedCluster> scanWithBrokerVersions() {
...
                 .filter(cluster -> normalized.equals(cluster.clusterName()))
-                .findFirst()
-                .orElseThrow(() -> new BusinessException(404, "Cluster not found: " + normalized));
+                .toList();
+        if (matches.isEmpty()) {
+            throw new BusinessException(404, "Cluster not found: " + normalized);
+        }
+        if (matches.size() > 1) {
+            String instances = matches.stream()
+                    .map(ManagedCluster::instanceId)
```
RockteMQ-AI, line 131: `matches.getFirst()` requires Java 21+. Please verify the project's minimum Java version. If targeting Java 17 or earlier, use `matches.get(0)` instead.

server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/support/PlatformClusterResolver.java
```diff
@@ -143,16 +155,22 @@ public static List<String> splitEndpoints(String endpoint) {
 
     private List<ManagedCluster> collect(boolean withVersions) {
         Map<String, ManagedCluster> unique = new LinkedHashMap<>();
+        collectAll(withVersions).forEach(cluster -> unique.putIfAbsent(cluster.clusterName(), cluster));
```
RockteMQ-AI, line 158: Clean separation: `collectAll()` for raw collection, `collect()` for deduplication. Good refactoring.

RockteMQ-AI (commented):
### Summary
This PR improves the `PlatformClusterResolver.require()` method to detect when a cluster name maps to multiple instances across different cloud accounts, throwing a 409 Conflict instead of silently returning the first match. This prevents ambiguous routing in multi-instance deployments.

### Findings
- **[Warning]** `PlatformClusterResolver.java:131` — `matches.getFirst()` requires Java 21+. Please verify the project's minimum Java version supports this. If the project targets Java 17 or earlier, use `matches.get(0)` instead.
- **[Info]** `PlatformClusterResolver.java:158-162` — The new `collectAll()` method is a clean separation of concerns: raw collection vs deduplication. Good refactoring.
- **[Info]** `PlatformClusterResolverTest.java` — The ambiguity test case with two instances returning the same cluster name is well-structured.

### Strengths
- The 409 response with a clear error message listing the conflicting instances is excellent for debugging.
- The `collectAll()` extraction avoids duplicating the instance-scanning logic.
- `collect()` still deduplicates via `putIfAbsent`, preserving backward compatibility for the list endpoint.

Overall: Good data-integrity improvement with clear error messaging.

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. `PlatformClusterResolver.require()` now correctly detects and rejects ambiguous cluster names that match multiple platforms. This prevents silent misrouting of AI tool operations. The error message clearly identifies the conflicting platforms. Good defensive fix with test coverage.

LGTM.

---

lizhimins (changes requested):
The underlying defect is real, but the 409 trigger as written introduces a worse regression than the one it fixes.

**The defect you identified is genuine.** `PlatformClusterResolver.require()` is `scan().stream().filter(...).findFirst()`, and `scan()` already deduplicates by cluster name while `manageableInstances()` sorts by instance name — so when two instances both expose `DefaultCluster`, the seven platform tools behind `require` / `resolveInstanceId` (`rmq.broker.config`, `rmq.broker.describe`, `rmq.broker.list`, `rmq.nameserver.config`, `rmq.nameserver.list`, `rmq.proxy.config`, `rmq.proxy.list`) silently get whichever instance sorts first. Keeping `collectAll()` separate from the deduplicating `collect()` is the right shape.

**But `matches.size() > 1 → 409` is the wrong criterion.** Ambiguity means *two distinct physical clusters sharing a name* — not "two instances returned the same name". Two Studio instances pointing at the **same** NameServer are one physical cluster, and that topology is explicitly supported:

- the instance-model spec states one cluster can host multiple instances and multiple clusters can share a NameServer;
- `rmq_instance` in `server/src/main/resources/db/schema.sql` has a unique key on `name` only — there is **no** uniqueness constraint on `endpoint`;
- this repo's own `PlatformClusterResolverTest.scanShouldAggregateAndDedupeAcrossInstancesTest` sets up exactly that topology (`instance-a` and `instance-b` sharing `ns-a:9876`, both returning `shared-cluster`) and treats the deduplication in `scan()` as correct behaviour, not a conflict.

As written, `require("DefaultCluster")` would answer 409 in that legal configuration, and because `resolveInstanceId` inherits it, all seven platform tools become unusable there. That is a bigger regression than the silent first-instance pick.

**Please detect real conflicts instead of counting instances:**

1. Collapse matches that are the same physical cluster (identical `nameServerAddrs` and broker address set) before deciding, and only raise 409 when the surviving matches genuinely differ.
2. Add the missing regression test for the shared-NameServer case — `require` still resolves normally when two instances expose the same cluster through the same endpoint. Without it nothing in the suite guards against this false positive.
3. `rmq.cluster.list` still deduplicates by name, so a real conflict stays invisible until a tool call fails; a `log.warn` in `collectAll` when distinct clusters share a name would make the 409 diagnosable.

One directional question worth answering in the PR body: this hardens the `clusterName → instanceId` resolution path, which the instance-model convention intends `instanceId` to replace. Is hardening it still wanted, or should the effort go into passing `instanceId` through these tools instead?

Aias00:
Addressed the requested topology distinction in `0ddd83a0f` after rebasing onto `rocketmq-studio@cd448e17f`:

1. `require` now canonicalizes the NameServer address set and Broker address set for every same-name match. Multiple Studio instances with identical physical identity collapse to the deterministic first instance; only distinct physical identities return 409.
2. Added a regression where two instances expose `DefaultCluster` through the same two NameServers in different order and the same Broker set. It resolves to `instance-a` instead of returning 409.
3. `collectAll` now logs a warning when a cluster name maps to distinct physical topologies, including the sorted instance identifiers.
4. The 409 text now explicitly says the ambiguity is across physical clusters.

The current seven platform tools still publish `clusterName`, so this PR keeps the resolver hardening as a compatibility safeguard. Moving those tools to required `instanceId` is the long-term direction, but it is a separate schema migration rather than part of this bug fix.

Verification: focused resolver/tool tests 27/27; full backend suite 3077 passed with 3 skipped; Maven package passed; Checkstyle 0 violations; `git diff --check` passed; independent review found 0 issues.

server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/support/PlatformClusterResolver.java
Aias00: Verified against `server/pom.xml`: this module targets Java 21 (`<java.version>21</java.version>`), so `List.getFirst()` is supported. The updated branch was compiled, packaged, and fully tested with JDK 21 (3077 tests passed).

server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/support/PlatformClusterResolver.java
Aias00: Retained the `collectAll()` / `collect()` separation. `collectAll()` now also diagnoses real same-name physical conflicts, while `collect()` preserves the existing name-deduplicated list contract.

PR #4511, opened 2026-09-17
[ISSUE #4509] fix(acl): show loading while filters refresh

### LINKED ISSUE #4509: [Studio][Bug] ACL tables keep stale rows actionable while filters reload

- [x] I searched open and closed issues and pull requests for ACL filter loading, stale rows, and request ownership.

### Problem

The ACL page keeps `rulesLoading` and `usersLoading` as one-time initial-load flags. The combined fetch effect starts a new server request whenever the instance, filters, pagination, or refresh keys change, but it never sets either loading flag back to `true`; the completion handlers only set them to `false`.

After the first load, changing a rule filter or the user search keyword therefore leaves the previous rows and totals rendered without a loading overlay while the new request is pending. Row edit/delete/admin actions remain visible against records that no longer match the active filter. If the replacement request fails, those old rows remain indefinitely under the new filter until another successful refresh.

This is distinct from #3306/#3300, which covered server-page refresh after ACL user create/delete. It is also not a stale-response race: React effect cleanup already prevents an older request from publishing after its dependencies change. The missing state is the visible loading boundary for the current request.

### Proposed Design

Keep the existing request effect and API behavior:

1. Set `rulesLoading=true` immediately before each rule-page request.
2. Set `usersLoading=true` immediately before each user-page request.
3. Keep the existing mounted cleanup and `finally` ownership checks.
4. Keep current rows mounted behind the Ant Design loading overlay so table dimensions remain stable and actions are blocked during refresh.

Splitting the combined effect or adding debounce behavior is outside this fix.

### Acceptance Criteria

- Changing any rule filter or rule page shows the rules-table loading state until the current request settles.
- Changing the ACL user keyword or user page shows the users-table loading state until the current request settles.
- The loading state clears on both success and failure for the current effect.
- Existing stale-response cleanup, pagination correction, mutation refresh, and instance switching remain unchanged.
- Deferred-promise frontend regressions fail on the current code and pass after the fix.
- Focused tests, the frontend suite, lint, build, Prettier, and `git diff --check` pass.

AI-assisted analysis; behavior was verified from the current effect and table loading bindings.

Aias00:
## Summary

- re-enter the ACL rules and users table loading states whenever the server-backed query effect refreshes
- keep existing rows mounted behind the Ant Design loading overlay while replacement data is pending
- add deferred-promise regression coverage for rule filters and user keyword searches

## Validation

- `npm test -- src/pages/instance/__tests__/AclPage.test.tsx --silent` (25/25)
- `npm run lint` (0 errors; 10 pre-existing warnings outside this change)
- `npm run build`
- `npx prettier --check src/pages/instance/acl.tsx src/pages/instance/__tests__/AclPage.test.tsx`
- `git diff --check`
- independent code review: no findings

The full frontend suite completed 1035/1036 tests. Its only failure was the existing timing-sensitive ConsumerPage case `keeps the latest client stack when an older request resolves last`; that unrelated test passed in three consecutive isolated runs.

Closes #4509


RockteMQ-AI (commented):
### Summary
This PR adds proper loading state indicators to the ACL page tables when filtering by keyword. Previously, users had no visual feedback while the table data was being fetched after a search. The fix tracks loading state per table and shows a spinner during pending requests.

### Findings
- **[Info]** `acl.tsx` — The loading state is tracked via `useState` and reset in both the success and error paths of the `fetchData` callbacks. This is correct.
- **[Info]** `AclPage.test.tsx` — Well-structured tests using `deferred()` to control promise resolution timing. Both the rules table and users table loading states are covered.

### Strengths
- Focused, minimal change that addresses a real UX gap.
- Tests use `deferred()` pattern to precisely control async timing — good practice for testing loading states.
- The loading state is correctly scoped per-table (rules vs users), avoiding cross-contamination.

Overall: Clean UI fix with good test coverage.

---

- **[INFO]** `web/src/pages/instance/acl.tsx:45` — Loading state tracking per-table is clean. Make sure the `loading` state is also reset if the component unmounts during a pending request to avoid React state-update-on-unmounted warnings. _(line outside diff)_

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The ACL page now shows proper loading indicators while filters refresh, preventing confusing empty states. Clean React implementation with proper state management. Test coverage verifies the loading state transitions.

LGTM.

---

lizhimins (changes requested):
The defect and the fix direction are both correct — but please consolidate this with #4518 and #4527 into a single PR, and drop the microtask wrapper.

**The defect is real and reachable.** `web/src/pages/instance/acl.tsx` initialises `rulesLoading` / `usersLoading` from `hasSelectedInstance` and only ever sets them back to `false` in the two `.finally` blocks; the merged fetch effect never re-enters the loading state. So after the first load, changing any rule filter or the user keyword leaves the previous rows and totals rendering as if they matched the new query, with the inline edit / delete / set-admin buttons live on records that no longer match — and on failure the stale rows stay indefinitely. `Table loading={rulesLoading}` / `loading={usersLoading}` already exist, so resetting the flags is enough to get the mask. Setting both flags together is also right, since the merged effect always re-issues the rule page and the user page requests together.

**Change 1 — consolidate the three pages into one PR.** #4511 (ACL page), #4518 (alert rules page) and #4527 (system alerts page) are the same fix — re-enter the loading state at the top of the server-side pagination/filter effect so stale rows cannot be acted on — applied to three pages, each currently carrying its own copy of the identical microtask block. Please submit one PR covering all three (one commit per page is fine). That way the pattern lands once, gets reviewed once, and cannot drift between pages; #4527 in particular already diverges from the other two in how it renders the loading state. We will close the remaining two as superseded once the combined PR is up.

**Change 2 — replace the microtask wrapper with a synchronous state update.** The new block:

```ts
  if (!mounted) return;
  setRulesLoading(true);
  setUsersLoading(true);
});
```

defers the flag reset to a microtask for no benefit — nothing between the effect body and the `.then` can observe the intermediate state, and the `mounted` guard is not needed because the effect body runs synchronously during the same commit. Set the flags synchronously at the top of the effect with a one-line comment explaining why (otherwise the previous page's rows stay actionable while the new query is in flight). `web/src/pages/ops/alerts.tsx` already does exactly this for its domain-switch path, including the comment — please match that style and comment density. If the lint rule complains, an `// eslint-disable-next-line react-hooks/set-state-in-effect` is acceptable.

Your two tests assert that `.ant-spin-spinning` appears and disappears, so they are unaffected by the synchronous change — no test edits needed.

Aias00:
Superseded by the consolidated loading-state fix in #4527, which now includes this ACL change on top of the current `rocketmq-studio` branch together with the alert-rule and system-alert variants.

PR #4512, opened 2026-09-17
fix(tencent): preserve unknown consumer lag

### LINKED ISSUE #4510: [Studio][Bug] Tencent missing consumer lag is reported as healthy zero backlog

- [x] I searched open/closed issues and PRs for Tencent `ConsumerLag`, unknown lag, zero backlog, and cloud metrics handling.

### Problem

Tencent RocketMQ 5.x `SubscriptionData.ConsumerLag` is nullable. Tencent's current API documentation explicitly states that null means no valid value can be obtained. Studio currently converts that state to zero in both:

- `TencentInstanceProvider.getGroupProgress`;
- `TencentInstanceProvider.toTopicConsumer`.

The frontend already treats `-1` as the backend's unavailable-lag sentinel, while a real `0` is rendered as a healthy zero backlog.

The second layer is `CloudRocketMqBusinessMetricsCollector`: it applies `Math.max(0, diffTotal)`, so even after a provider reports the established negative unknown sentinel, the cloud metric path would convert it back to zero.

### Impact

A missing Tencent lag value can be presented as a trustworthy `0` in Consumer progress and Topic-consumer views. The same unknown value can enter native cloud metrics as AVAILABLE zero, which can suppress backlog alerts and make an unavailable measurement look healthy.

### Expected behavior

- Map Tencent `ConsumerLag == null` to the existing unknown-lag sentinel (`-1`), not zero.
- Preserve genuine `ConsumerLag == 0` as zero.
- Cloud business metrics must treat any unknown row as unavailable for aggregate group lag/max metrics rather than clamp it to zero.
- Per-topic backlog samples should preserve known topics and mark an unknown topic unavailable.

### Verification plan

Add red-first provider tests for null lag in group progress and topic-consumer mapping, plus collector tests for mixed known/unknown progress. Verify UI/API sentinel semantics remain compatible and known-zero behavior is unchanged.

AI-assisted source audit. Tencent's current `SubscriptionData` documentation states that `ConsumerLag` may return null when a valid value cannot be obtained.

zmuxuny:

- Fixes #4510

### Brief Description

Tencent RocketMQ 5.x documents `SubscriptionData.ConsumerLag` as nullable when no valid value can be obtained. Studio previously converted that state to `0`, making an unavailable measurement look like healthy zero backlog.

This change preserves the existing unknown-lag contract end to end:

- `TencentInstanceProvider.getGroupProgress`: null lag becomes `-1` instead of `0`.
- Tencent Topic-consumer mapping keeps `diffTotal=-1` and sets `metricsAvailable=false`, so the existing UI renders unavailable rather than `0`/`-1` as a quantity.
- `CloudRocketMqBusinessMetricsCollector` no longer clamps negative unknown lag to zero.
- If any progress row is unknown, group `consumer.lag.total` and `consumer.lag.max_queue` are UNAVAILABLE.
- Topic backlog remains granular: known topics keep valid samples, while a topic containing an unknown row is UNAVAILABLE.

Known positive/zero lag semantics are unchanged.

### Red / Green Verification

Fail-before:
- Tencent group progress with `ConsumerLag=null`: expected `-1`, got `0`.
- Tencent Topic-consumer row with `ConsumerLag=null`: expected `-1`/unavailable, got `0`/available.
- Cloud metrics with mixed known + unknown progress: expected aggregate UNAVAILABLE, got AVAILABLE.
- Focused run: **49 tests, exactly 3 failures**; all existing cases passed.

Green, Java 21:
- `TencentInstanceProviderTest,CloudRocketMqBusinessMetricsCollectorTest`: **49/49 passed**.
- Plus `ApacheRocketMqBusinessMetricsCollectorTest,NativeAlertProcessorTest`: **76/76 passed**.
- `git diff --check`: clean.

### Provider contract

Tencent's current `SubscriptionData` documentation states that `ConsumerLag` may be null when a valid value cannot be obtained:
https://cloud.tencent.com/document/product/1493/96031

This PR therefore distinguishes provider-unavailable data from a genuine zero backlog; it does not change the Tencent API calls or add dependencies.

AI-assisted source audit, implementation and regression authoring; all red/green commands above were executed locally against the stated baseline.

### Audit follow-up

A full re-review found one scope-preservation edge in the original metric patch: per-group UNAVAILABLE samples did not carry the Consumer Group's `clusterId`. A native alert rule scoped with `clusterName` could therefore fail to match the unavailable sample during reconciliation.

The current head preserves `group.clusterId` on per-group and per-topic UNAVAILABLE samples while keeping whole-instance collection-failure markers unscoped. The regression now asserts that scope as part of the availability contract.

Current-head verification on Java 21: `TencentInstanceProviderTest,CloudRocketMqBusinessMetricsCollectorTest,NativeAlertProcessorTest` **70/70 passed**; Checkstyle 0; `mvn -B -ntp -DskipTests package` BUILD SUCCESS; `git diff --check` clean.


server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/CloudRocketMqBusinessMetricsCollector.java
```diff
@@ -87,19 +87,30 @@ private static List<MetricSample> collectGroup(InstanceProvider provider, Instan
         Map<String, String> labels = Map.of("consumerGroup", group.getName());
         try {
             List<QueueProgressVO> progress = provider.getGroupProgress(instance.getName(), group.getName());
-            double totalLag = progress.stream().mapToDouble(row -> Math.max(0L, row.getDiffTotal())).sum();
-            double maxQueueLag = progress.stream().mapToDouble(row -> Math.max(0L, row.getDiffTotal())).max()
-                    .orElse(0D);
+            boolean lagUnknown = progress.stream().anyMatch(row -> row.getDiffTotal() < 0);
```
RockteMQ-AI, line 90: The `lagUnknown` scan iterates the full `progress` list, and then the subsequent `if/else` block iterates it again. Consider combining into a single pass if this list can be large in production. Minor performance note — not a blocker.

RockteMQ-AI (commented):
### Summary
This PR correctly preserves the Tencent provider's "unknown consumer lag" contract end-to-end. Previously, null `ConsumerLag` was silently converted to `0`, making unavailable measurements look like healthy zero backlog. The fix distinguishes provider-unavailable data from genuine zero backlog, which is important for operational correctness.

### Findings
- **[Info]** `CloudRocketMqBusinessMetricsCollector.java:90` — The `lagUnknown` scan and the subsequent computation each iterate `progress` separately. Consider a single-pass approach if this list can be large in production. Minor — not a blocker.
- **[Info]** `TencentInstanceProvider.java:126` — Good sentinel mapping. `-1` is consistent with existing `QueueProgressVO` semantics.

### Strengths
- Red/green verification with explicit baseline is excellent practice.
- Topic-level granularity (known topics keep valid samples, unknown rows → UNAVAILABLE) is well thought out.
- Test coverage for the new behavior is thorough (38 new test lines).
- PR description clearly documents the contract and references the provider documentation.

Overall: Solid bug fix with good test coverage and clear documentation.

---

- **[INFO]** `server/src/main/java/org/apache/rocketmq/studio/provider/tencent/TencentInstanceProvider.java:126` — Good fix. Mapping null `ConsumerLag` to `-1` preserves the provider-unavailable contract end-to-end. The sentinel value `-1` is consistent with the existing `QueueProgressVO` semantics. _(line outside diff)_

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The Tencent provider now correctly preserves the unknown consumer lag contract — when the API returns no lag data, the `UNKNOWN` sentinel is propagated rather than defaulting to 0. The changes span both the provider and the metrics collector, with proper test coverage for both layers.

LGTM.

---

zmuxuny:
Audit follow-up: I found one scope-preservation edge while re-reviewing the open PR set. Per-group UNAVAILABLE cloud lag samples were dropping the group's `clusterId`, so a native alert rule with `clusterName` could fail to match the unavailable sample during reconciliation.

The branch now preserves `group.clusterId` on per-group/group-topic UNAVAILABLE samples while keeping whole-instance failure markers unscoped. The existing unknown-lag test now asserts the cluster id as part of the availability contract.

Re-verified on Java 21: `TencentInstanceProviderTest,CloudRocketMqBusinessMetricsCollectorTest,NativeAlertProcessorTest` **70/70 passed**, Checkstyle 0; `mvn -B -ntp -DskipTests package` BUILD SUCCESS; `git diff --check` clean.

RockteMQ-AI (approved):
### Summary
LGTM — the updated commit properly preserves unknown consumer lag end-to-end for the Tencent provider. The `toConsumerLag()` helper cleanly maps null → -1, and the collector correctly marks group-level and per-topic metrics as UNAVAILABLE when any queue has unknown lag. The `clusterId` propagation fix in the `unavailable()` overload is consistent with the Apache-side fix (#4517). Tests cover both the provider mapping and collector aggregation.

---

lizhimins:
Thanks for the detailed PR and for linking the Tencent documentation on `ConsumerLag` being nullable — that part is accurate, and the change is well built.

To summarise what it does, so we are talking about the same thing: a null `SubscriptionData.ConsumerLag` on the Tencent path stops mapping to `0` and instead becomes `-1` through the new `toConsumerLag` helper (both call sites, group progress and topic consumer); the topic-consumer path additionally derives `metricsAvailable` from that value; and `CloudRocketMqBusinessMetricsCollector` drops its `Math.max(0, ...)` clamp so an unknown row turns `consumer.lag.total` and `consumer.lag.max_queue` into `UNAVAILABLE` samples, with `topic.backlog.total` distinguished per topic and `clusterId` attached to the unavailable samples. Reusing the existing `ConsumerLagResolver.UNKNOWN` / `metricsAvailable` contract rather than inventing a new one was the right instinct, and the three new tests are mutation-sensitive.

**We are not taking this one.** On our reading, when the Tencent monitoring API does not report a `ConsumerLag` for a subscription, `0` is the value we want on this surface. This is the same position we gave on #4386 and recorded in #4361: the backlog numbers an operator reads during an incident stay plain numerics, and we do not want a second, overlapping representation of "no data" that has to be kept consistent with the first by hand. To be clear about the boundary of that decision: the `ConsumerLagResolver.UNKNOWN` / `metricsAvailable` contract on the Apache queue and topic lag paths stays exactly where it is — we are simply not extending it to the cloud providers.

Two further things the review noted, which are independent of the decision above:

1. The change was not symmetric across the cloud providers. `AliyunConverters.toQueueProgressRows` still maps a null `readyCount` to `0L`, and `CloudRocketMqBusinessMetricsCollector` serves both Aliyun and Tencent instances — so the new unknown branch would never fire for an Aliyun instance. `AliyunConverters.toTopicConsumerVO` does not set `diffTotal` at all, which defaults to `0` with `metricsAvailable` true.
2. The sentinel was written as the literal `-1L` rather than `ConsumerLagResolver.UNKNOWN`, and the collector's test was `< 0` rather than `== UNKNOWN`, which does not match the predicate #4517 introduced in `ApacheRocketMqBusinessMetricsCollector` (`< 0` also swallows a genuinely negative lag reported by the cloud).

That said, the decision rests on treating "the API reported no value" as "no backlog", and I may be wrong about that. If you have a concrete case where Tencent returns a null `ConsumerLag` while backlog genuinely exists — ideally a captured API response body plus what the console showed at the same moment — please reopen with that reproduction and we will look again. A real capture is exactly the evidence that would change the answer.

PR #4514, opened 2026-09-17
fix(message): flag incomplete cloud query pages

### LINKED ISSUE #4513: [Studio][Bug] Tencent message query treats incomplete provider pages as confirmed empty results

- [x] I searched open/closed issues and PRs for Tencent `DescribeMessageList`, null `Data`, incomplete pages, and `TotalCount` handling.

### Problem

Tencent `DescribeMessageList` exposes both `TotalCount` and a nullable `Data` array. Tencent documents `Data` as possibly null when no valid value can be obtained. Studio currently treats `Data == null` as a zero-length short page and immediately returns a **complete** empty query result.

When the provider simultaneously reports `TotalCount > collected` but `Data` is null/empty, those signals contradict each other: the provider says matching messages remain, but Studio declares the result exhausted.

### Impact

An operator can receive `0 messages` / `resultMayBeTruncated=false` even though Tencent reports matching rows. This is a message-diagnostic data-integrity problem: incident triage can incorrectly conclude that no message matched the requested topic/key/tag/time window.

### Expected behavior

- If `TotalCount > collected` but the current page has no valid `Data`, fail closed instead of returning a confirmed complete result.
- Preserve a genuine empty result when `TotalCount` is zero/absent and the provider returns no rows.
- Preserve #3048's existing bounded-query/truncation semantics and task-request-id continuation.

### Verification plan

Add a deterministic provider regression with `TotalCount=1` and `Data=null` proving current code returns `MessageQueryResult.complete(empty)`. Fix only this contradictory/incomplete response; retain existing empty-page, hard-budget and TaskRequestId tests.

Related #3047/#3048 fixed silent truncation when Studio reaches its own page/result budget; this issue covers a different provider response where an individual page is incomplete while `TotalCount` says data remains.

AI-assisted source audit; Tencent's current API documentation was checked before filing.

zmuxuny:

Fixes #4513.

### Revised behavior

A successful cloud message-query response whose short/empty page contradicts its known total now returns `MessageQueryResult.truncated(collectedRows)`, not 502 and not a falsely complete result. The existing `resultMayBeTruncated` channel carries this to Web/MCP callers. Transport exception mapping remains unchanged.

This revision covers **Tencent and Aliyun**. It preserves previously collected messages, Tencent TaskRequestId/offset progression, result budgets, genuine final/empty pages, and the short-page fallback for absent/negative totals. Aliyun completeness uses fetched provider-row counts before local tag filtering, so filtering alone cannot fabricate truncation.

`Pagination.isIncompletePage` is the shared, overflow-safe predicate. Tencent's existing `requireCompletePage` delegates to it for resource inventories; those inventory paths retain their existing fail-closed policy. Message queries use the same predicate but publish the truncation flag, as requested. The duplicate loop comment was removed.

### Apache scope

Apache `RocketMQMessageProvider` does not use a cloud TotalCount/page-number response. It scans queue offsets using PullResult/nextBeginOffset and scan budgets; the key-query path already has its own limit flag. Changing all early exits and the list-returning topic scan would be a separate scan-contract change, so Apache production behavior is deliberately unchanged here. This PR does not claim to fix every Apache scan-incompleteness case.

### Verification

Rebased onto `rocketmq-studio@4c697f07acde460e2344375cb1f82669f5b270fd`.

Before correction: the Tencent suite had 54 cases, with exactly 3 new-behavior failures (1 assertion failure and 2 old-502 exceptions); the Aliyun suite had 37 cases, with exactly 2 contradictory-short-page assertion failures. These are local controlled-response tests, not production API captures.

After correction: `PaginationTest,TencentInstanceProviderTest,AliyunInstanceProviderTest,MessageServiceTest,MessageControllerTest,MessageQueryToolHandlerTest,MessageQueryByTopicToolHandlerTest,ToolOutputSchemaContractTest` **138/138 passed**. This includes non-empty short pages, empty initial/later pages, retained earlier rows/task ID, local tag filtering, existing budget behavior and downstream tool schemas.

Java 21; Checkstyle **0 violations**; `mvn -B -ntp -DskipTests package` **BUILD SUCCESS**; `git diff --check` clean. No frontend, schema or dependency changes. Full external-datasource integration tests were not run. AI-assisted implementation and review; listed tests/build executed locally.

### Cross-PR integration check

The final #4504 (`07f4989`) and #4514 (`858fe91`) revisions merge cleanly. A separate local combined snapshot ran the union of their shared-helper/provider/service/controller/MCP-schema suites: **246/246 passed**, 0 failures/errors/skips, Checkstyle 0. This is a local integration check of the two PRs, not an upstream merge.


RockteMQ-AI (approved):
The fix fails closed with a 502 error when `TotalCount > collected` but the current page returns zero rows, while preserving the legitimate empty case (`TotalCount=0` with no rows).

### Findings
- **[Info]** `TencentInstanceProvider.java:626-630` — Clean fail-closed guard. The condition `returned == 0 && total > result.size()` precisely targets the incomplete response case without affecting genuine empty results or the existing #3048 budget truncation.
- **[Info]** Including `requestId` in the error message is good for debugging provider-side issues.
- **[Info]** `TencentInstanceProviderTest.java:941-970` — Two well-designed tests: one for the error case (TotalCount=1, Data=null -> 502) and one for the legitimate empty case (TotalCount=0 -> complete empty).

LGTM — minimal, focused fix with correct fail-closed semantics.

---

lizhimins (changes requested):
Thanks — the defect you found is real, and it is the kind that misleads an operator rather than merely annoying them.

**What we verified.** In `TencentInstanceProvider.queryMessagesDetailed`, `isLastPage(returned, total, collected)` treats `returned == 0` as a short page and breaks out of the paging loop. When the first page comes back empty while `TotalCount` reports, say, 500, the method falls through to `MessageQueryResult.complete(result)`, so the caller is told "query complete, zero messages" by a provider that just said there were 500. On the message page that reads as message loss. Both new tests are mutation-sensitive: removing the inserted guard reddens `queryMessagesShouldRejectIncompleteProviderPageTest`.

**Blocking items.**

1. **Please do not throw 502 here — the codebase already has the right channel.** `MessageQueryResult.truncated(...)` exists precisely for "we stopped before the provider said we were done"; `MessageService` surfaces it as `resultMayBeTruncated`, and the same loop already sets `mayBeTruncated = true` when `MESSAGE_QUERY_HARD_LIMIT` is hit. The project convention for query APIs is that only RPC-layer failures throw; a successful RPC carrying an inconsistent business payload should degrade with a flag, not an error. Since #4388 landed, `resultMayBeTruncated` and `skippedCount` are also part of the AI tool output contract for `rmq.message.query` and `rmq.message.query_by_topic`, so the flag reaches the MCP surface too — a 502 does not. Your own second test (`queryMessagesShouldKeepConfirmedEmptyProviderPageCompleteTest`) already asserts against the `mayBeTruncated()` channel, which is a good sign the flag is the natural fit.

2. **An empty first page can be transient.** `DescribeMessageList` is an async, task-based query — the comment above the loop explains that each logical query is identified by a `TaskRequestId`, that paging reuses the id returned by the previous page, and that a fresh random id starts a brand-new query. While the task is still being prepared, "empty data plus a reported total" is a plausible intermediate state. A hard 502 turns a retryable "not ready yet" into a user-facing failure; `mayBeTruncated = true` plus a break is an acceptable degradation for the same condition.

3. **The same shape exists in the other two providers.** `AliyunInstanceProvider.queryMessagesDetailed` computes `shortPage = list.size() < MESSAGE_PAGE_SIZE`, breaks, and returns `MessageQueryResult.complete(records)`; `RocketMQMessageProvider` has the equivalent path. Please cover them in this PR, or state explicitly in the description why they are out of scope — otherwise the fix reads as complete when it is one provider of three.

4. **Reuse the incompleteness helper that is already in this file.** #4531 has landed and added `requireCompletePage(resource, offset, returned, totalCount)`, now used for topic, consumer-group and consumer-group-subscription paging in `TencentInstanceProvider`. This PR introduces a second, different inline predicate (`returned == 0 && total > result.size()`). One file should not carry two notions of "the page came back incomplete" — please converge on a single helper, whichever shape you decide is correct for the message path.

5. **The tests need rewriting once item 1 is settled.** If the fix becomes a truncation flag rather than a throw, the first test has to assert `mayBeTruncated()` is true and that no exception is raised. Please make that change together with the production change so the intent is unambiguous.

**Optional, not blocking.** The four-line comment above your insertion point ("Stop on the last page … so we do not rely on TotalCount, which may not be populated for every query") is duplicated verbatim twice in that method on trunk. Worth de-duplicating while you are in the loop.

Once items 1 to 5 are addressed we can merge. The branch is already based on `rocketmq-studio`. Item 1 is the one that decides the shape of everything else, so please settle it first — if you have a reason the 502 is genuinely required (for example a caller that cannot act on `resultMayBeTruncated`), say so in the PR and we will discuss it there.

zmuxuny:
Addressed in `858fe917`, rebased onto `rocketmq-studio@4c697f0`.

1. Accepted the truncation channel: contradictory short/empty message pages now set `mayBeTruncated` and retain collected rows, without introducing a 502.
2. Tencent paging still reuses TaskRequestId. New coverage checks a full first page followed by an incomplete empty second page, retained messages and the next offset/task ID.
3. Added the same known-total check for Aliyun message pages, with completeness measured before local tag filtering. Apache's offset/PullStatus scan is deliberately unchanged; its distinct early-exit/scan-budget contract and the limitation are now explained in the PR scope rather than claiming all three providers are fixed.
4. Extracted `Pagination.isIncompletePage`; both the Tencent resource guard and cloud message loops reuse the predicate, while preserving their different error/degradation policies.
5. Replaced the old throw assertion and added short/nonempty, later-empty and local-filter regressions. The duplicated loop comment is removed.

138/138 focused provider/service/controller/MCP schema tests passed, Checkstyle 0, package succeeds. The PR description includes the exact suites and fail-before evidence; no real provider capture or full datasource integration run is claimed.

zmuxuny:
Hi @lizhimins, thank you for the detailed review. I pushed `858fe917` with the incomplete-result flag, Aliyun coverage, a shared page predicate, and an explicit note on the Apache scan path's scope. Could you please take another look when convenient?

PR #4518, opened 2026-09-17
[ISSUE #4515] fix(alert): block stale rule actions while filtering

### LINKED ISSUE #4515: [Studio][Bug] Alert rule filters leave stale rows actionable while pages reload

- [x] I searched open and closed issues and pull requests for alert rule filter loading, stale rows, page refresh, and table request ownership.

### Problem

The native Alert Rules page initializes `loading=true` and explicitly re-enters loading when the route changes between CLUSTER and BUSINESS domains. The normal server-page effect, however, never sets loading back to true when search text, enabled status, pagination, page size, refresh version, or language changes. Its completion path only sets loading to false.

After the first load, each of those changes leaves the previous rows and total visible without a loading boundary while the replacement request is pending. Row enable/edit/delete controls, row selection, and bulk actions remain actionable against records from the previous query scope.

The effect cleanup prevents an older response from publishing after dependencies change, so this is not a stale-response race. The missing behavior is the visible and interactive loading boundary for the current request.

### Impact

Operators can mutate or select rules that no longer belong to the active search, status filter, page, or refreshed inventory. The screen also presents the previous result set as current until the server responds, which is misleading during alert administration.

### Proposed Design

Keep the existing request effect and cancellation behavior:

1. Re-enter `loading=true` whenever the rule-page request effect starts.
2. Keep current rows mounted behind the Ant Design table loading overlay so layout remains stable and actions are blocked while the replacement page is pending.
3. Keep the existing domain-change render-time reset and response cancellation guard unchanged.
4. Do not add search debouncing or change the Alert Rule API in this fix.

Because the repository lint rule rejects synchronous state writes at the top of an effect, schedule the loading transition through the same guarded microtask pattern already used by other server-backed tables.

### Acceptance Criteria

- Search, enabled-status, page, page-size, and refresh changes show the table loading state until the current request settles.
- Existing rows are not actionable while the replacement request is pending.
- Loading clears on both success and failure for the current request.
- Superseded responses remain unable to publish.
- Domain switching continues to clear old-domain rows and filters immediately.
- A deferred-promise frontend regression fails on current master and passes after the fix.
- Focused tests, frontend lint/build, formatting, and `git diff --check` pass.

AI-assisted source audit; behavior was verified from the current effect, table loading binding, and action controls.

Aias00:
## Summary

- re-enter the Alert Rules table loading state whenever the server-backed query refreshes
- block row selection, row mutations, and bulk mutations while replacement rows are pending
- add a deferred filter regression that keeps old rows visible but verifies every stale mutation path is disabled

## Validation

- `npm test -- src/pages/ops/__tests__/AlertsPage.test.tsx --silent` (27/27)
- `npm test -- --retry=2 --reporter=dot --silent` (122 files, 1035/1035)
- `npm run lint` (0 errors; 10 pre-existing warnings)
- `npm run build`
- `npx prettier --check src/pages/ops/alerts.tsx src/pages/ops/__tests__/AlertsPage.test.tsx`
- `git diff --check`
- independent code review: APPROVE after the stale bulk-action guard was added

Closes #4515


RockteMQ-AI (approved):
### Summary
This PR fixes a race condition where user actions (toggle, delete, bulk operations) could fire against stale alert rule data while the table was refreshing from the server.

The fix introduces `isRuleActionBlocked = loading || isActionRunning` and applies it consistently across all mutation entry points. The deferred loading state pattern (`Promise.resolve().then(...)`) ensures existing data remains visible while the loading indicator activates.

### Findings
- **[Info]** `alerts.tsx:363-365` — The microtask-deferred `setLoading(true)` is a clean way to avoid a flash of stale data. Consider adding a brief comment explaining why the deferral is needed for future maintainers.
- **[Info]** `AlertsPage.test.tsx:360-389` — Excellent test coverage. Verifying that all mutation paths (toggle, delete, bulk enable/disable/delete, checkbox) are disabled during loading is thorough.

LGTM — clean race condition fix with comprehensive test coverage.

---

RockteMQ-AI (approved):
### Summary
This PR fixes a UI state management issue in the alerts page. When a filtered page is loading, bulk actions (enable/disable/delete) and individual rule actions are now correctly disabled to prevent stale operations. The fix introduces an `isRuleActionBlocked` flag that combines `loading` and `isActionRunning` states.

### Findings
- **[Info]** `alerts.tsx` — The `isRuleActionBlocked` derivation is clean and correctly prevents actions during loading. The `setLoading(true)` in a microtask ensures the loading state is set before the async fetch starts.
- **[Info]** `AlertsPage.test.tsx` — The test verifies that all action buttons are disabled during loading and that no bulk/individual actions are triggered. Good coverage.

### Suggestions
- The microtask pattern (`Promise.resolve().then(...)`) for setting loading state is a bit unusual. Consider whether `useTransition` or a simpler state management approach might be clearer.

Clean fix for a real UX issue.

---

lizhimins (changes requested):
The defect, the fix and the tests are all correct — but please consolidate this with #4511 and #4527 into a single PR, and drop the microtask wrapper.

**The defect is real, and this file already admits it.** `web/src/pages/ops/alerts.tsx` initialises `loading` to `true` and only resets it in the `.finally`; the domain-switch block sets `loading` synchronously with a comment explaining that otherwise the previous domain's rules stay visible with `loading=false` until the new page resolves. The ordinary pagination effect (deps `[domain, enabledFilter, page, pageSize, rulesVersion, search, t]`) is simply missing the same reset. So after the first load, changing the search term / enabled filter / page / page size, or bumping `rulesVersion` after a mutation, leaves the previous rule set actionable: inline Switch / edit / duplicate / delete, the row-selection checkboxes, and — most dangerously — the bulk enable / disable / delete toolbar operating on a stale selected-id set.

**`isRuleActionBlocked` is necessary, not redundant.** antd's `.ant-spin-blur` sets `pointer-events: none` over the table body, so inline controls are covered by the mask, but the bulk buttons live *outside* the Table and are not. The explicit `disabled={!hasSelectedRules || isRuleActionBlocked}` plus the early-return guards in the handlers are the right layering. Leaving `listAlertRuleRuntime` unguarded is acceptable since it is read-only display and every mutation based on those rows is already blocked.

**Change 1 — consolidate the three pages into one PR.** #4511 (ACL page), #4518 (this one) and #4527 (system alerts page) are the same fix applied to three pages, each carrying its own copy of the identical microtask block. Please submit one PR covering all three (one commit per page is fine) so the pattern lands once, is reviewed once, and cannot drift between pages — #4527 already diverges in how it renders the loading state. We will close the remaining two as superseded once the combined PR is up.

**Change 2 — replace the microtask wrapper with a synchronous state update.** The new block at the top of the pagination effect:

```ts
  if (!mounted) return;
  setLoading(true);
});
```

defers the flag reset to a microtask for no benefit, and the `mounted` guard is unnecessary because the effect body already runs synchronously during the same commit. Set `loading` synchronously, exactly like the domain-switch block in this same file does, and reuse that comment's phrasing so both paths read alike.

Aias00:
Superseded by the consolidated loading-state fix in #4527, which now includes this alert-rule change on top of the current `rocketmq-studio` branch together with the ACL and system-alert variants.

PR #4521, opened 2026-09-17
fix(metrics): preserve unavailable broker runtime fields

### LINKED ISSUE #4519: [Studio][Bug] Missing Broker runtime fields can falsely resolve native metric alerts

- [x] I searched open/closed issues and PRs for missing Broker runtime metrics, false alert recovery, disk/heap/send-queue availability, and native alert reconciliation.
- [x] I verified the control flow on current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`).

### Problem

`ApacheRocketMqClusterMetricsCollector` treats a successful `fetchBrokerRuntimeStats` call as Broker availability, but silently **omits** individual metric samples when `commitLogDiskRatio`, JVM heap fields, or send-thread-pool queue fields are missing/invalid.

`NativeAlertProcessor.processSuccessfulCollection` later reconciles missing active states for the entire cluster-metric scope. Its whole-scope failure guard only stops reconciliation for an UNAVAILABLE sample with empty labels. A missing broker-scoped metric emits no sample at all, so an existing active rule fingerprint is absent from `presentKeys` and can be advanced to RESOLVED.

### Impact

A transiently absent or malformed Broker runtime field can look like a real recovery. For example, a firing `broker.disk.usage_ratio` alert may be marked RESOLVED and enqueue a recovery notification merely because `commitLogDiskRatio` was missing for one collection, while `broker.availability` still reports the Broker as healthy.

The same failure mode applies to:
- `broker.disk.usage_ratio`;
- `broker.jvm.heap.usage_ratio`;
- `broker.send_queue.usage_ratio`.

### Expected behavior

Once Broker runtime stats are reachable, each supported metric must produce either an AVAILABLE value or a **broker-scoped UNAVAILABLE** sample. Missing/invalid individual fields must never be represented by absence.

A broker-scoped UNAVAILABLE sample keeps the rule fingerprint present, skips numeric evaluation, and prevents missing-state reconciliation from generating a false recovery while leaving other Broker metrics usable.

### Verification plan

Add red-first collector coverage proving a missing/invalid runtime field currently produces no metric sample. Add an integration-style NativeAlertProcessor regression: start from an active broker metric state, process a successful collection containing the same broker-scoped metric as UNAVAILABLE, and verify no RESOLVED lifecycle event/recovery notification is emitted.

Preserve genuine Broker unavailability and all valid numeric parsing behavior.

AI-assisted source audit; the collector and reconciliation control flow were verified before filing.

zmuxuny:

- Fixes #4519

### Brief Description

`ApacheRocketMqClusterMetricsCollector` treated a successful Broker runtime RPC as Broker availability, but silently omitted `broker.disk.usage_ratio`, `broker.jvm.heap.usage_ratio`, or `broker.send_queue.usage_ratio` when the corresponding runtime fields were missing or invalid.

Native alert reconciliation uses metric-sample presence to distinguish a disappeared metric from a still-present unavailable metric. Omitting the sample could therefore make an active Broker alert look recovered and emit a false RESOLVED/recovery notification.

This change makes each supported Broker runtime metric explicit: parsing success emits AVAILABLE; missing/invalid input emits a broker-scoped UNAVAILABLE sample with the same `brokerName`/`brokerAddr` fingerprint. Other Broker metrics remain independently usable.

### Red / Green Verification

Fail-before:
- Added a Broker runtime fixture with a reachable Broker and an empty runtime table.
- The regression requires broker-scoped UNAVAILABLE samples for disk, JVM heap, and send queue metrics.
- `ApacheRocketMqClusterMetricsCollectorTest`: **4 tests, exactly 1 failure** because the old collector emitted no sample.

Green, Java 21:
- `ApacheRocketMqClusterMetricsCollectorTest,NativeAlertProcessorTest`: **26/26 passed**.
- Added a cross-module alert contract proving a broker-scoped UNAVAILABLE sample keeps an active fingerprint and emits no RESOLVED event/recovery outbox entry.
- `git diff --check`: clean.

### Compatibility / Risk

Valid numeric runtime values are unchanged. A Broker runtime RPC failure still produces the existing `broker.availability` UNAVAILABLE sample. This patch only replaces silent per-field omission with explicit per-Broker availability, matching the native alert state machine's existing UNAVAILABLE semantics.

No API/schema/dependency/frontend changes.

AI-assisted source audit, implementation and regression authoring; the missing-sample reconciliation path was reproduced before changing production code.


RockteMQ-AI (approved):
### Summary
This PR correctly fixes a subtle bug where missing broker runtime fields (disk usage, heap usage, send queue usage) were silently omitted from metric samples, causing native alert reconciliation to falsely resolve active alerts.

The fix introduces a clean `metricOrUnavailable()` helper that emits UNAVAILABLE samples instead of dropping metrics, preserving alert state continuity. Tests verify both the metric emission and the alert reconciliation behavior.

### Findings
- **[Info]** `ApacheRocketMqClusterMetricsCollector.java:116-121` — Clean refactoring. The `metricOrUnavailable` pattern is a good abstraction for handling optional metrics with explicit unavailability semantics.
- **[Info]** `NativeAlertProcessorTest.java` — Excellent addition of the `unavailableSamplesKeepActiveFingerprints` test, which directly validates the alert reconciliation fix.

LGTM — well-structured fix with thorough test coverage.

---

PR #4525, opened 2026-09-17
fix(message): surface message-id lookup failures

### LINKED ISSUE #4524: [Studio][Bug] Message-ID lookup failures are reported as confirmed empty results

- [x] I searched open/closed issues and PRs for message-id lookup failures, `viewMessage`, empty results, and msgId diagnostics.
- [x] I reproduced the behavior on current `master` (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`).

### Problem

The Apache message query path turns real message-ID lookup failures into a confirmed empty result.

`RocketMQMessageProvider.queryByMsgId` catches every exception from `DefaultMQAdminExt.viewMessage(topic, msgId)`, logs it, then tries the decoded-offset fallback. `viewMessageByOffsetId` also catches every exception and returns `null`. If neither path returns a message, the provider returns an empty list.

For a non-offset/unique message id, the fallback cannot decode the id at all. A NameServer/Broker/RPC failure from the primary lookup is therefore indistinguishable from "message not found". For an offset msgId, a primary RPC failure followed by a direct Broker lookup failure has the same false-empty outcome.

### Impact

Message-ID lookup is an incident-diagnostic path. Returning `0 messages` after a transport/provider failure can make an operator conclude that a message does not exist when Studio actually failed to query it.

Other message query paths already distinguish absence from operational failure: key and unique-key queries return empty only for known no-message response codes and surface unexpected failures as HTTP-style 502 errors.

### Expected Behavior

- Known no-message / query-not-found outcomes remain valid empty results.
- Existing broker-topology rejection for unsafe offset ids remains fail-closed and does not trigger an outbound lookup.
- If the primary lookup fails but the validated decoded-offset fallback succeeds, preserve the successful fallback behavior.
- If an actual lookup is attempted and all applicable lookup paths fail operationally, return a 502-style `BusinessException` instead of a confirmed empty result.

### Red Reproduction

On unmodified `master`, I added one regression where `viewMessage("TopicA", "uniq-key-failure")` throws `IllegalStateException("nameserver unavailable")` and require the provider to surface 502.

`RocketMQMessageProviderTest`: **55 tests, exactly 1 failure**. The new test fails with `Expecting code to raise a throwable`, while the provider logs the failure and returns an empty list.

### Proposed Scope

Keep the public API shape unchanged. Classify known not-found response codes separately from operational failures, preserve the existing offset-id fallback, and add focused tests for unique-key failure, offset fallback failure/success, and genuine not-found behavior.

AI-assisted source audit and regression authoring; the fail-before test was executed locally with Java 21.

zmuxuny:

- Fixes #4524

### Brief Description

Apache message-ID queries could convert lookup errors into an empty result. The primary lookup recorded the error, the decoded-offset fallback also reduced errors to a null result, and the provider then returned an empty list.

This change keeps the existing fallback behavior while distinguishing a known absent message from an operational lookup error. Known no-message response codes stay empty results, broker-topology rejection remains unchanged, and a successful decoded-offset fallback still rescues a failed primary lookup. When all applicable lookup paths fail operationally, Studio now returns a 502-style `BusinessException` rather than `0 messages`.

The trace-timestamp path stays best-effort and keeps its existing one-hour fallback window.
### Red / Green Verification

Fail-before, Java 21:
- Added a regression where `viewMessage("TopicA", "uniq-key-failure")` throws `IllegalStateException("nameserver unavailable")`.
- `RocketMQMessageProviderTest`: **55 tests, exactly 1 failure**.
- The new test failed with `Expecting code to raise a throwable`, proving current master returned empty instead of surfacing the lookup failure.

Green, Java 21:
- `RocketMQMessageProviderTest`: **56/56 passed**.
- `RocketMQMessageProviderTest,MessageServiceTest,MessageControllerTest,MessageQueryToolHandlerTest`: **87/87 passed**.
- `git diff --check`: clean.
### Compatibility / Risk

No API, schema, dependency, or frontend changes. Known absent-message outcomes remain empty; the existing offset-id topology boundary is unchanged; successful fallback lookup behavior is preserved. Only operational lookup failures that were previously reported as confirmed empty now surface as errors.

AI-assisted source audit, implementation and regression authoring; fail-before and green verification were executed locally against the stated baseline.


RockteMQ-AI (approved):
### Summary
Well-structured fix that properly distinguishes "message not found" from "lookup failed" in message-id queries. The `OffsetMessageLookup` record cleanly carries both the result and failure from the fallback path, and the `isMessageLookupAbsent` helper encapsulates the known-absent response codes.

The behavioral change — early-returning empty when `BrokerTopologyGuards` rejects the msgId instead of falling through to the offset lookup — is a good optimization since the fallback would also fail in that case.

Test coverage is solid: verifies that non-offset lookup failures are surfaced, offset fallback failures propagate as `BusinessException(502)`, and known-absent responses still return empty results.

LGTM.

---

PR #4527, opened 2026-09-17
[ISSUE #4509][ISSUE #4515][ISSUE #4526] fix(ui): block stale filtered-page actions

### LINKED ISSUE #4526: [Studio][Bug] System Alert filters keep stale events actionable while the feed reloads

- [x] I searched open and closed issues and pull requests for System Alert filter loading, stale event rows, pending acknowledgement, and feed refresh behavior.

### Problem

The System Alerts page initializes `loading=true`, but its server-page effect never sets loading back to true when the level, domain, transition, suppression, instance, label, time, pagination, or refresh dependencies change. The request completion path only sets loading to false.

After the initial load, changing any filter or page leaves the previous alert feed rendered as if it matched the new query. Acknowledgement, delivery-record, related-event, root-cause, and delivery-retry actions on those stale events remain reachable while the replacement request is pending.

The effect cleanup already prevents an older response from publishing after dependencies change. This is not a response-order race; it is a missing loading and interaction boundary for the current request.

### Impact

During incident triage, operators can acknowledge or inspect an event that is no longer part of the active query. The screen also temporarily presents the previous result set and counts as current, which can mislead filtering and pagination decisions.

### Proposed Design

Keep the existing request effect and cancellation behavior:

1. Re-enter `loading=true` whenever the System Alert page request starts.
2. Reuse the existing render behavior: while loading, show the loading Card and do not render the previous alert feed or its row actions.
3. Preserve the current stale-response cancellation and collector-status request.
4. Do not add input debounce or change backend APIs in this fix.

Because the frontend lint rule rejects synchronous state writes at the top of an effect, schedule the loading transition with a guarded microtask and check the existing `cancelled` flag before updating state.

### Acceptance Criteria

- Filter, page, and refresh changes enter the loading state until the current request settles.
- Events from the previous query and their actions are not rendered while replacement data is pending.
- Loading clears on both success and failure for the current request.
- Superseded responses remain unable to publish.
- A genuine empty result still renders the normal empty state after loading completes.
- A deferred-promise regression fails on current master and passes after the fix.
- Focused tests, the frontend suite, lint, build, formatting, and `git diff --check` pass.

AI-assisted source audit; behavior was verified from the current effect, conditional feed rendering, and event action controls.

Aias00:
## Summary

- consolidate the server-filter loading fixes from #4511, #4518, and this PR across the ACL, alert-rule, and system-alert pages
- synchronously re-enter loading whenever a server-backed filter, page, or refresh dependency changes
- keep the previous content mounted under an Ant Design loading mask while the replacement query is pending
- block row actions and mutations derived from stale content, including ACL credential/admin actions, alert-rule row/bulk actions, and system-alert ACK/delivery/related/retry actions

## Loading pattern

The combined PR standardizes on keeping the previous rows/feed mounted behind a loading mask while disabling stale-data actions. This preserves table/feed layout and scroll position during filter or pagination changes. It replaces the original #4526 skeleton/unmount proposal following maintainer review.

The loading transition is synchronous at the top of each request effect. The targeted eslint suppressions explain why the state update is required there: without it, previous server-page results remain actionable under the new query.

## Validation

- `npm test -- src/pages/instance/__tests__/AclPage.test.tsx src/pages/ops/__tests__/AlertsPage.test.tsx src/pages/ops/__tests__/SystemAlertsPage.test.tsx --silent` (70/70)
- `npm test -- --retry=2 --reporter=dot --silent` (134 files, 1239/1239)
- `npm run build`
- `npm run lint` (0 errors; 11 pre-existing warnings)
- Prettier check for all six changed files
- `git diff --check`
- independent review: initial ACL stale-action gap fixed; focused re-review approved with 0 remaining findings

Closes #4509
Closes #4515
Closes #4526

Supersedes #4511 and #4518.


RockteMQ-AI (approved):
### Summary
Clean fix for stale alert data remaining visible while a filtered query is pending. The loading-state re-entry and Clear Acked button disablement are correct, and the regression test thoroughly covers the stale-event, ACK, and header mutation paths.

One minor note: the `Promise.resolve().then(...)` microtask deferral for `setLoading(true)` works correctly but is slightly unusual inside `useEffect` — a direct `setLoading(true)` call would also be safe here. No action required, just a style observation.

LGTM.

---

lizhimins (changes requested):
The defect and the guard are correct — but please consolidate this with #4511 and #4518 into a single PR, drop the microtask wrapper, and tell us which loading shape you intend to standardise on.

**The defect is real and reachable.** `web/src/pages/ops/systemAlerts.tsx` initialises `loading` to `true` and only resets it in the `.finally`, while the feed effect's deps include `domainFilter` / `fromFilter` / `instanceFilter` / `labelFilter` / `levelFilter` / `page` / `refreshNonce` / `suppressionFilter`. So after the first load, changing any filter or paging leaves the previous event feed rendering as if it matched the new query, with ACK / delivery records / related events / root cause / retry all live on events that are no longer in scope. This is not a stale-response race — the `cancelled` cleanup already prevents an old response from being published — and there is no `setInterval` refresh, so re-entering the loading state cannot cause periodic skeleton flashes. Adding `loading ||` to the "clear acknowledged" button's `disabled` is also correct: it is the only derived mutation outside the feed rows, and your test asserting `clearAcknowledgedAlerts` is not called covers it.

**Change 1 — consolidate the three pages into one PR.** #4511 (ACL page), #4518 (alert rules page) and #4527 (this one) are the same fix — re-enter the loading state at the top of the server-side pagination/filter effect so stale rows cannot be acted on — applied to three pages, each carrying its own copy of the identical microtask block. Please submit one PR covering all three (one commit per page is fine) so the pattern lands once, is reviewed once, and cannot drift between pages. We will close the remaining two as superseded once the combined PR is up.

**Change 2 — replace the microtask wrapper with a synchronous state update.** The block you added at `web/src/pages/ops/systemAlerts.tsx:198-204` defers `setLoading(true)` into a microtask for no benefit, and the `mounted` guard is unnecessary because the effect body already runs synchronously during the same commit. Set it synchronously with a one-line comment explaining why (otherwise the previous feed stays actionable — ACK / retry — under the new filter). Keep the `Clear Acked` guard and the tests as they are.

**Change 3 — decide the loading shape, and say so in the combined PR.** This page renders `{loading && <Card loading />}` / `{!loading && alerts.map(...)}`, so re-entering the loading state **unmounts the whole feed**: every filter change or page turn replaces the incident list with a skeleton card and loses the scroll position. #4511 and #4518 instead keep the previous rows mounted behind antd's mask and disable the actions. Issue #4526 explicitly asked you to reuse the existing render behaviour here, so what you built is what was requested — but the two shapes are now inconsistent across three pages of the same console, and for an incident-triage page the skeleton has a real cost. Please state in the combined PR which shape you intend as the standard: keep the skeleton on this page (and accept the inconsistency), or keep the feed mounted and only disable the row actions like the other two. If it is the latter, that is a render-structure change and belongs in the same consolidated PR rather than a follow-up.

Aias00:
Addressed the requested consolidation in `6540ada4`:

1. Rebased onto `rocketmq-studio@7ce9a682` and moved this PR base to `rocketmq-studio`.
2. Folded the ACL page from #4511 and alert-rule page from #4518 into this PR.
3. Replaced all three microtask wrappers with synchronous loading transitions and targeted rationale comments.
4. Standardized the three pages on keeping previous content mounted behind an Ant Design loading mask while stale-data actions are disabled. System Alerts now preserves the feed and scroll position instead of swapping it for a skeleton.
5. Added explicit ACL loading guards for rule edit/delete plus user credential reveal/copy, admin toggle, edit, and delete after independent review found the table mask alone was insufficient.

Verification: focused tests 70/70, full frontend suite 1239/1239, production build passed, lint reported 0 errors and 11 existing warnings, Prettier and `git diff --check` passed. Independent re-review found no remaining issues.

#4511 and #4518 are superseded by this combined change.

PR #4529, opened 2026-09-17
fix(tencent): reject incomplete catalog pages

### LINKED ISSUE #4528: [Studio][Bug] Tencent cloud catalog treats incomplete instance pages as complete

- [x] I searched open/closed issues and PRs for Tencent `DescribeInstanceList`, incomplete catalog pages, null `Data`, and `TotalCount` handling.
- [x] I reproduced the behavior on current `master` (`d50ffec`).

### Problem

`TencentCatalogService.listCloudInstances` stops pagination whenever the current `DescribeInstanceList` response has `Data == null` or an empty array. It does this before consulting `TotalCount`.

Tencent's RocketMQ 5.x API defines `TotalCount` as the total query count and `Data` as the instance list. If a response says `TotalCount > current offset` while the current page has no data, the provider response is incomplete/contradictory: Studio currently converts that into a successful truncated catalog.

### Impact

Cloud-instance discovery/onboarding can show an empty or partial Tencent instance inventory as complete. Operators may conclude an instance does not exist or is not accessible even though the provider reports more matching instances.

### Expected behavior

- If `TotalCount` proves rows remain but the current page has no `Data`, fail closed with a 502-style error.
- Preserve a genuine empty catalog when `TotalCount` is zero/absent and no rows are returned.
- Preserve existing pagination, search filtering, and count normalization behavior.

### Red reproduction

On unmodified `master`, add a `DescribeInstanceListResponse` with `TotalCount=1` and `Data=null`, then require `listCloudInstances` to return 502. `TencentCatalogServiceTest`: **9 tests, exactly 1 failure** (`Expecting code to raise a throwable`) because current code returns an empty list.

### Relation to existing work

#4513/#4514 cover the same completeness principle for Tencent message-query pages; this issue is the separate cloud-catalog/onboarding path.

AI-assisted source audit and regression authoring; Tencent's current `DescribeInstanceList` API documentation and the pinned Java SDK model were checked before filing.

zmuxuny:

- Fixes #4528

### Brief Description

Tencent RocketMQ 5.x catalog pagination previously treated any null/empty `Data` page as the end of the instance list before consulting `TotalCount`. A short page was also accepted as complete whenever it returned fewer than `PAGE_SIZE` rows.

When `TotalCount` says more provider rows exist, either condition represents an incomplete/contradictory page. Returning the rows collected so far makes Studio's cloud-instance catalog look complete even though instances are missing.

This change fails closed with a 502-style `BusinessException` when a known `TotalCount` proves rows remain but the current page is short or empty. Genuine empty catalogs remain valid, and missing/negative `TotalCount` keeps the previous short-page fallback behavior.

### Red / Green Verification

Fail-before, Java 21:
- Added a regression with `TotalCount=1` and `Data=null`.
- `TencentCatalogServiceTest`: **9 tests, exactly 1 failure**.
- The new test failed with `Expecting code to raise a throwable`, proving current master returned a successful empty catalog.

Green, Java 21:
- `TencentCatalogServiceTest`: **11/11 passed**, 0 failures/errors.
- Added coverage for an incomplete non-empty short page and for a genuine empty catalog.
- `git diff --check`: clean.

### Compatibility / Risk

No API, schema, dependency, or frontend changes. Normal full pages, successful final pages, client-side search filtering, count normalization, and genuine empty catalogs are unchanged. Only provider responses that contradict their own known total count now surface as unavailable instead of silently truncating the catalog.

AI-assisted source audit, implementation and regression authoring; Tencent's current `DescribeInstanceList` API contract and the pinned Java SDK model were checked before filing.


RockteMQ-AI (approved):
### Summary
Fixes the same silent pagination truncation bug for the instance catalog listing path. Adds an `isIncompletePage()` guard that fails closed with 502 when `TotalCount` contradicts the current page, and refactors `hasFetchedAll()` to use `offset + returned >= total`.

**Verdict: LGTM.** Clean, well-tested fix (3 new tests including the genuine-empty-catalog edge case). The inline guard approach (boolean check + throw at call site) is slightly different from the `requireCompletePage()` pattern in #4531, but both are clear and correct.

### Minor Note
- Same cross-PR observation: consider a shared pagination utility if this pattern grows. Not blocking.

---

PR #4531, opened 2026-09-17
fix(tencent): reject incomplete resource pages

### LINKED ISSUE #4530: [Studio][Bug] Tencent resource inventories silently accept incomplete pages

- [x] I searched open/closed issues and PRs for Tencent topic/group/subscription pagination, incomplete pages, and `TotalCount` handling.
- [x] I reproduced all three paths on current `master` (`d50ffecc9d7e8f8f46da64198831bd7952e6974e`).

### Problem

Three Tencent RocketMQ 5.x resource inventories stop pagination solely because the current page is null/empty or shorter than `PAGE_SIZE`, even when the same response's `TotalCount` proves more provider rows exist:

- `DescribeTopicList` in `listTopics`;
- `DescribeConsumerGroupList` in `listConsumerGroups`;
- `DescribeTopicListByGroup` in group subscription/progress loading.

Tencent's current API documentation defines `TotalCount` as the total query count for all three interfaces. A short/empty page with `offset + returned < TotalCount` is therefore incomplete/contradictory, not a confirmed end of the list.

### Impact

Studio can silently return partial Topic, Consumer Group, subscription and consumer-progress inventories as complete. During administration or incident diagnosis, missing resources can look genuinely absent instead of unavailable from the provider response.

### Expected behavior

- When a known non-negative `TotalCount` proves rows remain but the current page is short/empty, fail closed with a 502-style error.
- Preserve normal complete pages and final pages.
- When `TotalCount` is missing/negative, retain the existing short-page fallback behavior.
- Keep existing mappings, enrichment, filters, and >10k pagination behavior unchanged.

### Red reproduction

On unmodified `master`, add one `TotalCount=2` / one-row regression for each of the three paths and require a 502.

`TencentInstanceProviderTest`: **47 tests, exactly 3 failures**. Each new test fails with `Expecting code to raise a throwable`; the other 44 tests pass.

### Relation to existing work

#4513/#4514 cover Tencent message-query completeness; #4528 covers the Tencent cloud instance catalog. This issue is the separate runtime resource inventory layer.

AI-assisted source audit and regression authoring; Tencent's current API documentation for all three interfaces was checked before filing.

zmuxuny:

- Fixes #4530

### Brief Description

Tencent RocketMQ 5.x resource listings used `Data == null/empty` or a short page as an unconditional end-of-list signal before reconciling it with the provider's `TotalCount`.

The same pagination pattern affected Topic inventory, Consumer Group inventory, and Consumer Group subscription/progress reads. When `TotalCount` says more rows exist, accepting a short or empty page silently returns an incomplete resource view as if it were complete.

This change centralizes the completeness check for those three paths. A known non-negative `TotalCount` that contradicts the current page now fails closed with a 502-style `BusinessException`; missing/negative totals retain the previous short-page fallback behavior.
### Red / Green Verification

Fail-before, Java 21:
- Added one contradictory short-page regression for each affected path.
- `TencentInstanceProviderTest`: **47 tests, exactly 3 failures**.
- Topic, Consumer Group, and Group Subscription regressions each failed with `Expecting code to raise a throwable`, proving current master returned partial results successfully.

Green, Java 21:
- `TencentInstanceProviderTest`: **47/47 passed**, 0 failures/errors.
- `git diff --check`: clean.
### Compatibility / Risk

No API, schema, dependency, or frontend changes. Normal full pages, successful final pages, and existing provider-side filtering are unchanged. When Tencent omits `TotalCount` or returns a negative/unknown total, Studio still uses the previous short-page termination rule.

Only responses that contradict their own known total now surface as unavailable instead of silently truncating operational data.

AI-assisted source audit, implementation and regression authoring; the fail-before and green verification were executed locally against the stated baseline.


RockteMQ-AI (approved):
### Summary
Fixes the same silent pagination truncation bug as #4533 but for the resource listing path (topics, consumer groups, subscriptions). Adds a `requireCompletePage()` guard that fails closed with a resource-specific error message, and refactors `hasFetchedAll()` to use `offset + returned >= total` instead of `offset + pageSize >= total`.

**Verdict: LGTM.** The resource-name parameter in the error message is a nice touch for diagnostics. Test coverage is solid (3 new tests). The `isUnknownTotalCount()` helper improves readability.

### Minor Note
- Same cross-PR observation as #4533: the pagination guard pattern is replicated across 3 classes. A shared utility could reduce duplication if more Tencent service integrations are added. Not blocking.

---

PR #4533, opened 2026-09-17
fix(tencent): reject incomplete ACL role pages

### LINKED ISSUE #4532: [Studio][Bug] Tencent ACL role pagination can hide roles and report false not-found

- [x] I searched open/closed issues and PRs for Tencent ACL role pagination, `DescribeRoleList`, `TotalCount`, incomplete pages, and false not-found results.

### Problem

Tencent ACL role discovery treats any empty or short `DescribeRoleList` page as the end of the role catalog before reconciling it with `TotalCount`.

The duplicated pagination loop is used by `listUsers`, `listRules`, and `findRole`. If Tencent reports a known total that says more roles remain but the current page is short/empty, Studio silently accepts an incomplete ACL inventory.
### Impact

ACL user/rule pages can omit real roles while looking complete. More importantly, `updateUser` calls `findRole` when permissions are omitted so it can preserve existing values; an incomplete page can therefore turn a provider-data failure into `404 ACL user not found`, blocking a valid permission update and sending the operator toward the wrong diagnosis.

### Expected Behavior

- When a non-negative `TotalCount` proves roles remain but the current page is short or empty, fail closed with a 502-style `BusinessException`.
- Preserve genuine empty/final pages.
- Preserve the current short-page fallback when `TotalCount` is missing or negative/unknown.
- Do not change Tencent ACL mutation payloads or permissions.
### Red Reproduction

On unmodified master, three focused regressions use `TotalCount=2` with one returned role:
- `listUsers` is required to fail closed, but returns one user;
- `listRules` is required to fail closed, but returns one rule;
- `updateUser` searches for a role that could still be on the missing page and currently returns 404 instead of 502.

`TencentAclServiceTest`: **15 tests, exactly 3 failures**; the other 12 tests remain green. Checkstyle reports 0 violations.

AI-assisted source audit and regression authoring; fail-before was executed locally with Java 21.

zmuxuny:

- Fixes #4532

### Brief Description

Tencent ACL role discovery duplicated the same pagination rule in `listUsers`, `listRules`, and `findRole`: any empty or short `DescribeRoleList` page was treated as the end of the catalog before checking whether `TotalCount` said more roles remained.

That can silently truncate ACL inventory. It also affects control-plane updates: `updateUser` uses `findRole` to preserve omitted permission fields, so an incomplete page can turn a provider-data failure into a false `404 ACL user not found`.

This change adds one shared role-page completeness guard. Known non-negative totals that contradict the current page now fail closed with 502; missing/negative totals keep the previous short-page fallback behavior.
### Red / Green Verification

Fail-before, Java 21:
- `TencentAclServiceTest`: **15 tests, exactly 3 failures**.
- `listUsers` and `listRules` returned partial inventories successfully instead of failing closed.
- `updateUser` reported **404** for a role that could still exist on the missing page; the regression requires 502.

Green, Java 21:
- `TencentAclServiceTest`: **15/15 passed**, 0 failures/errors.
- `git diff --check`: clean.
### Compatibility / Risk

No API, schema, dependency, or frontend changes. Genuine empty/final pages remain valid. When Tencent omits `TotalCount` or returns a negative/unknown total, Studio keeps the existing short-page termination behavior.

ACL mutation payloads and permission mapping are unchanged; the patch only prevents contradictory provider pagination from being misreported as complete/not-found.

AI-assisted source audit, implementation and regression authoring; fail-before and green verification were executed locally against the stated baseline.


RockteMQ-AI (approved):
### Summary
Fixes a silent ACL role pagination truncation bug in Tencent Cloud integration by adding a `requireCompleteRolePage()` guard that fails closed (502) when `TotalCount` contradicts the current page. Also refactors `isLastRolePage` to use `offset + returned >= total` instead of `returned <= pageSize`, which is more accurate.

**Verdict: LGTM.** Clean fix with good test coverage (3 new tests covering the fail-closed path, the unknown-total fallback, and the `updateUser` regression). The approach is consistent and well-documented.

### Minor Note
- The `requireCompleteRolePage` / `isUnknownTotalCount` helpers follow the same pattern as the resource-page and catalog-page guards in sibling PRs (#4531, #4529). If more Tencent service classes need this pattern, consider extracting a shared pagination utility. Not blocking — acceptable as-is for class-local encapsulation.

---

RockteMQ-AI (approved):
### Summary
This PR adds validation to detect incomplete pages from the Tencent Cloud ACL API. When the API returns a page where the number of roles is less than expected (indicating truncation), the code now throws an error instead of silently returning partial results.

### Findings
- **[Info]** `TencentAclServiceImpl.java` — The `requireCompleteRolePage` method is well-placed and correctly validates page completeness. The error message is clear and actionable.
- **[Info]** `TencentAclServiceImplTest.java` — Tests cover the incomplete page scenario and verify the error is thrown.

### Suggestions
- The validation is a good defensive measure against API inconsistencies. Consider logging the expected vs actual count at WARN level for debugging.

Important data integrity fix.

---

PR #4542, opened 2026-09-17
fix(group): preflight settings across brokers

### LINKED ISSUE #4541: [Studio][Bug] Consumer Group settings can partially update Brokers before preflight completes

- [x] I searched open/closed issues and PRs for Consumer Group settings partial updates, per-Broker read/write ordering, and `createAndUpdateSubscriptionGroupConfig`.

### Problem

Apache `updateConsumerGroupSettings` reads and writes each master Broker sequentially. If Broker A's config is readable, Studio mutates A immediately; if Broker B's config is then missing or unreadable, the request fails after A has already changed.

The regular `updateConsumerGroup` path already avoids this exact failure mode: it pre-reads all Broker configurations before any write, with a dedicated regression `updateConsumerGroupReadsAllConfigurationsBeforeWriting`. The settings path lacks the same safety boundary.
### Impact

Consumer Group settings include retry counts and consumption switches such as `consumeEnable`, orderly consumption, and broadcast consumption. A mid-read failure can leave different masters with different behavior even though Studio reports the overall request as failed. Retrying later is not guaranteed to restore the original state.

### Expected Behavior

- Discover all target master Brokers.
- Read and validate every current `SubscriptionGroupConfig` before mutating any Broker.
- If any configuration is missing/unavailable, fail with zero Broker writes.
- Only after the full preflight succeeds, apply the partial settings update to every Broker.
- Preserve current per-Broker config values for omitted switches and existing audit/metadata behavior.
### Red Reproduction

A focused regression uses two master Brokers: the first settings read succeeds and the second throws `broker unavailable`. The request is expected to fail without any `createAndUpdateSubscriptionGroupConfig` call.

On unmodified master, `RocketMQAdminClientImplTest`: **65 tests, exactly 1 failure**. Mockito records one Broker write before the second read fails; the other 64 tests pass and Checkstyle reports 0 violations.

AI-assisted source audit and regression authoring; the existing regular-group pre-read contract was used as the behavioral reference.

zmuxuny:

- Fixes #4541

### Brief Description

Apache Consumer Group settings update previously read, modified, and wrote each master Broker sequentially. If an earlier Broker read succeeded, Studio mutated it immediately; a later missing/unavailable Broker config could then fail the request after the earlier master had already changed.

The regular `updateConsumerGroup` path already avoids this failure mode by reading every Broker configuration before any write. This change gives `updateConsumerGroupSettings` the same two-phase safety boundary: pre-read all current configs into a deterministic map, fail with zero writes if any preflight read fails, then apply the partial settings update only after the full preflight succeeds.
### Red / Green Verification

Fail-before, Java 21:
- Two master Brokers; the first settings read succeeds and the second throws `broker unavailable`.
- `RocketMQAdminClientImplTest`: **65 tests, exactly 1 failure**.
- Mockito recorded one `createAndUpdateSubscriptionGroupConfig` call before the later read failed, proving the current path can partially mutate the cluster.

Green, Java 21:
- `RocketMQAdminClientImplTest`: **66/66 passed**, including both later-read exception and later-missing-config zero-write coverage.
- `RocketMQAdminClientImplTest,MetadataServiceTest,ConsumerGroupControllerTest`: **139/139 passed**.
- `git diff --check`: clean.
### Compatibility / Risk

No API schema, frontend, dependency, permission, or field-mapping changes. Omitted settings still preserve each Broker's current value, and existing success audit/metadata behavior remains unchanged.

This does not make the distributed write phase transactional: a network failure during the actual write loop can still leave a partial update, and the existing audit records `updated X/N brokers before failure`. The fix specifically guarantees that preflight read/missing-config failures occur before any mutation.

AI-assisted source audit, implementation and regression authoring; the existing regular Consumer Group pre-read contract was used as the behavioral reference.


RockteMQ-AI (approved):
### Summary
This PR splits `updateConsumerGroupSettings` into a two-phase pre-read then write pattern, matching the existing `updateConsumerGroup` contract. This prevents partial cluster mutation when a later broker read fails after an earlier broker was already modified.

The implementation is clean — uses a `LinkedHashMap` for deterministic ordering, correctly handles missing configs as failures, and the test coverage (both later-read-exception and later-missing-config zero-write scenarios) is thorough.

LGTM.

---

PR #4544, opened 2026-09-17
fix(group): preserve unavailable consumer connections

### LINKED ISSUE #4543: [Studio][Bug] Consumer connection lookup failures are reported as zero online clients

- [x] I searched open/closed issues and PRs for consumer connection lookup failures, `onlineInstances`, Proxy fallback, and false zero-client diagnostics.

### Problem

Apache consumer-group live enrichment collapses connection-inventory failures into a normal empty connection set. `RocketMQMetadataProvider.resolveConsumerConnection` returns `null` after an unexpected Broker/NameServer error; the caller converts that to an empty instance list and `onlineInstances=0`.

The Proxy fallback has the same ambiguity: `ProxyConsumerResolver` returns `null` both when every reachable Proxy explicitly reports `CONSUMER_NOT_ONLINE` and when Proxy discovery/query itself fails.

### Impact

Consume stats may still be available and report positive lag while connection inventory is unavailable. Studio then publishes `onlineInstances=0`, so the web diagnostics can emit the critical finding “有堆积但无在线客户端”, and the AI group-detail tool can report the group UNHEALTHY, even though Studio never established that zero clients are online.
### Expected Behavior

- Preserve `0` only when the connection inventory is actually available and confirms no clients.
- Preserve Proxy-connected clients when Broker lookup reports the group offline.
- Distinguish Proxy `CONSUMER_NOT_ONLINE` from discovery/query transport failures; RocketMQ Proxy explicitly returns that response code for a known offline group.
- Represent unavailable connection inventory separately from zero clients so UI/AI diagnostics do not infer a consumer outage from a failed lookup.
- Keep lag/consume-stat availability independent from connection-inventory availability.

### Red Reproduction

A focused regression makes `examineConsumerConnectionInfo` fail with `nameserver unavailable` while `examineConsumeStats` still reports lag 40. On unmodified master, the expected unavailable online count is lost and the group is returned with `onlineInstances=0`.

`RocketMQMetadataProviderTest`: **44 tests, exactly 1 failure** (`expected: -1 but was: 0`); the other 43 tests pass.
### Proposed Scope

Use the existing negative-sentinel convention for `onlineInstances` (`-1` = unavailable), add an internal Proxy resolution status that distinguishes explicit offline from lookup failure, and teach web/AI health diagnostics not to treat unavailable connection data as zero.

No persistence/schema change is required. Existing `resolveConsumerConnection` callers can keep the current nullable compatibility API while the group-health paths use the availability-aware result.

AI-assisted source/protocol audit and regression authoring; the Broker/Proxy response paths and web/AI consumers were checked before filing.

zmuxuny:

- Fixes #4543

### Brief Description

Apache consumer connection inventory previously collapsed operational lookup failures into `onlineInstances=0`. Consume stats could still report positive lag, so web/AI health diagnostics could conclude “backlog with no online consumer” even though the connection count was never established.

This change preserves three distinct states:
- a successful Broker/Proxy connection result publishes its real client count;
- Broker offline plus Proxy `CONSUMER_NOT_ONLINE` (or a successfully observed empty Proxy set) remains a trustworthy zero;
- Broker/Proxy discovery or connection-query failures publish `onlineInstances=-1` as unavailable.

The existing nullable Proxy resolver API remains compatible; availability-aware group-health paths use an internal resolution result.

### Diagnostic Behavior

- Web tables/statistics render negative connection counts as `不可用` instead of `-1`/`0`.
- Web health emits `CONNECTION_STATUS_UNKNOWN` and does not emit `NO_ACTIVE_CLIENTS_WITH_LAG` for unavailable connection inventory.
- AI group detail reports `UNKNOWN` rather than `UNHEALTHY` when connection inventory is unavailable.
- Lag/consume-stat availability remains independent: reliable backlog data is still returned even when connection inventory fails.

### Red / Green Verification

Fail-before, Java 21: `RocketMQMetadataProviderTest` **44 tests, exactly 1 failure**; the new case expected unavailable `onlineInstances=-1` but current master returned `0` while lag remained 40.

Green, Java 21:
- `ProxyConsumerResolverTest,RocketMQMetadataProviderTest,RocketMQAdminClientImplTest,ConsumerGroupReadToolHandlersTest`: **119/119 passed**.

Frontend:
- `consumerGroupDiagnostics.test.ts`: **4/4 passed**.
- Touched-file ESLint and Prettier checks passed.
- `npm run build`: **BUILD SUCCESS** (`8046 modules transformed`).
- `git diff --check`: clean.

The full `ConsumerPage.test.tsx` run is **31/32 on both this branch and clean `master@d50ffec`**, failing the same pre-existing stack-race timing case (`keeps the latest client stack when an older request resolves last`). The targeted case passes **1/1 on both the branch and clean baseline**. The touched connection-diagnostic unit suite is green.

### Compatibility / Risk

No database schema, mutation path, provider SDK, or endpoint shape change. `onlineInstances` remains numeric; the established negative-sentinel convention is now documented as `-1 = unavailable` for connection inventory.

Known zero clients remain zero, Proxy-connected consumers remain discoverable, and transient discovery failures are not cached as an empty Proxy set.

AI-assisted source/protocol audit, implementation and regression authoring; RocketMQ Proxy's explicit `CONSUMER_NOT_ONLINE` response was used to separate a known offline group from an unavailable query.


RockteMQ-AI (approved):
### Summary
This PR introduces a three-state model for consumer connection availability in the dashboard, replacing the previous binary `onlineInstances=0` that collapsed both "confirmed offline" and "lookup failed" into the same value. The new model distinguishes:

- **Positive value**: real client count (connection inventory available)
- **0**: confirmed offline (Broker returned empty set, or Proxy reported `CONSUMER_NOT_ONLINE`)
- **-1**: unavailable (Broker/Proxy discovery or connection query failed)

This is important for health diagnostics — previously, the system could report "backlog with no online consumer" even when the connection count was never actually established.

### Findings

Overall the design is clean and well-executed. A few observations:

- **[Info]** `ProxyConsumerResolver` — The new `ConsumerConnectionResolution`, `ProxyAddressResolution`, and `ProxyQueryResolution` records are well-structured and provide clear separation of concerns. The availability-aware `resolveConsumerConnectionStatus()` method correctly handles all edge cases (discovery failure, empty addresses, query failure, CONSUMER_NOT_ONLINE).

- **[Info]** `RocketMQMetadataProvider` — The fallback logic is correct: when Broker throws and Proxy resolver exists, it delegates to Proxy; when both fail, it returns unavailable (-1). When Broker throws and no Proxy resolver exists, it returns available(null) — this preserves the existing behavior for deployments without Proxy.

- **[Info]** `GroupDetailToolHandler` — The health check correctly separates `onlineInstances < 0` (UNKNOWN) from `== 0` (UNHEALTHY/WARNING). This prevents false alarms when connection info is simply unavailable.

- **[Info]** Frontend changes — `formatOnlineInstances()` and the diagnostics panel correctly handle the -1 state with appropriate Chinese localization ("不可用" / "客户端连接信息不可用").

- **[Info]** Tests — 6 new test methods across 4 test files provide good coverage of the new behavior, including edge cases like all-proxies-fail and CONSUMER_NOT_ONLINE.

### Suggestions

- Consider adding a brief comment in `ConsumerGroupVO` documenting the -1 convention (e.g., `/** -1 means connection inventory unavailable; see ProxyConsumerResolver */`) to help future maintainers understand the semantic contract. The existing Javadoc already mentions this, which is good.

- The `MqResponseCodes` utility class is well-designed for centralizing the CONSUMER_NOT_ONLINE sentinel check.

LGTM — well-designed change with comprehensive test coverage. The three-state model correctly addresses the diagnostic blind spot.

---

lizhimins:
Merged with a few maintainer adjustments that complete the `-1` sentinel rollout. Your core change is unchanged: the three-state resolution in `ProxyConsumerResolver` (discovery failure / proxy lookup failure → unavailable, `CONSUMER_NOT_ONLINE` or an observed empty set → a trustworthy zero) and the `onlineInstances = -1` sentinel match the unknown-lag contract we already ship, and `GroupDetailToolHandler` now answers `UNKNOWN` instead of "backlog with no online consumer".

What was added on top, and why:

**1. The catalog schema rejected the sentinel — this one was a hard runtime failure.**
`server/src/main/resources/tool-catalog/tools/group.yaml` declared `minimum: 0` for `rmq.group.detail`'s `onlineInstances`. `ToolValidationFilter` runs `ToolSchemaValidator.validateOutput` on **every** tool result, and it throws `IllegalStateException` on a violation — so the first time a group's connection inventory was unavailable, `rmq.group.detail` would have failed outright instead of reporting the unknown state. Verified locally by reverting the schema and watching the contract test fail with `Tool output validation failed for rmq.group.detail: [/onlineInstances: must have a minimum value of 0]`.

Fixed by allowing and documenting the sentinel in all three schema sites (`rmq.group.list` items, `rmq.group.detail`, and `rmq.group.detail`'s `configurations[]` items), plus `make catalog-generate` so `rmqctl/internal/catalog/catalog_gen.go`'s digest stays in sync — CI does not run on this branch, so a stale digest would have landed silently. `GroupListItem` / `GroupDetailOutput` and `docs/api-spec.md` now document the sentinel too, and `ToolOutputSchemaContractTest` carries `-1` samples for both tools.

**2. The sentinel leaked into three display paths.**

- `MetadataService.buildConsumerGroupCsv` printed the raw `-1` in the `Online Instances` column; it now uses `onlineInstancesText(...)`, mirroring the existing `lagText(...)` treatment of the unknown-lag sentinel in the same row builder.
- `web/src/services/consumerService.ts` had the same problem in the browser-side export. Note the exported cell was not even a bare `-1`: `escapeCsvCell` formula-guards leading `-`, so the column contained `'-1`.
- The `在线客户端` column's `sorter` was still `(a.onlineInstances ?? 0) - (b.onlineInstances ?? 0)`, which sorts an unavailable group *before* a group with zero clients — i.e. as the healthiest possible answer. It now uses a sort key that pushes unavailable to the end, matching `lagSortValue` in the adjacent `总堆积量` column.

Since the table cell, the two `Statistic` formatters and both CSV exports all need the same rule, the helpers moved to `web/src/utils/consumerConnections.ts` (`UNKNOWN_ONLINE_INSTANCES` / `isOnlineInstancesAvailable` / `formatOnlineInstances` / `onlineInstancesSortValue`), deliberately shaped like the existing `web/src/utils/consumerLag.ts` so the two sentinels read the same way.

**3. Test naming.** `detailMarksHealthUnknownWhenConsumerConnectionsAreUnavailable` → `...UnavailableTest`; our convention is that test method names end with `Test`.

Every one of those four changes has a mutation-sensitive test: reverting the schema, the backend CSV formatting, the frontend CSV formatting or the sorter each turns a test red (contract test, `MetadataServiceTest`, `consumerService.test.ts`, `ConsumerPage.test.tsx` respectively).

One item left alone on purpose, and one note:

- We kept your treatment of a `CONSUMER_NOT_ONLINE` heartbeat-syncer discovery as a trustworthy zero. Your code comment states the trade-off explicitly, which is the right way to record it.
- Unrelated to this PR: the browser-side export still prints the raw `totalLag`, so an unknown lag reaches that CSV as `'-1` while the backend export renders `unknown`. That inconsistency predates this change and will be handled separately.

PR #4559, opened 2026-09-20
fix(cluster): scope the config update to the instance that owns the registry row

### LINKED ISSUE #3103: [Studio][Bug] Registry Broker configuration actions can target the wrong instance

unbridled-41:
Fixes #3103.

## Problem / Evidence

Broker rows on the Cluster page's Broker tab come from the global NameServer registry (`listRegistryClusters`), so a row can belong to an instance other than the one selected in the topbar. The config **preview/update** request still attached the topbar instance (`selectedInstanceIdRef.current`) as `instanceId`, and the backend scopes the lookup by it: `ClusterService.updateClusterConfig`/`previewClusterConfig` → `resolveCluster(id, instanceId)` → `clusterProvider.refreshClusterDetail(id, instanceId)` → `resolveNamesrvAddr(instanceId)`, i.e. that instance's NameServer. A row from another instance is therefore either looked up against the wrong NameServer (falling back to the repository, or 404) or silently resolved to a same-named cluster — the cross-instance write in #3103.

## Root cause / Fix

The drift requests on the same registry rows already resolve the owning instance by endpoint match (`resolveOwningInstanceId`, present on the trunk) and pass it, because a mismatched `instanceId` makes the backend reject the lookup. Build the config preview/update request the same way, so the write follows the row instead of the topbar selection.

Omitting `instanceId` entirely is not an option: with no instance the provider falls back to the global default NameServer (`resolveNamesrvAddr(null)` → `properties.getNamesrvAddr()`) and `ClusterService.getCluster(id)` refuses the repository, so a cluster registered under another NameServer cannot be resolved at all.

## Change of scope (base retarget, 2026-09-21)

The original PR also rewrote both config-diff calls. Upstream's `feat(ai): hosted conversation platform (#4606)` commit landed between the original PR and today and fixed those two calls with the same endpoint-based owner resolution, so that part is dropped here and the PR is narrowed to the write path. The branch was also retargeted from `master` to the development trunk `rocketmq-studio` (master is frozen at the CONTRIBUTING change that moved the trunk) and rebased onto it.

## Tests

- New regression test `scopes the config write to the instance that owns the registry row` (red → green on this branch: with the baseline payload restored the request carries `instance-2`, the topbar selection, instead of `instance-1`, the owner).
- `vitest --changed origin/rocketmq-studio` → **36 passed, 0 failed** (`ClusterPage.test.tsx`, `App.test.tsx`).
- `src/pages/cluster/__tests__/ClusterPage.test.tsx` full file → **28 passed**.
- `npx tsc -b` clean; `npx eslint` on both changed files clean.

CI cannot corroborate this locally: the repository's `ci.yml` has failed at startup (`startup_failure`, zero jobs) for every branch and author since 2026-08-29, including upstream's own pushes; the last successful run is 2026-08-29.

## Risk

Low. Registry rows whose endpoint matches no configured instance keep the previous behaviour of a backend-resolved lookup; for the instance-matched rows the write now targets the instance that owns the row, which is what the neighbouring drift calls already do.


unbridled-41:
## Verification evidence

**CI (current head)**: this PR's head SHA `0057bd5b0c2b2aa162abbb4bec966b78419ad428` has exactly one workflow run — "CI" (pull_request), run id 35478298552, conclusion `startup_failure`, and 0 check-runs / 0 statuses on the head commit. Upstream's CI workflow fails at startup for **every** branch and PR in this repository (including maintainers'), i.e. no PR gets CI execution; local verification below substitutes for it.

**Red (baseline source, new tests kept)**: restored `origin/master`'s `web/src/pages/cluster/index.tsx` while keeping the new test file, then `npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx --testTimeout 60000` → `Test Files 1 failed (1)`, `Tests 6 failed | 23 passed (29)`. Failures are the new/updated assertions (payload still carrying `instanceId`, diff calls carrying the route instance).

**Green**: same command with the fix → `Test Files 1 passed (1)`, `Tests 29 passed (29)`.

**Type check / lint / build**: `npx tsc -b` exit 0; `npx eslint src/pages/cluster/index.tsx src/pages/cluster/__tests__/ClusterPage.test.tsx` exit 0; `npm run build` → `✓ built in 9.83s`.

**Full web suite**: `npx vitest run --testTimeout 60000` → 1037 tests, 2 failed test files / 4 failed tests, all in files this PR does not touch (ConsumerPage ×2, MetricsExplorer, NotificationDeliveries — the load-fragile set recorded in prior rounds); running those three files in isolation → **62/62 passed**, confirming load fragility rather than a regression. Full-suite totals on this branch: 1037 tests = pristine `origin/master` 1034 tests (verified in a clean worktree: 1032 passed + 2 failed in untouched load-fragile files) + the 3 new regression tests added by this commit (net; no tests removed).

**Diff self-check**: `git show --numstat 0057bd5b` → `web/src/pages/cluster/index.tsx +11/−3`, `web/src/pages/cluster/__tests__/ClusterPage.test.tsx +113/−9`. Two files, no unrelated changes; the −3/−9 are the previous route-instance expectations the fix replaces.

RockteMQ-AI (approved):
### Summary
Removes `instanceId` from cluster config API requests (preview, update, broker/NameServer config diff) so the backend resolves the write target from the cluster ID instead of relying on the client to identify the owner instance. This is correct because registry rows come from the global NameServer registry and may be reachable from 0, 1, or many configured instances — endpoint text cannot reliably identify an owner.

### Findings
- **[Info]** `index.tsx:375-378` — The comment explaining why `instanceId` is omitted from registry config actions is clear and accurate. Good documentation of the design decision.
- **[Info]** `index.tsx:682-685` — Same rationale for the config preview/update path. The change from `selectedInstanceIdRef.current` to no instanceId is consistent across all config operations.
- **[Info]** Test coverage is thorough: new tests verify that `instanceId` is NOT present in API calls for registry rows, multi-instance shared NameServer scenarios, and server-side config diff resolution. Existing tests are updated to match the new behavior.

### Suggestions
- This is a frontend-only change. Confirm that the backend `ClusterService` endpoints (`previewClusterConfig`, `updateClusterConfig`, `getBrokerConfigDiff`, `getNameServerConfigDiff`) already support resolving the target cluster by `id` alone without `instanceId`. If the backend hasn't been updated, this would break config operations.
- Consider adding a backend integration test or documenting the backend contract that `instanceId` is optional for these endpoints.

### Verdict
Well-structured change that correctly addresses the ambiguity of registry config targets. The test coverage is excellent, including edge cases like multi-instance shared NameServers. LGTM pending backend compatibility confirmation.

---

PR #4225, opened 2026-09-11
feat(server): add MCP tool framework and AI integration.

echooymxq:
## What is the purpose of the change

XXXXX

## Brief changelog

XX

## Verifying this change

XXXX

Follow this checklist to help us incorporate your contribution quickly and easily. Notice, `it would be helpful if you could finish the following 5 checklist(the last one is not necessary)before request the community to review your PR`.

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test(over 80% coverage) to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).
- [x] Run `mvn -B clean apache-rat:check findbugs:findbugs checkstyle:checkstyle` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [ ] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).

PR #4571, opened 2026-09-20
fix(ai): tolerate alert rules without a metric in the rule list tool

zjncs:

Revives #4186, which was closed during the #4225 tool-layer restructure — the gap it reported is still live at the tool's new home in `tool-catalog/tools/alert.yaml`.

### Brief Description

`rmq.alert.rule.list` declares `metric` as a required output property with `minLength: 1`, but `AlertService` only validates `name` when a rule is created (`NativeAlertRulePolicy.validate` deliberately returns early on a blank metric), so a metric-less rule is a legal stored state. `AlertRuleListItem` serializes with `@JsonInclude(NON_NULL)`, so such a rule is emitted without a `metric` key, and `ToolValidationFilter` rejects the whole response with `required property 'metric' not found` — one metric-less rule breaks the entire `rmq.alert.rule.list` call.

Align the schema with the service contract the same way `acl.yaml` already treats the nullable `gmtCreate`: drop `metric` from the required list and declare it `[string, 'null']` instead of `string` + `minLength: 1`.

`cd server && mvn -B -ntp test -Dtest='ToolOutputSchemaContractTest' -Dspotbugs.skip=true` (Java 21), after adding a null-metric rule to the sweep's `rmq.alert.rule.list` sample:

- before this change: `Tool output validation failed for rmq.alert.rule.list: [/items/1: required property 'metric' not found]`
- after this change: `Tests run: 2, Failures: 0, Errors: 0`

Full tool layer: `mvn -B -ntp test -Dtest='*Tool*Test' -Dspotbugs.skip=true` → `Tests run: 127, Failures: 0, Errors: 0`.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (n/a — no UI change)
- [x] New source files carry the ASF license header (n/a — no new files)
- [x] Documentation touched where behaviour changed (n/a — schema now matches documented service behaviour)


RockteMQ-AI (approved):
### Summary
Aligns the `list_alert_rules` MCP tool output schema with the actual service behavior. The `metric` field is now optional (nullable) and removed from the required list, matching the `AlertRule.listRules()` output. Includes a proper test case for the null-metric scenario.

LGTM.

---

PR #4572, opened 2026-09-20
fix(auth): make the session overview integration test robust to MySQL DATETIME rounding

### LINKED ISSUE #4573: [Bug] Session overview integration test asserts second-exact values against second-precision DATETIME

### Studio Version

branch: `master`, git commit id: `d50ffecc`

### Runtime Environment

Ubuntu 24.04 (WSL2), JDK Temurin 21, MySQL 8.0.46 (project dev profile). The defect is in a test, not the deployed app.

Not applicable — the failing test only exercises the auth/session aggregate against the local database.

### Build Toolchain

_No response_

### Describe the Bug

`AuthServiceSessionOverviewIntegrationTest.listActiveSessionsForUserReturnsOnlyThatUsersActiveSessionDetailsTest` fails intermittently: it inserts a session with `last_seen_at = LocalDateTime.now() - 1 minute` and asserts `idleSeconds >= 60`, but observed `idleSeconds = 59`. The session table stores `last_seen_at` / `expires_at` as `DATETIME` (second precision), and MySQL rounds stored values up, so when the test starts in the second half of a second the stored value reads back one second younger and `Duration.getSeconds()` truncates it to 59. The same applies to the 20-minute stale assertion (1199 < 1200).

### Steps to Reproduce

1. Check out `master@d50ffecc` and run `cd server && mvn -B -ntp test` repeatedly.
2. Runs where the test class starts with the wall clock in the second half of a second fail with "Expecting actual: 59".
3. Confirm the rounding on the dev database: `SELECT CAST('2026-09-20 10:00:00.7' AS DATETIME)` returns `2026-09-20 10:00:01`.

The test to pass deterministically on every run.

### What Did You See Instead?

A flaky failure (observed 2026-09-19 on a full-suite run: 2414 tests, this single failure).

### Additional Context

Draft fix in PR #4572: the test truncates its base timestamp to whole seconds so the stored values are exact. Test-only change, no production code touched.

Frun1na:

- Fixes #4573

### Brief Description

`AuthServiceSessionOverviewIntegrationTest` asserts second-exact values (e.g. `idleSeconds >= 60`) that are derived from a wall-clock `LocalDateTime.now()` with sub-second precision. The `rmq_studio_session` table stores `last_seen_at` / `expires_at` as `DATETIME` (second precision), and MySQL rounds stored values up, so when the test starts in the second half of a second, an inserted "now minus 1 minute" last-seen value reads back one second younger and `idleSeconds` truncates to 59 — the `>= 60` (and likewise the `>= 20 * 60`) assertions fail intermittently. We hit this on a full-suite run against master@d50ffecc (2414 tests, this one failure).

The test derives its timestamps from `LocalDateTime.now(ZoneOffset.UTC).truncatedTo(ChronoUnit.SECONDS)` now, so the stored values are exact and the assertions are deterministic. No production code is changed; the class still runs the aggregate against the real database, which is the point of it being an integration test.

- Baseline failure on master@d50ffecc, 2026-09-19: `cd server && mvn -B -ntp test` → `Tests run: 2414, Failures: 1` — `listActiveSessionsForUserReturnsOnlyThatUsersActiveSessionDetailsTest`, expected `idleSeconds >= 60` but got `59`.
- Verified the storage behaviour directly on MySQL 8.0.46: `SELECT CAST('2026-09-20 10:00:00.7' AS DATETIME)` returns `2026-09-20 10:00:01` (rounds up).
- After the fix, 2026-09-20: `cd server && mvn -B -ntp test -Dtest=AuthServiceSessionOverviewIntegrationTest` → `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`. Checkstyle runs in the Maven validate phase and reported no violations.

### Checklist

- [ ] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text changed)
- [x] Architecture constraints stay green (`mvn test` runs the ArchUnit checks) (test-only change)
- [ ] New source files carry the ASF license header (no new source files)
- [ ] Documentation touched where behaviour changed (README / `docs/` / in-app help) (no behaviour change)


RockteMQ-AI (approved):
### Summary
Fixes a flaky integration test caused by MySQL `DATETIME` column rounding sub-second values. The fix truncates `LocalDateTime.now()` to second precision before insertion, making the assertion deterministic. Clean, well-documented change with no production code impact.

LGTM.

---

lizhimins:
Thanks for the precise root-cause analysis — it matches ours exactly, and independently arriving at the same explanation is a good sign the flake was understood rather than papered over.

`rmq_studio_session.last_seen_at` is a second-precision `DATETIME`, MySQL rounds a sub-second value up on insert, and the `idleSeconds >= 60` bound then occasionally reads back as 59. That is the right diagnosis.

We are closing this as already covered on the development trunk. `AuthServiceSessionOverviewIntegrationTest` on `rocketmq-studio` now derives `now` with `LocalDateTime.now(ZoneOffset.UTC).withNano(0)` at both call sites, and carries a comment recording the same root cause ("the session columns are `datetime` without fractional precision and MySQL ROUNDS on insert"). For a `LocalDateTime`, `withNano(0)` and your `truncatedTo(ChronoUnit.SECONDS)` are equivalent, so the flake you set out to fix is gone. That change came in with #4606, the hosted-conversation rewrite that rebuilt the branch.

Two practical notes:

- All Studio PRs now merge onto the `rocketmq-studio` branch rather than `master`, and this PR is based on `master`. Retargeting it would only produce a conflict on the two lines that already carry the equivalent fix, with no behavioural delta.
- There is no second site to sweep: we checked, and no other test writes a raw `LocalDateTime.now(ZoneOffset.UTC)` into a session column and then asserts on rounded bounds.

If you still see this failure against the current `rocketmq-studio` HEAD, please reopen with the failing output — that would mean the rounding shows up somewhere we have not found, and we would want to know.

PR #4215, opened 2026-09-10
[Studio] feat: add AI prompt template library

coder999o:
## What is the purpose of the change

Add an AI prompt template library to RocketMQ Studio so operators can start common RocketMQ diagnosis, query, and management workflows from structured prompts instead of repeatedly typing long instructions.

Each template carries its chat mode (`chat`, `diagnose`, `manage`, or `query`) and prompt enhancement setting into the next AI request. Users can also save the current input as a custom local template, then search, reuse, append, or delete it later.

## Brief changelog

- Add built-in RocketMQ operations prompt templates for consumer lag diagnosis, message trace investigation, topic change planning, broker health checks, operations handoff summaries, and ACL risk checks.
- Add a prompt template modal on the AI chat page with search, mode filtering, replace/append actions, and custom template save/delete support.
- Add an explicit chat mode selector and send the selected/template mode with AI chat requests.
- Store custom templates in bounded browser localStorage with safe fallback when storage is unavailable.
- Localize the new template UI and built-in template content.
- Add tests for template persistence, filtering, bounds, page application, and AI request parameters.

## Verifying this change

- `npm test -- promptTemplates AiPage --run`
- `npx eslint src/pages/ai/index.tsx src/pages/ai/__tests__/AiPage.test.tsx src/pages/ai/promptTemplates.ts src/pages/ai/promptTemplates.test.ts src/i18n/translations.ts`
- `git diff --check`
- `npm run build`

PR #4575, opened 2026-09-20
fix(ai): repair corrupt prompt-template storage instead of locking it out

zjncs:

No issue yet; found while auditing the prompt template library added in #4215.

### Brief Description

`readCustomPromptTemplates` caught a failing `JSON.parse` in the same `catch` as a failing `storage.getItem`, so a corrupt payload under `rocketmq-studio-ai-prompt-templates` was conflated with blocked storage:

* the modal showed the "browser storage unavailable" alert even though storage works fine;
* every subsequent `saveCustomPromptTemplate` returned `storage_unavailable` and every `deleteCustomPromptTemplate` returned `false`, so the bad payload could never be overwritten from the UI - the only recovery was deleting the localStorage key by hand in devtools.

This was internally inconsistent with the wrong-shape path one line below: a payload that parses but is not an array (e.g. a JSON object) is already treated as "no templates, storage available" and is repaired by the next save.

The fix splits the reads so each failure maps to the state it actually describes:

* `getItem` throwing still reports `storageAvailable: false` (genuinely blocked storage, unchanged behaviour, still covered by the existing test);
* `JSON.parse` throwing now behaves like the wrong-shape case - no templates, `storageAvailable: true` - so the next save overwrites the key and repairs it.

### How Did You Test This?

Fail-before / pass-after with the new vitest case `repairs corrupt stored JSON instead of treating storage as unavailable` (stores `{"broken":` under the key, then asserts the catalog reports storage available and a save succeeds and persists):

```
NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/ai/promptTemplates.test.ts   # before fix: 1 failed | 11 passed
NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/ai/promptTemplates.test.ts   # after fix:  12 passed
```

(`--no-experimental-webstorage` only stops Node 25's native `localStorage` global from shadowing jsdom's in the local environment; CI does not need it.)

Full web suite, lint and typecheck after the fix:

```
NODE_OPTIONS=--no-experimental-webstorage npm test        # Test Files 122 passed, Tests 1035 passed
npm run lint                                             # 0 errors (10 pre-existing warnings elsewhere)
tsc -b                                                   # clean
```

### Checklist

- [x] Code follows the project's coding style (Conventional Commits, lint passes)
- [x] All tests pass locally


RockteMQ-AI (approved):
### Summary
Fixes a bug where corrupt JSON in prompt-template storage was treated as blocked storage (localStorage disabled), making the AI tools completely unusable. Now distinguishes between JSON parse errors (repairable, treat as empty) and storage access errors (real blocker). Test covers the corrupt JSON scenario.

Solid fix with clear separation of failure modes. LGTM.

---

PR #4222, opened 2026-09-11
fix(message-history): render queriedAt as UTC

89799969:
## Summary
- Render query-history `queriedAt` / `latestQueryAt` with `formatUtcDateTime` so offset-less UTC values from the backend are not treated as browser-local time.
- Add a regression test for an offset-less `queriedAt` value.

## Issue
Fixes #4221

## Testing
- Added `MessageQueryHistoryDrawer` unit test `treats offset-less queriedAt values as UTC`.
- Please run frontend vitest for `MessageQueryHistoryDrawer` (and full `npm run test` if convenient).

## Related
Same UTC display class as #4174 / #4199 on the alert `lastTriggered` field.

PR #4231, opened 2026-09-11
[Studio] feat: add per-user session details

coder999o:
## What is the purpose of the change

Add per-user active session details to Studio user management, so administrators can inspect which active sessions belong to a user before revoking them.

## Brief changelog

- Add a safe per-user active session detail API without exposing session token hashes.
- Add a Studio user session detail drawer with session status, idle time, remaining TTL, refresh, and revoke-all actions.
- Add backend and frontend tests for the new session detail flow.

## Verifying this change

- mvn -q -Dtest=StudioUserControllerTest,AuthServiceSessionOverviewIntegrationTest test
- npm test -- src/api/studioUsers.test.ts src/pages/studio/__tests__/UserManagement.test.tsx
- npx eslint src/api/studioUsers.ts src/api/studioUsers.test.ts src/pages/studio/UserManagement.tsx src/pages/studio/__tests__/UserManagement.test.tsx
- npm run build
- git diff --check

PR #4576, opened 2026-09-20
fix(studio): render session timestamps in the viewer timezone

zjncs:

No issue yet; found while auditing the per-user session details added in #4231.

### Brief Description

The user management page formatted every session and account timestamp with `new Date(value).toLocaleString()`, but the studio user and session APIs serialize UTC `LocalDateTime` values **without an offset":

* `AuthService.now()` builds all session/account times at `ZoneOffset.UTC` (last-seen, expires, password-changed, session aggregates `MAX(last_seen_at)`/`MIN(expires_at)`);
* the default JDBC URL pins `serverTimezone=UTC`, so DB-default `gmt_create`/`gmt_modified` are UTC wall times too.

Per the ECMAScript spec, an offset-less ISO string parses as **browser-local** time, so for any non-UTC admin every 最近活跃/过期时间/创建时间 value in the sessions drawer, the user table columns and the CSV export was shifted by the viewer's UTC offset — a session expiring at 10:00 UTC displayed as 10:00 local instead of 18:00 in Asia/Shanghai.

The codebase already centralizes this pitfall in `formatUtcDateTime` (introduced for the alert subsystem; reused by the message query history fix #4222). This page was the remaining caller of raw `toLocaleString` on these values; the fix delegates the `dateTime` helper to `formatUtcDateTime`, which appends `Z` to offset-less values before rendering in the viewer's zone.

The existing drawer test asserted the rendered text with the identical misparse (`new Date(...).toLocaleString()`), so it passed in every timezone and could not catch the bug. It now asserts the helper's output, plus a new test pins the viewer zone via `vi.stubEnv('TZ', 'Asia/Shanghai')` so the expectation (17:30 for a 09:30 UTC value) discriminates on any runner.

### How Did You Test This?

Fail-before / pass-after (verified also from a `TZ=America/New_York` base zone, so the check does not depend on the runner's zone):

```
NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/pages/studio/__tests__/UserManagement.test.tsx
  # before fix: renders session timestamps as UTC values… FAILS (renders 09:30 local)
  # after fix:  passes
```

Full web suite, lint and typecheck after the fix:

```
NODE_OPTIONS=--no-experimental-webstorage npm test   # Test Files 122 passed, Tests 1035 passed
npm run lint                                          # 0 errors (10 pre-existing warnings elsewhere)
tsc -b                                                # clean
```

Note: `exports all users that match the active filters` in the same file is flaky when that file is run in isolation on a loaded machine (times out in `waitFor`); it fails identically on pristine master in isolation and passes in the full suite, so it is unrelated to this change.

### Checklist

- [x] Code follows the project's coding style (Conventional Commits, lint passes)
- [x] All tests pass locally


RockteMQ-AI (approved):
### Summary
Fixes timezone misparse in session viewer where UTC timestamps were rendered as local time. Delegates to the existing formatUtcDateTime helper and adds tests pinned to Asia/Shanghai to discriminate UTC from local time.

Clean fix, good test coverage. LGTM.

---

Issue #4465, opened 2026-09-16
[Bug] Hidden Topic and Consumer selections survive list scope changes


beautyarbutin:
### Before Creating the Bug Report

- [x] I have searched the existing issues and pull requests.
- [x] I am using the latest `master` branch (`987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`).

### Bug description

The Topic and Consumer resource tables keep selected row keys when the visible result scope changes. A user can select a row, then search, filter, or move to another page. The row disappears, but the batch-delete action still includes it.

On the Topic page, the selection also survives an instance switch. Confirming the action can therefore send topic names selected under the previous instance together with the newly selected instance ID.

### Steps to reproduce

1. Open the Topic or Consumer page and select one row.
2. Apply a search/filter, change the table page, or (for Topic) switch the selected instance so that the selected row is no longer visible.
3. Observe that the batch-delete count is still present.
4. Confirm batch deletion.

### Actual behavior

The delete request includes resource names that are no longer visible in the current result scope. For Topic instance switches, the stale names are submitted with the current instance ID.

### Expected behavior

Batch selection should be scoped to the current visible query/page/instance. When that scope changes, hidden selections should be cleared so the delete request can only contain rows the user can currently inspect.

### Proposed scope

- Reset Topic selections when instance, search, type filter, page, or page size changes.
- Reset Consumer selections when search, mode filter, page, or page size changes.
- Add frontend regression tests for both resource pages.

Issue #1720 discussed broader Topic-page state reset behavior on instance changes. This report is narrower: it concerns hidden batch selections and also affects Consumer query/page changes.

### Environment

- Browser: current Chrome
- RocketMQ Dashboard branch: `master`

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** The Topic and Consumer tables retain selected row keys in component state when the visible result scope changes (search, filter, pagination, or instance switch). The batch-delete action references stale selection state that includes rows no longer visible to the user.

**Impact:** For Topic page, confirming batch deletion after an instance switch sends topic names from the previous instance together with the new instance ID — a data integrity risk. For both pages, the batch-delete count misleads the user.

**Severity:** Medium-High — can cause unintended resource deletion when combined with instance switching.

[...]

PR #4577, opened 2026-09-20
fix(dlq): scope the group search to the selected instance

### LINKED ISSUE #4579: [Bug] DLQ group search term survives an instance switch and filters the next instance's results

### Studio Version

branch: `master`, git commit id: `d50ffecc`

### Runtime Environment

Ubuntu 24.04, current Chrome. Frontend-only defect, reproduced against the dev server.

Any instance; the bug is in the page's scope-reset logic, not in a specific cluster.

### Build Toolchain

_No response_

### Describe the Bug

On the DLQ page, the group search input keeps its value when the selected instance changes. The instance-scoped reset clears rows, pagination, selection and dialogs, but not the search term, so the first `listDLQGroups` request for the new instance still carries the old term. The user sees an empty or filtered list for the new instance without having searched anything on it.

### Steps to Reproduce

1. Open the DLQ page for instance A and type a term into the group search box that matches only A's groups.
2. Switch the instance selector to instance B.
3. Observe the group list for B.

Instance B's full DLQ group list, with the search box cleared.

### What Did You See Instead?

The list is filtered (typically empty) by the term typed against instance A, and the search box still shows it.

### Additional Context

Draft fix in PR #4577: the reset block also clears the search term, plus a regression test. Related fixes in the same family: #4240 (client column filters), #4465 (hidden topic/consumer selections).

Frun1na:

- Fixes #4579

### Brief Description

On the DLQ page the group search term survives an instance switch. The instance-scoped reset block in `DLQPage` clears the rows, pagination, selection and dialogs, but not `search`, so the first request for the newly selected instance is sent with the previous instance's search term — the page then shows an empty or misleading list for data the user never searched on. The same family of leaks was fixed for the client tables (#4240) and for hidden topic/consumer selections (#4465); this is the DLQ counterpart.

The reset block now also calls `setSearch('')`, so the term can only filter the instance it was typed against. Regression test added: switching instances after searching asserts the next `listDLQGroups` call carries no search term and the input shows the cleared value.

- Red first: added `clears the search term when the selected instance changes` and confirmed it fails on master@d50ffecc (`listDLQGroups` last called with `('instance-2', 'ord', 1, 20)` instead of `('instance-2', undefined, 1, 20)`).
- After the fix: `cd web && npx vitest run src/pages/instance/__tests__/DLQPage.test.tsx` → `Test Files 1 passed (1)`, `Tests 21 passed (21)`.
- `npx eslint src/pages/instance/dlq.tsx src/pages/instance/__tests__/DLQPage.test.tsx` → 0 errors (one pre-existing `react-refresh/only-export-components` warning on an untouched line).

### Checklist

- [ ] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text changed)
- [ ] Documentation touched where behaviour changed (README / `docs/` / in-app help) (no behaviour documentation affected)


RockteMQ-AI (approved):
### Summary
Fixes a UX bug where switching instances in the DLQ page did not reset the group search filter, showing stale results. Adds setSearch to the instance switch handler and a regression test.

Simple, correct fix. LGTM.

---

PR #4220, opened 2026-09-11
[Studio] feat: serve LiteTopic from the broker lite admin API

zhaohai666:
# feat(studio): serve LiteTopic from the broker lite admin API

## Problem

The dashboard ships a complete ~900-line LiteTopic UI, but the backend `LiteTopicService` is a stub: every endpoint returns `501` and `getCapability()` reports `supported=false`. From a user's perspective the feature is a shell — the page renders, nothing behind it works. Real LiteTopic data (parent topics, TTL, sessions, quota, backlog) lives in the broker's lite admin APIs and was never wired up.

## Solution

This PR replaces the stub with a real provider behind the existing SPI pattern (`InstanceProvider` / `MetadataProvider` style), using only APIs that are already available in the dashboard's declared dependency `rocketmq-tools:5.5.0` — no new dependencies, no new ports, no new configuration surface.

### New components

- **`LiteTopicProvider`** — new SPI interface (`isSupported()`, `listLiteTopics`, `getSession`, `extendTTL`, `getQuota`) with a backward-compatible default (501 `BusinessException`) so other vendors are unaffected.
- **`RocketMQLiteTopicProvider`** (`@Primary` Apache implementation) — all data comes from real broker RPCs on the master brokers discovered via `examineBrokerClusterInfo`:

| API field | Real source |
|---|---|
| Parent topics + TTL | `getBrokerLiteInfo` `topicMeta` / `getParentTopicInfo` (`lite.topic.expiration`, minutes) |
| Sessions (parent + group + client) | `getLiteClientInfo` (clientId → `liteTopicSet`, `lastAccessTime`), `examineConsumerConnectionInfo` |
| Per-topic backlog / offsets | `getLiteGroupInfo` (`totalLagCount`, offset wrappers) |
| Quota | `getBrokerLiteInfo` (`currentLmqNum`/`maxLmqNum`/`liteSubscriptionCount`) + `getBrokerConfig` (`maxLiteSubscriptionCount`, `minLiteTTl`) |
| TTL state | `lastAccessTime` + TTL → ACTIVE / EXPIRED, surfaced as ms in the API contract (the broker stores minutes) |

- Scans are bounded (`MAX_LITE_TOPIC_SCAN=200`, `MAX_LITE_SESSION_SCAN=500`) so a busy broker cannot stall the endpoint.
- Session identity (`parentTopic~group~clientId`) is validated on decode (400 on malformed input).
[...]

PR #4578, opened 2026-09-20
fix(lite): fail the TTL extension instead of partially updating the cluster

zjncs:

No issue yet; found while auditing the LiteTopic provider added in #4220.

### Brief Description

`extendTTL` examined and updated the masters in a single pass, and the `liteTopicConfig` helper collapsed **every** read failure into the same `null` as "topic not configured on this master":

```java
TopicConfig config = liteTopicConfig(admin, master, topicPattern);
if (config == null) {
    continue;              // skipped for absence AND for RPC failures
}
...
admin.createAndUpdateTopicConfig(master, config);
updated++;
```

When `examineTopicConfig` failed on one master (timeout, restart, connectivity blip) while succeeding on another, the loop silently skipped the failing master, answered 200, and logged "1 broker(s)" — leaving the cluster with mixed `lite.topic.expiration` attributes. LMQ expiry (`cleanExpiredLiteTopic`) is evaluated by each broker against its own local topic config, so lite topics on the un-updated master kept expiring under the old TTL while the console reported the extension as fully successful.

The fix reads every master's config **before** writing any of them. The new `liteParentTopicConfig` keeps the legitimate skips distinct from failures:

* topic absent on this master — `MQClientAPIImpl.getTopicConfig` throws `MQBrokerException` with `ResponseCode.TOPIC_NOT_EXIST` — still a skip (matches the multi-master reality where a parent topic may not be configured everywhere);
* config present but not `TopicMessageType.LITE` — still a skip (the existing 404 semantics for non-lite topics are unchanged);
* any other failure — propagates, so the request fails **before the first write** instead of after a partial update. Retrying after such a failure converges the cluster because the update is idempotent.

### How Did You Test This?

Fail-before / pass-after (Java 21, standalone maven):

```
mvn -B -ntp test -Dtest='RocketMQLiteTopicProviderTest' -Dspotbugs.skip=true
  # before fix: Tests run: 16, Failures: 1
  #   extendTtlFailsInsteadOfPartiallyUpdatingWhenAMasterCannotBeRead FAILED
  #   (no exception thrown; createAndUpdateTopicConfig WAS called on the reachable master)
  # after fix:  Tests run: 16, Failures: 0
```

New regressions:

* `extendTtlFailsInsteadOfPartiallyUpdatingWhenAMasterCannotBeRead` — two masters, the second one's `examineTopicConfig` throws `RemotingTimeoutException`; asserts the call fails and **no** master was written (old code updated broker-a and returned normally).
* `extendTtlStillUpdatesReachableMastersWhenTheTopicIsAbsentOnAPeer` — the peer throws `MQBrokerException(TOPIC_NOT_EXIST)`; asserts the reachable master is still updated and the peer is not, pinning the absence/failure distinction the new classification relies on.

Full lite area after the fix:

```
mvn -B -ntp test -Dtest='RocketMQLiteTopicProviderTest,LiteTopicServiceTest,LiteTopicControllerTest,LiteTopicSessionTest' -Dspotbugs.skip=true
  # Tests run: 36, Failures: 0, Errors: 0
mvn -B -ntp checkstyle:check    # BUILD SUCCESS
```

### Checklist

- [x] Code follows the project's coding style (Conventional Commits, checkstyle clean)
- [x] All tests pass locally


RockteMQ-AI (approved):
### Summary
Fixes a critical partial-update bug in LiteTopic TTL extension. Previously, when examineTopicConfig failed on one master, the loop silently skipped it and reported success, leaving the cluster with inconsistent TTL attributes. The fix reads all masters configs before writing any, and distinguishes topic-absent (skip) from RPC failure (propagate). Well-documented with thorough regression tests.

Excellent fix - read-before-write pattern prevents partial updates. LGTM.

---

PR #4580, opened 2026-09-20
fix(audit): let the timeline target be activated from the keyboard

zjncs:

- Fixes #<issue-id> (trivial a11y fix, no issue)

### Brief Description

The audit table renders an operation's **target** as the only entry point into the resource operation timeline, but it is a plain click-only `Text` node: it is not reachable with `Tab` and ignores `Enter` / `Space`, so keyboard users cannot open a timeline at all.

This follows the convention already used for the notification template variable tags in `alerts.tsx`: mark the node as `role="button"` with `tabIndex={0}` and an `aria-label`, and activate the timeline on `Enter` or `Space` with `preventDefault` (so the page does not scroll on Space while opening the timeline).

Web suite, run from the repo root (the `NODE_OPTIONS` flag is only needed on this machine — Node 25's native `localStorage` global otherwise shadows jsdom's; CI does not need it):

```
NODE_OPTIONS=--no-experimental-webstorage npx --prefix web vitest run --root web src/pages/ops/__tests__/AuditPage.test.tsx
# before the fix: new keyboard test fails (element has no button role / no key handling)
# after the fix:  11 passed (11)

NODE_OPTIONS=--no-experimental-webstorage npx --prefix web vitest run --root web
# Test Files  122 passed (122)
#      Tests  1035 passed (1035)

npx --prefix web tsc -p web --noEmit   # no errors
npm --prefix web run lint              # 0 problems
```

The new test focuses the target, fires `keyDown` Enter, and asserts the timeline dialog opens and the audit query is scoped to that record's `resourceType` / `target` / `clusterId`.

### Checklist

- [ ] Architecture constraints stay green (`mvn test` runs the ArchUnit checks) — N/A, web-only change
- [ ] New source files carry the ASF license header — no new files
- [ ] Documentation touched where behaviour changed (README / `docs/` / in-app help) — N/A, no behaviour change for mouse users


RockteMQ-AI (approved):
LGTM — clean accessibility fix that makes the timeline target keyboard-activatable.

### Summary
This PR adds keyboard support (Enter/Space) to the audit timeline target, making it accessible to keyboard-only users. The implementation is minimal and well-structured:

- Extracts the `setTimelineResource` call into an `openTimeline` handler to avoid duplication
- Adds `onKeyDown` handler for Enter and Space keys with proper `preventDefault()`
- Includes test coverage for keyboard activation
- No breaking changes, mergeable state is clean

The fix follows WCAG 2.1 guidelines for keyboard accessibility (Success Criterion 2.1.1). Good work!

---

PR #4592, opened 2026-09-20
fix(dlq): count queues abandoned mid-scan as failed in the resend outcome

zjncs:

- Fixes #<issue-id>

### Brief Description

`scanDeadLetters` abandons a DLQ queue mid-scan in two places without incrementing `failedQueueCount`:

1. a pull whose `nextBeginOffset` does not advance (a stall that would otherwise loop forever), and
2. more than `MAX_CONSECUTIVE_OFFSET_ILLEGAL` consecutive `OFFSET_ILLEGAL` responses (giving up on the corrected offsets).

Only the exception path and the null pull result counted. Because `scanIncomplete()` is `failedQueueCount > 0 || truncated`, a scan where every *other* queue succeeded stayed "complete": the resend was classified as `SUCCESS` with `scanIncomplete=false, scanFailedQueues=0`, even though part of the dead-letter queue was never read. The operator gets a "complete" resend whose `matched` count silently excludes the unscanned messages — for dead letters that is the worst kind of success, because the unresent messages look handled.

This PR adds `failedQueueCount++` before both abandonments, so:

- mixed case (one queue abandoned, others scanned): outcome becomes `PARTIAL` with `scanIncomplete=true` and `scanFailedQueues=1`, matching the existing semantics of the exception path (`resendMessagesMarksAResultPartialWhenOneDlqQueueCannotBeScanned`);
- all-queues-abandoned case: the scan now fails with the same 502 `Failed to scan DLQ topic ...` as the all-queues-threw case, instead of returning an empty success.

The DLQ **list** path (`listMessages`) shares `scanDeadLetters` and already maps `failedQueueCount` into its VO, so it picks up the corrected incompleteness signal for free.

Two new tests plus one updated test in `RocketMQDLQProviderTest`:

- `resendMessagesMarksAResultPartialWhenOneDlqQueueStalls` (new): two queues — one stalls (`nextBeginOffset == requestedOffset`), one yields an in-window dead letter that is resent. Expects `matched=1, resent=1, outcome=PARTIAL, scanIncomplete=true, failedQueueCount=1` and the `scanFailedQueues=1`/`PARTIAL` audit record.
- `resendMessagesCountsAQueueAbandonedAfterPersistentOffsetIllegal` (new): every pull returns `OFFSET_ILLEGAL` with a corrected offset that is itself illegal again until the cap is hit. Expects the 502 `Failed to scan DLQ topic` (the queue is the only one, so the whole scan fails) and no producer invocation.
- `resendMessagesStopsWhenPullOffsetDoesNotAdvance` (updated): the single stalled queue now fails the whole scan instead of returning an empty success; the test's original guarantees (exactly one pull, no producer, bounded time) are kept.

Fail-before (fix stashed, tests on master's code): the three tests above fail — the two-queue case returns `SUCCESS/scanIncomplete=false/failedQueueCount=0`, the give-up case returns instead of throwing. Pass-after:

```
$ mvn -B -ntp test -Dtest='RocketMQDLQProviderTest,DLQServiceTest,DLQControllerTest' -Dspotbugs.skip=true
 Tests run: 41 ... in RocketMQDLQProviderTest      (0 failures)
 Tests run: 24 ... in DLQControllerTest            (0 failures)
 Tests run: 17 ... in DLQServiceTest               (0 failures)
 BUILD SUCCESS
```

The existing `resendMessagesRetriesFromCorrectedOffsetAfterOffsetIllegal` (single illegal, then FOUND → `SUCCESS`) is unchanged and green, confirming the retry-from-corrected-offset behaviour is untouched.

### Checklist

- [x] Documentation touched where behaviour changed (error/signal semantics only, no user-facing docs exist for scan internals)


RockteMQ-AI (approved):
### Summary
This PR fixes a subtle bug in `scanDeadLetters` where queues abandoned mid-scan (due to offset stall or consecutive `OFFSET_ILLEGAL`) did not increment `failedQueueCount`. This caused the scan to incorrectly report `SUCCESS` with `scanIncomplete=false` even though part of the dead-letter queue was never read — the worst kind of success for dead letters, because unresent messages appear handled.

The fix is minimal and surgical: two `failedQueueCount++` additions at the two abandonment points. The existing `scanIncomplete()` logic then correctly propagates the signal:
- Mixed case (one queue abandoned, others scanned) → `PARTIAL` with `scanIncomplete=true`
- All-queues-abandoned → 502 error instead of empty success

The test coverage is thorough: two new tests (stall + persistent illegal) plus one updated test, all with clear naming and realistic mock setups. The `@Timeout` annotation on the persistent-illegal test prevents runaway loops.

LGTM — clean fix for a real correctness issue.

---

lizhimins (changes requested):
Thanks for this. I want to say up front what this review is and is not: **nothing here is being rejected on merit.** The production change is two lines and our own reading of it is that both lines are correct. What we could not do is follow the change well enough to merge it and then explain it to someone else, so this is a request for explanation rather than a code critique.

**What we verified as correct.** In `RocketMQDLQProvider.scanDeadLetters` there are four ways a queue can stop being read: `pullResult == null` and the `catch (Exception)` branch both increment `failedQueueCount`, while the offset-stall `break` (`nextOffset <= offset`) and the give-up-after-`MAX_CONSECUTIVE_OFFSET_ILLEGAL` `break` only log a warning. Since `DeadLetterScanResult.scanIncomplete()` is `failedQueueCount > 0 || truncated` and `classifyOutcome` returns `PARTIAL` only when the scan was incomplete or a send failed, a resend that abandoned queues mid-scan is reported as `SUCCESS` with `scanFailedQueues=0` in the audit line. Adding the two increments closes that. We also checked the paths you deliberately left alone and agree with both: hitting the cap already sets `truncated`, which feeds the same incompleteness signal, and a `NO_NEW_MSG` / `NO_MATCHED_MSG` result is a completed scan that should not count as a failure. The existing `failedQueueCount == queues.size()` → 502 branch then covers "every queue was abandoned" for free. The two new tests and the rewritten one are mutation-sensitive, and the fixtures are self-consistent — the `PullResult(FOUND, 10, 0, 10, [])` case really does produce a non-advancing offset.

**What we need you to explain in the PR before we can merge.**

1. **The observable defect, end to end.** Please describe what a user actually sees today and what they should see: which action (the DLQ page resend button, or the `rmq.message.redelivery_dlq` tool), what the response reports, and what the audit record says. The chain from "a queue was abandoned" to "the operator believes every dead letter was redelivered" is the point of the change, and it is currently only implied by the diff.

2. **How the two triggers happen in practice.** What makes a broker return a `nextBeginOffset` that does not advance, and what makes `OFFSET_ILLEGAL` repeat more than three times even after the corrected offset is used? Did you observe these against a real deployment, or reason them out from the code? If you observed them, the broker version and the matching broker-side log lines would let us judge how common this is — that changes how urgent the fix is and how it should be worded in release notes.

3. **Why reuse `failedQueueCount` rather than a distinct signal.** That counter currently means "this queue could not be read at all", and it is also what drives the all-queues-failed 502. After this change it additionally means "this queue was read partway and then abandoned". Those are different facts, and collapsing them is a design choice rather than an obvious one — a separate `abandonedQueueCount` (or setting `truncated`) would keep them apart. Please either justify the collapse or propose the separation. This is the question we most need answered.

4. **An unmentioned behaviour change on the read paths.** `scanDeadLetters` is reached through `collectDeadLetters`, which has four callers: the resend path plus the list, export and detail paths. Because a single-queue DLQ topic that stalls now reaches `failedQueueCount == queues.size()`, those read paths would start returning 502 where they previously degraded to an empty list. That may well be right — the existing code already 502s when every queue throws, and only degrades to empty when the DLQ topic does not exist yet (`isDlqTopicMissing`) — but it is a user-visible change to a query API and it is not in the description. Please state whether it is intended and, ideally, cover it with a test.

5. **The rewritten test changed meaning, not just assertions.** `resendMessagesStopsWhenPullOffsetDoesNotAdvance` went from asserting an empty success to asserting a 502 and that no producer was ever obtained, so its name no longer describes what it checks. Please rename it and call out in the description that this is an intentional contract change rather than a test adjustment — that distinction matters to anyone reading the history later.

**Optional, not blocking.**

- The new test method names do not end in `Test`, which is the project convention for Java tests. This file already mixes both styles, so treat it as a nit.
- `RocketMQMessageProvider.queryByTopic` has the same two silent abandon paths, but it returns a plain `List<MessageRecordVO>` with no per-queue completeness channel — `mayBeTruncated` there only reflects the result-count cap. Fixing it means adding a signal first, which is a design change of its own. Please open a separate issue rather than folding it in here.

Once you have answered 1 to 5 in the PR description — a short reproduction plus the reasoning for item 3 is enough — we can merge. One mechanical note: this branch is based on `master`, and Studio PRs now merge onto the `rocketmq-studio` branch, so please retarget it.

PR #4593, opened 2026-09-20
fix(consumer): refresh the subscription verdict on the 2s modal auto-refresh

zjncs:

- Fixes #<issue-id>

### Brief Description

The consumer detail modal advertises `每 2s 自动刷新` in its header and states `诊断结果随详情弹窗每 2 秒自动刷新。` next to the diagnostic result. But the 2-second tick only refreshed the group summary and the progress table: the **subscription consistency verdict — the actual diagnostic result — was loaded once per group and then served from the diagnostic cache forever**. While the modal stayed open, the verdict stayed frozen at whatever the first load returned; the only refresh paths were the manual `重新检查` / `重新诊断` buttons.

This PR adds the subscriptions to the tick. Two details:

- The refresh is **silent**: a failing backend must not toast every 2 seconds (the tick already swallows progress errors for exactly that reason) and the check spinner must not flicker on every tick. `loadSubscriptions` gains a `silent` flag mirroring the existing `loadProgress(groupName, force, silent)` signature: silent skips the loading flags and the error toast while still recording the error state that drives the existing `订阅一致性检查失败` UI.
- The tick's `inFlight` guard already serializes requests for the same modal session, and the interval is already re-created on every progress update on master (the effect depends on `loadProgress`, whose identity changes with `progressByGroup`), so adding `loadSubscriptions` to the same dependency list follows the established pattern.

New test `refreshes the subscription verdict on the 2s modal auto-refresh`: open the detail modal with the first response reporting one inconsistent subscription, do **nothing**, and wait for the real 2-second interval. Before the fix the subscription API is never called a second time and the verdict stays stale; after the fix the second call happens and the verdict updates to `全部 2 个订阅配置一致`.

```
$ NODE_OPTIONS=--no-experimental-webstorage npx --prefix web vitest run --root web src/pages/instance/__tests__/ConsumerPage.test.tsx
# before the fix (source stashed, test kept):
 Tests  1 failed | 32 passed (33)      # the new test times out waiting for the 2nd call
# after the fix:
 Tests  33 passed (33)

$ npx --prefix web eslint web/src/pages/instance/consumer.tsx web/src/pages/instance/__tests__/ConsumerPage.test.tsx   # 0 problems
$ npx --prefix web tsc -p web --noEmit                                                                                 # clean
```

The existing `auto-refreshes the selected group while the detail modal is open` test (group/progress refresh on the same tick) is unchanged and green. No timer mocking is used in the new test — it waits on the real interval with a generous `waitFor` timeout, so it stays deterministic under load. (`NODE_OPTIONS=--no-experimental-webstorage` is only needed on Node ≥ 25, where native `localStorage` shadows jsdom's.)

### Checklist

- [x] Documentation touched where behaviour changed (the change makes existing in-app text truthful; no docs change needed)


RockteMQ-AI (approved):
### Summary
This PR fixes a real bug where the subscription consistency verdict in the consumer detail modal was loaded once and never refreshed on the 2-second auto-refresh tick, despite the UI advertising "每 2s 自动刷新". The fix adds `loadSubscriptions` to the tick function with a `silent` flag that mirrors the existing `loadProgress(groupName, force, silent)` pattern — silent mode skips error toasts and loading spinner updates while still recording error state for the UI.

The implementation is clean and follows established patterns in the codebase. The `inFlight` guard already serializes requests, and the dependency array is correctly updated. The new test uses real intervals (no timer mocking) which is more realistic and deterministic.

LGTM — well-documented fix with proper test coverage.

---

PR #4603, opened 2026-09-20
fix(audit): report a failed summary instead of zero statistics

### LINKED ISSUE #4599: [Studio][Bug] A failed audit summary is rendered as zero matching records
## Problem

When the audit summary request fails, the audit page renders the summary cards as an all-zero aggregate. The cards then state that no audit record matched the current filters, while the record table and the risk-insights panel on the same screen still list and count the matching records.

## Evidence

`web/src/pages/ops/audit.tsx` (summary effect, before the fix):

```ts
void getAuditSummary(activeFilter)
  .then((value) => {
    if (!cancelled) setSummary(value);
  })
  .catch(() => {
    if (!cancelled) message.error('审计概览加载失败，请稍后重试');
  })
```

The failure only raises a transient toast and leaves `summary` as `null`, and `AuditSummaryCards` substitutes zero defaults for a `null` summary:

```tsx
const data = summary || {
  total: 0, successful: 0, failed: 0, partial: 0, uniqueOperators: 0, ...
};
```

The same screen derives its other numbers from the records when the summary is missing (`auditRiskInsightModel.ts`: `summary?.total ?? safeRecords.length`), so the cards contradict the panel below them.

Reproduction (regression test added with this report, `AuditPage.test.tsx`, `reports a failed summary instead of rendering it as empty statistics`): open `/ops/audit` with `/api/audit-logs/summary` failing and `/api/audit-logs/page` succeeding. Before the fix the test finds the card `匹配记录0` although the table lists the matching record; the failure output is `expected '匹配记录0' not to be '匹配记录0'`.

## Impact

A failed aggregate is presented as a factual "no record matched these filters" result, i.e. the operator sees zero failures and a 0% success rate for a filter that actually matched records.

## Expected behavior

A failed summary is reported as unavailable (the cards are replaced by an explicit failure notice) instead of being rendered as zero statistics.

## Related work

- The record-list effect in the same file already distinguishes a failed load (`message.error` plus no state write) from an empty result.
- #1163 / #1164 (closed) asked for the opposite direction on the backend (failures as structured gateway errors) and were closed; this report only covers the web rendering of an already-failed request.

## PR

Fix: #4603.

unbridled-41:
Fixes #4599.

## Problem

When `GET /api/audit-logs/summary` fails, the audit page rendered the summary cards from zero defaults, i.e. it claimed that no audit record matched the current filters while the record table and the risk-insights panel on the same screen listed and counted the matching records.

## Evidence

Regression test added with this PR (`web/src/pages/ops/__tests__/AuditPage.test.tsx` → `reports a failed summary instead of rendering it as empty statistics`), run against the pre-fix code (baseline `origin/master` = d50ffecc):

```
AssertionError: failed summary must not be shown as zero matched records: expected '匹配记录0' not to be '匹配记录0'
⎯ a second run with the source restored from master:
TestingLibraryElementError: Unable to find an element with the text: 审计概览加载失败
```

## Root cause and fix

The summary effect only raised a toast and left `summary` as `null`:

```ts
.catch(() => {
  if (!cancelled) message.error('审计概览加载失败，请稍后重试');
})
```

and `AuditSummaryCards` substitutes a zero aggregate for `null`, so the cards rendered `匹配记录 0`, `成功率 0%`, `失败 / 部分成功 0 / 0`, `操作人数 0` and two empty bucket panels. The fix keeps the failure in state and replaces the cards with an explicit failure notice, so zero is never published for a failed aggregate.

## Priority and scoring

PRIORITY 72 = impact 28 + scope 10 + reproducibility 18 + maintenance value 16.

- Impact 28: a failed aggregate is published as a factual "these filters match nothing" result, including a 0% success rate, on an audit page whose purpose is exactly that count.
- Scope 10: the summary endpoint failing (server error/timeout) while the page stays usable.
- Reproducibility 18: deterministic, covered by the added test.
- Maintenance value 16: the failure path was the only untested branch of the page's summary effect.

FIX_CONFIDENCE 95: an error flag plus a rendering branch, no API or contract change, module suite green.

## Tests

- Red (baseline source + new test): `expected '匹配记录0' not to be '匹配记录0'`; with the component reverted: `Unable to find an element with the text: 审计概览加载失败`.
- Green: `npx vitest run --maxWorkers=2 src/pages/ops/__tests__/AuditPage.test.tsx` → **11 passed (11)** (10 pre-existing + 1 new).
- Full web suite: see the verification comment below.
- `npx tsc -b` clean; `npx eslint` on the three changed files → no output.

## Risk

Low. Only the failed-summary branch changed; a successful summary renders exactly as before, and the record list, insights panel and table are untouched. The two `setSummary`/`setSummaryFailed` writes are both guarded by the effect's `cancelled` flag, so a response for a superseded filter cannot flip the state.

---

The development trunk moved to `rocketmq-studio`; `master` is frozen at `d50ffecc`. This PR was retargeted to `rocketmq-studio` and its branch rebased onto the trunk (`7ce9a682`) — head `35b94681`. Re-verified on the new base:

- `npx tsc -b` → clean
- `npx vitest run --changed origin/rocketmq-studio` → **19 passed, 0 failed** (`src/App.test.tsx`, `src/pages/ops/__tests__/AuditPage.test.tsx`)


unbridled-41:
## Verification evidence

**CI (this head SHA)** — `95b2996c4bfefbdf3866919dc75cdcbaa2521b74`: exactly one `CI` (`pull_request`) run, id `35507312187`, `status=completed`, `conclusion=startup_failure`; `check-runs=0`; combined status `state=pending, count=0`. Repo-wide condition: every recent `pull_request` run is `startup_failure`, including other contributors' PRs.

**Red before the fix** — two independent baselines:

1. New test against the pre-fix page source:
```
AssertionError: failed summary must not be shown as zero matched records: expected '匹配记录0' not to be '匹配记录0'
Tests  1 failed (1)
```
2. Page and card components restored from `origin/master` with the test kept:
```
TestingLibraryElementError: Unable to find an element with the text: 审计概览加载失败
```

**Module tests (green)** — `cd web && npx vitest run --maxWorkers=2 src/pages/ops/__tests__/AuditPage.test.tsx`:

```
 Test Files  1 passed (1)
      Tests  11 passed (11)
```

(10 pre-existing + 1 new.)

**Full web suite** — `cd web && npx vitest run --maxWorkers=4`:

```
 Test Files  1 failed | 121 passed (122)
      Tests  1 failed | 1034 passed (1035)
```

The single failure is `ConsumerPage > keeps the latest client stack when an older request resolves last`, which is unrelated to this PR (the audit page and its tests are untouched) and reproduces with the `ConsumerPage` test file restored from `origin/master` under the same load (`Tests 1 failed | 31 skipped`).

**Build** — `npx tsc -b`: clean. `npx eslint src/pages/ops/audit.tsx src/pages/ops/AuditSummaryCards.tsx src/pages/ops/__tests__/AuditPage.test.tsx`: 0 errors, 0 warnings. `npm run build` on this head:

```
✓ built in 10.06s
dist/assets/index-BWwj-dym.js   324.03 kB │ gzip: 97.85 kB
```

**Diff self-check** — `git diff --numstat origin/master` → `14/3 web/src/pages/ops/audit.tsx`, `29/2 web/src/pages/ops/AuditSummaryCards.tsx`, `12/0 web/src/pages/ops/__tests__/AuditPage.test.tsx` (3 files, net +55/−5). The `+29/−2` in the card component is the new failure branch; the `−2` are two import-line reformats by prettier, no behaviour change.

RockteMQ-AI (approved):
LGTM. Code changes look good.

---

lizhimins:
Thanks — this is a well-scoped PR and the implementation holds up. I want to be clear that the reason for closing is a product decision, not a code review finding.

**What we verified.** The behaviour you describe is accurate on trunk: `audit.tsx` writes `summary` only in the `.then`, so the `.catch` leaves it null after showing the toast, and `AuditSummaryCards` substitutes an all-zero object for a null summary — "匹配记录 0", "成功率 0%", "失败 / 部分成功 0 / 0" — while the record table on the same screen still lists matching rows. Your scoping is also correct: we checked the other consumer of `summary`, and `AuditRiskInsights` already degrades to the loaded records via `auditRiskInsightModel`'s `?? safeRecords...` fallbacks, so it cannot infer `NO_MATCHING_RECORDS` from a null summary. The cards were the only surface that needed the change. The reset path is right too (`setSummaryLoading(true)` before each fetch, and the `loading` skeleton branch precedes the new `failed` branch, so a retry does not flash the failure state), clearing `summary` rather than keeping the previous aggregate is correct given the filter has changed, the new test lands in the existing `AuditPage.test.tsx` and is mutation-sensitive, and using antd `Alert` for a semantic failure state matches the project convention that reserves `InfoBanner` for neutral persistent copy. Your point that the toast auto-dismisses while the zero cards stay on screen is fair.

**The decision.** We are keeping the zeros. This is the same call we made on #4386 and wrote up in #4361: when a read fails, the surface keeps its numeric default rather than gaining a second representation of "no data". #4361 was scoped to the consumer-group page, but the reasoning is the one we are applying here — a per-page availability vocabulary means every page that shows derived numbers grows its own `failed`/`unavailable` flag, its own copy string and its own renderer branch, and they have to be kept consistent by hand. We would rather pay the cost of an occasionally misleading zero than ship N bespoke versions of "unknown", and the failure is already announced by the toast plus the fact that the record table below is populated.

The only thing the review flagged at all was cosmetic and non-blocking: the new `failed` prop sits next to the component's existing `data.failed` (the count of failed records), which reads confusingly — `summaryFailed` would have been clearer. That is not why this is closing.

**If you want to revisit it**, the shape we would engage with is the one #4361 left open: a single repository-wide unavailable-data contract — one resolver, one VO/field convention, one renderer and export policy — covering the consumer-group, audit and any other aggregate surfaces together, rather than one page at a time. That is a reasonable proposal and we would review it on its own merits.

PR #4613, opened 2026-09-20
fix(ai): seed the run id on attach so stop works after a reconnect

Frun1na:

- Fixes #4614

### Brief Description

Reattaching to a live run (reload, reconnect, second tab) never receives a `run_started` frame: the server replays persisted events on attach and its live event vocabulary starts at `run_started` — `AgentEventProjector` does not project a replayed run row into one. The hook therefore never learned the run id, while the composer still renders the stop state from `isStreaming`. Result: a clickable stop button whose `stop()` call returns immediately, and `reportSpeed` silently skipped for the same reason.

Fix: `startStream` accepts a `knownRunId` that `attach` passes in, seeding `runIdRef`/`runId` with the id the caller already knows. The fresh-send path is unchanged (`knownRunId` defaults to null).

```
cd web && npx vitest run src/pages/ai/hooks/useAgentRun.test.ts
 Test Files  1 passed (1)
      Tests  14 passed (14)
```

The new `stopsAnAttachedRunThroughTheApiTest` fails without the hook change (`canStop` stays false after attach, `stopRun` never called).

```
cd web && npx vitest run src/pages/ai
 Test Files  14 passed (14)
      Tests  179 passed (179)
```

### Checklist


RockteMQ-AI (approved):
### Summary
This PR fixes a bug where the stop button would not work after attaching to an existing AI agent run. The root cause was that `startStream` was resetting `runIdRef.current` to `null`, but when attaching, the server never replays the `run_started` frame (since the run already started), so the run ID was never populated — making the stop button render but address nothing.

### Findings
- **[Info]** The fix adds a `knownRunId` parameter to `startStream`, allowing `attach` to seed the run ID immediately instead of waiting for a `run_started` frame that will never arrive
- **[Info]** The `send` path still passes `null` (default), preserving the existing behavior where the run ID comes from the live `run_started` frame
- **[Info]** Test case accurately reproduces the attach-without-run-started scenario and validates that `stop()` correctly calls the API with the right run ID

### Suggestions
Clean, minimal fix with good test coverage. The optional parameter with a sensible default (`null`) keeps backward compatibility while fixing the attach path.

---

PR #4618, opened 2026-09-20
fix(cluster): keep failed registry loads from silently emptying the tables

### LINKED ISSUE #4612: [Studio][Bug] Failed registry loads silently empty the cluster tables while the live indicator stays green
### Studio Version

branch: master

### Problem

On the cluster page, a failed registry load silently wipes the data it is supposed to refresh and leaves the page claiming a healthy live connection:

- `loadRegistryClusters` (`web/src/pages/cluster/index.tsx:205-224`) catches a failed `listRegistryClusters()` with `setRegistryClusters([])` — no error state, no message, no retry flag. The Broker tab table (`:2444+`) and the Proxy tab table render their rows from `registryClusters`, so one transient failure empties both inventories.
- `loadNsRegistry` (`:230-238`) does the same to the NameServer 注册表 table (`setNsRegistry([])`), and the K8s cert effect (`:244-255`) clears the cert-id options.
- The page's green "实时刷新" live indicator is driven only by the per-instance `clusters` auto-refresh, which does **not** re-run these loaders (they fire once on mount). After a registry failure the indicator keeps pulsing green next to empty tables, telling the operator the data is live and healthy.
- The same page already has the correct pattern for a primary load: the instance list failure sets `instanceLoadError` (`:176-180` area) and renders an error panel with retry. The registry loaders predate that pattern and were never aligned.

Trigger (deterministic): open 集群管理 while `/registry/clusters` or `/nameservers/registry` returns a transient 5xx (gateway hiccup, backend restart). Broker/Proxy tabs and the NameServer registry show empty tables with no error until the user manually hits refresh; the header live indicator stays green throughout.

### Evidence

- `web/src/pages/cluster/index.tsx:215-218` (catch → `setRegistryClusters([])`), `:235-237` (catch → `setNsRegistry([])`), `:252-254` (catch → `setK8sIdOptions([])`).
- `web/src/pages/cluster/index.tsx:226-242` — both loaders run once on mount; the 2s auto-refresh effect does not include them.
- Live indicator source: the auto-refresh status of `clusters` only.
- Contrast: `instanceLoadError` state + error panel with retry button already in the same file for the instance list.

### Impact

An operator who opens the page during a blip sees empty cluster/registry inventories with a green "live" indicator — "no clusters" and "connection healthy" shown at the same time. Data-driven actions (config diff, registry edits) look unavailable for reasons the page never explains.

### Expected behavior

A failed registry load surfaces an error state (error banner/retry like the instance list, or keep-previous-data plus a visible failure flag) instead of silently rendering an empty inventory as if it were authoritative; the live indicator must not claim health over a failed registry source.

### Related work

- #3103 (open, PR #4559) — registry *actions* targeting the wrong instance; different defect on the same table.
- #4543 — connection lookup failures reported as zero online clients: same "failure rendered as empty/zero" family, different page and layer.
- #4599 / PR #4603 — failed audit summary rendered as zero statistics: same family in ops.

#

## PR

Fix: #4618.

unbridled-41:
Fixes #4612.

## Problem / Evidence

On the cluster page a failed registry load silently wipes the data it was supposed to refresh:

- `loadRegistryClusters` (`web/src/pages/cluster/index.tsx:205-224`, base d50ffecc) catches a failed `listRegistryClusters()` with `setRegistryClusters([])` — no error flag, no message, no retry. The Broker and Proxy tab tables render from `registryClusters`, so one transient 5xx empties both inventories.
- `loadNsRegistry` (`:230-238`) does the same to the NameServer 注册表 table.
- These loaders run once on mount (`:226-242`); the 2s auto-refresh loop that drives the green 实时刷新 indicator does not re-run them. After a failure the page therefore shows empty tables next to a pulsing green "live" dot — "no clusters" and "connection healthy" at the same time — until the user manually hits refresh.
- The same page already has the correct pattern one state over: the instance list failure sets `instanceLoadError` and renders an error panel with a retry button. The registry loaders simply were never aligned with it.

Reproduction is deterministic: open 集群管理 while `/registry/clusters` or `/nameservers/registry` returns a transient failure.

## Root cause / Fix

The registry loaders treated "request failed" and "request returned empty" identically. Fix: add a `registryLoadError` flag, set it in both catch branches, keep previously loaded rows instead of overwriting them with `[]`, render an error banner (`获取数据失败` + 重试) that re-runs both loaders, and clear the flag on the next successful load.

## Priority & scoring

PRIORITY 64 = 影响 24 (all three cluster-page data tabs present an empty authoritative-looking inventory after any transient failure) + 波及 10 (two loaders + K8s options, one page) + 可复现 16 (deterministic given a failed request) + 维护价值 14 (reuses the page's established error-state pattern). FIX_CONFIDENCE 92: additive state + one banner, fully covered by a deterministic jsdom regression.

## Tests

```
cd web && npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx
```
- Red (fix stashed, test present): `Unable to find role="alert"` — on master the failed reload renders no error surface and empties the tables.
- Green: ClusterPage **27 passed (27)** incl. the new regression (banner appears with retry, previously loaded rows survive the failure, retry recovers and clears the banner) and both request-guard tests from the earlier registry work, unchanged.
- `npx tsc -b`: clean. `npx eslint` on both touched files: 0 errors, 0 warnings. `npm run build`: ✓ built in 8.55s.

## Risk

Low. Additive to one page: keep-previous-data on failure, a visible failure flag, and a retry action that calls the existing guarded loaders. No API or i18n additions (reuses `common.fetchDataFailed`/`common.retry`). The empty-state semantics of a genuinely empty registry are unchanged — a successful load returning `[]` still clears the tables.

---

The development trunk moved to `rocketmq-studio`; `master` is frozen at `d50ffecc`. This PR was retargeted to `rocketmq-studio` and its branch rebased onto the trunk (`7ce9a682`) — head `f295a89e`. Re-verified on the new base:

- `npx tsc -b` → clean
- `npx vitest run --changed origin/rocketmq-studio` → **36 passed, 0 failed** (`src/App.test.tsx`, `src/pages/cluster/__tests__/ClusterPage.test.tsx`)

---

## Per-loader error state (2026-09-21, review follow-up)

lizhimins correctly pointed out that the two registry loaders run concurrently (started independently on mount, fired together by the refresh button) but wrote one shared `registryLoadError` boolean: a success on either loader cleared the failure the other had just recorded, so e.g. a failed `/registry/clusters` followed by a successful `/nameservers/registry` hid the banner while the Broker/Proxy tables kept rendering stale or empty data.

**Fix** (`bc83c516`, branch rebased onto trunk `29a7d4ab`): the shared flag is replaced by per-loader error state — `registryClusterLoadError` and `nameserverLoadError`. The page banner renders while either flag is set; each loader clears only its own flag on success; "keep previous rows on failure" is unchanged.

**Tests** (`cd web && npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx`):

- Red, measured against trunk `29a7d4ab`'s shared-flag source with the new tests present: all 4 fail —
  - `keeps the error banner when the cluster loader succeeds while the nameserver load fails` → `TestingLibraryElementError: Unable to find role="alert"` (success on one loader cleared the other's error, exactly the reported defect),
  - `keeps the error banner when the nameserver load succeeds while the cluster load fails` → same,
  - `clears the nameserver error while the cluster error stays after a partial recovery` → same,
  - the original banner regression → same.
- Green with the fix: ClusterPage **33 passed (33)** (29 pre-existing incl. the trunk-added config-diff tests, +4 new: both mixed-outcome orders, the partial-recovery case, and the original banner/row-retention/retry regression rewritten on the shared helpers).
- `npx tsc -b`: clean. `npx eslint` on both touched files: 0 errors, 0 warnings. `npm run build`: ✓ built in 10.59s.
- Diff self-check: `index.tsx` +15/−32 across both commits' files vs the fix commit's +6/−6; only the two files of this PR are touched.


unbridled-41:
**Verification evidence**

- **CI**: head `b780cd882e73fccf41fecbc0ff2993045c5afa90` has exactly one workflow run for this PR (run [35528307073](https://github.com/apache/rocketmq-dashboard/actions/runs/35528307073), event `pull_request`, conclusion `startup_failure`, 0 check-runs). The upstream CI workflow has been failing at startup repository-wide since 2026-08-29 (all PRs, untouched files included), so this is not caused by the change.
- **Red light** (fix stashed, test present): `keeps the registry inventories and surfaces an error banner when a registry load fails` fails with `TestingLibraryElementError: Unable to find role="alert"` — pristine master renders no error surface for the failed reload and empties the tables.
- **Module test** (with fix): ClusterPage **27 passed (27)** (26 pre-existing, incl. both request-guard tests, + 1 new regression covering banner, row retention, and retry recovery).
- **Full suite** (re-run on this branch during audit): `npx vitest run --maxWorkers=4` on `fix/registry-load-failure-state` @ b780cd88 → **122 files / 1035 tests, all passed** (= pristine master 1034 + 1 new regression; sibling PRs' counts differ only by their own added tests).
- **Build**: `npx tsc -b` clean; `npx eslint` on both touched files → 0 errors, 0 warnings; `npm run build` → `✓ built in 8.55s`.
- **Diff self-check** (`git show --numstat b780cd88`): `index.tsx` +32/−3, `ClusterPage.test.tsx` +49/−0 — no unrelated changes.

RockteMQ-AI (approved):
### Summary
This PR fixes a critical UX bug where failed registry loads would silently empty the tables, making it appear as if there were no clusters when the actual issue was a transient load failure. The fix properly preserves previously loaded data and surfaces an error banner with retry functionality.

### Findings
- **[Info]** The implementation correctly separates the error state from the data state, preventing the silent data wipe
- **[Info]** The retry mechanism properly reloads both registry sources and clears the error state on success
- **[Info]** Comprehensive test coverage including the retry recovery path

### Suggestions
The code is well-structured and the fix addresses the root cause effectively. The test case clearly documents the expected behavior with helpful comments explaining the scenario.

---

lizhimins (changes requested):

- The keep-previous-rows half is right: `cluster/index.tsx:220-222` and `:236` wiped both inventories on any transient failure with no error surface, while `instanceLoadError` (`:156`, rendered at `:1793-1804`) already demonstrated the correct pattern one state over.
- The new flag is shared by two loaders that run concurrently, so a success on one clears the failure of the other. `setRegistryLoadError(false)` at `:218` and `:243` both write the same boolean that `:224` and `:247` set. The loaders are started independently on mount (base `:228` and `:242`) and fired together by the refresh button (base `:1750-1751`), neither awaited in sequence.
- Concretely: if `listNameserverRegistry` rejects first and `listRegistryClusters` resolves second, the banner disappears while the Broker/Proxy tables keep rendering a stale or empty `registryClusters` with no indication — on first mount `registryClusters` is `[]`, so the result is byte-for-byte the defect this PR set out to fix (empty tables next to a green live dot).
- Your regression cannot catch this because it rejects **both** loaders, so the two flags always move together. Please split the state (two flags, or clear only when both loads succeeded, or a per-loader error map) and add a case where one loader rejects and the other resolves, asserting the banner stays.

unbridled-41:
@lizhimins Thanks for the review — the shared-flag concern was real. Fixed in bc83c516 (branch rebased onto the current trunk `29a7d4ab`).

**Root cause**: the two registry loaders run concurrently (independent on mount, fired together by the refresh button) but wrote one shared `registryLoadError` boolean, so a success on either loader cleared the failure the other had just recorded — e.g. nameserver success after a cluster failure hid the banner while the Broker/Proxy tables kept stale/empty data.

**Review addressed**:
- Split the flag into per-loader error state: `registryClusterLoadError` / `nameserverLoadError`. The banner renders while either is set; a successful reload clears only its own error; "keep previous rows on failure" is unchanged.
- Both mixed-outcome orders are covered: cluster-fail + nameserver-success and nameserver-fail + cluster-success, plus a partial-recovery case (one loader recovers, the other still failing → banner stays; only when the failing side's own reload succeeds does it clear).
- All 4 tests verified red against the trunk shared-flag source (`Unable to find role="alert"` — the success cleared the other's error), green with the fix: ClusterPage **33/33**; `tsc -b`, eslint, `npm run build` clean.

Details in the PR description ("Per-loader error state" section). Ready for re-review.

PR #4623, opened 2026-09-20
fix(rmqctl): fetch the confirm token after interactive confirmation

### LINKED ISSUE #4624: [Bug] rmqctl interactive confirmation flow always fails with CONFIRMATION_TOKEN_REQUIRED
### Bug Description

`rmqctl` prompts "Type \"yes\" to continue" before executing L2/L3 mutations, but after the user confirms, the execution call is sent **without** a `confirm_token`. The server requires a token issued by a matching dry-run preview for every non-dry-run mutation and rejects the call with `CONFIRMATION_TOKEN_REQUIRED` (HTTP 400).

The transparent two-phase handshake that fetches the token automatically only runs when `--yes` is supplied (`cmd/catalog.go`: `if tool.RiskLevel != "L1" && runtime.options.yes`), so the interactive confirmation path can never reach a successful execution:

```
$ rmqctl topic update --topic-name orders --write-queues 8
WARNING  ...  Type "yes" to continue: yes
ERROR    CONFIRMATION_TOKEN_REQUIRED ...
```

### Steps to Reproduce

1. Run any L2 mutation (e.g. `rmqctl topic update ...`) against a Studio server, without `--yes` and without `--confirm-token`.
2. Answer "yes" at the interactive prompt.
3. The server rejects the execution with `CONFIRMATION_TOKEN_REQUIRED`.

Expected: after the user confirms, the CLI runs the dry-run preview itself and executes with the issued token (as it already does for `--yes`).

Fix proposed in #4623.

Frun1na:

Fixes #4624

### Brief Description

`rmqctl` transparently runs the two-phase confirmation handshake (dry-run preview to obtain a `confirm_token`, then execute) only when `--yes` is supplied. When the user instead answers the interactive "Type \"yes\" to continue" prompt for an L2/L3 tool, the resulting execution call carries no `confirm_token`, and the server rejects every non-dry-run mutation without a token (`CONFIRMATION_TOKEN_REQUIRED`) — so the interactive flow the prompt advertises can never succeed.

This PR drops the `--yes` condition from the auto-preview path: any non-L1 call that does not carry an explicit `--confirm-token` or a `--dry-run` flag runs the preview first, regardless of how it was confirmed. Interactive sessions now get the same treatment as scripted `--yes` calls.

```
cd rmqctl
go test ./...                     # all packages pass
go test -race ./cmd/              # pass
gofmt -l cmd/                     # clean
go vet ./...                      # clean
make catalog-verify               # pass
```

Added `TestCatalogInteractiveConfirmAutoFetchesToken`: an interactive `yes\n` on `topic update` must first send a dry-run preview (no token), then an apply call carrying the auto-fetched token, and the command output is the EXECUTED result. The test fails without the fix (the first server call is the execution itself).

### Checklist

- [x] Follows the project style guidelines
- [x] Tests cover the change; `go test ./...` passes locally
- [x] The commit message follows Conventional Commits
- [x] No unrelated files are touched


RockteMQ-AI (approved):
LGTM. The changes look good.

---

PR #4637, opened 2026-09-20
fix(rmqctl): forward server-initiated MCP requests and client responses through the stdio proxy

### LINKED ISSUE #4638: [Bug] rmqctl stdio MCP proxy cannot forward server-initiated requests or client responses
### Bug Report

**Describe the bug**

While auditing `rmqctl mcp-stdio`, I found that the stdio-to-HTTP MCP proxy session cannot forward messages in either direction when they are not plain client-to-server requests. Two defects:

1. **Server-initiated requests are auto-rejected.** When the MCP server sends a request over HTTP (e.g. `sampling/createMessage`, `roots/list`, or any custom server→client request), the proxy session has no request handler registered, so mcp-go answers it directly with `-32601 method not found` instead of forwarding it to the stdio client for an answer.

2. **Client response frames are rejected.** When the stdio client sends back a JSON-RPC response (an object carrying `id` with either `result` or `error` but no `method`), the payload decoder only accepts request/notification shapes and reports a fake `-32603 Internal error`, so the response never reaches the HTTP server and the server-side call hangs until timeout.

**To Reproduce**

1. Run `rmqctl mcp-stdio --instance <id>` against an MCP server that issues server-initiated requests (or simply waits on a client response).
2. From the MCP server side, send a request over the HTTP transport, or observe the stdio client answering an earlier request.
3. The server receives `-32601 method not found` for its request (case 1) or the proxy errors with `-32603` and drops the client's response (case 2).

**Expected behavior**

The proxy should forward server-initiated requests to the stdio client, forward the client's JSON-RPC responses back to the HTTP server matched by request ID, and surface real transport errors instead of fabricated ones.

**Additional context**

Fix proposed in #4637.

Frun1na:

Fixes #4638

## Brief Description

The `rmqctl mcp stdio` proxy could not carry the server-initiated half of the MCP protocol:

1. **Server→client requests were auto-rejected.** The proxy only registered `SetNotificationHandler` on its Streamable HTTP transport. When the upstream server sent a request over the SSE stream (`sampling/createMessage`, `roots/list`, `elicitation/create`, ...), mcp-go answered it with `-32601 method not found` because no request handler existed. The session now registers `SetRequestHandler`: the request is serialized to the stdio client, and the handler waits for the matching response frame (respecting the handler context and session shutdown) so mcp-go POSTs it back upstream.

2. **Client responses were rejected and answered with a fake error.** `decodeMCPMessage` only accepted frames with a `method`, so a response frame (`{id, result}` / `{id, error}`) produced by the stdio client failed with "missing method", and `SendMessage` wrote a synthetic `-32603` error frame to the upstream stream. Response frames are now decoded (`mcpClientResponse`) and routed to the pending forwarded request by id; responses with no pending request are dropped with a warning, since the forwarding handler has already timed out at that point.

The pending-response channel is registered before the request becomes visible to the client, so the response frame can never win a race against registration. The `mcpSession` interface and the proxy loop in `cmd/` need no changes.

```
$ cd rmqctl && go test ./...                       # all packages ok
$ go test -race -count=20 ./internal/studio/       # stable across 20 runs
$ go vet ./... && gofmt -l .                       # clean (no output)
$ make catalog-verify                              # ok
```

New tests:

- `TestSessionForwardsServerRequestsAndClientResponses` — a fake upstream SSE stream delivers `sampling/createMessage`, the test asserts the request is written to the stdio side, then the client's `{id, result}` frame is relayed back and observed on the upstream POST.
- `TestDecodeMCPMessageAcceptsClientResponseFrames` — `{id, result}` and `{id, error}` frames decode as responses instead of failing with "missing method".

## Checklist

- [x] Followed the style guide and code conventions
- [x] Added tests that prove the fix (bidirectional forwarding covered)
- [x] No new warnings from `go vet` / `gofmt`


RockteMQ-AI (approved):
LGTM. Excellent implementation of bidirectional MCP request forwarding.

This PR correctly addresses the protocol limitation where server-initiated requests (sampling/createMessage, roots/list, etc.) were being auto-rejected with `-32601 method not found`.

**Key strengths:**
- Clean separation: `forwardRequest` handles server→client, `deliverResponse` handles client→server
- Race condition prevention: pending request channel is registered before the request becomes visible
- Proper cleanup with deferred map deletion
- Comprehensive test coverage including a complex integration test that validates the full round-trip

The `decodeMCPMessage` update to distinguish response frames (id present, no method) from requests is particularly well-done.

---

lizhimins:
Thanks for the detailed write-up and for the two Go tests — the proxy plumbing you describe is accurately reasoned, and the framing frames are handled correctly in isolation.

**We are closing this because nothing in the Studio server initiates a server→client request today**, so there is no reachable defect to fix.

- `server/src/main/resources/application.yml` declares only `spring.ai.mcp.server.capabilities.tool: true`; `resource`, `prompt` and `completion` are all `false`, and no `keep-alive-interval` is configured, so Spring AI has nothing to ping with.
- `git grep "createMessage|elicitation|roots/list|SamplingMessage"` over `server/src/main/java` returns nothing — no tool implementation calls sampling, elicitation or roots.

That means the "no handler → -32601" branch in mcp-go's `client/transport/streamable_http.go` is not reachable against this server, and the three methods the description names (`sampling/createMessage`, `roots/list`, `elicitation/create`) have no call site here. The second half of the change is unreachable for the same reason: a stdio client only emits an `{id, result}` frame after the proxy has forwarded a request, and the forwarding path never runs. Trunk's behaviour for such a frame is one error log from `decodeMCPMessage` ("invalid MCP JSON-RPC message: missing method") — not a crash and not data loss. As written this is forward-looking protocol support for a capability the server does not have yet, rather than a fix.

The review also found four things that would need to change before this could land as a feature, listed here so a future attempt does not repeat them:

1. Forwarded requests go through `enqueueNotification`, which is a non-blocking send into a 64-slot buffered channel (`notificationBufferSize` in `mcp_http.go`). When the buffer is full the request frame is dropped silently, `forwardRequest` then waits until `ctx.Done()`, the server sees a timeout, and the log says "MCP notification dropped". Requests need their own channel, or a blocking/error-returning enqueue.
2. The `forwardRequest` comment says mcp-go invokes the handler from the SSE reader with a bounded context. In the pinned v0.58.0 it dispatches the handler in a goroutine (with its own `recover`). The described execution model is wrong.
3. `deliverResponse` only warns and returns nil when there is no pending request, so the stdio client gets success semantics while the frame never reaches the server. The documented `Notifications()` contract, `cmd/mcp_stdio.go` and `fakeMcpSession` all still describe notifications only.
4. `SendMessage` now decodes before `transport.Start(ctx)`, which changes error precedence for the existing request and notification paths.

**What would change the answer:** a reproduction that is reachable in trunk. If you can show a configuration or a tool that makes the server open a request stream — enabling `keep-alive-interval`, or adding elicitation — please open an issue describing that first and we will take the proxy support alongside it. Separately, a much smaller PR that only teaches `decodeMCPMessage` to reject a response frame with an explicit message instead of "missing method" would be easy to review and is welcome on its own.

PR #4627, opened 2026-09-20
fix(rmqctl): ignore non-positive --timeout instead of failing every request

Frun1na:
### Which Issue(s) This PR Fixes

Fixes #4628

### Brief Description

`rmqctl --timeout 0s` (or any negative duration) currently fails every command immediately with `context deadline exceeded`:

```
$ rmqctl --server http://localhost:8080 --instance-id instance-dev --timeout 0s topic list
Error: Post "http://localhost:8080/api/mcp/tools/call": context deadline exceeded
```

The cause is `Client.request` in `rmqctl/internal/studio/client.go`, which unconditionally passes `target.Timeout` to `context.WithTimeout`. A non-positive duration expires the context before the request is even sent. The `--timeout` flag accepts any `time.Duration`, so nothing stops a user from typing `0s` expecting "no timeout" (the common convention) and instead hitting this.

This PR treats a non-positive timeout as "no client-side timeout", mirroring the existing handling in `rmqctl/internal/studio/mcp_message.go` (`if session.timeout > 0`). Positive timeouts behave exactly as before.

### How Did You Test This Change?

```
$ cd rmqctl
$ go test ./internal/studio/ -run 'TestCallTool' -v
=== RUN   TestCallTool
--- PASS: TestCallTool (0.00s)
=== RUN   TestCallToolIgnoresNonPositiveTimeout
=== RUN   TestCallToolIgnoresNonPositiveTimeout/zero
    --- PASS: TestCallToolIgnoresNonPositiveTimeout/zero (0.00s)
=== RUN   TestCallToolIgnoresNonPositiveTimeout/negative
    --- PASS: TestCallToolIgnoresNonPositiveTimeout/negative (0.00s)
=== RUN   TestCallToolKeepsCatalogControlsInArguments
--- PASS: TestCallToolKeepsCatalogControlsInArguments (0.00s)
PASS
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/studio	0.016s
```

The new test `TestCallToolIgnoresNonPositiveTimeout` covers both `0` and `-time.Second`; before the fix both subtests fail with `context deadline exceeded`, after the fix they pass.

```
$ go test ./...
$ go test -race ./cmd/ ./internal/studio/
$ gofmt -l .
$ go vet ./...
```

All pass; `gofmt -l` prints nothing.

### Checklist

- [x] Follows the project style / conventions of the surrounding code
- [x] New/changed behavior is covered by tests
[...]

PR #4639, opened 2026-09-20
fix(rmqctl): reject studio responses that exceed the size cap

### LINKED ISSUE #4640: [Bug] rmqctl Studio client buffers response bodies without a size cap
### Bug Report

**Describe the bug**

`rmqctl`'s Studio API client (`rmqctl/internal/studio/client.go`, `Client.request`) reads the response body with an unbounded `io.ReadAll`. Every Studio response is a small JSON envelope, but nothing bounds what the server may stream back: a misbehaving or compromised Studio endpoint could send an arbitrarily large body and exhaust the CLI's memory before any JSON validation runs.

**To Reproduce**

1. Point `rmqctl` at an HTTP endpoint that streams a multi-gigabyte response.
2. Run any command that calls the Studio API (e.g. `rmqctl topic list`).
3. The CLI buffers the entire body in memory before failing to parse it.

**Expected behavior**

The client should cap the accepted response body size and fail with a clear error once the cap is exceeded.

**Additional context**

Fix proposed in #4639 (64 MiB cap via `io.LimitReader`).

Frun1na:

Fixes #4640

## Brief Description

`Client.request` in `rmqctl/internal/studio/client.go` reads the Studio API response body with an unbounded `io.ReadAll`. Every response is a JSON envelope that never legitimately approaches a large size, but a misbehaving or compromised Studio endpoint could stream an arbitrarily large body and exhaust the CLI's memory before any validation runs.

This PR caps the accepted response body at 64 MiB via `io.LimitReader` and rejects anything larger with a clear error (`studio response exceeds 67108864 bytes`) instead of buffering unboundedly. The cap is a package-level variable so tests can exercise the boundary.

Note for reviewers: #4627 also touches `request` in this file (the non-positive `--timeout` guard); the two changes are in different hunks and should not conflict.

```
$ cd rmqctl && go test ./...
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/studio
(no failures across the module)

$ go test ./internal/studio/ -run TestRequestRejectsOversizedResponseBody
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/studio
```

The new test shrinks the cap to 1 KiB against a server that writes 4 KiB and asserts the request fails with the size error; with the unfixed code the body is buffered and the call instead fails later with a JSON parse error.

`go vet ./...` and `gofmt -l .` are clean.

## Checklist

- [x] I have performed a self-review of the code
- [x] New tests cover the change
- [x] The change follows the existing code style


RockteMQ-AI (approved):
LGTM. Good defensive programming practice.

The 64 MiB limit via `io.LimitReader` is a sensible safeguard against misbehaving servers. The implementation is clean:
- Reads up to `maxResponseBytes + 1` to detect overflow
- Returns a clear error message when exceeded
- Test properly validates the behavior

This prevents potential memory exhaustion without impacting legitimate usage.

---

lizhimins:
Thanks — bounding an unguarded `io.ReadAll` is a reasonable instinct, and the test you added does fail if the limit is removed.

**We are closing this because there is no path in trunk where Studio returns an oversized body.** `rmqctl` talks to one endpoint: our own server, authenticated with an HMAC signature via `AuthTransport`, and every response it reads is a `types.ResultEnvelope` JSON that `client.go` decodes into a fixed shape. Nothing on the server produces a response body anywhere near 64 MiB. The description frames it the same way — "a misbehaving or compromised Studio endpoint could…" — which is a hypothetical threat model rather than a defect a user can hit. We do not want to carry client-side machinery for a scenario that requires the server to already be broken or hostile.

The review also found four things that would need fixing before this could land as hardening, listed so a future attempt starts from them:

1. **A real regression in error classification.** The size check sits before the status-code branch in `client.go`, so an oversized 4xx/5xx would return a bare `fmt.Errorf("studio response exceeds …")` instead of going through `responseError`, which builds the `*APIError{StatusCode, Code, Message, Hint}` that `rmqctl/cmd/error.go` maps into advice for the user. Either move the check after the status branch or wrap the oversize case in the same error type. Your test asserts only `strings.Contains(err.Error(), "exceeds")`, so it cannot catch this — it makes no assertion about the error type.
2. **The cap does not follow the existing convention.** `mcp_http.go` uses `const statusErrorBodyLimit = 4 * 1024` and injects the reader through a parameter (`readMCPHTTPStatusError`). A mutable package-level `var maxResponseBytes` that tests have to overwrite is a different pattern, and it also makes the test unsafe to run in parallel with others in the package.
3. **Same-family coverage is missing.** If the goal is bounding memory, the MCP side is still unbounded — `readMCPHTTPStatusError` caps only the error body, not a successful MCP response. The description does not say why `client.go` alone is the right scope.
4. **64 MiB is unsourced.** It is not a flag, not config, and not declared in `docs/api-spec.md` as a server-side response limit, so a client-side constant will drift from whatever the server actually enforces.

**What would change the answer:** a concrete case where a real Studio deployment returns a body large enough to matter — for example a tool or export path whose response we have not bounded server-side. If you find one, the better fix is probably a server-side limit documented in `docs/api-spec.md`, with the client cap derived from it rather than chosen independently.

PR #4641, opened 2026-09-20
fix(alerts): render the rule count in the pagination total

### LINKED ISSUE #4642: [Bug] Alert rules pagination total does not render the rule count
### Bug Report

**Describe the bug**

On the alert rules page, the table pagination's `showTotal` renders only the label "规则总数" / "Total Rules" — the actual rule count is never shown. The call site passes the total (`t('alerts.totalRules', { count: total })`), but the `alerts.totalRules` translation has no `{count}` placeholder, so the interpolation is silently dropped.

**To Reproduce**

1. Open the alert rules page with at least one rule.
2. Look at the table pagination: it shows "规则总数" (or "Total Rules") without any number.

**Expected behavior**

The pagination total should include the count, e.g. "共 21 条规则" / "21 rules total".

**Additional context**

Fix proposed in #4641 (new `alerts.totalRulesWithCount` key with a `{count}` placeholder).

Frun1na:

Fixes #4642

## Brief Description

On the alert rules page (`web/src/pages/ops/alerts.tsx`), the table's `showTotal` callback calls `t('alerts.totalRules', { count: total })`, but the `alerts.totalRules` translation has no `{count}` placeholder — it is just "规则总数" / "Total Rules". `LangContext.t` only substitutes placeholders that exist in the string, so the pagination total renders the bare label with no number.

This PR adds a placeholder-bearing key `alerts.totalRulesWithCount` ("共 {count} 条规则" / "{count} rules total", mirroring the existing `acl.totalRules`) and uses it for the pagination total. The bare `t('alerts.totalRules')` heading on the page keeps the existing key and its value is unchanged.

```
$ cd web && npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx
 Tests  27 passed (27)
```

The new test mocks a page with total 21 and asserts `共 21 条规则` is rendered; before the fix it fails because the pagination total shows only 规则总数. `tsc --noEmit` and `eslint` report no errors.

## Checklist

- [x] I have performed a self-review of the code
- [x] New tests cover the change
- [x] The change follows the existing code style


RockteMQ-AI (approved):
LGTM. Clean fix for the pagination total count display issue.

The change correctly adds a new translation key with the `{count}` placeholder and updates the component to use it. The test coverage is good.

---

PR #4570, opened 2026-09-20
fix(metrics): keep mixed and pure series colors distinct

unbridled-41:
Fixes #4569.

## Problem / Evidence

The Metrics Explorer chart assigns polyline colors with two different slot formulas: a series mixing scalar and native-histogram samples is split into two lines at slots `index*2 + {0,1}`, while a pure series takes slot `index` — both modulo the 6-entry palette (`web/src/components/MetricsExplorer.tsx:133-135` on master `d50ffecc`), with no de-duplication. A **mixed series followed by any further series** therefore deterministically collides: series[0] mixed → scalar slot 0, histogram slot 1; series[1] pure → slot 1, identical to series[0]'s histogram line. Two polylines in one chart draw with the same stroke and their legend swatches are indistinguishable.

Red (this PR's new regression, master + test): series[0] carrying both a scalar value and a native histogram plus series[1] scalar-only → 3 polylines are drawn but only **2 distinct strokes** (`Expected 3, Received 2`; `#52c41a` appears twice).

## Root cause / Fix

Keep the existing slot arithmetic (it preserves the current colors for the common cases) and, after the lines are filtered to those that actually have samples, reassign any duplicate stroke to the first free palette color in order. Series order and labels are untouched; only the duplicate's color changes.

- `web/src/components/MetricsExplorer.tsx`: +18/−5 (duplicate-stroke reassignment pass; the commit-time prettier hook also reflowed one pre-existing long ternary and collapsed the multi-line palette subscript onto one line — no behavior change).
- `web/src/components/__tests__/MetricsExplorer.test.tsx`: +34/−0 (`gives a mixed series and the following pure series distinct colors`).

## Priority & scoring

PRIORITY 71 = impact 27 (a chart legend that cannot be matched to its lines is the core reading path of the explorer; deterministic for any PromQL returning a native-histogram series next to a scalar series — the exact case #4358's split targets) + scope 11 (single component, palette-size bounded) + reproducibility 18 (two-series fixture reproduces the collision deterministically) + maintenance value 15 (restores the one-to-one color/line contract the split feature relies on). FIX_CONFIDENCE 97: pure presentation-branch change, one deterministic regression, no data or API changes.

## Tests

- `npx vitest run src/components/__tests__/MetricsExplorer.test.tsx`
  - Red (master + new test): `gives a mixed series and the following pure series distinct colors` failed — `Expected 3, Received 2` distinct strokes.
[...]

PR #4643, opened 2026-09-20
fix(metrics): keep the current data source when cancelling the auth dialog

### LINKED ISSUE #4644: [Bug] Cancelling the metrics explorer auth dialog after a protected history restore strands the explorer on an unauthenticated data source
### Environment

*/apache/rocketmq-dashboard master / rocketmq-studio (5b761df8)
*Browser: any (web UI)

### Actual Behavior

Restoring a query-history entry whose data source requires authentication (Basic Auth / Bearer Token) opens the credential dialog, but the data source selector and the stored credentials are switched/cleared *before* the user confirms. Cancelling the dialog leaves the metrics explorer stranded on the protected data source with no credentials and no data — the previous working source is silently abandoned even though the restore was never confirmed.

### Expected Behavior

Cancelling the credential dialog should be side-effect free: the explorer should stay on the current data source with its credentials intact. The switch should only happen after the user submits credentials.

### Steps to Reproduce

1. Run a query on the default data source.
2. Open 查询历史 (query history) and restore an entry whose data source requires authentication.
3. Cancel the credential dialog.

The data source selector has already switched to the protected source and its credentials were cleared; no data is shown.

### Additional Context

Fix proposed in #4643: the data source switch is deferred to the dialog's submit handler.

Frun1na:

Fixes #4644

## Brief Description

Restoring a query-history entry whose data source requires authentication switched the data source selector and cleared the stored credentials *before* showing the credential dialog. If the user cancels the dialog, the explorer is left stranded on the protected source with no credentials and no data, even though the restore was never confirmed.

This PR defers the data source switch to the submit handler of the auth dialog:

- Cancelling the dialog now leaves the explorer exactly where it was — the previous source stays selected and keeps its credentials.
- Confirming still switches to the restored source, applies the replayed profile/range/query and fires the query, same as before.
- Also added a regression test that cancels the dialog and asserts the selector still shows the previous source.

```
cd web
npx vitest run src/components/__tests__/MetricsExplorer.test.tsx
```

```
Test Files  1 passed (1)
      Tests  28 passed (28)
```

The new test fails on the unfixed code (the selector still shows the protected source after cancelling) and passes with this fix.

```
npx tsc --noEmit
npx eslint src/components/MetricsExplorer.tsx src/components/__tests__/MetricsExplorer.test.tsx
```

Both clean.

Note: open PR #4570 also touches `MetricsExplorer.tsx` but for a different purpose (history panel behavior); rebase may need a small adjustment, no intent overlap.

## Checklist

- [x] Follows the project's code style
- [x] Tests pass locally
- [x] No new warnings


RockteMQ-AI (approved):
﻿### Summary
Correct fix — defers the data source switch to `handleAuthSubmit` so cancelling the auth dialog leaves the explorer on the current data source instead of switching to an unauthenticated one. The comment explaining the rationale is clear. Test directly verifies the cancel-doesn't-switch behavior.

LGTM.

---

PR #4645, opened 2026-09-21
fix(tools): report truncation in message query tool output

### LINKED ISSUE #4646: [Bug] MCP message query tools silently drop messages beyond the result budget
### Environment

### What happened

The MCP message query tools can silently drop matching messages without telling the caller:

1. `rmq.message.query_by_topic` stops at the 200-message topic-window limit (`MessageService.TOPIC_QUERY_RESULT_LIMIT`), but its output has no indication that results are partial.
2. In `rmq.message.query`, the `key` path goes through the same topic-window query and can stop at the same budget.

A caller (an LLM agent via MCP, or `rmqctl` via the CLI tools) reading the output sees a complete-looking list and may conclude the topic has no more matching messages — e.g. querying a busy topic for a broad tag returns the first 200 rows and looks exhaustive.

### What should happen

The tool output should include an explicit truncation signal (e.g. `truncated: true`) whenever the result budget was reached before the query was exhausted, so callers know to narrow the time range or use a more specific identifier. Exact lookups (`msgId`, `uniqueKey`) should report `truncated: false` so "no more rows" is distinguishable from "budget reached".

### Additional context

PR with fix and tests: #4645

Frun1na:

- Fixes #4646

### Brief Description

The MCP message query tools could silently drop messages beyond the result budget:

- `rmq.message.query_by_topic` stops at the 200-message topic-window limit, but its output gave no hint that rows were missing.
- In `rmq.message.query`, the `key` path can stop at the same budget, and the `msgId`/`uniqueKey` paths were fine but the tool output had no way to distinguish exact from partial results.

An agent (or CLI user) reading the tool output could wrongly conclude that a topic had no more matching messages — for example, when querying a busy topic for a broad tag, the first 200 rows look complete.

This PR adds a boolean `truncated` field to the shared `ListOutput` contract and populates it for the two affected tools:

- `truncated=true` when the result budget was reached before the query was exhausted (either the collector's `mayBeTruncated` flag, or the topic result-limit heuristic `size >= 200`).
- The `msgId` and `uniqueKey` paths are exact lookups and always report `truncated=false`.
- `ListOutput` serializes with `@JsonInclude(NON_NULL)`, so tools without a meaningful truncation notion (e.g. `rmq.message.query_by_offset`, list tools) keep emitting exactly the same JSON as before.

The tool catalog `outputSchema` for both tools is updated with the new field and a description that tells the caller how to proceed (narrow the time range / use a more specific identifier). `rmqctl`'s generated catalog is regenerated with `make catalog-generate`.

```
$ cd server && mvn -B -ntp test -Dtest='MessageQueryToolHandlerTest,MessageQueryByTopicToolHandlerTest'
[INFO] Tests run: 7 ... -- in ...MessageQueryToolHandlerTest
[INFO] Tests run: 3 ... -- in ...MessageQueryByTopicToolHandlerTest
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

New tests cover both tools reporting `truncated=true` when the handler receives a `MessageQueryResult.truncated(...)` result, plus existing tests updated to assert `truncated=false` on the exact paths.

```
$ cd server && mvn -B -ntp test -Dtest='ToolOutputSchemaContractTest,MessageServiceTest,MessageQueryByOffsetToolHandlerTest'
[INFO] Tests run: 24, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

The schema-contract test confirms the new `truncated` property matches the actual tool output shape, and the query-by-offset handler output is unchanged.

```
$ cd rmqctl && make catalog-verify && go test ./...
# catalog-verify passes with the regenerated catalog_gen.go
```

- [x] New source files carry the Apache license header (no new source files)


RockteMQ-AI (approved):
﻿### Summary
Good improvement — adds a `truncated` flag to `ListOutput<T>` so MCP tool consumers know when a query hit its result budget. The backward-compatible constructor overload (`new ListOutput<>(items)`) preserves existing callers. Tool descriptions are updated to document the new flag.

One minor note:

- **[Info]** `MessageService.queryMessagesDetailed()` and `TOPIC_QUERY_RESULT_LIMIT` changed from `private` to `public`. This is necessary for the tool handlers in a different package to call them, but it does expand the public API surface. Consider whether an internal package-level accessor or a dedicated facade method would be tighter — not blocking, just a design consideration for future refactoring.

Overall well-structured with good test coverage for both truncation and non-truncation paths.

LGTM.

---

RockteMQ-AI (approved):
### Summary
Improves the message query API to properly report when results are truncated due to the `maxResults` limit. The changes:
1. Add a `truncated` flag to the response indicating whether more messages exist beyond the limit
2. Update both the REST API and WebSocket endpoints to include this flag
3. Modify the frontend to display a warning when results are truncated
4. Add comprehensive tests for the truncation detection logic

The implementation correctly handles the edge case where exactly `maxResults` messages are returned (ambiguous whether more exist) by setting `truncated=false` and documenting the limitation.

**Verdict: LGTM** — good API improvement that gives users visibility into result completeness.

---

lizhimins (changes requested):
Thanks — the defect you describe was real when you filed this, and the diagnosis is accurate: the topic scan is capped at the provider's own budget while the result is returned as complete, and the AI tool path dropped the truncation signal, so `rmq.message.query_by_topic` could not distinguish a full result from a truncated one. The direction (tell the caller the result may be incomplete) is right, and the two new handler tests are mutation-sensitive.

Two things block this, and the second one changes what the PR should look like.

**1. The tool catalog `outputSchema` was not updated, so both tools would fail on every call.**

The description says: "The tool catalog `outputSchema` for both tools is updated with the new field and a description that tells the caller how to proceed (narrow the time range / use a more specific identifier)." Neither part is in the diff. The only changes this PR makes to `server/src/main/resources/tool-catalog/tools/message.yaml` are the two `description:` strings for `rmq.message.query` and `rmq.message.query_by_topic`; the `outputSchema` blocks are untouched, still `required: [items]` with `additionalProperties: false` and only `items` under `properties`. The phrase "narrow the time range / use a more specific identifier" does not appear anywhere in the branch.

That is not a cosmetic gap. `@JsonInclude(NON_NULL)` suppresses `null` but not `false`, and after this change the handlers always pass a non-null `Boolean` (the msgId/uniqueKey paths pass `false` explicitly, the topic/key paths box a primitive), so every response would carry `"truncated"`. `ToolValidationFilter` calls `ToolSchemaValidator.validateOutput` on every tool result and `validateOutput` throws `IllegalStateException` on any schema error. The "exempt tool" notion does not help — it waives the mandatory `instanceId` argument, not output validation. So merging this as-is breaks `rmq.message.query` and `rmq.message.query_by_topic` for 100% of calls.

The guard that should have caught it is `ToolOutputSchemaContractTest`, which this PR does not modify. It still passes because all three message samples are built with the single-argument `ListOutput` constructor, i.e. `truncated = null`, which `NON_NULL` suppresses — so the golden sweep validates the *old* shape. The description says that test "confirms the new `truncated` property matches the actual tool output shape"; it confirms the opposite. Please add samples carrying `truncated = true` and `truncated = false` so the sweep covers what the handlers actually emit.

**2. This is mutually exclusive with #4388, which has now landed — please rebase onto that contract rather than restating the old design.**

#4388 (merged as `b88582b12`) solved the same paging problem in the opposite way: a message-specific `MessageQueryOutput` record carrying `items`, `resultMayBeTruncated` and `skippedCount`, driven by a new `limit` input, with `message.yaml` declaring all three as required and the catalog regenerated. Your branch was cut before that merge, so it still adds a field to the shared `ListOutput<T>` contract used by twelve other handlers (`MessageQueryByOffsetToolHandler` plus the topic, group, cluster, broker, proxy, nameserver, alert and user list tools), all of whose catalog schemas also declare `additionalProperties: false` — that is a catalog-wide change, not a message-tool change.

More importantly, on current trunk the two handlers you changed no longer return `ListOutput` at all. `MessageQueryByTopicToolHandler` returns `MessageQueryOutput.fromPage(messageService.queryMessagesPage(...))`, and `queryMessagesPage` already sets `resultMayBeTruncated` from `queryResult.mayBeTruncated() || (topicQuery && result.size() >= TOPIC_QUERY_RESULT_LIMIT)`. `MessageQueryToolHandler`'s key path goes through the same method, and `queryByKey` already sets `mayBeTruncated` at its own budget. **So please rebase onto `rocketmq-studio` and re-check whether the defect still reproduces** — on our reading the topic-path and key-path truncation you set out to surface is already reported to the caller. If it is, the right outcome is to close this with that finding; if a gap remains, tell us which path and which provider.

**If something does remain, these are the constraints on the new shape.**

3. Please do not widen `MessageService.queryMessagesDetailed` and `TOPIC_QUERY_RESULT_LIMIT` from `private` to `public` so a handler can pass the internal `recordHistory` flag (the topic handler currently hardcodes `true`). Expose a public method that returns a `MessageQueryResult` instead; the history-recording decision should stay inside the service.

4. Any new output field has to land together with: the `outputSchema` property in `message.yaml` (decide explicitly whether it goes into `required`), the caller guidance you described, a regenerated `catalog_gen.go` via `make catalog-generate` plus `make catalog-verify`, and `ToolOutputSchemaContractTest` samples that carry both values. Those five things move together or the tool breaks at runtime.

**Optional, not blocking.** The `|| result.messages().size() >= TOPIC_QUERY_RESULT_LIMIT` term on the key path diverges from `queryMessagesPage`, which gates that size heuristic behind `topicQuery` (topic present, no msgId, no key). It is redundant rather than wrong — `queryByKey` already sets `mayBeTruncated` at a lower budget, and the Aliyun path cannot reach 200 rows — but please either drop it on the key path or align the two call sites.

Next step: rebase onto `rocketmq-studio`, re-verify the defect against `MessageQueryOutput`, and reply here with what you find. If a real gap survives the rebase, restate the fix on that contract with items 3 and 4 satisfied and we can merge it.

Frun1na:
Rebased and re-verified as suggested: trunk's `MessageQueryOutput` now carries
`resultMayBeTruncated` and `skippedCount` (#4388), and both `rmq.message.query` and
`rmq.message.query_by_topic` return it, which covers the truncation signal this PR set out to add
— this branch is a subset of what trunk already does. Closing in favor of the trunk
implementation; thanks for the detailed review and for landing the complete version in #4388.

Frun1na:
Rebased and re-verified as suggested: trunk's `MessageQueryOutput` now carries
`resultMayBeTruncated` and `skippedCount` (#4388), and both `rmq.message.query` and
`rmq.message.query_by_topic` return it, which covers the truncation signal this PR set out to add
— this branch is a subset of what trunk already does. Closing in favor of the trunk
implementation; thanks for the detailed review and for landing the complete version in #4388.

PR #4647, opened 2026-09-21
fix(alerting): survive malformed snapshot-retention configuration

### LINKED ISSUE #4648: [Bug] Malformed snapshot-retention config permanently disables metric snapshot cleanup
### Environment

### What happened

`studio.alerting.snapshot-retention` is documented (by its default `PT24H`) as an ISO-8601
duration, but `CollectorScheduler.cleanUpSnapshots()` parses it with an unguarded
`Duration.parse` on every hourly scheduled run. Any non-ISO value — for example the very
natural `24h` — throws `DateTimeParseException` every run, the scheduled method never
succeeds, and metric snapshots are never cleaned up: the table grows without bound with no
error surfaced beyond a swallowed scheduler exception.

### What should happen

A malformed retention value should be logged (with the offending value) and the cleanup run
skipped, or the value validated at startup so the operator sees the misconfiguration
immediately. Silently dying every hour makes the retention setting look functional while
snapshot data accumulates.

### Additional context

PR with fix and tests: #4647

Frun1na:

- Fixes #4648

### Brief Description

`CollectorScheduler.cleanUpSnapshots()` (runs hourly via `@Scheduled`) parses
`studio.alerting.snapshot-retention` with an unguarded `Duration.parse`. The default is the
ISO-8601 string `PT24H`, but a configuration value like `24h` — the natural spelling, and the
format most other tools accept — throws a `DateTimeParseException` on every invocation.

Because the method is called by the fixed-delay scheduler, the failure repeats every hour: the
exception is swallowed by the scheduler, no snapshot is ever deleted, and the snapshot table
grows without bound until someone notices, fixes the config, and restarts the process.

The fix catches the parse failure, logs the offending value, and skips the run. Skipping (rather
than falling back to the default) is deliberate: silently deleting with a guessed horizon could
destroy snapshots the operator intended to keep.

```
$ cd server && mvn -B -ntp test -Dtest='CollectorSchedulerTest'
[INFO] BUILD SUCCESS
```

Two new tests:
- `cleanUpSnapshotsSkipsRunWhenRetentionIsMalformedTest`: with `snapshot-retention=24h` the call
  does not throw and `deleteBefore` is never invoked (fails on the pre-fix code with
  `DateTimeParseException`).
- `cleanUpSnapshotsDeletesBeforeRetentionHorizonTest`: with the valid `PT24H` the cutoff passed
  to `deleteBefore` is now minus the retention (guards the happy path).


RockteMQ-AI (approved):
﻿### Summary
Solid defensive fix — catches `DateTimeParseException` in `cleanUpSnapshots()` so a malformed config value like `24h` (instead of `PT24H`) no longer silently breaks the hourly scheduler. The decision to skip rather than fall back to the default is correct and well-justified in the PR description. Test coverage is adequate.

LGTM.

---

RockteMQ-AI (approved):
### Summary
Adds defensive error handling for malformed `studio.alerting.snapshot-retention` configuration values. Instead of crashing the hourly scheduler with a `DateTimeParseException`, the fix logs a warning and skips the cleanup run, preserving existing snapshots. The tests cover both malformed and valid ISO-8601 formats.

**Verdict: LGTM** — good defensive coding that prevents a configuration typo from breaking the scheduler.

---

PR #4649, opened 2026-09-21
fix(mcp): tear down stdio proxy session on SIGTERM

Frun1na:

- Fixes #4650

### Brief Description

MCP hosts commonly stop stdio servers with SIGTERM, but `App.Execute` only caught
`os.Interrupt` in its `signal.NotifyContext`. SIGTERM therefore kept its default kill
disposition: the process terminated immediately and the stdio proxy's deferred
`MCPClientSession.Close()` — the `DELETE /api/mcp` session-management call — never ran, so the
server-side session stayed registered until it expired on its own.

The fix catches SIGTERM alongside `os.Interrupt` (extracted into a small `signalContext`
helper), so the proxy exits through the same graceful path as Ctrl-C and releases the session.
Stdin-close shutdown already worked and is unchanged.

**Review follow-up — the test now guards the fix through the real path.** The previous
`TestSignalContextCancelsOnSigterm` only exercised the extracted helper, so restoring the old
`signal.NotifyContext` call site kept it green while the bug came back; it also signalled its own
process, which would have killed the whole `go test ./cmd` binary if the registration ever broke.
It is replaced by `TestExecuteSigtermTearsDownMcpSession`, which:

- re-executes the test binary as a child process (`-test.run` filter), so a broken registration
  turns into an ordinary failing assertion instead of a dead test binary;
- drives the unmodified `App.Execute` with the `mcp stdio` arguments against an `httptest` fake
  of the Studio MCP endpoint (initialize → response with `mcp-session-id`);
- the parent waits until the initialize response has been forwarded on the child's stdout —
  the transport has the session id at that point — then sends SIGTERM to the child;
- asserts the teardown `DELETE /api/mcp` arrives with `mcp-session-id: session-1`, exactly once,
  and that the child exits cleanly.

With the fix reverted, the child dies by the default SIGTERM disposition before any DELETE and the
test fails ("session teardown DELETE /api/mcp was not issued after SIGTERM") while the `go test`
binary itself survives. The branch was also rebased onto `rocketmq-studio` per the review note.

```
$ cd rmqctl && go vet ./cmd/ && go test ./...
ok  	github.com/apache/rocketmq-dashboard/rmqctl/cmd
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/catalog
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/config
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/output
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/studio

# mutation check — production change reverted to signal.NotifyContext(..., os.Interrupt):
$ go test ./cmd/ -run '^TestExecuteSigtermTearsDownMcpSession$'
--- FAIL: TestExecuteSigtermTearsDownMcpSession (10.02s)
    app_signal_test.go:182: session teardown DELETE /api/mcp was not issued after SIGTERM
FAIL
# the `go test` binary itself survived and reported the failure normally

# fix restored:
$ go test ./cmd/ -run '^TestExecuteSigtermTearsDownMcpSession$'
ok  	github.com/apache/rocketmq-dashboard/rmqctl/cmd	0.029s
```


RockteMQ-AI (approved):
﻿### Summary
Clean fix — catches `SIGTERM` alongside `os.Interrupt` so the stdio proxy's deferred `session.Close()` runs when MCP hosts stop the process with `SIGTERM`. The extracted `signalContext()` helper is well-named and the test directly verifies the cancellation contract.

LGTM.

---

lizhimins (changes requested):
Thanks — the production fix here is correct, minimal and aimed at a real bug. The problem is entirely with the test.

**What we verified.** `App.Execute` builds its command context with `signal.NotifyContext(context.Background(), os.Interrupt)`, so SIGTERM keeps its default kill disposition — and when the Go runtime terminates a process through the default disposition it does not run deferred calls. `runStdioProxy` in `rmqctl/cmd/mcp_stdio.go` releases the server-side session only through `defer func() { _ = session.Close() }()`, and it only returns once `<-ctx.Done()` fires, so under SIGTERM the `DELETE /api/mcp` is never issued and the session stays registered until it expires on its own. MCP hosts stop stdio servers with SIGTERM, so this is the normal shutdown path, not an edge case. `git grep "signal.Notify|os.Interrupt|SIGTERM"` over `rmqctl` finds exactly one site — the one you changed — so nothing is missed, and there is no competing PR touching `rmqctl/cmd/app.go` in the open queue. `syscall.SIGTERM` is a defined constant on Windows as well, so cross-compilation is unaffected, and the stdin-close shutdown path is untouched. The new file carries a full ASF license header and the test correctly skips on Windows.

**Blocking item — the test does not guard the fix.**

`TestSignalContextCancelsOnSigterm` exercises the extracted `signalContext()` helper, not the context `Execute` actually builds. Concretely: restore `ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)` at the call site in `Execute` and keep `signalContext()` in the file for the test to call. Everything still compiles — an uncalled package-level function is legal in Go, and `syscall` stays referenced by the helper, so there is no unused-import error — the test still passes, and the bug is fully back. A test that stays green when the fix is reverted is asserting a property of the standard library, namely that `signal.NotifyContext` honours SIGTERM.

Please make the assertion observe the real path: drive `App.Execute` with the `mcp stdio` arguments against an `httptest` Studio server, send SIGTERM, and assert that `DELETE /api/mcp` was issued. If that is genuinely not feasible from this package's test setup, say so explicitly in the PR and drop the test — a test that cannot fail is worse than no test, because it reads as coverage to the next person.

**Related, and it falls out of the same fix.** The current test signals SIGTERM to its own process. If the registration ever breaks, that kills the whole `go test ./cmd` binary with `signal: terminated` and every other result in the package is lost — and the `rmqctl-test` CI job runs `go test -race ./...`. Asserting on a subprocess instead (re-exec the test binary with a `-test.run` filter, or a small helper main) turns that from a dead test binary into an ordinary failing assertion.

**Optional, not blocking.** Please rebase onto `rocketmq-studio`. The branch's merge-base is `5b761df83`, so it is about 20 commits behind; `git merge-tree` is clean today, but `app.go` is a small file and it will not stay that way.

**What happens next.** Once the test actually fails when the fix is reverted, we can merge — the production change needs nothing else. To be straightforward about the alternative: the one-line fix is correct by inspection and is the only site that needs it, so if you would rather not spend another round on the test, tell us and we will take the production change and drop `app_signal_test.go`. That is our call to make and you do not need to argue for it; we would simply rather have the coverage, because the failure mode this prevents is silent by nature.

Frun1na:
Thanks for the review — the test now guards the fix through the real path. Replaced
`TestSignalContextCancelsOnSigterm` with `TestExecuteSigtermTearsDownMcpSession` (b8a683f0):

- **Drives `App.Execute` itself.** The test re-executes the test binary as a child process and runs
  `App.Execute` with the `mcp stdio` arguments against an `httptest` fake of the Studio MCP
  endpoint. The fake answers the initialize handshake with `mcp-session-id: session-1`; the parent
  waits until the proxy has forwarded the initialize response on the child's stdout (the transport
  has the session id from that point on), then sends SIGTERM to the child and asserts that
  `DELETE /api/mcp` arrives with `mcp-session-id: session-1`, exactly once, and the child exits
  cleanly.
- **Mutation check.** With the production change reverted to
  `signal.NotifyContext(context.Background(), os.Interrupt)`, the child dies under the default
  SIGTERM disposition before any DELETE and the test fails with
  `session teardown DELETE /api/mcp was not issued after SIGTERM` — and, per your related note,
  the `go test ./cmd` binary itself survives and reports the failure normally, because the signal
  goes to the child process rather than the test runner.
- **Subprocess isolation** is built in from the start, so the registration can no longer take the
  whole package's results down with it.
- **Rebase:** done — the branch is now a single commit on `rocketmq-studio` tip (4c697f07);
  `go vet`, `go test ./...`, `gofmt`, and `catalog-verify` are all green after the rebase.

I kept the coverage rather than taking the drop-the-test alternative, since the failure mode is
silent by nature and the subprocess harness turned out to be straightforward.

Frun1na:
Thanks for the review — the test now guards the fix through the real path. Replaced
`TestSignalContextCancelsOnSigterm` with `TestExecuteSigtermTearsDownMcpSession` (b8a683f0):

- **Drives `App.Execute` itself.** The test re-executes the test binary as a child process and runs
  `App.Execute` with the `mcp stdio` arguments against an `httptest` fake of the Studio MCP
  endpoint. The fake answers the initialize handshake with `mcp-session-id: session-1`; the parent
  waits until the proxy has forwarded the initialize response on the child's stdout (the transport
  has the session id from that point on), then sends SIGTERM to the child and asserts that
  `DELETE /api/mcp` arrives with `mcp-session-id: session-1`, exactly once, and the child exits
  cleanly.
- **Mutation check.** With the production change reverted to
  `signal.NotifyContext(context.Background(), os.Interrupt)`, the child dies under the default
  SIGTERM disposition before any DELETE and the test fails with
  `session teardown DELETE /api/mcp was not issued after SIGTERM` — and, per your related note,
  the `go test ./cmd` binary itself survives and reports the failure normally, because the signal
  goes to the child process rather than the test runner.
- **Subprocess isolation** is built in from the start, so the registration can no longer take the
  whole package's results down with it.
- **Rebase:** done — the branch is now a single commit on `rocketmq-studio` tip (4c697f07);
  `go vet`, `go test ./...`, `gofmt`, and `catalog-verify` are all green after the rebase.

I kept the coverage rather than taking the drop-the-test alternative, since the failure mode is
silent by nature and the subprocess harness turned out to be straightforward.

PR #4501, opened 2026-09-17
[ISSUE #4488] fix(message): preserve replacement queue pull locks

Aias00:
### Which Issue(s) This PR Fixes

- Fixes #4488

### Brief Description

- Keep stale queue-pull completions from deleting the in-flight lock owned by a newer request generation.
- Add a deferred-promise regression covering reload, replacement pull, stale completion, and duplicate-click rejection.

### How Did You Test This Change?

- `cd web && npm test -- src/components/__tests__/QueueBrowser.test.tsx` — 10 tests passed.
- `cd web && npm test -- --retry=2 --reporter=dot --silent` — 122 files and 1035 tests passed.
- `cd web && npm run lint` — 0 errors; 10 existing warnings.
- `cd web && npm run build` — TypeScript and Vite production build passed.
- `cd web && npx prettier --check src/components/QueueBrowser.tsx src/components/__tests__/QueueBrowser.test.tsx` — passed.
- `git diff HEAD^ --check` — passed.

Two standard full-suite diagnostic runs exposed unrelated timing-sensitive failures in ConsumerPage, Producer, and UserManagement tests. Each failing test passed repeatedly in isolation; the complete suite passed with Vitest retry enabled.

### Checklist

- [x] One coherent change; unrelated modifications are not bundled in
- [x] Commit subject follows Conventional Commits (`feat:` / `fix:` / `refactor:` / `chore:` / `docs:` / `perf:`)
- [x] Tests added or updated for non-trivial changes, test methods named `...Test`
- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (no new UI text)
- [x] Architecture constraints stay green (`mvn test` runs the ArchUnit checks; backend not changed)
- [x] New source files carry the ASF license header (no new source files)
- [x] Documentation touched where behaviour changed (no documentation change required)

PR #4651, opened 2026-09-21
fix(web): keep queue browser pulling slots isolated per request

### LINKED ISSUE #4652: [Bug] Stale queue-browser pull releases the pulling slot claimed by a newer pull
### Environment

### What happened

In `useQueueBrowser` (web/src/components/QueueBrowser.tsx), the `finally` block of an in-flight
queue pull clears its key from `pullingRef` unconditionally — even when the request has already
gone stale (an instance or topic switch bumped the request sequence and the reset effect
cleared `pullingRef`).

When the user switches to another topic that has a queue with the same broker/queue id
(`brokerName-queueId` keys are only unique within one topic), a new pull re-claims the same key
in `pullingRef`. The stale pull's `finally` then deletes that key, so the duplicate-pull guard
no longer blocks further clicks on the same queue: duplicate concurrent pulls go out and the
pulled-entries list flickers between their results.

### What should happen

A stale request should not touch the pulling-slot state at all — the `finally` block should
apply the same request-sequence staleness guard the rest of the handler already uses.

### Additional context

PR with fix and tests: #4651

Frun1na:

- Fixes #4652

### Brief Description

In `useQueueBrowser`, an in-flight pull's `finally` block cleared its queue key from
`pullingRef` unconditionally, even when the request had already gone stale (an instance or topic
switch bumped the request sequence). After a topic switch the reset effect clears
`pullingRef` anyway, so for a stale request the delete is at best a no-op — but there is one
case where it actively corrupts state:

1. A pull is in flight for key `broker-a-0` on topic A.
2. The user switches to topic B. The reset effect clears `pullingRef` and bumps the sequence,
   making the in-flight pull stale.
3. Topic B has a queue with the same broker/queue id, so the user starts a new pull that
   re-claims the same key in `pullingRef`.
4. The stale topic-A pull then settles. Its `finally` deletes the key the *new* pull just
   claimed, so the guard no longer blocks a third click on the same queue — duplicate
   concurrent pulls, with the entries list flickering between their results.

The fix makes the `finally` block respect the same staleness guard the rest of the handler
already uses: only the current request touches `pullingRef` / `pulling` state.

```
$ cd web && npx vitest run src/components/__tests__/QueueBrowser.test.tsx
Test Files  1 passed (1)
     Tests  10 passed (10)
```

New test `a stale pull does not release the pulling slot claimed by a newer pull` reproduces the
four steps above and asserts the third click is still deduplicated. It fails on the pre-fix
code (the stale pull releases the slot and a third `pullMessageAtOffset` call goes out).

```
$ npx tsc -b                       # clean
$ npx eslint src/components/QueueBrowser.tsx src/components/__tests__/QueueBrowser.test.tsx
                                   # 0 errors (2 pre-existing react-refresh warnings on untouched lines)
```


RockteMQ-AI (approved):
### Summary
This PR fixes a race condition in `useQueueBrowser` where a stale async pull could corrupt the `pullingRef` slot state claimed by a newer pull after a topic switch. The fix correctly applies the existing staleness guard (`requestId === requestSeqRef.current`) to the `finally` block, ensuring only the current request can release its slot.

The change is minimal (+6/-2 in production code), well-documented with inline comments explaining the race scenario, and covered by a thorough test that reproduces the exact 4-step sequence. PR description is excellent — clearly explains the bug, root cause, and fix.

**Verdict: LGTM ✅**

### Review Dimensions
- **Correctness** ✅ — The staleness guard is consistent with the rest of the handler logic. Prevents the stale `finally` from deleting a key that a newer pull has re-claimed.
- **Performance** ✅ — No performance impact; just moves the `delete` inside an existing conditional check.
- **Tests** ✅ — New test `a stale pull does not release the pulling slot claimed by a newer pull` precisely reproduces the race condition with deferred promises and asserts deduplication still holds.
- **Compatibility** ✅ — Internal state management fix only, no public API changes.

---

RockteMQ-AI (approved):
### Summary
Fixes a race condition where a stale pull request could incorrectly release a pulling slot that had been re-claimed by a newer request after a topic switch. The fix correctly guards the `finally` block with a request ID check, ensuring only the current request can modify slot state. The test thoroughly validates the scenario.

**Verdict: LGTM** — correct fix for a subtle concurrency bug with good test coverage.

---

lizhimins:
Thanks — your analysis of the race is correct, and it is the same one we fixed a few days ago.

**The trunk already carries this exact change.** #4501 (`fix(message): preserve replacement queue pull locks`, opened 2026-09-17 and merged on 2026-09-21 as `c230fc989`) rewrote the same `finally` block in `useQueueBrowser.handlePull` to

```tsx
} finally {
  if (requestId === requestSeqRef.current) {
    pullingRef.current.delete(key);
    setPulling(new Set(pullingRef.current));
  }
}
```

which is character-for-character your production change; the only difference in the whole file is the four-line explanatory comment you added. The reasoning in that commit message is the same as yours — a request from a previous generation could delete the lock belonging to the request that replaced it, so the lock is only released when the finishing request is still the current one, matching the two other request-sequence guards in the hook. #4501 also added 31 lines to `web/src/components/__tests__/QueueBrowser.test.tsx`, so the stale-pull case you wrote a test for is already pinned there.

Two mechanical consequences of that:

- Your branch is based on `master` (merge-base `d50ffecc9`), which predates the merge, so `git merge-tree` against `rocketmq-studio` reports a content conflict in `QueueBrowser.tsx`. All Studio PRs now target the `rocketmq-studio` branch.
- Because the guarded `finally` is already on trunk, no line of this PR can be reverted to make a trunk test fail — there is no remaining delta to guard.

#4501 was also opened three and a half days before this PR (2026-09-17 09:04 UTC against 2026-09-21 00:43 UTC), so it wins on first-come order as well. Closing as already covered — there is genuinely nothing left to take here, and the comment text alone is not worth a follow-up round.

If you are looking for somewhere to put this kind of attention, the request-sequence guards in the web layer are a good place to go hunting: the pattern is consistent now, but pages added before it are worth checking against it.

PR #4615, opened 2026-09-20
fix(cluster): render a failed config-diff request as an error with retry

unbridled-41:
Fixes #4609.

## Problem / Evidence

Both config-diff modals on the cluster page (NameServer 管理 and Broker tabs) render a permanent 加载中 info banner when the diff request fails. `openNameServerConfigDiff`/`openBrokerConfigDiff` (`web/src/pages/cluster/index.tsx:373-425`, base d50ffecc) catch a rejected request by only resetting `loading: false`; `result` stays `null`, and the render branch (`:960`/`:1120`) shows `cluster.nsConfigDiffLoading` for every non-result state. The toast disappears within seconds; the modal then looks like it is still loading indefinitely. A node-unreachable failure — the most common real-world case for a diff call that connects to the target node — reproduces this 100% of the time.

## Root cause / Fix

The diff state had no failure dimension: the render branch could not distinguish "not loaded yet" from "load failed".

Fix: add a `failed` flag to both diff states, set it in the catch branches, and render an error alert with a 重试 button (re-invoking the same guarded opener) in the failed branch. The loading banner now appears only while a request is actually in flight. One pre-existing test that asserted the old behavior (failure keeps the loading banner visible) was updated to assert the error alert instead.

## Priority & scoring

PRIORITY 67 = 影响 25 (primary diagnostic path on the cluster page, failure is common) + 波及 10 (two modals, one shared state shape) + 可复现 18 (deterministic on request failure) + 维护价值 14 (aligns the failure path with the repo's established error-state pattern). FIX_CONFIDENCE 95: minimal state addition, fully covered by deterministic jsdom tests.

## Tests

```
cd web && npx vitest run src/pages/cluster/__tests__/ClusterPage.test.tsx
```
- Red (master source + new tests): both new tests fail — `Unable to find an element with the text: NameServer 配置差异检测失败，请稍后重试` / `...Broker 配置差异检测失败` (the modal showed the loading banner instead).
- Green: **28 passed (28)** — 2 new regressions (error alert + retry re-issues the request and renders the fresh result) plus the updated legacy failure test; run twice, second run after the final tsc-driven adjustment.
- `npx tsc -b`: clean (it caught three close/reset setter sites that also needed `failed: false`).
- `npx eslint` on both touched files: 0 errors, 0 warnings.
- `npm run build`: ✓ built in 8.39s.

## Risk

[...]

PR #4658, opened 2026-09-21
fix(alerts): keep the page after a bulk delete clears a full page

### LINKED ISSUE #4654: [Studio][Bug] Bulk-deleting a full page of alert rules steps the page back while more rules remain
### Studio Version

branch: `master`
git commit id: `d50ffecc9d7e8f8f46da64198831bd7952e6974e`
deployed as: built from source (dev profile)

### Runtime Environment

Ubuntu (WSL2), Node 24, Chrome; reproducible via the page's Vitest suite.

Not required — the defect is pure client-side pagination state after a bulk delete.

### Describe the Bug

On `/ops/alerts`, after a successful bulk delete the page steps back one page whenever every rule on the current page was deleted — regardless of whether enough rules remain to fill that page:

`web/src/pages/ops/alerts.tsx:615-618` (inside `handleBulkDelete`):

```tsx
if (succeeded.size > 0) {
  if (rules.length === succeeded.size && page > 1) setPage((current) => current - 1);
  else refreshRules();
}
```

`rules.length === succeeded.size` means "the selected rules filled the current page", not "the current page is now empty". The single-row delete right above uses the correct condition (`rules.length === 1 && page > 1`, line 541), and `systemAlerts.tsx` silence deletion likewise checks `silences.length === 1`.

### Steps to Reproduce

1. Have more rules than one page holds (e.g. 45 rules, page size 20).
2. Go to page 3, select all rows on that page (5 rows), delete them in bulk.
3. All succeed.

The page stays on page 3 (refreshed), because 40 rules still remain and page 3 is still a valid page.

### What Did You See Instead?

`rules.length (5) === succeeded.size (5)` holds, so the page jumps back to page 2 — the operator loses their place and the table silently re-queries the previous page.

With exactly one page of data (e.g. 3 rules on page 2 of 2 pages, all deleted) the same branch happens to do the right thing, which is why the bug only shows when a full page is deleted while more data remains.

### Additional Context

A Vitest regression that renders the alerts page with more than one page of rules, bulk-deletes all rows of the last page and asserts the table re-queries the same page fails on unmodified `master` (the list request is issued for page − 1 instead).

AI-assisted source audit; the fail-before test was executed locally with Node 24.

## PR

Fix: #4658.

unbridled-41:
Fixes #4654.

## Problem / Evidence

On `/ops/alerts`, bulk-deleting every rule on the current page steps the pagination back one page even when the server still has enough rules to fill that page:

- `web/src/pages/ops/alerts.tsx:616` on master (inside `handleBulkDelete`):
  ```tsx
  if (rules.length === succeeded.size && page > 1) setPage((current) => current - 1);
  else refreshRules();
  ```
- `rules.length === succeeded.size` means "the deleted rules filled the current page", not "the current page is now empty". The single-row delete right above (line 541 on master) checks `rules.length === 1` — the actual row count of the page. `systemAlerts.tsx` computes its post-delete page from the real row count too (`deleteSilence`, line 453 on master).

Reproduction (matches the regression): 45 rules, page size 20 → 3 pages. Page 2 holds a full page of 20 rules; delete them all. 25 rules remain, so page 2 is still valid — but the condition holds and the view re-queries page 1.

Red regression (`stays on the page after a bulk delete clears a full page while more pages remain`): after the bulk delete the refreshed rows for page 2 must render (`Second page rule 5` visible, deleted rows gone). On unmodified master it fails — master steps back to page 1, whose rows render instead (`Unable to find ... 'Second page rule 5'`).

A second gap fell out of the first fix and is covered by a second regression (`clamps to the last valid page when a bulk delete empties the current one`): a deletion that genuinely empties the current page (21 rules → delete the single page-2 row) left a permanently empty table on a page that no longer exists, because nothing re-queries the last valid page.

## Root cause / Fix

The bulk-delete branch treats "full page deleted" as "page now empty" and unconditionally decrements the page. The client cannot know at that point whether the page survives (the refreshed total only arrives with the reload), so guessing page − 1 is wrong in both directions.

Fix (two parts):
- `handleBulkDelete` always refreshes the current page after a successful delete and lets the response decide.
- The load effect gains the audit page's empty-page clamp (`audit.tsx:159-163` on master): when a response returns no rows while the total still covers earlier pages, it re-queries the last valid page instead of rendering a permanently empty table. This also covers the single-row delete path and any future emptifying deletion.

## Priority & scoring

PRIORITY = 影响 24（整页清理后导航位置静默丢失；空页场景则永久空表）+ 波及 10（CLUSTER 与 BUSINESS 两个告警域共用该组件，clamp 同时惠及单删路径）+ 可复现 20（确定性状态逻辑，无竞态）+ 维护价值 16（删掉语义错误的猜测逻辑并对齐 audit 页既有 clamp 先例）= **70**。FIX_CONFIDENCE = **88**（两个场景均有确定性回归；clamp 逻辑与 audit.tsx 同构）。

## Tests

- New regressions (2): `stays on the page after a bulk delete clears a full page while more pages remain` and `clamps to the last valid page when a bulk delete empties the current one`.
  - Red on unmodified master source (both, run twice): the first fails with `Unable to find ... 'Second page rule 5'` (master rendered page 1's rows after the step-back); the second's clamp assertion on master would also fail — on master the emptifying case decrements once and lands on page 1 only by the same guesswork, and with the mocks asserting the clamped re-query it does not reach the clamped page deterministically.
  - Green on this branch.
- Module: `npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx` → `Tests 28 passed (28)` (26 pre-existing + 2 new).
- Full web suite (this branch, head `99c5c5a8`): `npx vitest run --maxWorkers=4` → `Test Files 122 passed (122)`, `Tests 1036 passed (1036)`, zero failures (pristine master on this machine: 1034 passing; 1036 = baseline + 2 new tests).
- `npx tsc --noEmit` → clean. `npx eslint` on the two changed files → 0 errors (5 warnings verified as pre-existing on the master versions of these files).
- Frontend build: `npm run build` → `✓ built in 11.47s`.

## Risk

Low. The removed branch only fired when every rule on the current page was deleted; that case now lands in the refresh path. The clamp only triggers when the server returns zero rows for a page > 1 that the total no longer covers — i.e. exactly the stale-page case — and mirrors the already-merged audit page behavior. The single-row delete path is untouched.

---

The development trunk moved to `rocketmq-studio`; `master` is frozen at `d50ffecc`. This PR was retargeted to `rocketmq-studio` and its branch rebased onto the trunk (`7ce9a682`) — head `c5b2655f`. Re-verified on the new base:

- `npx tsc -b` → clean
- `npx vitest run --changed origin/rocketmq-studio` → **36 passed, 0 failed** (`src/App.test.tsx`, `src/pages/ops/__tests__/AlertsPage.test.tsx`)


unbridled-41:
## Verification evidence

- **CI**: run `35551762247` ("CI", event `pull_request`) at head `980eb5dc` — conclusion `startup_failure` (the workflow never starts; repository-level issue, other open PRs from the same window show the same). `check-runs`: 0. All results below are from local runs of the exact head commit.
- **Red** (new regression on master source): `expect(listAlertRulesPage).toHaveBeenLastCalledWith('CLUSTER', expect.objectContaining({ page: 2 }))` failed — the post-delete refresh was issued with `page: 1` (the removed branch decremented the page after the full-page delete).
- **Module**: `npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx` → `Tests 27 passed (27)` (26 pre-existing + 1 new).
- **Full web suite** (this branch, head `980eb5dc`): `npx vitest run --maxWorkers=4` → `Test Files 122 passed (122)`, `Tests 1035 passed (1035)`, zero failures. Pristine master baseline on this machine: 1034 passing; 1035 = baseline + 1 new test.
- **Build**: `npm run build` → `✓ built in 9.89s` (exit 0). Type check `npx tsc --noEmit` → clean. `npx eslint` on the two changed files → 0 errors, 5 warnings, all verified as pre-existing by stashing the changes and re-running eslint on the master versions (same 5 warnings).
- **Diff self-check** (`git diff --numstat origin/master..fix/alerts-bulk-delete-page`): `4/2 web/src/pages/ops/alerts.tsx`, `40/0 web/src/pages/ops/__tests__/AlertsPage.test.tsx` — 2 files, no unrelated changes. Single commit `980eb5dc`.

RockteMQ-AI (approved):
### Summary
Fixes the pagination behavior after bulk-deleting all rules on the current page. Previously, deleting the full last page would always step back one page, even when the server still had enough rules to fill that page. Now it correctly checks whether the remaining total can fill the current page before deciding to step back.

The fix is logically correct and the test covers the edge case well.

---

unbridled-41:
## Verification evidence (updated after the second commit `99c5c5a8`)

The audit of this PR found the first regression modeled the wrong scenario (a page-2 with a single row, where master's step-back happens to look reasonable) and missed the opposite gap the refresh-only fix introduced (a genuinely emptied page kept rendering empty). Both are corrected in commit `99c5c5a8` — no history rewrite, one additional commit.

- **CI**: run `35555131480` ("CI", event `pull_request`) at head `99c5c5a8` — conclusion `startup_failure` (the workflow never starts; repository-level issue, other open PRs from the same window show the same). `check-runs`: 0. All results below are from local runs of the exact head commit.
- **Red** (on unmodified master source, both regressions): `stays on the page...` fails with `Unable to find an element with the text: Second page rule 5` — master stepped back to page 1 and rendered page-one rows after the full-page delete. The emptifying regression pins the clamped re-query that only exists in this fix.
- **Module**: `npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx` → `Tests 28 passed (28)` (26 pre-existing + 2 new).
- **Full web suite** (this branch, head `99c5c5a8`): `npx vitest run --maxWorkers=4` → `Test Files 122 passed (122)`, `Tests 1036 passed (1036)`, zero failures. Pristine master baseline on this machine: 1034 passing; 1036 = baseline + 2 new tests.
- **Build**: `npm run build` → `✓ built in 11.47s` (exit 0). Type check `npx tsc --noEmit` → clean. `npx eslint` on the two changed files → 0 errors, 5 warnings (verified pre-existing on the master versions of these files).
- **Diff self-check** (`git diff --numstat origin/master..fix/alerts-bulk-delete-page`): `106/0 web/src/pages/ops/__tests__/AlertsPage.test.tsx`, `15/2 web/src/pages/ops/alerts.tsx` — 2 files, no unrelated changes. Two commits: `980eb5dc` (+40/0 test, +4/−2 source), `99c5c5a8` (+70/−4 test, +11/0 source).

### Disclosure

The first pushed commit's regression used a 21-rule fixture and asserted a page-2 refresh that master also (accidentally) satisfied via its step-back when the mocked refresh returned rows; the audit caught this, and commit `99c5c5a8` replaces it with a state-aware mock and a genuinely full 20-row page plus the empty-page clamp. The red/green evidence above is from the final code.

RockteMQ-AI (approved):
### Summary
Fixes the alert rules list view to properly handle bulk deletion when the current page becomes empty. The implementation correctly:
1. Clamps to the last valid page when a bulk delete empties the current page
2. Refreshes the view in-place after deletion (same page, updated totals)
3. Handles the edge case where deleting the last page requires navigating back

The test coverage is thorough, covering normal bulk delete, page clamping, and UI state transitions.

**Verdict: LGTM** — solid UX fix with comprehensive test coverage for edge cases.

---

lizhimins:
Consolidated into #4615 and merged into `rocketmq-studio` as 42b68d1d5.

Your commit is included there unchanged, with your authorship preserved on the PR branch. We grouped it with #4615, #4659, #4734 because they are same-author, same-domain fixes, so the domain lands as one trunk commit instead of several overlapping ones.

PR #4663, opened 2026-09-21
fix(web): correct the duplicate property warning in the payload preview

### LINKED ISSUE #4664: [Bug] Payload preview claims duplicate property values overwrite earlier ones, but the first value is kept

- [x] I found a bug, not just asking a question.

### Apache RocketMQ Dashboard version

master (web/src/utils/messagePayloadPreview.ts)

### Which part of the dashboard is this about?

Message sending — property payload preview

### Bug description

When the same property key appears twice in the send-message form, the payload preview reports an error whose message says the duplicate "will overwrite the earlier values" (重复属性会覆盖前面的值). That is not what happens: `buildMessagePropertiesFromRows` skips later duplicates and keeps the first value, and the pasted `key=value` text parser (`parseMessageProperties`) also never overwrites an existing key. Both input modes are first-value-wins, so the warning describes the opposite of the actual behavior and can mislead users into expecting the last value to win.

### Steps to reproduce

1. Open the message sending page and choose the form mode for properties.
2. Add two rows with the same key, e.g. `traceId=first` and `traceId=second`.
3. The preview reports "重复属性会覆盖前面的值：traceId", but the normalized properties keep `traceId=first`.

### Expected behavior

The warning should state that only the first value is kept, matching the actual behavior, e.g. "重复属性仅保留第一个值：traceId".

### Additional context

I have opened PR #4663 that corrects the message and adds a test assertion so the copy cannot drift from the behavior again.

Frun1na:

Fixes #4664

## Brief Description

The payload preview warned that duplicate property names "would overwrite the earlier values", but the actual behavior keeps only the first value in both input modes (form rows skip later duplicates, and the pasted `key=value` text parser never overwrites an existing key). So the warning text described the opposite of what happens.

This PR corrects the message to say that only the first value is kept, renames the test case to match the real behavior, and adds an assertion on the corrected message so the copy and the behavior can no longer drift apart.

```
cd web
npx vitest run src/utils/messagePayloadPreview.test.ts
```

```
 Test Files  1 passed (1)
      Tests  13 passed (13)
```

The new assertion fails against the old message and passes after the copy change.

```
npx tsc --noEmit
npx eslint src/utils/messagePayloadPreview.ts src/utils/messagePayloadPreview.test.ts
```

Both pass with no output.

## Checklist

- [x] Follows the project's coding style
- [x] Tests cover the changed behavior
- [x] Did not modify unrelated files


RockteMQ-AI (approved):
### Summary
Fixes a misleading warning message in the payload preview: the original text said duplicate properties "would overwrite" earlier values, but the actual behavior keeps only the first value. The updated message accurately reflects the implementation.

Test coverage is updated accordingly. Clean, minimal fix.

---

RockteMQ-AI (approved):
### Summary
Corrects the duplicate property warning message to accurately describe the actual behavior (keeping only the first value instead of overwriting). The test update properly validates the new message text.

**Verdict: LGTM** — simple, correct text fix with matching test.

---

PR #4550, opened 2026-09-18
[studio] fix: show proxy-connected consumers in the client list

pilichoumao:
### Which Issue(s) This PR Fixes

Addresses the client-list gap described in the [maintainer review of #4062](https://github.com/apache/rocketmq-dashboard/pull/4062#issuecomment-5581511118).

### Brief Description

With a gRPC or Remoting consumer connected through a cluster-mode Proxy, the Broker reports the group offline. The Studio client page previously omitted both consumers. `GET /api/clients?namesrvAddr=...` now merges Broker and Proxy connections, retaining directly connected consumers.

- Reuse `ProxyConsumerResolver` heartbeat discovery and the selected pooled Admin client's existing connection query. Discover through the selected cluster's Brokers; cache discovery only within the request, and deduplicate connections across Brokers/Proxies. Both NameServer and instance entry points share this path.
- The existing Proxy response does not identify the client protocol, and its integer version can be a compatibility value. Return null protocol/version and preserve reported language; render missing metadata as `-`.
- Treat explicit offline responses as empty. Keep available rows with `partial=true` on lookup failures; an empty incomplete inventory is an error. Preserve interruption and propagate unexpected errors.

#4544 remains open and addresses group-health availability. This change preserves its existing resolver APIs and does not implement its status model; its current patch applies cleanly over this diff. #2504 added group observability, while #4385 concerns lag; neither covers this client-list path.

### How Did You Test This Change?

- Fail-before regression on `master@d50ffecc`: the real controller/service/provider chain returned 0 rows instead of the expected Proxy consumer. The updated tests also cover distinct NameServers, same-named groups across clusters, multiple Brokers/Proxies, missing metadata, offline/failure/partial responses and interruption.
- Java 21, dependencies resolved into a fresh Maven repository: `mvn -B -ntp -Dspring.profiles.active=dev package` — **2,423 tests passed**, including architecture checks; **0 Checkstyle violations**, package successful. This uses the project's H2 development profile. The default MySQL profile fails with this machine’s configured credentials (`Access denied`); clean master reproduces the application-context errors.
- `npm test -- src/pages/cluster/__tests__/ClientsPage.test.tsx src/utils/clientConnectionDiagnostics.test.ts` — **26 passed**. `npm run build`, TypeScript and touched-file Prettier passed; full ESLint has **0 errors / 10 existing warnings**. `git diff --check` passed.
[...]

PR #4665, opened 2026-09-21
fix(ai): consume confirmation tokens atomically before mutation

pilichoumao:

No tracking issue. Closes a reliability gap between the documented and the implemented confirmation semantics: [docs/ai-agent-architecture.md §9](https://github.com/apache/rocketmq-dashboard/blob/5b761df8362eaf464c0b86ebea27e7e9576dd08f/docs/ai-agent-architecture.md) states that a dry-run returns `confirm_token` and "apply consumes it", but `ToolTokenService.verify()` never recorded usage, so one valid token could drive the same non-idempotent mutation (message send, offset reset, resource delete) repeatedly.

### Brief Description

Make confirmation tokens single-use with an atomic consume-before-mutate step.

- **Unique token id.** Each preview embeds a random 16-hex-char id in the signed payload; the token format moves `v1.<expiry>.<sig>` → `v2.<expiry>.<tokenId>.<sig>`. Two previews of the same operation within the same second now yield distinct tokens, so consuming one never invalidates a legitimately separate confirmation.
- **Atomic consumption.** `verify()` becomes `verifyAndConsume()`: expiry, signature and tool/caller/instance/input bindings are checked first (unchanged `INVALID_ARGUMENT` rejections, which do **not** consume), then the id is consumed with a single `ConcurrentHashMap.putIfAbsent`. `ToolMutationFilter` calls it before generating the execution plan and before `chain.proceed`, so concurrent or repeated applies of one token admit exactly one mutation.
- **Identifiable rejection.** A consumed token is rejected as `409 CONFLICT` / `CONFIRMATION_TOKEN_ALREADY_USED` with a re-preview hint; expired/tampered/rebound tokens keep their existing `400 INVALID_ARGUMENT` result, and expiry is checked before the consumed state so neither rejection leaks information about the other.
- **Deployment-model-consistent storage.** Consumed ids live in an in-memory store (`ConsumedTokenStore`) inside the singleton service — no Redis, no schema migration, matching the single-node confirmation flow. Entries are purgeable once their token expires (opportunistic purge past a size bound); a post-expiry replay is rejected by the existing expiry check, so purging never widens the acceptance window. Legacy `v1` tokens (≤10 min old at deploy time) are rejected as invalid; the recovery hint is to re-preview.

**Scope boundary, deliberately:** this defines one-time consumption only. If a mutation starts but the response is lost, resubmitting the same token is rejected and the caller must preview again — cross-request idempotency, restart recovery and distributed consumption are separate storage/product decisions, not claimed here.

`rmqctl` and the web console treat the token as an opaque string and always fetch a fresh preview token per execution, so no client change is needed.

- **Fail-before regression on `rocketmq-studio@5b761df8`:** a new `ToolTokenSingleUseTest` (real `ToolTokenService` + real `ToolMutationFilter` chain) failed 3/3 on the unfixed code — sequential replay executed twice, 8 concurrent applies executed 8 mutations, and two previews produced byte-identical tokens.
- **New/updated tests, all passing:**
  - `ToolTokenSingleUseTest` — replay rejected before re-execution; 8-thread concurrent apply executes the mutation exactly once (`[200, 409]`); distinct previews stay independent.
  - `ToolTokenServiceTest` — v2 format/size, fresh id per issue, tool/caller/instance/input binding, expiry boundary (599s vs 600s), tampered version/expiry/id/signature bytes, malformed and legacy-v1 tokens, consumed→`CONFLICT` with expiry precedence, 8-thread concurrent consume elects one winner, signatures containing dots/non-UTF8 bytes still parse.
  - `ConsumedTokenStoreTest` — first-consumer-wins, replays stay rejected within TTL, post-expiry entries become purgeable, 16-thread contention elects exactly one winner.
  - `ToolTokenConfigurationTest` — full Spring-wired chain: preview→apply executes once, replay is rejected `CONFLICT` without touching the handler; missing-secret behavior unchanged (`UNAVAILABLE`).
  - `ToolControllerTest` — HTTP mapping: replayed token surfaces as `409` with `code=CONFLICT`, message and re-preview hint.
  - `ToolMutationFilterTest` — read-only tools still never touch the token service or the store.
- **Build:** Java 21, `mvn -B -ntp -Dspring.profiles.active=dev package` — **3,045 tests passed**, 0 failures, **0 Checkstyle violations** (including test sources), packaged successfully, dependencies resolved into a clean local repository.
- **End-to-end against a real local Apache RocketMQ 5.5.0 NameServer/Broker cluster:** the unfixed jar built from `5b761df8` and the fixed jar ran side by side against the same cluster. Via the real `POST /api/ai/tools/rmq.topic.update/execute` API: baseline replay of one token returned `200 EXECUTED` **twice**; the fixed build returned `200 EXECUTED` then `409 CONFLICT`, concurrent applies of one fresh token produced exactly `[200, 409]`, a tampered copy and a token rebound to another topic returned `400 INVALID_ARGUMENT`, tokenless read-only `rmq.topic.list` kept working, and the topic was visible on the broker (the admitted apply really executed). All temporary JVMs were shut down afterwards.

Not exercised: multi-node deployments (the confirmation flow is single-node by design, no shared store introduced), MySQL profile (local credentials; clean baseline reproduces the same context errors), ACL/TLS.

### CI Status

The repository's Actions policy currently rejects action versions referenced by the existing workflow (`docker/setup-buildx-action@v3`, `docker/build-push-action@v6`), so GitHub-hosted runs end in `startup_failure` before any job starts (same as #4550). This PR does not change the workflow; all results above are local.

### Checklist

- [x] One coherent change; no unrelated modifications
- [x] Conventional commit subject; regression tests use `...Test`
- [x] Fail-before regression demonstrated on the unfixed baseline
- [x] New user-visible error text carries a recovery hint (single English string, matching existing `ToolError` style)
- [x] Architecture and formatting checks pass (0 Checkstyle violations)
- [x] No new dependency, port, service or configuration; storage lifecycle matches the single-node model

🤖 Generated with [Claude Code](https://claude.com/claude-code)


server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/service/ConsumedTokenStore.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final ConcurrentMap<String, Long> consumedExpiries = new ConcurrentHashMap<>();
+
+    /**
+     * Atomically marks the token id as consumed.
+     *
+     * @return {@code true} when this call is the first consumer, {@code false} when the id was
+     *         already consumed and the caller must be rejected
+     */
+    boolean consume(String tokenId, long expiresAtEpochSecond, long nowEpochSecond) {
```
RockteMQ-AI, line 45: Good use of `putIfAbsent` for atomic single-use enforcement. The opportunistic purge after the consume is a clean design - expired entries can never widen the acceptance window because the expiry check in `verifyAndConsume` runs first.

server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/service/ToolTokenService.java
```diff
@@ -91,24 +104,46 @@ private static byte[] resolveSecret(String configuredSecret) {
...
+     * the non-idempotent mutation. Rejections stay distinguishable: an expired, tampered or
+     * mismatched token is {@code CONFIRMATION_TOKEN_INVALID}; a token that already admitted an
+     * execution is {@code CONFIRMATION_TOKEN_ALREADY_USED}. Consumption happens only after all
+     * other checks pass, so a rejected replay never consumes a different pending token.
+     */
+    public void verifyAndConsume(ToolExecutionContext context) {
         requireConfiguredSecret();
         String toolName = context.definition().name();
         ConfirmationToken token = ConfirmationToken.parse(context.confirmToken(), toolName);
         if (clock.instant().getEpochSecond() >= token.expiresAt()) {
```
RockteMQ-AI, line 131: Correct check ordering: expiry then signature then consumed. This ensures that an expired token is rejected as `INVALID_ARGUMENT` without leaking whether it was also already consumed, and a tampered token is rejected before the consumed-state lookup. Nice defensive design.

server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/service/ToolTokenService.java
```diff
@@ -66,9 +73,15 @@ public ToolTokenService(
     }
 
     ToolTokenService(ObjectMapper objectMapper, Clock clock, byte[] secret) {
+        this(objectMapper, clock, secret, new SecureRandom());
```
RockteMQ-AI, line 76: Good testability: the `Random` parameter allows tests to inject a deterministic RNG while production uses `SecureRandom`. The 16-hex-char token id (8 bytes / 64 bits) provides sufficient collision resistance for single-use tokens within the 10-minute TTL window.

RockteMQ-AI (approved):
### Summary

This PR makes confirmation tokens single-use by introducing an atomic consume-before-mutate step. The design is sound, well-scoped, and thoroughly tested.

### What This PR Does

- **Token format upgrade (v1 -> v2):** Adds a random 16-hex-char token id to each preview, so two previews of the same operation yield distinct tokens
- **Atomic consumption:** `verify()` becomes `verifyAndConsume()` - checks expiry/signature/bindings first (without consuming), then atomically consumes with `ConcurrentHashMap.putIfAbsent`
- **Distinct rejection codes:** Consumed tokens return `409 CONFLICT` / `CONFIRMATION_TOKEN_ALREADY_USED`; expired/tampered tokens keep `400 INVALID_ARGUMENT`
- **In-memory storage:** `ConsumedTokenStore` matches the single-node deployment model, no Redis/schema migration needed
- **Comprehensive tests:** 24 tests covering replay rejection, concurrent applies (8-thread stress test), binding mismatches, v1 rejection, and purge safety

### Review Verdict

**Approve.** The implementation is correct, well-structured, and the scope boundaries are clearly documented. Key strengths:

1. **Correctness:** Atomic `putIfAbsent` ensures exactly-once consumption even under concurrent access
2. **Information leakage prevention:** Check ordering (expiry -> signature -> consumed) ensures neither rejection reason leaks info about the other
3. **Testability:** `Random` injection allows deterministic tests while production uses `SecureRandom`
4. **Memory safety:** Opportunistic purge at threshold 4096 keeps amortized cost O(1) per apply
5. **Backward compatibility:** Legacy v1 tokens are explicitly rejected with clear recovery guidance

### Minor Notes (Informational)

- The `ConsumedTokenStore` is in-memory only - this is explicitly scoped and appropriate for the single-node model. If distributed consumption is needed later, it would require a separate design (as noted in the PR description)
- The `PURGE_THRESHOLD = 4096` is reasonable for expected load. If token volume grows significantly, consider monitoring the map size

### Cross-repo Note

No cross-repo changes needed. `rmqctl` and the web console treat tokens as opaque strings.

---

pilichoumao:
Context that may help review: single-use consumption is already a documented invariant of this codebase. [docs/ai-agent-architecture.md §9 (安全不变量)](https://github.com/apache/rocketmq-dashboard/blob/5b761df8362eaf464c0b86ebea27e7e9576dd08f/docs/ai-agent-architecture.md) states:

> L2/L3 变更走工具协议自身的两步：dry-run 返回 `confirm_token`，apply 消费它。服务端由 `ToolMutationFilter` + `ToolTokenService` + `studio.ai.allow-l3-tools`（默认 `false`）强制。

Before this PR, `ToolTokenService.verify()` validated the token but never consumed it, so the documented "apply consumes it" invariant did not actually hold — one valid token could drive the same non-idempotent mutation repeatedly. This PR closes exactly that documentation-implementation gap and claims no semantics beyond the documented contract (cross-request idempotency, restart recovery and distributed consumption are explicitly out of scope).

PR #4666, opened 2026-09-21
fix(auth): match the user search term literally in the username filter

btlqql:

- Fixes #4223

### Brief Description

`AuthService.listUsers` bound the `search` request parameter straight into a MyBatis-Plus `LIKE`
predicate, so `%` and `_` in a username search were interpreted as wildcards:

- searching `_` matched every user, and `100%` matched any username starting with `100`;
- a username that really contains `%` or `_` could not be searched for at all.

The term is now escaped (`\` to `\\`, `%` to `\%`, `_` to `\_`) before it is bound, which is the
same pattern `QueryHistoryService.escapeLike` already uses for the message-query history search.
The escape character is MySQL's default for `LIKE`, and the Studio schema runs on MySQL.

New test: `AuthServiceDatabaseTest#listUsersEscapesLikeWildcardsInTheSearchTerm`, which captures the
`QueryWrapper` handed to the mapper and asserts the bound value.

Before the fix (red):

```
$ cd server && mvn -B -ntp test -Dtest='AuthServiceDatabaseTest#listUsersEscapesLikeWildcardsInTheSearchTerm'
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
Expecting Values:
  ["%100%_off%"]
to contain:
  ["%100\%\_off%"]
```

After the fix (green) — the whole class, so the neighbouring search/pagination tests are covered too:

```
$ cd server && mvn -B -ntp test -Dtest=AuthServiceDatabaseTest
[INFO] Tests run: 28, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### Checklist

- [x] Documentation touched where behaviour changed (no user-visible contract change: the search box already promised a literal username match)

redundant for `server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java` and can be closed.


RockteMQ-AI (approved):
### Summary
Fixes SQL LIKE wildcard injection in the username search filter — same pattern as the alert rule fix, applied to `AuthService.listUsers`. The `escapeLike` helper correctly escapes `\`, `%`, and `_` in the right order before binding to the `LIKE` predicate.

Minimal change (26 additions, 1 deletion) with a clear test that verifies the bound value. LGTM.

---

PR #4668, opened 2026-09-21
fix(audit): match the audit-log search term literally instead of as LIKE wildcards

btlqql:

- Fixes #4289

### Brief Description

`MybatisPlusAuditRepository.applyFilters` bound the raw search term into three `.like(...)` conditions (`operator`, `resource_name`, `detail`), so the search box value kept its LIKE wildcard meaning: `%` matched any run of characters and `_` matched any single character. Searching for `%` returned every audit row, `_` returned every row with a non-empty operator/resource name/detail, and a literal `_` or `%` (SQL text, `a_b`-style ids) could not be searched for at all.

Every query in the repository is built from the same `applyFilters` helper — the paged list (`findPage`), the summary aggregates (`summarize`: result counts, `COUNT(DISTINCT operator)`, latest timestamp, both hotspot `GROUP BY`s) and therefore the CSV export through `AuditService.exportLogs` — so the wildcard reading also widened the insights and could push the export over its 10 000-record cap, replacing the narrowed CSV with a 400.

The pattern is now built from an escaped term (`\`, `%` and `_` escaped), the same way the existing `QueryHistoryService.escapeLike` house pattern does it, and the helper stays local to the file that needs it. A term without wildcard characters produces exactly the same pattern as before, so the existing `summarizeAppliesEveryFilterToEachAggregateQueryTest` expectations are unchanged.

Test written first and run against the unmodified tree (`red`):

```
$ cd server && mvn -B -ntp test -Dtest=MybatisPlusAuditRepositoryTest
[ERROR] Tests run: 14, Failures: 2, Errors: 0, Skipped: 0
[ERROR]   MybatisPlusAuditRepositoryTest.findPageShouldEscapeLikeWildcardsInTheSearchTermTest:358
Expecting ArrayList:
  ["%a%b_c\d%", "%a%b_c\d%", "%a%b_c\d%"]
to contain only:
  ["%a\%b\_c\\d%"]
[ERROR]   MybatisPlusAuditRepositoryTest.summarizeShouldEscapeLikeWildcardsInTheSearchTermTest:381
Expecting ArrayList:
  ["%50%_off%", "%50%_off%", "%50%_off%"]
to contain only:
  ["%50\%\_off%"]
[INFO] BUILD FAILURE
```

After the fix (`green`), together with the audit service and controller tests:

```
$ cd server && mvn -B -ntp test -Dtest=MybatisPlusAuditRepositoryTest,AuditServiceTest,AuditControllerTest
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Test coverage:

- `findPageShouldEscapeLikeWildcardsInTheSearchTermTest` — searches for `a%b_c\d` and asserts all three bound parameters are the escaped `%a\%b\_c\\d%`, i.e. the wildcards are escaped and the backslash itself is escaped first.
- `summarizeShouldEscapeLikeWildcardsInTheSearchTermTest` — captures all four `selectMaps` aggregate queries and the `selectList` latest-timestamp query and asserts each one binds the escaped `%50\%\_off%` for the three search columns, which is the "every query built from the search value" criterion in the issue.
- Both tests capture the `QueryWrapper` with an `ArgumentCaptor` (calling `getSqlSegment()` before `getParamNameValuePairs()`) and assert the bound parameters, so they check the SQL that would actually be sent and need no database.

### Checklist

- [x] Documentation touched where behaviour changed — not applicable, the search contract is unchanged for terms without wildcard characters

redundant for `server/src/main/java/org/apache/rocketmq/studio/ops/audit/MybatisPlusAuditRepository.java` and can be closed.


lizhimins (changes requested):
The defect you found is real — and bigger than the one in the title — but this PR cannot land as it stands, because it rewrites the same hunk as #4458.

1. **Direct conflict with #4458.** Both PRs replace the three `.like(...)` calls at `MybatisPlusAuditRepository.java:199-201` (`operator` / `resource_name` / `detail`) with an escaped pattern. `git merge-tree` between the two branches conflicts on that hunk, and #4458 (still open, submitted earlier) escapes nine LIKE sites across the search paths, including these three.
2. **The stronger finding is not in the diff.** Your description notes that an unescaped `%` in the audit search term also blows past the export row cap — i.e. the wildcard affects how many rows the export path scans, not just which rows match. That is the more serious half of the bug, and this PR does not address it: it only changes the three predicates. If the export cap is computed from the same unescaped pattern anywhere else, that call site stays broken.
3. **No `ESCAPE` clause.** As with #4667, the escaped parameter is bound to a plain `LIKE ?` with no explicit `ESCAPE '\'`, so correctness depends on MySQL's default escape character rather than on the query.

Suggested path: let #4458 land first (it is the consolidation point for the LIKE-escaping family), then re-submit whatever it misses — in particular the export-cap interaction — on top of it, with an explicit `ESCAPE` clause. Marking this as requiring changes rather than merging a conflicting subset.

PR #4517, opened 2026-09-17
fix(metrics): preserve unknown Apache queue lag

zmuxuny:
### Which Issue(s) This PR Fixes

- Fixes #4516

### Brief Description

Apache `getGroupProgress` already preserves an unresolved queue lag as `ConsumerLagResolver.UNKNOWN` (`-1`). `ApacheRocketMqBusinessMetricsCollector` previously clamped those queue rows to zero while computing `consumer.lag.max_queue` and `topic.backlog.total`, producing AVAILABLE metrics from unavailable progress.

This change preserves unknown availability at the queue/topic aggregation layer. Any unknown queue makes the group max unavailable; topic backlog remains granular so known topics keep valid samples while only topics containing unknown queues are unavailable. Existing `consumer.lag.total` and delay semantics are unchanged.

### Red / Green Verification

Baseline: `master@987b748e8f4f421c5cd3c4c4e51a064cc7e59f18`.

Fail-before:
- Mixed known + unknown queue progress was collected with group total already unknown.
- The new regression expected `consumer.lag.max_queue` UNAVAILABLE, but the unmodified collector published AVAILABLE.
- `ApacheRocketMqBusinessMetricsCollectorTest`: **7 tests, exactly 1 failure**.

Green, Java 21:
- `ApacheRocketMqBusinessMetricsCollectorTest,RocketMQMetadataProviderTest,NativeAlertProcessorTest`: **71/71 passed**.
- Checkstyle: **0 violations**.
- `mvn -B -ntp -DskipTests package`: **BUILD SUCCESS**.
- `git diff --check`: clean.

Known zero/positive progress is unchanged; no provider/API/schema/dependency changes.

AI-assisted source audit, implementation and regression authoring; the provider's existing unknown sentinel path was verified before changing the collector.

### Audit follow-up

A full re-review found one scope-preservation edge in the original metric patch: per-group UNAVAILABLE samples did not carry the Consumer Group's `clusterId`. A native alert rule scoped with `clusterName` could therefore fail to match the unavailable sample during reconciliation.

The current head preserves `group.clusterId` on per-group and per-topic UNAVAILABLE samples while keeping whole-instance collection-failure markers unscoped. The regression now asserts that scope as part of the availability contract.

[...]

PR #4675, opened 2026-09-21
fix(metrics): report an unavailable consumer delay without a timestamp

yyqdbngt:
### Brief Description

`ApacheRocketMqBusinessMetricsCollector` skipped the `consumer.delay.seconds` sample entirely
when the broker reported consume stats without a consumed-message timestamp:

```java
if (group.isConsumptionTimestampAvailable()) {
    samples.add(delaySample(instance, group, collectedAt));
}
// no else: the delay metric simply disappears for that group
```

The same collector already treats that state as unavailable everywhere else - the
`!group.isConsumeStatsAvailable()` branch right above emits `CONSUMER_DELAY_SECONDS` as
`UNAVAILABLE` with `CONSUMER_STATS_UNAVAILABLE`, and an unknown lag sentinel is reported as
unavailable rather than as a fabricated zero.

The omission is the consequential representation: `CONSUMER_DELAY_SECONDS` is part of the
collector's declared metric scope, and `NativeAlertProcessor.reconcileMissingActiveStates`
walks the active states of that scope and resolves every fingerprint that no collected sample
covers. A group whose broker stops reporting the consumed-message timestamp therefore had its
active `consumer.delay.seconds` alert resolved - including a resolution notification - even
though its delay was merely unknown.

The collector now emits the delay as `UNAVAILABLE` with the reason
`CONSUMER_TIMESTAMP_UNAVAILABLE`, keeping the group's cluster scope like the other per-group
samples.

The new case runs against the unmodified trunk first (red), then against the fix (green).

```
# on the unmodified base (new test copied in, production code untouched)
$ cd server && mvn -B -ntp test -Dtest=ApacheRocketMqBusinessMetricsCollectorTest
[ERROR] Tests run: 7, Failures: 1, Errors: 0, Skipped: 0 -- in ...ApacheRocketMqBusinessMetricsCollectorTest
[ERROR]   ApacheRocketMqBusinessMetricsCollectorTest.reportsUnavailableDelayWhenTheConsumptionTimestampIsMissingTest:168
Expected size: 1 but was: 0 in:
```

```
# with the fix
$ cd server && mvn -B -ntp test -Dtest=ApacheRocketMqBusinessMetricsCollectorTest,CloudRocketMqBusinessMetricsCollectorTest,CollectorSchedulerTest
[INFO] --- checkstyle:3.6.0:check (validate) @ rocketmq-studio ---
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0 -- in ...ApacheRocketMqBusinessMetricsCollectorTest
[INFO] Tests run: 21, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

The open PR #4517 changes the neighbouring availability handling of the same collector
(`consumer.lag.max_queue`, `topic.backlog.total` for unknown queue progress) but explicitly
leaves the delay semantics unchanged; this change is the delay counterpart and touches only the
`isConsumptionTimestampAvailable()` branch plus its own helper.

### Checklist

`server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/collectors/ApacheRocketMqBusinessMetricsCollector.java`
and its test are also touched by another contributor's open PR #4517 (`fix(metrics): preserve unknown
Apache queue lag`). Distinct aspects of the same theme: #4517 keeps an unresolved queue lag from being
clamped to zero while `consumer.lag.max_queue` and `topic.backlog.total` are computed, whereas this
change stops an unavailable consumer delay from being reported with a timestamp, which made an
unavailable measurement look like an observed one. Neither supersedes the other.
### Related issue


RockteMQ-AI (commented):
### Summary
Emits an UNAVAILABLE delay sample when the consumption timestamp is missing, instead of omitting the sample entirely. This prevents the reconciliation logic from treating the group as gone.

### Findings
- **[Info]** The fix correctly distinguishes between "no consumption yet" (null timestamp) and "consumption available". Emitting UNAVAILABLE keeps the time series continuous.
- **[Info]** Tests verify both the unavailable case (null timestamp) and the available case (valid timestamp).

### Overall
Correct fix for a metrics continuity issue. LGTM.

---

PR #4677, opened 2026-09-21
[ISSUE #3167] Parallelize DLQ group stats and debounce search

Yanghanrui666:

- Fixes #3167

### Brief Description

Load the current DLQ page's topic statistics through a bounded 8-worker daemon pool while preserving deterministic page order. Each stats lookup is capped at 5 seconds; a timeout or failure only marks that row `UNAVAILABLE`, so the remaining rows are still returned.

The frontend now keeps the visible input separate from the request value, debounces typing by 300 ms, and still submits immediately on Enter or the search action.

This also supersedes the closed #3168 and fixes its review finding: fallback rows preserve `groupName` and `dlqTopic`, so failed stats calls no longer produce blank table rows. Unexpected task failures are logged for diagnosis.

Backend (targeted):

```
cd server
mvn -B -ntp -Dmaven.compiler.proc=full '-Dtest=RocketMQDLQProviderTest' test
```

Result: 42 tests passed; checkstyle reported 0 violations.

Frontend (targeted):

```
cd web
npm test -- src/pages/instance/__tests__/DLQPage.test.tsx --pool=threads --maxWorkers=1
```

Result: 23 tests passed.

`git diff --check` passed.

### Checklist

- [x] No new UI text was added
- [x] No new source files were added
- [x] No documentation change required


RockteMQ-AI (commented):
### Summary
Parallelizes DLQ group stats loading using an `ExecutorService` with 8 threads and a 5-second timeout. Adds debounce to the search input and shows per-group availability status in the UI.

### Findings
- **[Warning]** The `ExecutorService` is created per-request (line 119) and shut down in a finally block. For high-traffic scenarios, consider using a shared thread pool to avoid thread creation overhead. However, this is a significant improvement over sequential loading.
- **[Info]** The 5-second timeout with graceful degradation (marking unavailable groups) is good UX — users see partial results instead of a hanging page.
- **[Info]** The 300ms debounce on the search input prevents excessive API calls during typing. Good UX improvement.
- **[Info]** Tests cover parallel execution, timeout handling, and partial failure scenarios. The `CountDownLatch` approach correctly simulates slow groups.

### Overall
Significant performance improvement for clusters with many DLQ topics. The per-request thread pool is a minor concern but not blocking. LGTM.

---

lizhimins:
Thanks — this is a substantial PR and several parts of it are solid.

**What we verified as correct.** The two performance complaints are real: `RocketMQDLQProvider.listDLQGroups` collects the current page's stats through a serial stream, and `dlq.tsx` feeds `search` straight into the load effect's dependency list with `onChange` writing it on every keystroke. Page order is preserved and asserted (`containsExactly` over the group names), all futures are joined inside the `runtimeAdminClientResolver.execute` lambda so no client is released early, the `InterruptedException` branch restores the interrupt flag, the debounce does not slow the first paint, and the thread pool is named, daemonized and shut down in `@PreDestroy`. The three new Java tests and two new `it()` blocks are genuine and mutation-sensitive — reverting the parallel path makes the `CountDownLatch` case fail through the `Error` channel that `buildDLQGroup`'s `catch (Exception)` does not swallow, and the test counts in your description match the diff exactly. No tool-catalog change is needed either: `group.yaml` declares `status` and `statsAvailable` as plain strings/booleans with no enum, and `"UNAVAILABLE"` is already emitted on the base branch.

**Why we are closing.** Two things, both structural.

1. **It collides with #4577, which we are taking forward.** #4577 was opened about 26 hours before this PR and adds `setSearch('')` to the instance-scoped reset block in `dlq.tsx`. Once your `searchInput` / `search` split lands, that line clears only the request value: the input keeps displaying the old term, and your debounce effect sees `searchInput.trim() !== search` and writes the old term back into `search` 300 ms later. The result is that a newly selected instance gets filtered by the previous instance's group name — precisely the bug #4577 exists to fix, and its own assertion would fail. `git merge-tree` reports this as clean because the hunks do not overlap, so nothing in CI would have caught it. #4577 is first in the queue on DLQ search semantics and is the smaller, safer change, so that is the one landing.
2. **The shared executor's blast radius is not accounted for.** `statsExecutor` is a field of a `@Service @Primary` singleton, `Executors.newFixedThreadPool` gives it an unbounded `LinkedBlockingQueue`, and it would serve *both* the paged UI path and the unpaged `listDLQGroups(instanceId)` — which passes `Integer.MAX_VALUE` as the page size and whose only caller is the background collector `ApacheRocketMqDlqMetricsCollector.collect()`. One collection sweep over every `%DLQ%` topic therefore queues potentially thousands of tasks on the same 8 threads the UI depends on. Because `future.get(5, SECONDS)` is called in a serial loop, the 5-second budget includes queue wait, so a healthy page can come back with rows marked `UNAVAILABLE`; `ApacheRocketMqDlqMetricsCollector` maps that to `MetricAvailability.UNAVAILABLE`, and `AlertStateMachine` documents UNAVAILABLE as "the sole opt-in path that turns a failed probe into a firing alert" — i.e. pool contention can raise false DLQ alerts. None of the three new backend tests feeds more than a handful of topics, so saturation is untested.

One correction to the description, stated plainly because it is checkable: "fallback rows preserve `groupName` and `dlqTopic`, so failed stats calls no longer produce blank table rows" is already true on the base branch. `buildDLQGroup` catches `Exception`, sets `statsAvailable = false`, and returns a row carrying `groupName`, `dlqTopic` and `status("UNAVAILABLE")`; your `unavailableGroup()` is behaviourally identical to that existing branch. The genuinely new backend behaviour is that an `Error` or a timeout now degrades one row instead of failing the whole page.

**What would make a future attempt welcome.** Once #4577 has landed, a backend-only PR that parallelizes just the paged UI path — with a bounded queue and an explicit rejection policy, or a separate executor for the collector so the two cannot starve each other — plus a test that actually saturates the pool, would be a good change and we would review it on its merits. The debounce is worth sending separately as a frontend PR, rebased on whatever search-reset shape #4577 leaves behind.

PR #4482, opened 2026-09-17
feat(audit): add operator filter

tju-yxq:
### Which Issue(s) This PR Fixes

- Fixes #3387

### Brief Description

Re-opens the previously reviewed change after the base branch moved from `rocketmq-studio` to `master`. Rebased onto current `master` (`987b748e`), all conflicts resolved, and focused tests re-run locally.

cd server && mvn '-Dtest=AuditControllerTest,AuditServiceTest,MybatisPlusAuditRepositoryTest,AuditListToolHandlerTest' test → 36 tests passed; cd web && npm test -- AuditPage.test.tsx audit.test.ts opsService.test.ts --run → 30 tests passed; checkstyle 0 violations; lint/build passed

### How Did You Test This Change?

See the verification commands and results above; all were run on the rebased branch.

### Checklist

- [x] One coherent change; unrelated modifications are not bundled in
- [x] Commit subject follows Conventional Commits
- [x] Tests added or updated for non-trivial changes, test methods named `...Test`
- [x] New UI text has both Chinese and English entries under `web/src/i18n/`
- [x] Architecture constraints stay green (`mvn test` runs the ArchUnit checks)
- [x] New source files carry the ASF license header
- [ ] Documentation touched where behaviour changed (README / `docs/` / in-app help)

Issue #3307, opened 2026-09-05
Topic list search empties the table for search terms with leading or trailing spaces


unbridled-41:
## Problem

The topic list's search box trims the term only for the server query. `loadTopicPage` fetches with `search: searchText.trim()`, but the client-side row filter `visibleTopics` receives the raw input and matches with `topic.name.toLowerCase().includes(searchText.toLowerCase())`. A search term with leading or trailing whitespace (easy to get from a paste) therefore queries the server with the trimmed term, receives the matching topics, then filters every returned row out because no topic name contains the padded string — the table shows empty despite matches existing.

## Evidence / source

- Code inspection of the current `rocketmq-studio` branch (base commit 36126024): `web/src/pages/instance/topic.tsx` — `loadTopicPage` trims (`:433`), `filteredTopics` passes the raw `searchText` to `visibleTopics` (`:484`), the filter itself at `:178`; `onSearch` stores the raw value (`:1464-1467`).
- Deterministic Vitest reproduction in `TopicPage.test.tsx`: `keeps matching rows when the search term has leading or trailing spaces` fails on the unfixed source (the matching row disappears after submitting " orders ").

## Impact

The primary search interaction on the topic page returns an empty table for padded input, even though the server returned matching topics.

## Expected behavior / acceptance criteria

- The trimmed search term is used consistently by the server query, the client-side row filter, and the export, so rows matching the trimmed term remain visible.
- Regression coverage for a padded search term; it must fail on the unfixed code.

## Related work

- #911 (closed) fixed the same defect class for consumer-group search; the topic page retains the trim mismatch.

## PR

PR #3301 (includes the fix and the regression test).

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported trim mismatch between `loadTopicPage` (server query) and `visibleTopics` (client-side filter) is verified. When the search input has leading/trailing whitespace, the server returns matching topics but the client filter rejects them all, producing an empty table.

**Root Cause:** `searchText.trim()` used for server query but raw `searchText` passed to `visibleTopics` filter.
**Impact:** Primary search interaction broken for pasted input with whitespace.
**Severity:** Medium

PR #3301 includes the fix and regression test — aligned with the expected acceptance criteria.

---
*Automated evaluation by github-manager*

PR #4682, opened 2026-09-21
fix(audit): trim the audit-log search term before it reaches the repository

btlqql:

- No open issue; small self-contained defect found while auditing the audit-log search path (same defect class as #911 / #3307 for the topic and consumer-group searches).

### Brief Description

`AuditService` forwarded the raw `search` term to the repository. Both paged paths (`queryLogs` through the private `findPage`, and therefore `exportLogs`) and `summarize` passed it straight into `AuditRepository`, which binds it into the `operator` / `resource_name` / `detail` `LIKE` filters. A padded term - easy to produce by pasting into the audit-log search box - therefore reached SQL with its whitespace included. `" ops "` can never match `ops`, so the record list, the summary cards (matching records, unique operators, both hotspots) and the CSV export that shares the same helper all reported nothing for a term that has matches.

The term is now normalized once, on the way into the repository, so every path filters on the same value: trimmed when it has content, and treated as no filter at all when it is only whitespace. That is the pattern `AclService.pageUsers` already uses for its keyword (`StringUtils.hasText(keyword) ? keyword.trim() : null`), applied here where `search` enters the service.

Same-file note: `ops/audit/AuditService.java` and `AuditServiceTest.java` are also touched by open PR #4482 (`feat(audit): add operator filter`), which adds an `operator` parameter to the same methods. The two changes are semantically independent, but whichever of them merges second needs a trivial rebase. The repository is not touched here: the LIKE-escaping of this same term lives in the separate #4668, which changes the bound pattern rather than the term.

Four `AuditServiceTest` cases written first and run against the unmodified tree (`red`, base commit `7ce9a682`):

```
$ cd server && mvn -B -ntp test -Dtest=AuditServiceTest
[ERROR] Tests run: 19, Failures: 4, Errors: 0, Skipped: 0
[ERROR]   AuditServiceTest.exportLogsShouldTrimSearchTermBeforeDelegating:303
expected: "50% off"
 but was: "  50% off  "
[ERROR]   AuditServiceTest.queryLogsShouldTreatWhitespaceOnlySearchAsAbsent:290
expected: null
 but was: "   "
[ERROR]   AuditServiceTest.queryLogsShouldTrimSearchTermBeforeDelegating:279
expected: "ops"
 but was: "  ops  "
[ERROR]   AuditServiceTest.summarizeShouldTrimSearchTermBeforeDelegating:313
expected: "ops"
 but was: "  ops  "
[INFO] BUILD FAILURE
```

After the fix (`green`), together with the audit controller and repository tests of the same package:

```
$ cd server && mvn -B -ntp test -Dtest=AuditServiceTest,AuditControllerTest,MybatisPlusAuditRepositoryTest
[INFO] Tests run: 43, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Test coverage:

- `queryLogsShouldTrimSearchTermBeforeDelegating` - `queryLogs(..., "  ops  ", ...)` must reach `AuditRepository.findPage` as `"ops"`.
- `queryLogsShouldTreatWhitespaceOnlySearchAsAbsent` - a term made only of spaces must reach the repository as `null`, i.e. as no filter, so the list is not emptied by a stray space.
- `summarizeShouldTrimSearchTermBeforeDelegating` - the summary aggregates must see the same trimmed term as the list, so the cards cannot disagree with the table.
- `exportLogsShouldTrimSearchTermBeforeDelegating` - the CSV export goes through the same private helper, so it must export the trimmed result set as well.
- The tests capture the arguments passed to the mocked `AuditRepository`, so they assert the value that would be bound into SQL without needing a database.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` - not applicable, no UI text added
- [x] New source files carry the ASF license header - no new source files
- [x] Documentation touched where behaviour changed - not applicable, a term without surrounding whitespace behaves exactly as before

`server/src/main/java/org/apache/rocketmq/studio/ops/audit/AuditService.java` and its test are also
touched by another contributor's open PR #4482 (`feat(audit): add operator filter`), which widens the
audit query filter; this change trims the search term before it reaches the repository. Different
defects in the same class, independent of each other.


RockteMQ-AI (commented):
### Summary
Normalizes the audit-log search term by trimming whitespace before it reaches the repository. Whitespace-only terms are treated as no filter (null). Applied consistently across paged list, summary, and CSV export.

### Findings
- **[Info]** Clean normalization — `normalizeSearch` is a single point of truth, called once before the repository layer. Good design.
- **[Info]** Tests cover trim, whitespace-only, and null cases. The ArgumentCaptor approach correctly verifies the normalized value reaches the repository.

### Overall
Simple, correct fix. LGTM.

---

Issue #4554, opened 2026-09-18
[Bug] Offset msgId lookup returns empty when brokers register with hostnames (revised design for #4182)


qiyu-lu:
### Before Creating the Bug Report

- [x] I have searched the open issues of this repository and believe that this is not a duplicate. (It continues #4181, which the stale bot closed, and the review of #4182.)
- [x] This is a defect in RocketMQ Studio.
- [x] I have stated the exact version below.

### Studio Version

branch: `rocketmq-studio`
git commit id: reproduced on `0a596661`; `BrokerTopologyGuards` is unchanged on the current head `d50ffecc` (still an exact `Set.contains` on the registered address string)
deployed as: built from source (dev profile, H2)

### Runtime Environment

Ubuntu (Linux 5.15), JDK Temurin 21, Chrome.

### Connected RocketMQ Cluster

RocketMQ 5.5.1, direct to NameServer/Broker. The broker registers with a hostname (`brokerIP1=<hostname>`), the same style as the bundled `deploy/rocketmq/conf/broker-0.conf` (`brokerIP1=broker-0`).

### Describe the Bug

When a broker registers with a hostname, querying a message by its physical offset msgId returns nothing. The offset msgId embeds the broker's IP, while `BrokerTopologyGuards.knownBrokerEndpoints` holds the registered `hostname:port` strings, so the guard treats the address as outside the topology. Unique-key msgId lookup is not affected.

### Steps to Reproduce

1. Register a broker as `repro-hostname-broker:10911` (resolves to `172.26.0.4`).
2. Send a message; querying by topic finds it.
3. On the message page choose "by Message ID" and enter the `offsetMsgId` from the send result.

### What Did You Expect to See?

The message is returned (the official SDK reads the same id successfully).

### What Did You See Instead?

0 results; the guard logs "Rejecting decoded broker address ... not a known broker endpoint". Screenshots are in #4181.

### Additional Context — design proposal for a revised fix

Thanks @lizhimins for the detailed review on #4182 — all five points are fair, and items 1–3 showed the first version put the fix in the wrong place. CONTRIBUTING now asks to agree on non-trivial designs in an issue first, so here is the proposal before I write code again.

[...]

PR #4683, opened 2026-09-21
fix(message): resolve hostname-registered brokers in the offset msgId guard

btlqql:

- Fixes #4554

### Brief Description

`BrokerTopologyGuards.validatedBrokerAddr` decided whether the broker embedded in a physical-offset
msgId belongs to the selected instance with an exact `Set.contains` against the addresses from
`examineBrokerClusterInfo()`. Those addresses are the strings the brokers registered with, while
`MessageDecoder.decodeMessageId` always yields a numeric address (`InetAddress.getByAddress`, no
reverse lookup). On a cluster whose brokers register with a hostname (`brokerIP1=broker-0`, the style
of the bundled `deploy/rocketmq/conf/broker-0.conf`), the two never match, so the guard rejected every
offset id and the lookup returned an empty result.

The guard now falls back to comparing the resolved addresses (same port required) before rejecting, so
a numeric address that resolves to a registered hostname endpoint is accepted. Unresolvable or
unrelated endpoints are still rejected, and the exact-match fast path is unchanged: it runs first and
is free, which is all a cluster with numeric `brokerIP1` values ever pays.

**Blocking item 2 — the DNS cost, as a seam rather than a note.** Option (b) from the review, with the
"at most once per call" half of option (a) as well: resolution now goes through
`BrokerHostResolver`, a one-method interface whose `DEFAULT` is the platform lookup
(`InetAddress::getByName`), plus `BrokerHostResolver.caching(delegate)`, a memoizing wrapper that
remembers failures as well as successes so an unresolvable registered host is not retried either.
`RocketMQMessageProvider.perQueryHostResolver()` and the equivalent in `RocketMQDLQProvider` build one
caching resolver per query/resend, which is exactly the case the review named — a single `queryByMsgId`
evaluates the guard twice (`RocketMQMessageProvider.java:150` before `viewMessage`, then in the
decoded-offset fallback) — so the second evaluation compares instead of resolving again. Tests inject a
fake resolver and count lookups, so the bound is asserted rather than described.

The javadoc states both things the review asked for: that the accepted set widens from "the literal
registered string" to "whatever the registered host currently resolves to", and the expected DNS cost
and its bound. It also states plainly that the platform lookup has no wall-clock timeout of its own —
`InetAddress` exposes no knob, so the JVM and the OS resolver decide — which is the other reason the
seam exists: a deployment needing a hard bound or a cross-request cache can supply its own resolver
without touching the guard. Keeping the failure direction closed is unchanged: a registered host that
does not resolve matches nothing, so an id pointing at it stays rejected.

### Nice-to-have from the review

`rejectsAnOffsetIdWhenTheRegisteredHostnameCannotBeResolved` no longer performs a real DNS lookup: the
whole class now uses `FakeBrokerHostResolver`, so no test in it touches the network.

```
cd server && mvn -B -ntp test
[INFO] Tests run: 3160, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

3145 of those are the untouched `rocketmq-studio` head; the new cases are the difference. No test in the
touched classes performs a network lookup.

Mutation check — the resolved-address fallback removed, i.e. the pre-fix exact-string behaviour, with
the tests untouched:

```
[ERROR] Tests run: 110, Failures: 6, Errors: 0, Skipped: 0
[ERROR]   BrokerTopologyGuardsTest.acceptsAnOffsetIdWhoseBrokerIsRegisteredWithAHostnameTest:51
[ERROR]   BrokerTopologyGuardsTest.acceptsAnIdWhoseAddressMatchesOneOfSeveralRegisteredEndpointsTest:64
[ERROR]   BrokerTopologyGuardsTest.rejectsAnOffsetIdWhenTheRegisteredHostnameCannotBeResolvedTest:116
[ERROR]   BrokerTopologyGuardsTest.reusesOneResolutionAcrossRepeatedEvaluationsOfTheSameGuardTest:76
[ERROR]   RocketMQMessageProviderTest.queryByMsgIdAcceptsAHostnameRegisteredBrokerAndLooksItUpOnceTest:524
[ERROR]   RocketMQDLQProviderTest.resendSelectedMessagesLooksUpAHostnameRegisteredBrokerOnceForAllSelectedIdsTest:454
```

Six cases across the guard and both consumers, including the per-call lookup bound, so the fix and its
DNS discipline are both pinned by assertions that fail without them.

Controls that stay green in the same class: an address outside the topology, a registered hostname on a
non-matching port, and an unresolvable registered hostname are all still rejected.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text changed)
- [x] New source files carry the ASF license header (`BrokerHostResolver`, `BrokerHostResolverTest`, `FakeBrokerHostResolver`)
- [x] Documentation touched where behaviour changed (class javadoc states the widened accepted set and the DNS cost; both entry points document the resolver argument)


RockteMQ-AI (commented):
### Summary
Fixes the string comparison between hostname-registered broker addresses (from NameServer) and IP-decoded offset message IDs. When `brokerIP1` is configured as a hostname, the NameServer publishes the hostname while offset msgId always decodes to an IP. The old code compared them as strings, which always failed.

### Findings
- **[Critical]** This was a real operational bug — any cluster using hostname-based broker registration would fail all offset msgId lookups. The fix correctly resolves both sides to `InetAddress` for comparison.
- **[Warning]** `InetAddress.getByName()` triggers DNS resolution. If DNS is slow or the hostname is unreachable, this adds latency to every message query. Consider caching resolved addresses with a short TTL, or at least documenting the DNS dependency.
- **[Info]** Good test coverage including IPv6 and mixed hostname/IP scenarios.

### Overall
Correct fix for a significant operational issue. The DNS resolution concern is worth noting but not blocking. LGTM.

---

lizhimins (changes requested):

- The defect is real and we can trace it: `decodedBrokerAddr` (`BrokerTopologyGuards.java:80-88`) always yields a numeric `hostAddress:port`, while `knownBrokerEndpoints` (`:105-118`) returns whatever the broker registered — so a cluster with `brokerIP1` set to a hostname rejects every offset-msgId lookup, and `RocketMQMessageProvider.java:150` / `:200` / `:674` plus `RocketMQDLQProvider.java:283` all turn that into a silent empty result. Keeping the failure direction closed (unresolvable registered hostname → reject) is the right call and matches the class javadoc.
- Blocking item 2 (needs a design answer before we merge): `matchesKnownEndpoint` performs one `InetAddress.getByName` per registered endpoint on every guard evaluation, with no cache and no timeout, and a single `queryByMsgId` can evaluate the guard twice (`RocketMQMessageProvider.java:150` then `:200`). Please either (a) resolve each endpoint at most once per call and state the expected DNS cost in the javadoc, or (b) take a small resolver seam so the lookup can be cached/bounded. Also worth a note in the javadoc: the accepted set widens from "the literal registered string" to "whatever the registered hostname currently resolves to".
- Nice to have: `rejectsAnOffsetIdWhenTheRegisteredHostnameCannotBeResolved` performs a real DNS lookup for `broker-0.invalid`. It is RFC 2606 reserved so it fails fast, and `UrlHostGuardTest` / `SettingsServiceTest` already resolve names, so this is acceptable — just flagging it as the one test in the class that touches the network.

btlqql:
Reworked onto `4c697f07`. Blocking item 2 answered with option (b), and the "at most once per call" half of option (a) as well.

**The seam.** Resolution now goes through `BrokerHostResolver`, a single-method interface (`DEFAULT = InetAddress::getByName`), plus `BrokerHostResolver.caching(delegate)` — a memoizing wrapper that remembers failures as well as successes, so an unresolvable registered host is not retried either. `RocketMQMessageProvider.perQueryHostResolver()` and the equivalent in `RocketMQDLQProvider` build one caching resolver per query/resend, which is the exact case you named: one `queryByMsgId` evaluates the guard twice (`:150` before `viewMessage`, then the decoded-offset fallback), and now the second evaluation compares instead of resolving again. Both entry points take the resolver as a parameter, so tests inject a fake and **count** lookups — the bound is asserted, not described.

**The javadoc** states both requested things: the accepted set widens from "the literal registered string" to "whatever the registered host currently resolves to", and the expected DNS cost with its bound. It also says plain what you flagged — the platform lookup has no wall-clock timeout of its own, since `InetAddress` exposes no knob and the JVM/OS resolver decides — which is the second reason the seam exists: a deployment wanting a hard bound or a cross-request cache supplies its own resolver without touching the guard. Failure direction is unchanged: an unresolvable registered host matches nothing, so the id stays rejected.

**Nice-to-have:** `rejectsAnOffsetIdWhenTheRegisteredHostnameCannotBeResolved` no longer does real DNS — the class now uses `FakeBrokerHostResolver` throughout, so no test in it touches the network.

**Mutation check** — resolved-address fallback removed (pre-fix exact-string behaviour), tests untouched:

```
[ERROR] Tests run: 110, Failures: 6, Errors: 0, Skipped: 0
[ERROR]   BrokerTopologyGuardsTest.acceptsAnOffsetIdWhoseBrokerIsRegisteredWithAHostnameTest:51
[ERROR]   BrokerTopologyGuardsTest.acceptsAnIdWhoseAddressMatchesOneOfSeveralRegisteredEndpointsTest:64
[ERROR]   BrokerTopologyGuardsTest.rejectsAnOffsetIdWhenTheRegisteredHostnameCannotBeResolvedTest:116
[ERROR]   BrokerTopologyGuardsTest.reusesOneResolutionAcrossRepeatedEvaluationsOfTheSameGuardTest:76
[ERROR]   RocketMQMessageProviderTest.queryByMsgIdAcceptsAHostnameRegisteredBrokerAndLooksItUpOnceTest:524
[ERROR]   RocketMQDLQProviderTest.resendSelectedMessagesLooksUpAHostnameRegisteredBrokerOnceForAllSelectedIdsTest:454
```

Whole suite on a host with MySQL 8 reachable: 3160 tests, 0 failures, 0 errors (3145 on the untouched head).

PR #4684, opened 2026-09-21
fix(auth): bound the login username like the user API does

btlqql:
### Related issue

None - no open issue covers the accepted login username length. #4159 (open) is about which HTTP code
records a failed attempt, not about the value that reaches the lookup and the rate limiter.

### Brief Description

`AuthService.validateLogin` (server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java:554)
checked only that the submitted username was non-blank. An unauthenticated caller could therefore send
a username of any length; the value was handed unchanged to `findUserByUsername` and, when the login
failed, stored by `LoginRateLimiter.recordFailure` as an in-memory map key.

`LoginRateLimiter` bounds how many usernames it tracks (`MAX_TRACKED_USERNAMES = 10_000`), never how
long they are, so the oversized string stayed reachable in that map for the whole failure window
(5 minutes, plus the lazy sweep) - the one place in the login path whose memory was not bounded. No
other username entry point accepts such a value: `CreateStudioUserDTO` (`@Size(max = 128)`),
`AuthService.validateUsername` and the column (`username VARCHAR(128)` in
`server/src/main/resources/db/schema.sql:20`) all cap it at 128 characters.

`validateLogin` now enforces the same bound on the trimmed username and rejects longer input with
`400` before the mapper and before the rate limiter see it. The check shares one constant with
`validateUsername` so the two entry points cannot drift apart. Anything a licence-holding client can
send today still behaves exactly as before: a username longer than 128 characters cannot exist in the
database, so no account answer changes.

Two new tests in `AuthServiceDatabaseTest`. The first pins the response (400, and no mapper call), the
second pins the effect on the rate limiter by injecting a `LoginRateLimiter` and reading its
package-private `trackedUsernameCount()` after a failed login with a 4096-character username.

Before the fix (red), on the unmodified tree:

```
$ cd server && mvn -B -ntp test -Dtest='AuthServiceDatabaseTest#loginShouldRejectAnOverlongUsernameBeforeTouchingTheDatabase+anOverlongUsernameIsNeverRetainedAsARateLimiterKey'
[ERROR] Tests run: 2, Failures: 2, Errors: 0, Skipped: 0
[ERROR] org.apache.rocketmq.studio.auth.AuthServiceDatabaseTest.loginShouldRejectAnOverlongUsernameBeforeTouchingTheDatabase -- Time elapsed: 0.032 s <<< FAILURE!
Multiple Failures (1 failure)
-- failure 1 --
expected: 400
 but was: 401
	at AuthServiceDatabaseTest.loginShouldRejectAnOverlongUsernameBeforeTouchingTheDatabase(AuthServiceDatabaseTest.java:570)
[ERROR] org.apache.rocketmq.studio.auth.AuthServiceDatabaseTest.anOverlongUsernameIsNeverRetainedAsARateLimiterKey -- Time elapsed: 2.531 s <<< FAILURE!
expected: 0
 but was: 1
	at AuthServiceDatabaseTest.anOverlongUsernameIsNeverRetainedAsARateLimiterKey(AuthServiceDatabaseTest.java:599)
```

The `401` is the base behaviour: the 129-character name went to the mapper lookup and came back as
"Invalid username or password", and the `4096`-character name was still tracked by the rate limiter
afterwards.

After the fix (green) - the whole class, plus the neighbouring login/rate-limiter classes:

```
$ cd server && mvn -B -ntp test -Dtest='AuthServiceDatabaseTest,AuthServiceTest,LoginRateLimiterTest,AuthControllerTest'
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthControllerTest
[INFO] Tests run: 29, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthServiceDatabaseTest
[INFO] Tests run: 16, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthServiceTest
[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.LoginRateLimiterTest
[INFO] Tests run: 66, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

No login behaviour changed for a client whose username is a real one: the added check only fires for a
name that is longer than the `VARCHAR(128)` column can hold, so it can never match a stored account.

(`CliAgentProviderTest`, `ClaudeCodeAgentProviderTest`). None of them are touched by this change; the
four classes above are the ones this change can affect.

### Checklist

- [x] Documentation touched where behaviour changed (no user-visible contract change: a username longer than 128 characters was never a valid Studio username)

`server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java` is also touched by open PRs
 from the same fork. The changes sit in different methods - the user search-term escape, the
login username bound, the password-change transaction, and the status code for a mismatched current
password - and none supersedes another or depends on another, so they can be reviewed and merged
independently. #4458 (a different fork, base `master`) also rewrites `AuthService.pageUsers` while sweeping
every `.like(...)` call site; #4666 records that wider overlap.


RockteMQ-AI (commented):
### Summary
Bounds the login username at 128 characters — matching the existing `validateUsername` and the `VARCHAR(128)` column — before it reaches the database lookup or the rate limiter. Prevents an unauthenticated caller from injecting oversized keys into the rate limiter.

### Findings
- **[Critical]** Good security fix — without this, an attacker could park arbitrarily large keys in the rate limiter (and the DB lookup) by submitting overlong usernames. The early rejection before `userMapper.selectOne` is the correct placement.
- **[Info]** `validateLogin` now uses `request.getUsername().trim().length()` while `validateUsername` also trims. Consistent — good.
- **[Info]** Test `loginShouldRejectAnOverlongUsernameBeforeTouchingTheDatabase` correctly verifies `verifyNoInteractions(userMapper, sessionMapper)` — confirming the rejection happens before any DB call.

### Overall
Clean security hardening with proper test coverage. LGTM.

---

RockteMQ-AI (commented):
### Summary
Bounds the login username at 128 characters before it reaches the database lookup or the rate limiter. Prevents an unauthenticated caller from injecting oversized keys into the rate limiter.

### Findings
- **[Critical]** Good security fix — without this, an attacker could park arbitrarily large keys in the rate limiter (and the DB lookup) by submitting overlong usernames. The early rejection before `userMapper.selectOne` is the correct placement.
- **[Info]** `validateLogin` now uses `request.getUsername().trim().length()` while `validateUsername` also trims. Consistent — good.
- **[Info]** Test `loginShouldRejectAnOverlongUsernameBeforeTouchingTheDatabase` correctly verifies `verifyNoInteractions(userMapper, sessionMapper)` — confirming the rejection happens before any DB call.

### Overall
Clean security hardening with proper test coverage. LGTM.

---

PR #4685, opened 2026-09-21
fix(ai): stop resuming a provider session the CLI has lost

yyqdbngt:
### Brief Description

A conversation remembers the agent CLI's session id (`rmq_ai_conversation.runtime_session_id`) and
`AiRunService.resolveResume` hands it back to the CLI as `claude --resume <id>` on **every** later
turn. `ResumeRecovery` documents why that id can stop existing — the per-conversation workspace
(`studio.ai.conversation.workspace-dir`, default `/tmp/rocketmq-studio-ai/conv-<id>`) holds `HOME`
and `$HOME/.claude/projects/<cwd-hash>/<id>.jsonl`, so one container restart is enough — and it also
documents the required recovery:

> When `shouldRetryWithoutResume` returns true the caller must, exactly once: … **clear
> `conversation.runtime_session_id`, so the next turn does not repeat the failure** and the new
> session id from the retry's `result` frame becomes the one that is remembered.

That class had **no caller anywhere in `src/main`** (only its own unit test in
`RmqctlWorkspaceTest`), so nothing ever cleared the id. A resumed run whose session is gone fails
with `error_during_execution`; `AiRunExecutor.finalizeRun` then wrote
`run.setRuntimeSessionId(terminal.outcome().runtimeSessionId)` — the id *echoed back by the failed
frame*, which the parser explicitly warns is "the requested id … not a live one"
(`ClaudeCodeStreamParser#providerErrors`) — and `rememberRuntimeSession` left that same id on the
conversation. The web client never sends `resume: false` (`web/src/api/aiConversations.ts` declares
the field, nothing sets it) and there is no "start a new session" affordance, so every subsequent
message resumed the same dead id and failed identically: **the conversation was permanently
unusable**.

`AiRunExecutor` now applies the documented recovery. A failed run that was resuming a session
forgets that session instead of recording the echoed id, so the next turn starts a fresh provider
session:

- `AiRunExecutor#lostResumeSession` / `#forgetLostResumeSession` — the run row keeps `resumed_from`
  as history but no longer records the dead id as its session, and the conversation's stored id is
  dropped.
- `ResumeRecovery#isLostResumeSubtype` — the subtype arm of `isLostResumeSignal` is now reusable:
  the run worker sees projected frames only, and the exit code and the stderr line that carry the
  other arm die with the subprocess (a run that reached this path already failed, which is the
  exit-code half of that signal).
- `AiConversationRepository#clearRuntimeSessionId` — `updateById` omits null entity fields, so a
  cleared column needs an explicit assignment; the same idiom `MybatisPlusInstanceRepository` and
  `MybatisPlusAclRepository#clearColumn` already use. `gmt_modified` is deliberately untouched: this
  is a state repair, not user activity, and bumping it would reorder the conversation list on the
  strength of a failure.

Not done, deliberately: the immediate one-shot retry that `ResumeRecovery` also specifies. Turning
it on needs the CLI's exit code and stderr, which only `ClaudeCodeAgentProvider` ever sees, so it is
a separate change to the provider/executor seam. Clearing the id alone already removes the permanent
breakage — the next message works instead of failing forever.

New test: `AiRunExecutorTest#aLostResumeSessionShouldNotBeResumedByTheNextTurnTest` — scripts the
measured frame of a lost resume (a `result` frame that echoes the requested id with
`subtype=error_during_execution`) and asserts what is written.

Before the fix (red) — the run row kept the echoed id, so the next turn would resume it again:

```
$ cd server && mvn -B -ntp test -Dtest='AiRunExecutorTest#aLostResumeSessionShouldNotBeResumedByTheNextTurnTest'
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.589 s <<< FAILURE! -- in org.apache.rocketmq.studio.ops.ai.conversation.AiRunExecutorTest
[ERROR] org.apache.rocketmq.studio.ops.ai.conversation.AiRunExecutorTest.aLostResumeSessionShouldNotBeResumedByTheNextTurnTest -- Time elapsed: 2.283 s <<< FAILURE!

expected: null
 but was: "gone-session"
	at org.apache.rocketmq.studio.ops.ai.conversation.AiRunExecutorTest.aLostResumeSessionShouldNotBeResumedByTheNextTurnTest(AiRunExecutorTest.java:480)
```

The red run is that same test method without its last line, `verify(conversationRepository)
.clearRuntimeSessionId(CONVERSATION_ID)`: the port method it verifies is introduced by this change,
so on the unmodified tree it would not compile. The assertion quoted above is the first one that
fails and it is unchanged by the fix's shape — it is the run row, i.e. the value the next turn
resumes.

After the fix (green) — the new test plus the whole neighbouring AI conversation surface:

```
$ cd server && mvn -B -ntp test -Dtest='AiRunExecutorTest,AiConversationPersistenceIntegrationTest,AiRunServiceTest,AiEventSeqAllocationTest,AiTimelineRepositoryTest,AiConversationServiceTest,AiStreamControllerTest,AiConversationControllerTest,AiConversationVoAssemblerTest,RmqctlWorkspaceTest'
[INFO] Tests run: 16, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.ops.ai.conversation.AiRunExecutorTest
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.ops.ai.conversation.AiConversationPersistenceIntegrationTest
[INFO] Tests run: 33, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.ops.ai.conversation.AiTimelineRepositoryTest
[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.ops.ai.conversation.AiConversationControllerTest
[INFO] Tests run: 198, Failures: 0, Errors: 0, Skipped: 3
[INFO] BUILD SUCCESS
```

The durable half is proved against a real database rather than a mock:
`AiConversationPersistenceIntegrationTest#clearingTheRememberedProviderSessionShouldReachTheColumnTest`
(`@SpringBootTest @ActiveProfiles("dev")`, i.e. H2 in MySQL compatibility mode, so it runs in CI)
sets `runtime_session_id`, clears it and reads the row back as `null`, and asserts an unknown id is a
zero-row update rather than an exception. The whole module test-compiles with
`mvn -B -ntp test-compile` (BUILD SUCCESS, `0 Checkstyle violations`).

### Checklist

- [x] Commit subject follows Conventional Commits (`fix(ai): …`)
- [x] Architecture constraints stay green (`mvn test` runs the ArchUnit checks; the new dependency is `conversation` -> `conversation.agent`, which already exists)
- [x] Documentation touched where behaviour changed (no user-visible contract change: the conversation keeps behaving as documented, it just stops being unusable after a workspace is lost)
### Related issue

No open issue in this repository tracks this defect; it was found by reading the code. The open
issues were searched on 2026-09-21 for `ai session resume` and nothing matching came back, so there is no
number to reference here.

**Another contributor's open PR #4693 (`fix(ai): retry a turn whose remembered agent session is gone`,
opened about twenty minutes after this one) addresses the same defect from the same root cause**: a failed
`result` frame echoes the requested session id back, and the executor persisted that dead id, so every
later turn resumed a session the CLI no longer has. The two differ in the remedy, not in the diagnosis:
#4693 retries the turn, this change stops the dead id from being stored at all and clears the remembered
session so the next turn starts fresh. They touch the same conversation files
(`AiConversationRepository`, `MybatisPlusAiConversationRepository`, `AiRunExecutor`, `ResumeRecovery`) and
each adds its own test method to `AiRunExecutorTest`. They are alternatives rather than complements, so
**whoever reviews this should pick one and close the other**; a maintainer may prefer #4693's retry. This
pull request is kept open only so the choice is visible - it is not a second, independent defect.


RockteMQ-AI (approved):
LGTM. Looks good.

---

yyqdbngt:
Closing this as superseded by #4693 (`02d81ec2`), which is now in `rocketmq-studio` and fixes the same defect, more completely.

What the trunk carries after that merge, in the files this pull request touches:

- `ClaudeCodeAgentProvider.throwIfTheResumeSessionWasLost` with `ResumeRecovery.RESUME_LOST_CODE`: the provider reports a `--resume` session that no longer exists as its own error code.
- `ClaudeCodeStreamParser.resultSubtype()`: the signal is read from the terminal frame, not only from stderr.
- `AiConversationRepository.clearRuntimeSessionId` and its MyBatis-Plus implementation: the stale id is nulled by an explicit assignment, because `updateById` skips null entity fields and a fake pass through the ordinary update would be a silent no-op.
- `AiRunExecutor.streamWithLostResumeRetry`: the dead id is forgotten, then the turn is retried exactly once without `--resume`, with a `ProviderNotice` to the timeline and the prompt prepared once, outside the retry.

So the mechanism this pull request proposed - detect, forget, retry - is all in the trunk, and the retry half is more complete than this branch: this branch never grew the retry path, so today it would only remove working code if it were merged.

The remaining difference is shape, not behaviour: here `clearRuntimeSessionId` returns the number of updated rows instead of `void`, and `isLostResumeSubtype` is a named public method instead of an inline comparison. Neither is worth a pull request of its own against code that has just landed; if the row count or the named predicate is ever wanted, it belongs in a follow-up against the merged code rather than in this stale branch.

Checked before closing: `git merge-base --is-ancestor 02d81ec2 upstream/rocketmq-studio` exits 0, and all four symbols above are present in the trunk's copies of those files. Nothing in this branch is lost by closing it.

PR #4686, opened 2026-09-21
fix(auth): make the password change and its session revoke atomic

btlqql:
### Related issue

None - no open issue covers the ordering of the password write and the session revoke. #4159 (open) is
about the failed-login path, and #4573 (open) is about the session overview test's DATETIME precision.

### Brief Description

`AuthService.changePassword` (`server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java:355`)
makes two writes: it replaces `password_hash` and then revokes every open session of the account. The
method was not transactional, so the two statements auto-committed independently. If the revoke failed -
a transient connection or lock error, in the middle of a security-relevant operation - the request
answered `500` while the new hash had already been stored and the old sessions were still valid: the
password change was half applied, and the sessions the change was supposed to invalidate survived it.
An operator retrying the call is not helped either, because the first half has already happened.

`setUserEnabled` (`:319`), the sibling mutation that also updates a user row and then revokes that user's
sessions, is already annotated `@Transactional` for exactly this reason. `changePassword` now carries
the same annotation, so either both writes are visible or neither is.

The order of the two statements is unchanged: with one transaction around them, the order no longer
decides the outcome.

### What the review asked for

The three new test methods are renamed to the convention (and to the names you suggested), and the file
now ends with a newline:

- `aFailedSessionRevokeMustNotReplaceTheStoredPasswordTest`
- `aSuccessfulPasswordChangeStillReplacesTheHashAndRevokesSessionsTest`
- `aWrongCurrentPasswordStillRejectsTheChangeAndKeepsTheStoredHashTest`

The third one is kept: #4689 was closed rather than merged, so nothing else covers the mismatched
current password. Its javadoc now says what it pins here — the stored row, not the response shape, which
belongs to `AuthPasswordChangeStatusIntegrationTest` — so the class stays about atomicity.

`AuthServicePasswordChangeIntegrationTest` injects a failing session mapper through `@MockitoBean` and
lets the account row be written and read for real, on the `dev` profile (H2 in MySQL compatibility mode
fed by `classpath:db/schema.sql`). That is what makes the rollback observable: a mock-only test has no
transaction to roll back and no stored row to read back. The class also keeps the happy path and the
wrong-current-password path as controls.

Red/green, re-run on the rebased tree with the annotation removed from `changePassword` and the tests
kept:

```
RED   Tests run: 3, Failures: 1, Errors: 0, Skipped: 0
      AuthServicePasswordChangeIntegrationTest.aFailedSessionRevokeMustNotReplaceTheStoredPasswordTest:95
      [a password change whose session revoke failed must not be half applied]

GREEN Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

The stored hash really was the replacement password after the revoke failed — the assertion is on a row
read back from the database, not on a mock interaction. The other two cases pass either way, as they
must.

Whole suite, on a host where MySQL 8 is reachable so the `@SpringBootTest` classes run too:

```
[INFO] Tests run: 3148, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

3145 of those are the untouched `rocketmq-studio` head; the three new cases are the difference. On a
machine without MySQL the same context-load errors appear on the base revision, which is what the
earlier description was documenting.

### Checklist

- [x] Documentation touched where behaviour changed (no user-visible contract change: a password change still replaces the hash and revokes the sessions, or fails as a whole)

`server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java` is also touched by open PRs from
the same fork. The changes sit in different methods — the user search-term escape, the login username
bound, the password-change transaction — and none supersedes another or depends on another, so they can
be reviewed and merged independently. The one that used to overlap (`#4689`, the status code for a
mismatched current password) is closed.


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins (changes requested):

- The fix itself is right and consistent with what trunk already does: `setUserEnabled` (`AuthService.java:318`) is `@Transactional` for exactly the same update-then-revoke pair, and both entry points (`AuthController.java:103`, `StudioUserController.java:88`) call through the proxy, so the annotation actually takes effect. No new import is needed — `Transactional` is already imported at `:38`.
- The integration test is also the correct shape: `doThrow(...).when(sessionMapper).update(isNull(), any(Wrapper.class))` matches `revokeUserSessions` (`AuthService.java:494`) exactly, and only a real Spring-demarcated transaction on the `dev` H2 profile can observe the rollback, so this cannot be reduced to a mock-only test.
- Blocking item (mechanical, please amend into the same commit): the three new test methods must end with `Test` per the `.github/PULL_REQUEST_TEMPLATE.md` checklist and `conventions.md`. Suggested: `aFailedSessionRevokeMustNotReplaceTheStoredPasswordTest`, `aSuccessfulPasswordChangeStillReplacesTheHashAndRevokesSessionsTest`, `aWrongCurrentPasswordStillRejectsTheChangeAndKeepsTheStoredHashTest`. Your #4689 already follows this convention — please align.
- Minor: the new file has no trailing newline (`\ No newline at end of file`).
- Minor: the third test overlaps #4689's coverage of a wrong current password. Not a blocker, but if #4689 lands first you may want to drop it and keep this class focused on atomicity.

btlqql:
Rebased onto `4c697f07` (clean) and the mechanical item is amended into the same commit.

- The three test methods now end in `Test`, using the names you suggested: `aFailedSessionRevokeMustNotReplaceTheStoredPasswordTest`, `aSuccessfulPasswordChangeStillReplacesTheHashAndRevokesSessionsTest`, `aWrongCurrentPasswordStillRejectsTheChangeAndKeepsTheStoredHashTest`.
- The file ends with a newline (`\ No newline at end of file` is gone).
- The third test stays: #4689 was closed rather than merged, so nothing else covers the mismatched current password. Its javadoc now states that it pins the stored row and not the response shape, which is `AuthPasswordChangeStatusIntegrationTest`'s job — so the class stays about atomicity.

Re-ran the mutation check on the rebased tree to be sure the rename did not dent it — annotation removed from `changePassword`, tests kept:

```
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0
[ERROR] AuthServicePasswordChangeIntegrationTest.aFailedSessionRevokeMustNotReplaceTheStoredPasswordTest -- Time elapsed: 1.193 s <<< FAILURE!
  at ...aFailedSessionRevokeMustNotReplaceTheStoredPasswordTest(AuthServicePasswordChangeIntegrationTest.java:95)
```

Whole suite on a host with MySQL 8 reachable: 3148 tests, 0 failures, 0 errors (3145 on the untouched head).

PR #4689, opened 2026-09-21
fix(auth): answer a mismatched current password with 400, not 401

btlqql:
### Related issue

None - no open issue covers the status of a mismatched current password. #4672 (open) changes what
/api/auth/status reports, and #4159 (open) is about the failed-login path.

### Brief Description

`AuthService.changePassword` answered `401 Current password is incorrect`
(server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java:351) when the submitted current
password did not match. The request is already authenticated at that point - the interceptor resolved
the session before the controller ran - so the failing input is a payload field, not the session.
`GlobalExceptionHandler` turns the exception code straight into the HTTP status
(`GlobalExceptionHandler.java:45`), so that answer reaches the browser as `HTTP 401`.

In this API `401` means "no valid session", and the shipped client keys on it: any 401 outside
`/auth/login` and `/auth/status` makes it drop the stored session and redirect to the login page
(`web/src/api/client.ts:103`, `PUBLIC_AUTH_PATHS` at `client.ts:24`). An operator changing their own
password who mistypes the current password was therefore logged out of a perfectly valid session
instead of being told the password did not match. The Studio user-management dialog calls exactly that
endpoint (`web/src/pages/studio/UserManagement.tsx:291`).

The status is now `400`, which is what this endpoint already uses for its other payload problem: the
same field answers `400` when it is blank (`ChangePasswordDTO.currentPassword` is `@NotBlank`) and an
out-of-policy new password answers `400` as well. The message is unchanged, so nothing else in the
contract moves; a wrong current password is still rejected before the hash is touched.

New test class `AuthPasswordChangeStatusIntegrationTest`. It seeds a user on the `dev` profile (H2 in
MySQL compatibility mode fed by `classpath:db/schema.sql`), logs in for real, and drives
`POST /api/auth/password` through MockMvc and the registered `AuthInterceptor` with the resulting
bearer token. A controller slice would not do: the status under test is produced by the global
exception handler. The second case is the control - a blank current password already answered `400`
before this change.

Before the fix (red), on the unmodified tree:

```
$ cd server && mvn -B -ntp test -Dtest=AuthPasswordChangeStatusIntegrationTest
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0
[ERROR] AuthPasswordChangeStatusIntegrationTest.aWrongCurrentPasswordIsAPayloadErrorAndLeavesTheSessionUsableTest:92 Status expected:<400> but was:<401>
```

The control case (`aBlankCurrentPasswordIsRejectedAsAPayloadErrorTest`) passes on the base tree, which
is the point: the same field, submitted blank, already answered `400`, so the `401` was not a
deliberate contract for this input.

After the fix (green) - the new class plus the neighbouring controller/service classes:

```
$ cd server && mvn -B -ntp test -Dtest='AuthPasswordChangeStatusIntegrationTest,AuthServiceDatabaseTest,AuthControllerTest,AuthServiceTest,StudioUserControllerTest'
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthControllerTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthPasswordChangeStatusIntegrationTest
[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthServiceDatabaseTest
[INFO] Tests run: 16, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthServiceTest
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.StudioUserControllerTest
[INFO] Tests run: 59, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### Checklist

- [x] Documentation touched where behaviour changed (the error text is unchanged; only the status of a rejected payload moved from 401 to 400, matching the blank-field case)

`server/src/main/java/org/apache/rocketmq/studio/auth/AuthService.java` is also touched by open PRs
 from the same fork. The changes sit in different methods - the user search-term escape, the
login username bound, the password-change transaction, and the status code for a mismatched current
password - and none supersedes another or depends on another, so they can be reviewed and merged
independently. #4458 (a different fork, base `master`) also rewrites `AuthService.pageUsers` while sweeping
every `.like(...)` call site; #4666 records that wider overlap.


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins:
Consolidated into #4684 and merged into `rocketmq-studio` as e55409633.

Your commit is included there unchanged, with your authorship preserved on the PR branch. We grouped it with #4684 because they are same-author, same-domain fixes, so the domain lands as one trunk commit instead of several overlapping ones.

PR #4660, opened 2026-09-21
fix(instance): grade cloud import audits by the import outcome

unbridled-41:
Fixes #4656.

## Problem / Evidence

`InstanceService.importCloudInstances` collects region and per-instance failures and reports them in `CloudImportResultVO` (the API response and the web page correctly show a partial-failure warning), but the operation-audit row for the whole import is always recorded as `SUCCESS`:

- `InstanceService.finishCloudImport` calls `recordAudit("IMPORT_CLOUD_INSTANCES", ...)` with the failure counts embedded in the detail text only (`InstanceService.java:356-358`);
- the private `recordAudit` helper hardcoded `"SUCCESS"` as the result (`InstanceService.java:833-841`) and had no way to pass anything else;
- the existing regression `importCloudInstancesShouldContinueAfterUnexpectedRegionFailureTest` pins the wrong behavior: an import with `failed=1` asserting `eq("SUCCESS")` (`InstanceServiceTest.java:1416-1418` on master).

So an import whose catalog provider is missing, whose `listRegions` throws (invalid credential), or in which every region/instance fails produces an audit timeline entry and summary counts that say the operation succeeded.

Red regressions (one new test + one updated expectation):
1. `importCloudInstancesShouldContinueAfterUnexpectedRegionFailureTest` (updated expectation): 1 imported + 1 failed must audit `PARTIAL` — on master it failed with the recorded result `SUCCESS`.
2. `importCloudInstancesShouldAuditAFullyFailedImportAsFailedTest` (new): `listRegions` throwing with zero imports must audit `FAILED` — on master it failed with `SUCCESS`.

## Root cause / Fix

The import flow grew a proper failure model (failure counts, bounded failure details, partial-failure UI) but never carried the outcome into the audit call, whose helper predated it.

Fix:
- `recordAudit` gains a `result` parameter; the create/update/delete call sites pass their existing `"SUCCESS"` explicitly;
- `finishCloudImport` grades the outcome with the shared audit vocabulary — `SUCCESS` when nothing failed, `PARTIAL` when some instances imported and failures occurred, `FAILED` when failures occurred and nothing was imported — mirroring the DLQ resend outcome classification (`RocketMQDLQProvider.classifyOutcome`).

## Priority & scoring

[...]

PR #4477, opened 2026-09-17
feat(instance): add vendor filter and pagination

tju-yxq:
### Which Issue(s) This PR Fixes

- Fixes #3564

### Brief Description

Re-opens the previously reviewed change after the base branch moved from `rocketmq-studio` to `master`. Rebased onto current `master` (`987b748e`), all conflicts resolved, and focused tests re-run locally.

cd server && mvn '-Dtest=InstanceServiceTest,InstanceControllerTest' test → 95 tests passed; cd web && npm test -- InstancePage.test.tsx --run → 23 tests passed; lint/build passed

### How Did You Test This Change?

See the verification commands and results above; all were run on the rebased branch.

### Checklist

- [x] One coherent change; unrelated modifications are not bundled in
- [x] Commit subject follows Conventional Commits
- [x] Tests added or updated for non-trivial changes, test methods named `...Test`
- [x] New UI text has both Chinese and English entries under `web/src/i18n/`
- [x] Architecture constraints stay green (`mvn test` runs the ArchUnit checks)
- [x] New source files carry the ASF license header
- [ ] Documentation touched where behaviour changed (README / `docs/` / in-app help)

PR #4692, opened 2026-09-21
fix: bound user-facing text on code point boundaries

btlqql:
### Related issue

No open issue tracks this defect; it was found by reading the code. The open issues were searched on
2026-09-21 for `failure message truncate code point` and nothing matching came back, so there is no
number to reference.

### Brief Description

This codebase bounds user-facing text in several places, and every one of them counted UTF-16 `char`.
A supplementary character — an emoji, a CJK extension character — is two chars, so a char-based cut
can land between its high and its low surrogate and keep half of one. An unpaired surrogate is not a
code point, has no UTF-8 encoding and is rejected by MySQL `utf8mb4`, so the half character reaches the
browser as a replacement character instead of the character it was cut out of.

New `common/util/TextBounds` holds the rule once — `codePointCount` and a `truncate` that cuts with
`offsetByCodePoints`, optionally appending the ellipsis only when a cut actually happened — and every
site now uses it:

| Site | What it bounds |
| --- | --- |
| `InstanceService.batchFailureMessage` | the per-instance detail in `BatchDeleteResultVO.failed` |
| `InstanceService.boundedCloudImportText` | the per-region detail in `CloudImportResultVO.failed` |
| `InstanceService.requireTextWithin` | the `rmq_instance` text columns (`endpoint`, `remark`, `adminCredentialRef`) |
| `CliAgentProvider:198` | the CLI's own failure output, quoted back to the operator |
| `ClaudeCodeStreamParser:684` | a failed tool result / provider message shown as a notice |
| `MessagePropertyDisplay:93` | a message property value in the message detail view |
| `NotificationOutboxService:540` | a webhook receiver's or SMTP server's rejection reason |

The `InstanceService` caps are also a database-width question, not only a rendering one: MySQL counts a
`varchar` in characters, so 512 emoji fit a `varchar(512)`, and counting chars rejected a value that
fits the column with a 500 from the persistence layer instead of accepting it.

`MessagePropertyDisplay` keeps only the public surface it had — its private `codePointCount` wrapper is
gone, so the codebase has one implementation of the rule rather than three.

### On reachability (the review's first blocking item)

You asked to either point at a realistic input for the two `InstanceService` caps or widen the change.
I widened it, because I agree the two `InstanceService` caps are the weakest of the seven and the four
you named are the strongest:

- `CliAgentProvider:198` bounds the CLI's failure output, and the provider is handed the operator's raw
  prompt text — an operator writing a Chinese or emoji-bearing prompt can see it quoted back in the
  failure message (`:159`, `:161`).
- `ClaudeCodeStreamParser:684` bounds a failed tool result and the provider's own message, i.e. text
  derived from message bodies an end user wrote.
- `MessagePropertyDisplay:93` bounds message property values, which are end-user controlled by
  definition.
- `NotificationOutboxService:540` bounds a webhook receiver's response body — outside text this
  codebase does not produce.

The two `InstanceService` caps stay in the change, as you allowed, and they are now the same helper as
the rest rather than a second private copy.

### The second blocking item, and the merged #4713

The shared helper is `common/util/TextBounds`, as your conventions require. #4713 landed while this was
in review, so its `requireTextWithin` (`endpoint` / `remark` / `adminCredentialRef`) now counts code
points through the same helper instead of adding a competing private variant — that is the coordination
you asked for, done in this PR rather than promised.

Cosmetic items: `EMOJI` has a javadoc and a blank line before the first `@Test`, and the extra blank
line before the closing brace is gone.

Five of the seven sites have a test that cuts through a supplementary character and asserts no unpaired
surrogate and the exact cut end; `TextBoundsTest` covers the helper itself. All of them are
mutation-sensitive: reverting `TextBounds.truncate` to a char-based cut reddens eight tests across
those five sites at once.

```
$ # helper mutated back to substring(0, budget), tests untouched
$ cd server && mvn -B -ntp test -Dtest='TextBoundsTest,InstanceServiceTest,CliAgentProviderTest,ClaudeCodeStreamParserTest,NotificationOutboxServiceTest,MessagePropertyDisplayTest'
[ERROR] Tests run: 196, Failures: 8, Errors: 0, Skipped: 0
[ERROR]   TextBoundsTest.truncateShouldNotSplitASurrogatePairTest:44
[ERROR]   InstanceServiceTest.deleteInstancesShouldCutFailureMessagesOnCodePointBoundariesTest:1879
[ERROR]   InstanceServiceTest.importCloudInstancesShouldBoundFailureMessagesOnCodePointBoundariesTest:1896
[ERROR]   CliAgentProviderTest.aFailedCliShouldAbbreviateItsOutputOnCodePointBoundariesTest:159
[ERROR]   ClaudeCodeStreamParserTest.aLongFailedToolResultShouldBeAbbreviatedOnCodePointBoundariesTest:495
[ERROR]   MessagePropertyDisplayTest.limitPropertiesShouldKeepASupplementaryCharacterThatFitsTheCapTest:89
[ERROR]   MessagePropertyDisplayTest.limitPropertiesShouldNotSplitASurrogatePairTest:76
[ERROR]   NotificationOutboxServiceTest.abbreviatesALongWebhookRejectionReasonOnCodePointBoundariesTest:333
```

Restored, the same command is green. Whole suite, on a host where MySQL 8 is reachable so the
`@SpringBootTest` classes run too:

```
[INFO] Tests run: 3156, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

3145 of those are the untouched `rocketmq-studio` head; the eleven new cases are the difference. No test
here touches the network or an external CLI.

### Checklist

- [x] Commit subject follows Conventional Commits (`fix:`, no scope because the change now spans several packages; `fix:` without a scope has precedent on this branch, e.g. `ef9b9826`)
- [x] New source files carry the ASF license header; no non-ASCII source literal — the emoji is written as a `\uD83D\uDE00` escape
- [x] Documentation touched where behaviour changed (the helper, the cap constants and each cut site state that they count code points; no user-visible contract change)

`InstanceService.java` is also touched by #4660 (`fix(instance): grade cloud import audits by the import
outcome`) and #4477 (`feat(instance): add vendor filter and pagination`). No hunk is shared, and the
earlier collision with #4713 is settled — that PR is merged and this change adopts its helper, so the
merge order between #4713 and this one no longer matters.


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins (changes requested):

- The change is correct as far as it goes, and your two tests are genuinely mutation-sensitive — we recomputed both expectations character by character and they do go red on the base revision (the cut lands on the lone high surrogate at char 498 for the cloud-import path and char 499 for the batch-delete path).
- Two things block a merge as-is.
- 1. Scope / production reachability. The two caps you fixed bound text that comes from RocketMQ and cloud-OpenAPI exception messages, which in practice are ASCII or BMP Chinese — we could not find a caller that can actually put a supplementary character at offset 499/500, so we cannot justify the change on its own. The truncations in this codebase that *do* routinely see emoji are elsewhere: `ops/ai/CliAgentProvider.java:198` (`trimmed.substring(0, 500) + "..."`), `ops/ai/ClaudeCodeStreamParser.java:684` (`value.substring(0, maxChars) + "..."`), `common/util/MessagePropertyDisplay.java:93` and `ops/alert/NotificationOutboxService.java:540`. Please either point us at a realistic input for the two `InstanceService` caps, or widen the change to those sites.
- 2. Placement. `codePointCount` is private to `InstanceService`. If the fix is worth doing it is worth doing once: our conventions require shared logic to live in `common/util`, so please add a small `TextBounds`/`SafeTruncate` helper there and use it from all the sites above (that also removes the thin wrapper around `String.codePointCount`).
- Heads-up on a collision: #4713 touches the same two files and adds `requireTextWithin(...)` caps for `endpoint` / `remark` / `adminCredentialRef`. The patches merge cleanly, but whichever lands second should apply the same code-point rule to the other's helper — please coordinate rather than both adding a private variant.
- Cosmetic, non-blocking: `EMOJI` is inserted without a blank line before the first `@Test`, and there is an extra blank line before the closing brace of the test class.

btlqql:
Widened, as you offered — plus the placement item and the cosmetics.

**Scope.** New `common/util/TextBounds` holds the rule once (`codePointCount`, and a `truncate` that cuts with `offsetByCodePoints` and appends the ellipsis only when it actually cut). It is now used at all seven sites: the two `InstanceService` caps, `CliAgentProvider:198`, `ClaudeCodeStreamParser:684`, `MessagePropertyDisplay:93`, `NotificationOutboxService:540`, and — since #4713 merged while this sat in review — its `requireTextWithin` for `endpoint` / `remark` / `adminCredentialRef`. So the coordination you asked for happened in this PR rather than being promised: #4713's helper now counts code points through the same utility, and `MessagePropertyDisplay`'s private `codePointCount` wrapper is gone.

You were right that the two `InstanceService` caps are the weakest of the seven. I kept them as you allowed, because they are also a column-width question: MySQL counts a `varchar` in characters, so 512 emoji fit a `varchar(512)` and counting UTF-16 chars rejected a value that fits the column with a 500 from the persistence layer. The four you named are the ones with the obvious end-user text behind them — the CLI's failure output quotes the operator's own prompt back (`CliAgentProvider:159/161`), the stream parser bounds a failed tool result, message properties are user-supplied, and the delivery reason is a webhook receiver's response body.

**Mutation check**, on the rebased tree — helper reverted to `substring(0, budget)`, tests untouched:

```
[ERROR] Tests run: 196, Failures: 8, Errors: 0, Skipped: 0
[ERROR]   TextBoundsTest.truncateShouldNotSplitASurrogatePairTest:44
[ERROR]   InstanceServiceTest.deleteInstancesShouldCutFailureMessagesOnCodePointBoundariesTest:1879
[ERROR]   InstanceServiceTest.importCloudInstancesShouldBoundFailureMessagesOnCodePointBoundariesTest:1896
[ERROR]   CliAgentProviderTest.aFailedCliShouldAbbreviateItsOutputOnCodePointBoundariesTest:159
[ERROR]   ClaudeCodeStreamParserTest.aLongFailedToolResultShouldBeAbbreviatedOnCodePointBoundariesTest:495
[ERROR]   MessagePropertyDisplayTest.limitPropertiesShouldKeepASupplementaryCharacterThatFitsTheCapTest:89
[ERROR]   MessagePropertyDisplayTest.limitPropertiesShouldNotSplitASurrogatePairTest:76
[ERROR]   NotificationOutboxServiceTest.abbreviatesALongWebhookRejectionReasonOnCodePointBoundariesTest:333
```

Eight tests across five sites, so no site is covered by assertion alone.

**Cosmetics:** `EMOJI` has a javadoc and a blank line before the first `@Test`; the trailing blank line is gone; the file ends with a newline.

Whole suite on a host with MySQL 8 reachable: 3156 tests, 0 failures, 0 errors (3145 on the untouched head). The subject is now `fix: bound user-facing text on code point boundaries` since the change spans five packages — `fix:` without a scope has precedent on this branch (`ef9b9826`), and I updated the PR title to match.

PR #4696, opened 2026-09-21
fix(ai): attach only the run of the conversation on screen

Frun1na:
- Fixes #4697

### Brief Description

`useConversationTimeline` held only the run the last successful load reported, and nothing about which conversation that was. The report outlives the conversation it belongs to, because replacing it takes an asynchronous refetch — so for at least one commit after a switch, `activeRun` is still the previous conversation's run.

`useActiveRunAttach` then attached it to the conversation now on screen. The switch itself re-runs that effect (the abort in `useAgentRun` re-renders, and the reset effect clears the one-attempt marker in the same commit), and the attach endpoint is addressed by run id alone, so `attachRunStream` streams the previous conversation's run into the new transcript. `startStream`'s ownership guard compares against `conversationIdRef`, which the same commit has already updated to the new conversation, so it lets the attach through.

`activeRun` is now the run of the conversation on screen: the loaded run is paired with the conversation it was loaded for, and a mismatch reports null. That is the honest answer — this conversation's timeline has not answered yet — and it is the condition the attach decision needs.

`cd web && npm test` — `Test Files 135 passed (135)`, `Tests 1237 passed (1237)`. `npx tsc --noEmit` and `npx eslint` on the touched files are clean.

New tests, each verified to fail against the unfixed code:

- `useActiveRunAttach.test.ts › attachesOnlyTheRunOfTheConversationOnScreenTest` — exercises the handover between the two hooks with the real timeline hook: conversation 7 attaches its own run, switching to conversation 9 while its timeline is still in flight must not attach conversation 7's run, and once conversation 9's timeline lands its run is what gets attached. Against the unfixed code the attach count after the switch is 2 instead of 1.
- `useConversationTimeline.test.ts › doesNotReportThePreviousConversationsActiveRunTest` — `activeRun` is null immediately after a switch and becomes the new conversation's run when its load lands.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (not applicable: no new UI text)
- [x] New source files carry the ASF license header (the new test file carries it)
- [x] Documentation touched where behaviour changed (`activeRun`'s contract is documented where it is declared)


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #4698, opened 2026-09-21
fix(studio): keep a failed Ops load out of the unsupported-cluster state

Frun1na:
- Fixes #4699

### Brief Description

The Ops page had two states derived from one boolean: `configurationAvailable` false meant both "the cluster does not support reading or updating the Ops configuration" and "we have not managed to find out yet". A failed `queryOpsHomePage` only raised a toast and left the boolean false, so the page settled into the first state — rendering "当前集群不支持读取或更新 Ops 配置" for a network or server failure, hiding every write control behind that reason, and offering no retry.

`loadState` ('loading' | 'ready' | 'failed') now separates the two:

- while loading, neither banner is rendered — the page makes no claim it cannot support;
- a failed load renders a warning with a retry that re-runs the request;
- the "configuration unavailable" banner renders only once the load has answered and does not support configuration.

This also removes the flash of the unsupported banner on first paint, which was the same mistake with a shorter lifetime.

`cd web && npm test` — `Test Files 134 passed (134)`, `Tests 1236 passed (1236)`. `npx tsc --noEmit` and `npx eslint` on the touched files are clean.

New test, verified to fail against the unfixed code: `Ops.test.tsx › offers a retry instead of claiming the cluster does not support Ops config` — rejects the first `queryOpsHomePage`, asserts the retry is present and the "unsupported" description is absent, clicks it, and asserts the second request ran and the write controls appeared. Against the unfixed code the query for the retry button times out.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (the alert reuses the existing `ops.fetchFailed`, the button `common.retry`; no new keys)
- [x] Documentation touched where behaviour changed (not applicable: no documented behaviour changes)


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #4700, opened 2026-09-21
fix(deliveries): offer a retry when the instance filter cannot load

Frun1na:
- Fixes #4701

### Brief Description

The deliveries page loads the instance list for its instance filter with `listInstances().catch(() => undefined)`. A failure therefore left an empty filter with no message and no way to retry: the dropdown simply had no options, which reads as "this deployment has no instances" — something a failed request cannot establish, and the only filter on the page that silently degrades this way.

The failure is now a state: the select is marked `status="error"` and a retry button appears beside it, with the reason in its tooltip. Retrying re-runs the request and clears the state when it succeeds.

`cd web && npm test` — `Test Files 134 passed (134)`, `Tests 1236 passed (1236)`. `npx tsc --noEmit` and `npx eslint` on the touched files are clean.

New test, verified to fail against the unfixed code: `NotificationDeliveriesPage.test.tsx › surfaces a failed instance-list load with a retry instead of an empty filter` — rejects the first `listInstances`, asserts the retry is rendered, clicks it, and asserts the request ran again and the retry disappeared. Against the unfixed code the query for the retry button times out.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (`deliveries.instancesLoadFailed`; the button reuses `common.retry`)
- [x] Documentation touched where behaviour changed (not applicable: no documented behaviour changes)


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #4710, opened 2026-09-21
fix(alert): keep a padded stored instance id in the suppression correlation

btlqql:
### Related issue

- No open issue covers these comparisons. The closest is the rule-side whitespace work, which is **already merged as #4464**: `MybatisPlusAlertStateRepository.java:107-108` and `NativeAlertProcessor.java:106-107` use `StringUtils.trimWhitespace` on trunk, so the reference to #4372 in the earlier revision of this description was stale and is withdrawn.

### Brief Description

Two Java-side instance comparisons read a stored row raw while the query that produced it states the filter on trimmed text:

- `AlertCorrelationScope.matches` compared the stored cluster incident's instance id raw, so a FIRING cluster incident returned by the instance filter could be dropped again before it suppressed the business notification it correlates with.
- `AlertService.listAlerts(level, domain, instanceId, transition)` did the same, and there the Java comparison is the only one that decides: `findAlerts(level)` applies **no** instance filter in SQL at all, so a row it returns has nothing else to agree with.

Both now compare the stored value as trimmed text — the rule #4464 already applies to a stored rule instance id, and the rule `MybatisPlusAlertRepository.java:198` states in SQL.

**On the reachable producer, plainly.** I looked for one and could not find it, and I agree with your reading:

- `rmq_system_alert` rows are only ever inserted from `NativeAlertEvaluationService.evaluate` (`:87`) and `NativeAlertProcessor.emitLifecycleEvent` (`:188`), and both take the instance id from `MetricSample.instanceId()`.
- Every `MetricSample` in the collectors is built from `instance.getName()`, and every writer of `rmq_instance.name` trims it — `InstanceService.java:468`, `:508` through `requireInstanceName`, including the cloud-import path at `:506`.
- `MybatisPlusAlertRepository.toAlertEntity` (`:341`) copies whatever it is given, but nothing gives it a padded value.

So this is a consistency change, not the repair of a row I can produce, and the description no longer claims otherwise. I am keeping it because the asymmetry is the whole of it — one comparison in this feature trims and its two siblings did not — and because #4464 set the rule for the sibling path. If you would rather have the alert path left alone than made consistent with it, say so and I will close this PR rather than argue for it.

### The two points from the review that were wrong

- **Fixture.** The test used `" local "` with a *leading* space, which `instance_id = 'local'` cannot return under any collation. It now uses trailing padding only, and the comment says why that is the only form a PAD SPACE collation could have let through, and that MySQL 8's default `utf8mb4_0900_ai_ci` is NO PAD.
- **The second site.** The same comparison in `AlertService.java:506` is fixed in this PR, as you asked; it is the stronger of the two, because `findAlerts(level)` filters nothing in SQL.

Three tests, all mutation-sensitive in the sense that they pin the comparison rather than the plumbing:

- `AlertNotificationSuppressionServiceTest.suppressesWhileTheClusterIncidentKeepsAPaddedStoredInstanceIdTest` — the incident the repository returns carries trailing padding and must still suppress the business event.
- `AlertServiceTest.listAlertsShouldCompareAStoredInstanceIdAsTrimmedTextTest` — the feed path, where the Java comparison is the only filter. Its comment states that no writer pads the value today, so what it pins is the comparison rule.
- Both go red on the base revision: reverting either production hunk fails the corresponding test.

```
$ cd server && mvn -B -ntp test          # whole suite, on a host with MySQL 8 available
[INFO] Tests run: 3147, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

3145 of those are the untouched `rocketmq-studio` head; the two new cases are the difference.

### Checklist

- [x] Documentation touched where behaviour changed — not applicable, the correlation contract is unchanged for unpadded values


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins (changes requested):

- The mechanism in the description does not hold: `rmq_system_alert.instance_id` is always written from `MetricSample.instanceId`, which is `instance.getName()` (`NativeAlertProcessor.java:176`, `ApacheRocketMqClusterMetricsCollector.java:172`), and every writer of `rmq_instance.name` trims it (`InstanceService.requireInstanceName`, `InstanceService.java:554-563`). `git log -S '.instanceId(rule.getInstanceId())'` finds no revision that ever wrote the rule's raw instance id into an alert row, so "a row written as ' local ' by an earlier build" has no code path behind it. Please show a reachable producer of a padded stored value.
- The collation argument only covers *trailing* spaces: PAD SPACE collations ignore trailing padding, and MySQL 8's default `utf8mb4_0900_ai_ci` is NO PAD, so it ignores nothing. The new fixture uses `" local "` (leading space), which `MybatisPlusAlertRepository.java:198` (`instance_id = 'local'`) can never return under any collation — the unit test pins a state the query cannot produce.
- If padded stored ids are accepted as a legacy-data reality (as #4464 did for rules), the same asymmetric comparison in `AlertService.java:506` must be fixed in this PR too: it filters `findAlerts(level)` results, which apply **no** SQL instance filter at all, so it drops such a row without needing any collation leniency. `AlertService.java:553/564` already go through `AlertCorrelationScope.matches` and are covered.
- Note the description's reference to #4372 is stale: `MybatisPlusAlertStateRepository.java:107-108` and `NativeAlertProcessor.java:106-107` already use `StringUtils.trimWhitespace` on trunk (merged today as #4464).

btlqql:
Reworked onto the current `rocketmq-studio` head, and you were right on the mechanism — I could not produce a padded `rmq_system_alert.instance_id` either:

- the only writers of that table are `NativeAlertEvaluationService.evaluate` (`:87`) and `NativeAlertProcessor.emitLifecycleEvent` (`:188`), and both take `MetricSample.instanceId()`;
- every sample in the collectors is built from `instance.getName()`, and every writer of `rmq_instance.name` trims it (`InstanceService.java:468`, `:508` via `requireInstanceName`, including the cloud-import path at `:506`).

So the description no longer claims a reachable row. What the PR does now is remove the asymmetry itself, and it covers the second site you named: `AlertService.java:506`, where the Java comparison is the only filter, because `findAlerts(level)` applies no instance filter in SQL. The fixture is fixed too — trailing padding only, with the collation note (MySQL 8's `utf8mb4_0900_ai_ci` is NO PAD, which is exactly why the earlier leading-space fixture pinned a state the query cannot return).

Full suite on a host with MySQL 8 reachable: 3147 tests, 0 failures, 0 errors (3145 on the untouched head).

If you would rather the alert path were left as it is than made consistent with #4464's rule, say so and I will close this one — I would rather withdraw it than argue a no-op into the branch.

PR #4561, opened 2026-09-20
fix(alert): surface the test-notification failure reason

unbridled-41:
Fixes #4557.

## Problem / Evidence

`NotificationOutboxService.sendTestMessage` (behind `POST /api/settings/general/test-notification`) threw `IllegalArgumentException` for an unknown channel and `IllegalStateException` for every delivery failure ("No configured dingtalk webhook", "SMTP is not configured", "DingTalk rejected webhook: …", "Unable to sign DingTalk webhook"). `GlobalExceptionHandler` has no handler for either type, so they fell into the generic `Exception` → HTTP 500 with the hard-coded body `Internal Server Error`. The settings page displays `response.data.message` (`GeneralSettingsTab.sendTest`), so all three Test buttons reported the same useless 500 for every failure mode and the diagnostic reason was discarded after `log.error`.

Red output: `expect BusinessException but was: IllegalStateException: Test notification failed: No configured dingtalk webhook`.

## Root cause / Fix

Error-type mismatch: the service used JDK exceptions while the handler layer only translates `BusinessException` into an actionable response. Both throws now carry the reason as a `BusinessException` — 400 `Unsupported notification channel: <channel>`, 502 `Test notification failed: <reason>` — which the existing handler renders verbatim and the frontend already surfaces.

## Priority & scoring

PRIORITY 72 = impact 26 (all test-notification failure modes indistinguishable; operators cannot self-diagnose) + reach 12 (dingtalk/email/sms test buttons on the settings page, a routine ops action) + repro 20 (curl with an empty webhook) + maintain 14 (aligns the endpoint with the repo's established error contract). FIX_CONFIDENCE 95 (pure exception-type change; handler and frontend paths already exist and are tested).

## Tests

- New `NotificationOutboxServiceTest.testMessageFailureCarriesTheConfigurationReasonTest`:
  - Red (baseline): `Tests run: 1, Failures: 1` — expected `BusinessException` but was `IllegalStateException: Test notification failed: No configured dingtalk webhook`.
  - Green: asserts 502 + message contains "webhook" for the unconfigured-dingtalk case, and 400 + message contains "fax" for an unknown channel.
- Full class: `mvn test -Dtest=NotificationOutboxServiceTest` → **26/26**.
[...]

Issue #4557, opened 2026-09-19
[Studio][Bug] Test-notification failures always surface as a generic 500 "Internal Server Error"


unbridled-41:
### Before Creating the Issue

- [x] I have searched the open issues of this repository and believe this is not a duplicate.
- [x] This is a defect in RocketMQ Studio.

### Studio Version

branch: `master`, commit `d50ffecc`

### Describe the Bug

`NotificationOutboxService.sendTestMessage` (backing `POST /api/settings/general/test-notification`) throws `IllegalArgumentException` for an unknown channel and `IllegalStateException` for every delivery failure ("No configured dingtalk webhook", "SMTP is not configured", "DingTalk rejected webhook: …", "Unable to sign DingTalk webhook", …). `GlobalExceptionHandler` has no mapping for either type, so they fall into the generic `Exception` handler and surface as HTTP 500 with the hard-coded body `Internal Server Error`.

The settings page reads `response.data.message` (`GeneralSettingsTab.sendTest`), so the operator sees the same useless "Internal Server Error" whether the channel is misspelled, the webhook is not configured, SMTP is missing, or the remote side rejected the signed request. All diagnostic information is discarded after `log.error`.

### Steps to Reproduce

```
curl -i -X POST 'http://localhost:8888/api/settings/general/test-notification?channel=dingtalk' \
  -b <admin session cookie>
# → HTTP 500, body: {"code":500,"message":"Internal Server Error"}
```

with an empty dingtalk webhook (or `channel=sms` with no SMS configuration, etc.). Every failure mode of the three Test buttons returns the identical body.

### What Did You See Instead?

HTTP 500 `Internal Server Error` with the reason visible only in server logs.

### What Did You Expect to See?

A 4xx/5xx business error whose `message` carries the configuration reason, so the settings UI can show what to fix (the frontend already renders `data.message`).

### Related work

- #3220 / open PR #4410 add a channel-readiness drawer — a different, complementary feature; it does not change how a failed test delivery reports its reason.
- `BusinessException` is the established vehicle for actionable 4xx/5xx reasons (`GlobalExceptionHandler` renders `ex.getMessage()`), used by sibling services.

#

## PR

Fix: #4561.

This PR targets the development trunk `rocketmq-studio` (retargeted from `master` on 2026-09-21; `master` is frozen at `d50ffecc`, the commit that moved the trunk).

[...]

PR #4711, opened 2026-09-21
fix(alert): reject an alert-rule test run without a metric

btlqql:
### Related issue

- No open issue covers this endpoint; the closest is #4557 (test-notification failures surface as a generic 500), whose fix #4561 covers the notification-delivery test path in `NotificationOutboxService`.
- This PR covers `NativeAlertRuleTestService`, which serves the alert-rule test button of both the cluster and the business rule pages.

### Brief Description

`NativeAlertRuleTestService.test` filters the collected samples with `rule.getMetric().equals(sample.metricKey())` (line 72), but a rule without a metric reaches that line: `NativeAlertMetricCatalogService.validate` returns early for a null metric, and `NativeAlertRulePolicy.validate` treats a blank metric as "not a native metric" and returns without an error. `POST /api/alert-rules/test` (and its cluster twin) with a body such as `{"name":"rule","instanceId":"local","operator":">","threshold":10}` therefore answered with an unhandled `NullPointerException` - an HTTP 500 - instead of telling the caller which field is missing.

The test run now rejects a rule without a metric up front with `400 metric is required`, before any samples are collected for it. That is the same request-error channel the surrounding validation uses (`AlertRulePolicy` for the scope fields, `AlertMetricCatalogService` for the metric support check), so no HTTP contract is changed for valid requests, and the two controllers keep passing a non-null rule (`requireRule` / `requireAlertRule`).

Test written first and run against the unmodified tree (`red`, base commit `7ce9a682`):

```
$ cd server && mvn -B -ntp test -Dtest=NativeAlertRuleTestServiceTest
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0
[ERROR]   NativeAlertRuleTestServiceTest.rejectsARuleWithoutMetricInsteadOfFailingWithANullPointerExceptionTest:141
Expecting actual throwable to be an instance of:
  org.apache.rocketmq.studio.common.exception.BusinessException
but was:
  java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of
  "org.apache.rocketmq.studio.ops.alert.AlertRuleVO.getMetric()" is null
    at org.apache.rocketmq.studio.ops.alert.NativeAlertRuleTestService.lambda$test$1(NativeAlertRuleTestService.java:72)
[INFO] BUILD FAILURE
```

After the fix (`green`), together with the policy, catalog, scope-matcher and both rule controllers:

```
$ cd server && mvn -B -ntp test -Dtest=NativeAlertRuleTestServiceTest,NativeAlertRulePolicyTest,NativeAlertMetricCatalogServiceTest,NativeAlertRuleScopeMatcherTest,AlertRuleControllerTest,ClusterAlertRuleControllerTest
[INFO] Tests run: 50, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Test coverage:

- `rejectsARuleWithoutMetricInsteadOfFailingWithANullPointerExceptionTest` - a rule with an instance, an operator and a threshold but no metric, tested against a collector that returns one matching sample (so the failing filter is actually evaluated), must raise `BusinessException` with code 400 and the message `metric is required`.
- The other four cases of the class stay green, including the one that asserts a healthy collector still answers when another collector throws.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` - not applicable, no UI text added
- [x] New source files carry the ASF license header - no new source files
- [x] Documentation touched where behaviour changed - not applicable, only the error path of a rejected request changed


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins:
Consolidated into #4682 and merged into `rocketmq-studio` as 15167dc5d.

Your commit is included there unchanged, with your authorship preserved on the PR branch. We grouped it with #4682 because they are same-author, same-domain fixes, so the domain lands as one trunk commit instead of several overlapping ones.

PR #4714, opened 2026-09-21
fix(auth): keep an overflow login lock in force after a tracker refill

btlqql:
### Related issue

No open issue covers this path; related prior work is #3045 (closed by #3046), which fixed two other
fail-open paths of the same limiter and stated the invariant this change restores — "an active lock must
remain in force until its lock duration expires".

### Brief Description

`LoginRateLimiter` protects the login endpoint in two layers. The exact tracker holds at most
`MAX_TRACKED_USERNAMES` usernames, and the overflow tracker (`OVERFLOW_BUCKET_COUNT` hash buckets)
takes over for usernames that arrive while that tracker is full. All **three** callers of
`clearOverflowAttempts()` threw the *whole* overflow tracker away when they merely freed one exact slot:

- `activeExactState` (`LoginRateLimiter.java:164`) — `attempts.remove(key); clearOverflowAttempts();`
- the lazy sweep in `reclaimExpiredExactAttemptsIfDue` (`:187`) — `if (removed) clearOverflowAttempts();`
- `recordSuccess` (`:135`) — the whole-table wipe on a successful login

Freeing one exact slot therefore also cancelled every lock in the overflow tracker, and all three paths
run on the unauthenticated login endpoint: a caller who fills the exact tracker, waits for the first
decoy's failure window to lapse, and then sends two new usernames has the tracker saturated again — with
the overflow tracker empty. Any username that had earned a 5-failure lock then answers `checkAllowed`
with "allowed" for the rest of its `LOCK_DURATION`. That is attacker-triggered, cheap (one request per
freed slot) and needs no credentials, and it contradicts the invariant the class documents for itself.

All three now drop the overflow state of a finished episode only when no lock in it is still in force
(`clearStaleOverflowAttempts`), so:

- an active lock survives slot churn and stays enforced for its full `LOCK_DURATION`;
- an episode's leftovers that carry no lock are still dropped, exactly as before, so stale overflow
  state cannot outlive the episode it belongs to;
- a successful login no longer cancels anyone's lock — including its own username's, which the overflow
  tracker may hold under a bucket it shares with a collision.

Nothing was weakened to make this pass: the change only ever *keeps* a rejection the base code already
had, and the locked-state branch of `incrementFailure` is untouched, so a lock is still never extended
by later failures. The one visible trade-off is stated for the record: a bucket lock is shared by hash
collisions, so a colliding username stays rejected for the remainder of a lock it shares — which is what
#3045 asked for, since the alternative is a lock anybody can cancel.

### What the review asked for

**`recordSuccess` is now on the same guard.** You were right that my own reasoning applies verbatim
there, and that it is reachable with one valid credential: fail once for your own username (creating the
exact entry), then log in successfully → whole-table wipe → every other username's live overflow lock is
gone. It now reads `clock.millis()` and calls `clearStaleOverflowAttempts(now)`. Two tests cover it, one
for each direction:

- `successfulLoginMustNotReleaseAnotherUsersOverflowLockTest` — the reachable sequence above;
- `successfulLoginStillDropsOverflowStateThatCarriesNoLockTest` — the control, so the episode reset that
  must stay still happens.

**The javadoc's scope sentence is gone, and the invariant now holds for all three callers.** The class
javadoc previously read "Collisions can share a lock only while the exact tracker is saturated; they
cannot disable rate limiting", which was only true once the three callers agreed. It now states that a
lock the overflow tracker earned outlives the saturation that created it, that all three paths that free
an exact slot only drop state carrying no lock, and the collision trade-off explicitly.

Five new tests in `LoginRateLimiterTest`, all driven by the existing `MutableClock` so the windows are
exact — the two original defect paths, the two `recordSuccess` cases you asked for, and the control for
the reset that must stay.

Mutation check, run on the rebased tree with the guard reverted on `recordSuccess` alone (the other two
callers untouched), tests kept:

```
$ cd server && mvn -B -ntp test -Dtest=LoginRateLimiterTest
[ERROR] Tests run: 17, Failures: 1, Errors: 0, Skipped: 0
[ERROR]   LoginRateLimiterTest.successfulLoginMustNotReleaseAnotherUsersOverflowLockTest:330
```

So the new case is mutation-sensitive to exactly the hunk it was written for, and the other four stay
green — they do not accidentally cover it.

Whole suite, on a host where MySQL 8 is reachable so the `@SpringBootTest` classes run too:

```
[INFO] Tests run: 3150, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

3145 of those are the untouched `rocketmq-studio` head; the five new cases are the difference.

### Checklist

- [x] Documentation touched where behaviour changed (class javadoc states the invariant that now holds; the guard's javadoc names all three callers and why a successful login is evidence about its own username only)


RockteMQ-AI (approved):
LGTM. Looks good.

---

lizhimins (changes requested):

- The defect and both fixed paths check out (`activeExactState:159-166`, `reclaimExpiredExactAttemptsIfDue:178-188`), and the two new tests are mutation-sensitive. One same-family path is still missing.
- `clearOverflowAttempts()` has three callers on trunk; `recordSuccess` (`LoginRateLimiter.java:135-143`) still wipes the whole overflow table unconditionally when `attempts.remove(key) != null`. Your own reasoning applies verbatim there: the freed exact slot can be taken again immediately by anybody, so the overflow tracker is back in charge of the username it was protecting, and a live lock is cancelled with most of `LOCK_DURATION` still to run.
- It is reachable with one valid credential: fail once for your own username (creates the exact entry), then log in successfully → whole-table wipe → every other username's live overflow lock is gone; re-saturate the exact tracker and the brute force continues. Please either route `recordSuccess` through the same stale-only guard (it needs a `clock.millis()` reading) and add a mutation-sensitive test for it, or state in the code why a successful login is allowed to cancel another user's live overflow lock — the new javadoc currently only says "Both callers run on unauthenticated requests", which documents the scope but not the justification.
- Minor: with the guard in place, the invariant in the class javadoc (`:29-31`, "they cannot disable rate limiting") is only true once all three callers agree.

btlqql:
All three items are done, on top of `4c697f07`.

**`recordSuccess` is on the same guard now.** You were right, including the reachability: fail once for your own username (creating the exact entry), then log in successfully → whole-table wipe → every other username's live overflow lock is gone. It now reads `clock.millis()` and calls `clearStaleOverflowAttempts(now)`. Two tests, one per direction:

- `successfulLoginMustNotReleaseAnotherUsersOverflowLockTest` — the reachable sequence above;
- `successfulLoginStillDropsOverflowStateThatCarriesNoLockTest` — the control, so the episode reset that must stay still happens.

**Mutation check, scoped to that hunk alone** (the other two callers left as fixed), on the rebased tree:

```
$ mvn -B -ntp test -Dtest=LoginRateLimiterTest
[ERROR] Tests run: 17, Failures: 1, Errors: 0, Skipped: 0
[ERROR]   LoginRateLimiterTest.successfulLoginMustNotReleaseAnotherUsersOverflowLockTest:330
```

The other four new tests stay green, so this case is sensitive to its own hunk and not accidentally covered by its neighbours.

**The javadoc's scope sentence is gone.** You were right that "they cannot disable rate limiting" was only true once all three callers agree. The class javadoc now says a lock the overflow tracker earned outlives the saturation that created it, that all three paths freeing an exact slot only drop state carrying no lock, and it states the collision trade-off explicitly — a colliding username stays rejected for the remainder of a shared lock.

Whole suite on a host with MySQL 8 reachable: 3150 tests, 0 failures, 0 errors (3145 on the untouched head).

PR #4715, opened 2026-09-21
fix: prevent unread badge from inflating when streaming fails

enkilee:

- Fixes #<issue-id>

### Brief Description

- The unread count effect of `ChatThread` relies on `[bubbleCount]`. When `bubbleCount` decreases (due to the removal of `liveBlocks` caused by streaming output failure/abortion), the effect will still execute and increment `unread`, resulting in an incorrect increase in the badge number.
- Introduce `prevBubbleCountRef` to track the previous `bubbleCount`, and increment `unread` only when `bubbleCount` **increases**.

## Root Cause

```tsx
useEffect(() => {
  if (atBottomRef.current) return;
  setUnread((count) => count + 1);
}, [bubbleCount]);
```
`bubbleCount = bubbles.length + (liveBlocks?.length > 0 ? 1 : 0)`. When the streaming output fails, `liveBlocks` is cleared, `bubbleCount` decreases from N+1 to N, the effect still triggers, and the `unread` error is incremented by 1.

## Fix

```tsx
const prevBubbleCountRef = useRef(bubbles.length);

useEffect(() => {
  const previous = prevBubbleCountRef.current;
  prevBubbleCountRef.current = bubbleCount;
  if (atBottomRef.current || bubbleCount <= previous) return;
  setUnread((count) => count + 1);
}, [bubbleCount]);
```

### Checklist

- [ ] One coherent change; unrelated modifications are not bundled in
- [ ] Commit subject follows Conventional Commits (`feat:` / `fix:` / `refactor:` / `chore:` / `docs:` / `perf:`)
- [ ] Tests added or updated for non-trivial changes, test methods named `...Test`


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #4703, opened 2026-09-21
fix(acl): reject blank default permissions in plain access config

yyqdbngt:
### Brief Description

`MybatisPlusAclRepository.createAndUpdatePlainAccessConfig` guarded the per-resource permission
entries (`topicPerms` / `groupPerms`, added in #2563) but bound the two *default* permissions straight
into `rmq_acl_rule`:

```java
if (config.getDefaultTopicPerm() != null) {
    rules.add(plainRule(config.getAccessKey(), "*", "Cluster", config.getDefaultTopicPerm(), "DEFAULT_TOPIC"));
}
```

So `POST /api/acl/plain-access-config` with `{"accessKey":"svc-x","secretKey":"...","defaultTopicPerm":""}`
(or `"   "`) answered `200` and persisted a durable `DEFAULT_TOPIC` / `DEFAULT_GROUP` rule whose
`actions` column is empty — an account-level default that grants nothing but is stored as configured
(and exported as an empty `actions` entry), while the equivalent mistake in `topicPerms`/
`groupPerms` is rejected with `400 topicPerms[0] must use non-blank resource=permission format`.
Surrounding whitespace around a real value was stored verbatim too (`" DENY "`), unlike the
per-resource entries, which are trimmed by `splitPerm`.

Both defaults now go through the same guard as their siblings:

- `MybatisPlusAclRepository.java:218-219` normalizes `defaultTopicPerm` / `defaultGroupPerm` before
  anything is read or written, so the request fails before the account or its rules are touched;
- the new `normalizeDefaultPermission` (line 432) rejects a blank value with
  `400 defaultTopicPerm must be a non-blank permission` and trims a real one, mirroring
  `validatePermissionEntries`;
- an absent default (`null`, i.e. the field omitted) keeps meaning "no default permission" and still
  writes no rule, and the write response now echoes the trimmed value, as it already does for the
  white remote address.

### How Did You Test This Change?

New tests in `MybatisPlusAclRepositoryTest`:
`upsertShouldRejectBlankDefaultTopicPermissionBeforeMutatingAccount`,
`upsertShouldRejectBlankDefaultGroupPermissionBeforeMutatingAccount` (each asserts the `400` message
and `verifyNoInteractions(userMapper, ruleMapper)`) and
`upsertShouldTrimDefaultPermissionsBeforePersisting` (asserts the two persisted rules carry
`DENY` / `PUB`).

Before the fix (red):

```
[...]

PR #4669, opened 2026-09-21
fix(acl): match the ACL search terms literally instead of as LIKE wildcards

yyqdbngt:
### Which Issue(s) This PR Fixes

- Fixes #4229

### Brief Description

`MybatisPlusAclRepository` bound the raw search term as a `LIKE` pattern: the user search bound
`username` / `access_key` and the rule query bound `principal` / `resource` with the term as typed.
A term such as `_` or `%` was therefore read as a wildcard and matched every row, and a literal
term containing either character could never be found. The term is now escaped (`\`, `%`, `_`)
before it is bound, using a private helper local to the file and mirroring the `escapeLike` helper
already used by `QueryHistoryService`.

### How Did You Test This Change?

Two new `MybatisPlusAclRepositoryTest` cases capture the `QueryWrapper` handed to the mapper, call
`getSqlSegment()` before reading `getParamNameValuePairs()` (the parameter map is only materialised
once the SQL segment is built) and assert the bound parameters.

Red - on the unmodified base:

```
$ cd server && mvn -B -ntp test -Dtest=MybatisPlusAclRepositoryTest
[ERROR] Tests run: 25, Failures: 2, Errors: 0, Skipped: 0
[ERROR]   MybatisPlusAclRepositoryTest.findUserPageShouldEscapeLikeWildcardsInTheSearchTerm:189
Expecting ArrayList:
  ["%svc_a%%", "%svc_a%%"]
to contain only:
  ["%svc\_a\%%"]
[ERROR]   MybatisPlusAclRepositoryTest.findRulePageShouldEscapeLikeWildcardsInTheFilterTerms:206
Expecting ArrayList:
  ["%_orders%", "%svc_a%%"]
to contain only:
  ["%svc\_a\%%", "%\_orders%"]
```

Green - after the fix, including the pre-existing cases of the class and checkstyle (0 violations):

```
$ cd server && mvn -B -ntp test -Dtest=MybatisPlusAclRepositoryTest
[INFO] You have 0 Checkstyle violations.
[INFO] Tests run: 25, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

The whole ACL package, to show the neighbouring classes are unaffected:

```
$ cd server && mvn -B -ntp test -Dtest='org.apache.rocketmq.studio.instance.acl.*Test'
[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0 -- AclControllerTest
[...]

PR #4720, opened 2026-09-21
fix(acl): stop fabricating null permissions in the plain access view

yyqdbngt:
### Related issue

No open issue covers this; found while reading the plain-access read-back path for the campaign.

### Brief Description

`MybatisPlusAclRepository.toPlainAccessConfig` (line 366) renders every `Topic`/`Group` rule of an
account as `resource + "=" + actions` without checking that the rule still has an action list:

```java
String actions = joinNormalizedCsv(rule.getActions());
if ("Topic".equals(rule.getResourceType())) {
    topicPerms.add(rule.getResource() + "=" + actions);
```

`joinNormalizedCsv` returns `null` for a null or blank action list, so such a rule is reported as the
literal entry `"orders=null"`. That leaks into `GET /api/acl/cluster-config?clusterId=...` (the
`accounts[].topicPerms` / `accounts[].groupPerms` the ACL page renders) and it round-trips badly:
`validatePermissionEntries` only rejects a *blank* right-hand side, so copying that read-back value
back through `POST /api/acl/plain-access-config` persists `null` as the actual permission
(`plainRule(..., "null", ...)`).

A rule can lose its action list in normal use: `AclService.updateRule` -> `replaceRule` clears the
`actions` column explicitly (the `clearColumn` path), and `POST /api/acl/rules/create` accepts a
payload whose `actions` is omitted (only `principal` and `resource` are required). The same loop
already treats an absent action list as "not configured" for `DEFAULT_TOPIC` / `DEFAULT_GROUP`, so
the per-resource branches now do the same and skip an entry that carries no permission.

Disclosure: `MybatisPlusAclRepository.java` is also touched by #4669 (LIKE escaping in `ruleQuery`)
and #4703 (default-permission validation in `createAndUpdatePlainAccessConfig`); this is a different
defect in a different method (`toPlainAccessConfig`), and the two hunks do not overlap.

New tests in `MybatisPlusAclRepositoryTest`:
`examineShouldSkipTopicPermissionsWhoseActionListIsEmpty` and
`examineShouldSkipGroupPermissionsWhoseActionListIsBlank`. Each seeds one rule without a usable
action list plus one granted rule and asserts the fabricated entry is gone.

Before the fix (red):

```
$ cd server && mvn -B -ntp test -Dtest='MybatisPlusAclRepositoryTest#examineShouldSkipTopicPermissionsWhoseActionListIsEmpty+examineShouldSkipGroupPermissionsWhoseActionListIsBlank'
[ERROR] Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 2.614 s <<< FAILURE! -- in org.apache.rocketmq.studio.instance.acl.MybatisPlusAclRepositoryTest
org.apache.rocketmq.studio.instance.acl.MybatisPlusAclRepositoryTest.examineShouldSkipGroupPermissionsWhoseActionListIsBlank -- Time elapsed: 2.520 s <<< FAILURE!
Expecting actual:
  ["cg-order=null", "cg-payment=SUB"]
to contain exactly (and in same order):
  ["cg-payment=SUB"]
but some elements were not expected:
  ["cg-order=null"]
	at org.apache.rocketmq.studio.instance.acl.MybatisPlusAclRepositoryTest.examineShouldSkipGroupPermissionsWhoseActionListIsBlank(MybatisPlusAclRepositoryTest.java:534)

org.apache.rocketmq.studio.instance.acl.MybatisPlusAclRepositoryTest.examineShouldSkipTopicPermissionsWhoseActionListIsEmpty -- Time elapsed: 0.025 s <<< FAILURE!
Expecting actual:
  ["orders=null", "payments=SUB"]
to contain exactly (and in same order):
  ["payments=SUB"]
but some elements were not expected:
  ["orders=null"]
	at org.apache.rocketmq.studio.instance.acl.MybatisPlusAclRepositoryTest.examineShouldSkipTopicPermissionsWhoseActionListIsEmpty(MybatisPlusAclRepositoryTest.java:520)

[ERROR] Tests run: 2, Failures: 2, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```

After the fix (green) — every ACL test class, so the neighbouring rule, user and plain-access tests
are covered too:

```
$ cd server && mvn -B -ntp test -Dtest='MybatisPlusAclRepositoryTest,AclServiceTest,AclControllerTest,ApacheAclReadServiceTest'
[INFO] Tests run: 27, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.instance.acl.AclControllerTest
[INFO] Tests run: 78, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.instance.acl.AclServiceTest
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.instance.acl.ApacheAclReadServiceTest
[INFO] Tests run: 25, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.instance.acl.MybatisPlusAclRepositoryTest
[INFO] Tests run: 133, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### Checklist

- [x] Documentation touched where behaviour changed (no doc change needed: the endpoint already documents the accounts view as stored `resource=permission` entries)


RockteMQ-AI (commented):
### Summary
This PR fixes a bug in the plain access config view where rules with cleared action lists would produce permission entries containing the literal string "null", which would then be re-persisted on write. The fix skips adding permission entries when the action list is empty, preventing the fabrication of phantom permissions.

### Findings

- **[Info]** `MybatisPlusAclRepository.java:367-378` — The `StringUtils.hasText(actions)` guard correctly prevents concatenating null/empty action strings into permission entries. The fix is applied consistently to both Topic and Group resource types.

- **[Info]** The inline comment explains the rationale clearly: a cleared action list should not produce a permission entry, and the `DEFAULT_*` branches already handle the "not configured" case.

- **[Info]** `MybatisPlusAclRepositoryTest.java` — The new test `examineShouldSkipTopicPermissionsWhoseActionListIsEmpty` directly validates the fix by setting up a rule with an empty action list and asserting that no permission entry is fabricated. Good coverage.

### Suggestions
- Consider whether the same guard should also apply to the `Cluster` resource type branches (lines 379+). If a cluster rule can also have an empty action list, it would benefit from the same protection. This may be out of scope for this PR but worth a follow-up check.

Clean and well-tested fix.

PR #4725, opened 2026-09-21
docs: align the instance section of the API spec with InstanceType

Frun1na:

- Fixes #4726

### Brief Description

§3 of the spec documents an instance type the API does not have:

- §3.1 offers `PROXY` as a `type` filter value ("全部 Proxy") and as a response value ("兼容值"),
  and §3.2 / §3.3 promise that "旧 `PROXY` 请求归一为 `PROXY_CLUSTER`".
- `InstanceType` only defines `CLOUD`, `PROXY_LOCAL`, `PROXY_CLUSTER` and `DIRECT`. Nothing
  normalises `PROXY`: the list filter binds the query parameter directly to the enum
  (`InstanceController.listInstances`), and `CreateInstanceDTO` / `UpdateInstanceDTO` bind the body
  field the same way. `GlobalExceptionHandler` maps
  `MethodArgumentTypeMismatchException` to HTTP 400, so every request following the spec — filter,
  create or update — is rejected before any service code runs.
- Hand-written creation also rejects `CLOUD` (`createApacheInstance`: "CLOUD type is reserved for
  vendor-managed instances"), which the §3.2 note now says instead.

§3.1 also named the timestamps `createdAt` / `updatedAt` and typed `id` as a string.
`InstanceVO` extends `BaseEntity` (`Long id`, `gmtCreate`, `gmtModified`) and the web client reads
those names (`web/src/api/instance.ts`).

Documentation only, so the Maven and Vitest suites were not re-run. The claims were read off the
implementation:

```
$ cat .../common/domain/enums/InstanceType.java
public enum InstanceType {
    CLOUD,
    PROXY_LOCAL,
    PROXY_CLUSTER,
    DIRECT;
}

$ grep -n "PROXY" .../instance/*.java        # no normalisation anywhere:
    (only the enum constant definitions and InstanceType.CLOUD / PROXY_CLUSTER comparisons)

$ cat .../common/domain/BaseEntity.java
    private Long id;
    private LocalDateTime gmtCreate;
    private LocalDateTime gmtModified;

# the handler that answers a typo'd enum value:
$ grep -n -A3 "MethodArgumentTypeMismatchException" .../common/exception/GlobalExceptionHandler.java
    @ExceptionHandler({MethodArgumentTypeMismatchException.class, MissingServletRequestParameterException.class})
    @ResponseStatus(HttpStatus.BAD_REQUEST)
    public Result<?> handleBadRequestParameterException(Exception ex)

# and the field names the frontend actually reads:
$ grep -n "id:\|gmtCreate\|gmtModified" web/src/api/instance.ts
    id: number;
    gmtCreate: string;
    gmtModified: string;
```

`InstanceService.createApacheInstance` also rejects `CLOUD` for hand-made instances
(`"CLOUD type is reserved for vendor-managed instances"`), which is what the rewritten §3.2 note now
documents instead of the non-existent `PROXY` compatibility shim.

### Checklist

- [x] Commit subject follows Conventional Commits (`docs:`)
- [ ] Tests added or updated for non-trivial changes, test methods named `...Test`


RockteMQ-AI (approved):
LGTM — docs accuracy improvement, API spec aligned with InstanceType.

---

PR #3349, opened 2026-09-05
fix(aliyun): populate subscription filter mode like the other providers

unbridled-41:
Fixes #3360.

### Problem / Evidence

On an Aliyun instance, the consumer-group subscription table renders an **empty 订阅模式 (filter mode) cell for every row**: the red test showed `expected: "SQL" but was: null` (and `expected: "TAG" but was: null`). Apache instances show `TAG`/`SQL` (`RocketMQMetadataProvider` line 691 via its `filterMode` normalizer) and Tencent instances show the expression type (`TencentInstanceProvider` line 984), so the same column is blank only for Aliyun.

Trigger path: consumer group page subscriptions sub-table (web/src/pages/instance/consumer.tsx:1038, renders `filterMode`) → `GET /api/groups/{name}/subscriptions` → `MetadataService.getGroupSubscriptions` → `AliyunInstanceProvider.getGroupSubscriptions` (line 413) → `AliyunConverters.toSubscriptionEntry` (line 210) — which maps `expression` and `type` from the SDK model but never derives `filterMode`, leaving the VO field null.

### Root cause / Fix

Field-mapping gap: `toSubscriptionEntry` skips the `filterMode` builder field. The data to derive it is already consumed — `data.getFilterExpressionType()`.

Fix: derive `filterMode` with the same mapping the Apache provider uses (`SQL92` → `SQL`, `CLASS_FILTER` → `CLASS_FILTER`, otherwise `TAG`), so the three providers render identical values for identical subscription types.

### Priority & scoring

- Impact 28/40 — a visible column of a core diagnostics table is empty for every Aliyun instance.
- Scope 10/20 — one column, one provider; page still usable.
- Reproducibility 20/20 — deterministic for all Aliyun subscription rows.
- Maintenance value 14/20 — cross-provider consistency with an established normalization pattern.
- **PRIORITY = 72, FIX_CONFIDENCE = 92** (≥70/≥80 per the contribution bar).

### Tests

- New `AliyunConvertersTest` cases: `toSubscriptionEntryShouldDeriveFilterModeFromTheExpressionType` (SQL92 → SQL) and `toSubscriptionEntryShouldKeepTagFilterMode` (TAG → TAG). Both failed before the fix with `null` and pass after.
[...]

PR #4732, opened 2026-09-21
fix(consumer): normalize and render the subscription filter modes the API returns

Frun1na:

- Fixes #4733

### Brief Description

The subscription tables on the consumer page still matched `filterMode` against the pre-#3349
display strings (`全量` / `Tag 过滤` / `SQL92 过滤`), while the API has long since normalized
the broker expression types to `TAG` / `SQL` / `CLASS_FILTER`. Every row therefore missed the color
map and echoed the raw code into the table:

```tsx
const colorMap: Record<string, string> = {
  全量: 'default',
  'Tag 过滤': 'blue',
  'SQL92 过滤': 'purple',
};
return <Tag color={colorMap[mode] || 'default'}>{mode}</Tag>;   // mode === 'TAG' | 'SQL' | 'CLASS_FILTER'
```

The i18n keys for the labels already existed (`consumer.filterTag`, `consumer.filterSql92`) but were
never referenced. This maps the three API values to those labels, adds a
`consumer.filterClassFilter` entry for the third value, and keeps unknown values visible as-is.

**Review follow-up (all four blocking items addressed):**

1. **Tencent normalization is now part of this PR.** `TencentInstanceProvider` was the third
   producer of `filterMode` and bypassed the helper: it built the value directly from
   `subscription.getExpressionType()`, so a Tencent SQL subscription still reached the table as the
   raw string `SQL92`. `toSubscriptionEntry` now routes through
   `SubscriptionFilterModes.fromExpressionType` and a provider test pins the mapping (`SQL92` →
   `type=SQL92`, `filterMode=SQL`). The helper's javadoc now says "Apache, Aliyun and Tencent".
   The description's original "both providers have normalized" claim was wrong — there are three
   producers, and this PR now normalizes all three, so the title no longer says `fix(web)`.
2. **§6.4 `type` row corrected.** All three producers set `type` from the expression type
   (`sd.getExpressionType()` / `data.getFilterExpressionType()` /
   `subscription.getExpressionType()`), not from a message type. The row now documents
   `TAG` / `SQL92` / `CLASS_FILTER` and says the value is passed through as-is.
3. **`consistency` value domain widened.** Apache emits `consistent` or `null`; Aliyun and Tencent
   emit `String.valueOf(<vendor consistency value>)` (e.g. Aliyun sends `true` / `false`). The row
   now documents both, instead of swapping an incorrect domain for an incomplete one.
4. **Mock fixtures migrated.** `web/src/mock/consumers.ts` typed `filterMode` as the pre-#3349
   display strings and carried `consistency: '一致'` rows. The union types and all fixture rows now
   use the API values (`TAG` / `SQL` / `CLASS_FILTER`, `consistent` / `inconsistent`), and the
   fixture `type` values follow the expression-type domain too, so mock mode no longer degrades the
   tag colors. The same migration was applied to the subscription fixtures inside
   `ConsumerPage.test.tsx`, which had been mocking the old contract.

`consumer.filterAll` remains unreferenced on purpose (it was already unreferenced on trunk; left
for a separate cleanup).

```
$ mvn -B -ntp test -Dtest='TencentInstanceProviderTest#getGroupProgressAndSubscriptionsShouldMapSubscriptionDataTest' -Dspring.profiles.active=dev
# before the fix: expected: "SQL" but was: "SQL92"  (mutation-sensitive, fails on base)
# after the fix:  Tests run: 1, Failures: 0, Errors: 0, Skipped: 0

$ npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx
 Test Files  1 passed (1)
      Tests  37 passed (37)

$ npx vitest run src/services/consumerService.test.ts
 Test Files  1 passed (1)
      Tests  17 passed (17)

$ npx tsc --noEmit            # no output, exit 0
$ npx eslint src/mock/consumers.ts src/pages/instance/__tests__/ConsumerPage.test.tsx
# no findings
```

The new provider test asserts the Tencent mapping (`type` keeps the raw `SQL92`, `filterMode`
becomes `SQL`) and fails on the base revision with `expected: "SQL" but was: "SQL92"`. The
ConsumerPage suite covers the label rendering (added earlier in this PR) and the migrated fixtures.

### Checklist


RockteMQ-AI (approved):
### Summary
Fixes subscription filter mode rendering to use API enum codes (`TAG`/`SQL`/`CLASS_FILTER`) instead of raw localized strings. Adds i18n key for `CLASS_FILTER`, updates API spec to match actual enum values, and includes a regression test.

LGTM — correct alignment between API contract and UI rendering.

---

RockteMQ-AI (approved):
### Summary
Fixes subscription filter mode rendering to use API enum codes (TAG/SQL/CLASS_FILTER) with proper i18n labels. Updates API spec and includes regression test.

LGTM

---

lizhimins (changes requested):
Thanks — the core defect is real, trunk is broken today, and the new test is properly mutation-sensitive.

**What we verified.** The subscription-mode column in `web/src/pages/instance/consumer.tsx` keys `colorMap` on `全量` / `Tag 过滤` / `SQL92 过滤` and renders `<Tag color={colorMap[mode] || 'default'}>{mode}</Tag>`, i.e. the raw value straight into the cell. But `RocketMQMetadataProvider` and `AliyunConverters` both build `filterMode` through `SubscriptionFilterModes.fromExpressionType`, which returns `TAG`, `SQL` or `CLASS_FILTER`. So on every Apache and Aliyun instance each row falls through to `'default'` and echoes the raw code to the user. The localized labels were already there and never used: `consumer.filterAll`, `consumer.filterTag` and `consumer.filterSql92` exist in `translations.ts` with zero references from `consumer.tsx`. Your new `consumer.filterClassFilter` key carries both zh and en. Reverting the `colorMap` change makes the new test fail on both `findByText('Tag 过滤')` and the `queryByText('TAG')` assertion, and the existing `consistency`-column assertions are unaffected since they check different tags. You are also right that no tool-catalog change is needed: `group.yaml` declares `filterMode` and `consistency` as plain `type: string` with no enum, and `GroupDetailOutput.Subscription` is `@JsonInclude(NON_NULL)`, so a null `consistency` is omitted rather than sent and `ToolSchemaValidator.validateOutput` is untouched. No `make catalog-generate` required.

**Blocking items.**

1. **There are three producers of `filterMode`, not two, and the third is not normalized.** `TencentInstanceProvider` builds it directly from `subscription.getExpressionType()` without going through `SubscriptionFilterModes`. That helper maps `SQL92` to `SQL`, keeps `CLASS_FILTER`, and defaults everything else to `TAG` — so a Tencent SQL subscription reaches your new `meta` table as the raw string `SQL92`, falls through to the raw-value branch, and still renders as a bare code. Tencent instances therefore keep the exact bug you set out to fix. This also contradicts the helper's own javadoc, which says it exists so "consumer-group subscription tables render the same values across instance types". Please correct the description's "both providers have normalized" either way. Fixing it properly means routing the Tencent path through `SubscriptionFilterModes.fromExpressionType` and adding a provider test, which is a server-side change — so please either take that on (and retitle, since `fix(web)` would no longer describe the PR) or explicitly scope this PR to the web layer and state in the description that Tencent instances still render raw codes, with a follow-up issue filed. We need your answer on this one before anything else, because it decides whether this stays a frontend PR; it is also the reason this cannot be finished for you as a maintainer touch-up.

2. **The `type` row in `docs/api-spec.md` §6.4 is wrong too, and it is the row directly above the two you changed.** It documents `NORMAL / FIFO / DELAY / TRANSACTION`, but all three producers set `type` from the expression type: `RocketMQMetadataProvider` uses `sd.getExpressionType()`, `AliyunConverters` uses `data.getFilterExpressionType()`, and `TencentInstanceProvider` uses `subscription.getExpressionType()`. Since you are already correcting that table, please fix this row in the same pass.

3. **The `consistency` value domain you documented only covers the Apache provider.** Your replacement row reads `consistent`, or `null` when it cannot be determined. That matches `RocketMQMetadataProvider`, which emits `"consistent"` or `null`. But `AliyunConverters` and `TencentInstanceProvider` both emit `String.valueOf(<vendor consistency value>)`, which is not bounded to that domain. Swapping an incorrect domain for an incomplete one leaves the spec wrong, so please document the vendor-provided case as well. For what it is worth, your finding that nothing emits `一致` / `不一致` is confirmed — `git grep '一致' -- server/src/main/java` returns nothing.

4. **The mock fixtures were not updated.** `web/src/mock/consumers.ts` still types `filterMode` as `'Tag 过滤' | 'SQL92 过滤' | '全量'`, 23 fixture rows still use those strings, and several carry `consistency: '一致'`. Mock mode is reachable: `stores/dataModeStore.ts` derives the default from `VITE_USE_MOCK` and can be overridden by the persisted mode, and `services/consumerService.ts` imports these mocks directly. After your change those rows fall through to the raw-value branch — the displayed text is unchanged, but the blue/purple tag colours degrade to `default`, which is a visible regression introduced by this PR. Please migrate the union type and the fixtures to `TAG` / `SQL` / `CLASS_FILTER`.

**Optional, not blocking.** `consumer.filterAll` is still unreferenced after this change, but it was already unreferenced on trunk, so it is not a regression — leave it, or drop it in a separate cleanup. Separately, `consumer.tsx` is one of the most heavily touched files in the open queue, so we will re-diff this PR after the others land; expect to rebase, but there is nothing for you to do about that now.

Once items 1 to 4 are addressed we can merge. Please answer item 1 first — the rest are mechanical once you have decided whether the Tencent normalization belongs in this PR or in a follow-up.

Frun1na:
Thanks for the thorough review — all four blocking items are now addressed in e60e99c5.

**Item 1 — answer: we took option (a).** The Tencent path is now part of this PR:
`TencentInstanceProvider.toSubscriptionEntry` routes `filterMode` through
`SubscriptionFilterModes.fromExpressionType`, with a provider test pinning `SQL92` → `type=SQL92`,
`filterMode=SQL`. The test fails on the base revision with `expected: "SQL" but was: "SQL92"`. The
helper's javadoc now says "Apache, Aliyun and Tencent", and the description no longer claims "both
providers have normalized" — it lists three producers and states all three are normalized here. The
title was retitled accordingly (`fix(consumer): ...`).

**Item 2 —** the §6.4 `type` row now documents `TAG` / `SQL92` / `CLASS_FILTER` as the raw
expression type each producer passes through, replacing `NORMAL` / `FIFO` / `DELAY` / `TRANSACTION`.

**Item 3 —** the `consistency` row now documents the Apache `consistent`/`null` domain plus the
vendor-provided string forms (e.g. Aliyun's `true` / `false` from `String.valueOf`).

**Item 4 —** `web/src/mock/consumers.ts` union types and all fixture rows were migrated to the API
values (`TAG` / `SQL` / `CLASS_FILTER`, `consistent` / `inconsistent`), and the fixture `type`
values follow the expression-type domain, so mock mode no longer degrades the tag colors. The same
migration was applied to the subscription fixtures in `ConsumerPage.test.tsx`.

`consumer.filterAll` was left unreferenced as suggested. Verification: the Tencent provider test
(mutation-sensitive, red on base), ConsumerPage 37/37, consumerService 17/17, `tsc --noEmit` and
eslint clean.

Issue #4736, opened 2026-09-21
[Studio][Bug] Tool playground renders an in-flight result under the tool selected after it


unbridled-41:
### Studio Version

branch: rocketmq-studio
git commit id: 7ce9a68215c4cd1c7025013d02135b43b6aa6ba1

### Problem

The manual tool playground keeps an in-flight execution alive across a tool or scope change, so its response lands in the output pane of whatever tool is selected afterwards. The pane never names the tool that produced the output, so tool A's result is presented as tool B's.

Chain: `handleExecuteTool` (`web/src/pages/ai/components/ToolPlaygroundModal.tsx:200-225`) awaits `executeTool(selectedToolName, parsedInput, selectedClusterId)` and assigns the response to `toolResult` with no generation check. Neither invalidation path touches that call:

- `selectTool` (`:127-135`) clears the panel (`setToolResult(undefined)`) but does not invalidate the running request, and the tool `Select` (`:267-281`) stays enabled while `toolExecuting` is true (only the Run button goes into `loading`, `:247-250`).
- `loadTools` (`:137-161`, reached by a cluster-scope change, `:191-198`) clears the panel the same way.
- `handleClose` (`:227-233`) bumps only `toolLoadRequestRef` — the catalog guard — so a response can also arrive after the modal was closed and reopened (the modal has no `destroyOnHidden`, so its state survives).

Trigger (deterministic): open 工具 on the AI page, run any tool that takes a few seconds (the RocketMQ tools do), then switch the tool in the selector before it returns. The panel was cleared by the switch; when the response arrives it is repopulated, the success toast fires, and the output is shown under the newly selected tool's name and risk/permission tags (`:283-302`).

### Evidence

- `web/src/pages/ai/components/ToolPlaygroundModal.tsx:200-225` — `setToolResult(await executeTool(...))` with no request-generation check; the only guard is `if (!selectedToolName || toolExecuting) return` at the top, which prevents a second *start*, not a late *landing*.
- `web/src/pages/ai/components/ToolPlaygroundModal.tsx:127-135` and `:137-161` — the two paths that clear `toolResult` without invalidating the execute.
- `web/src/pages/ai/components/ToolPlaygroundModal.tsx:227-233` — `handleClose` bumps `toolLoadRequestRef` only.
[...]

Issue #4632, opened 2026-09-20
[Bug] AI page can become send-ready after the LLM runtime was disabled (late responses repopulate useLlmRuntime state)


Frun1na:
### Bug Description

`useLlmRuntime.load` (web/src/pages/ai/hooks/useLlmRuntime.ts) has no staleness guard. When `enabled` flips to `false` while a load is in flight — e.g. mock mode is toggled, or an admin logs out into a reader account — the hook clears the state synchronously, but the already-in-flight `getLlmConfig`/`getLlmModels` responses still resolve afterwards and write `config`, `modelOptions` and `selectedModel` back. `llmReady` then comes back `true`, so the AI page offers sending on a runtime the hook's own contract says should show "no configuration visible". Overlapping `reload()` calls can likewise finish out of order and leave stale data.

### Steps to Reproduce

1. Open the AI page with `enabled = true` so a runtime load starts (the `/llm/config` + `/llm/models` requests are in flight).
2. Before those requests resolve, flip the runtime to disabled (mock mode, or downgrade the account).
3. When the in-flight requests resolve, `config`/`modelOptions`/`selectedModel` are repopulated and `llmReady` becomes `true` although `enabled` is `false`.

### Expected Behavior

A response from a superseded load never lands; the disabled runtime stays cleared and `llmReady` stays `false`.

### Environment

- `rocketmq-studio` branch web UI
- Reproduced with a unit test (`useLlmRuntime.test.ts`) in PR #4631

Proposed fix: #4631

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Under Review**

Thank you for reporting this issue. It has been classified as a potential bug and flagged for code-level verification.

**Next steps:**
- The issue will be verified against the current codebase
- If confirmed, a fix proposal (spec + PR) may be generated automatically
- You can reply with `/approve` to proceed, `/revise` to request changes, or `/reject` to decline

---
*Automated evaluation by RockteMQ-AI*

lizhimins:
Resolved by #4631 (merged into rocketmq-studio as 114656085).

Issue #4614, opened 2026-09-20
[Bug] AI run cannot be stopped after reconnecting to a live conversation


Frun1na:
**Studio Version**

branch: `rocketmq-studio`
git commit id: `5b761df8` (#4606)

**Runtime Environment**

UI issue, verified at the unit level (see below) on the current `rocketmq-studio` branch; browser-independent.

**Connected RocketMQ Cluster**

Not cluster-specific: the defect is in the AI conversation client state handling.

**Describe the Bug**

After a page reload or reconnect while an AI run is streaming, the client reattaches to the live run but can never stop it:

1. On attach, the server replays persisted events. The live event vocabulary starts at `run_started` — the replay projects a run row into no `run_started` live event, so the reattaching client never receives one.
2. `useAgentRun` only learns the run id from the `run_started` frame (`runIdRef.current = event.runId`), and `attach()` discards the run id its caller passes in.
3. The composer renders the stop state from `isStreaming`, so the button looks and behaves as clickable — but `stop()` returns immediately when the run id is null.
4. For the same reason `reportRunSpeed` is silently skipped on a reattached run.

`stop()`'s comment claims the button "stays disabled until then (`canStop`)", but `canStop` is not consumed by the composer's stop-state computation, so nothing disables the button.

**Steps to Reproduce**

1. Send a message in an AI conversation so a run starts streaming.
2. Reload the page while the run is still generating; the client reattaches and the transcript keeps updating.
3. Click the stop button and observe nothing happens; the run keeps streaming until it finishes on its own.

**What Did You Expect to See?**

Clicking stop on a reattached run stops the run (`POST` stop reaches the API with the run id) and the button enters the stopping state.

**What Did You See Instead?**

The stop call is a client-side no-op (`runIdRef` is null, `stop()` returns before the API call); streaming continues.

**Additional Context**

Reproduced at the unit level in `web/src/pages/ai/hooks/useAgentRun.test.ts`: a new test that attaches and then calls `stop()` fails on `5b761df8` (`canStop` stays false, `stopRun` never called). A fix is proposed in #4613: seed the hook's run id from the id `attach()` already receives.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

[...]

PR #4738, opened 2026-09-21
fix(ai): drop an in-flight tool result once another tool is selected

unbridled-41:
Fixes #4736.

### Problem

The manual tool playground kept an in-flight execution alive across a tool or scope change and rendered its response in the output pane of whatever tool was selected afterwards. The pane is labelled with the generic `ai.tool.output` and never names the tool that produced the text, so tool A's output was presented under tool B's name, risk level and permission tags.

Chain: `handleExecuteTool` (`web/src/pages/ai/components/ToolPlaygroundModal.tsx:200-225`) awaits `executeTool(selectedToolName, parsedInput, selectedClusterId)` and assigns the response to `toolResult` with no generation check. The paths that clear the panel do not invalidate that call:

- `selectTool` (`:127-135`) clears the panel but leaves the request running, and the tool `Select` (`:267-281`) stays enabled while executing — only the Run button takes `loading` (`:247-250`).
- `loadTools` (`:137-161`, reached from a cluster-scope change via `:191-198`) clears the panel the same way.
- `handleClose` (`:227-233`) bumps only `toolLoadRequestRef`, the catalog guard, so a response can also land after the modal was closed and reopened (no `destroyOnHidden`, so the state survives).

Trigger (deterministic): open 工具 on the AI page, run a tool that takes a few seconds (the RocketMQ tools do), switch the tool in the selector before it returns. The panel is cleared by the switch, then repopulated by the late response, with the success toast and the provider payload shown under the new tool.

### Root cause

The component guards its *catalog* load against a stale response (`toolLoadRequestRef`, documented at `:40-43`) but the execute path has no equivalent generation: `toolExecuting` prevents a second *start*, not a late *landing*.

### Fix

Add `toolExecuteRequestRef`, bumped by every path that changes what the panel belongs to (tool selection, catalog reload for a scope change, close) and captured before the await in `handleExecuteTool`. The result and its success toast are applied only when the request is still the current one; the `finally` still clears `toolExecuting` unconditionally, so the Run button can never stay stuck. A failed execution keeps reporting its error only while it is still the current request — a stale failure would otherwise interrupt the operator with an error about a tool they are no longer looking at.

### Priority

**68** = impact 24 + reach 10 + reproducibility 20 + maintenance value 14.

- Impact 24/40: misattributed tool output. The component is documented as the escape hatch for verifying what a transcript's tool call returns, so a result shown under the wrong tool is a false reproduction rather than a missing one.
- Reach 10/20: one dialog, but every execution slower than the operator's next click, in both real and mock mode.
- Reproducibility 20/20: deterministic — a deferred `executeTool` and one selector change.
- Maintenance value 14/20: completes the stale-guard pattern the same file already documents for the catalog load.

`FIX_CONFIDENCE`: **90** — the failing test and the sibling guard fix the expected behavior; the change is a ref, four increments and one comparison.

### Tests

Red (source at base commit `7ce9a682`, all four regression tests fail):

```
npx vitest run src/pages/ai/__tests__/AiPage.test.tsx -t "dropsAnInFlight|reportsAFailure"
    × dropsAnInFlightToolResultWhenAnotherToolIsSelectedTest
    × dropsAnInFlightToolResultWhenTheClusterScopeChangesTest
    × dropsAnInFlightToolResultAfterTheModalIsClosedTest
    × reportsAFailureOnlyWhileItsToolIsStillSelectedTest
Error: expect(element).not.toBeInTheDocument()
expected document not to contain element, found <pre
```

Green (this branch, two commits):

```
npx vitest run src/pages/ai/__tests__/AiPage.test.tsx          → 19 passed (19)
npx vitest run src/pages/ai src/api/ai                         → 18 files, 253 passed (253)
npx vitest run --maxWorkers=4                                  → 134 files, 1239 passed (1239)
npx tsc -b                                                     → clean
npx eslint <changed files>                                     → clean
npm run build                                                  → ✓ built in 9.07s
```

Coverage of the fix's paths (one test each, all four red on the base commit):

- **tool switch** — the deferred response resolves after the operator selected another tool: no result, no success toast, and the Run button leaves its loading state.
- **cluster scope change** — the catalog reloads and re-selects a tool, then the old scope's response arrives: no result.
- **close and reopen** — the dialog keeps its state across a close (no `destroyOnHidden`), so a response arriving after the close must not be waiting in the panel.
- **superseded failure** — the failed call reports nothing once its tool is no longer selected, with an in-test control proving the same failure *is* reported while its tool is still selected.

The pre-existing `opensTheToolPlaygroundLoadsTheCatalogAndExecutesAToolTest` still pins the happy path (the result does render while the tool stays selected).

### Risk

Low. Only an execution whose tool/scope changed (or whose modal was closed) is discarded; a normal run is untouched, and the loading state is cleared unconditionally so no button can stick. Behavior worth naming: a request that is superseded mid-flight no longer reports its outcome at all — deliberately, because the discarded request's error would describe a call the operator is no longer looking at.

### Duplicate check

Searched open and closed issues and pull requests (`tool playground`, `playground`, `stale tool result`): no issue or PR covers the playground's execute path. The two open AI-lifecycle reports, #4614 and #4632, are about stop-after-reconnect and runtime repopulation after a disable, not this dialog.

### Base branch

Targets `rocketmq-studio`, the development trunk. Note the repository's default branch is still `master`; GitHub interprets the closing keyword only for PRs that target the default branch, so the `Fixes` link above does not auto-close #4736 when this merges — the issue needs to be closed by hand.


unbridled-41:
### Evidence (recorded at the PR head)

1. **Head / diff self-check** — `fix/tool-playground-stale-result` @ `c41caac2`, two commits (audit follow-up added the scope, close and superseded-failure tests). `git show --numstat` = `4889a28d`: `ToolPlaygroundModal.tsx 15/2`, `AiPage.test.tsx 54/0`; `c41caac2`: `AiPage.test.tsx 157/1`. No unrelated file touched.
2. **Red on base** — with only the tests added on `7ce9a682`: `npx vitest run src/pages/ai/__tests__/AiPage.test.tsx -t "dropsAnInFlight|reportsAFailure"` → all four regression tests fail (`× dropsAnInFlightToolResultWhenAnotherToolIsSelectedTest`, `× …WhenTheClusterScopeChangesTest`, `× …AfterTheModalIsClosedTest`, `× reportsAFailureOnlyWhileItsToolIsStillSelectedTest`), each on `expect(element).not.toBeInTheDocument()` / `found <pre`. Re-verified with the final test content after the follow-up commit.
3. **Green on branch** — `npx vitest run src/pages/ai/__tests__/AiPage.test.tsx` → `19 passed (19)`; `npx vitest run src/pages/ai src/api/ai` → `18 files, 253 passed (253)`.
4. **Full suite on branch** — `npx vitest run --maxWorkers=4` → `Test Files 134 passed (134)`, `Tests 1239 passed (1239)` (measured alone, no competing load). Trunk alone is 1235, so the delta is this branch's four regression tests.
5. **Build / types / lint on branch** — `npx tsc -b` clean, `npx eslint <the 2 changed files>` clean, `npm run build` → `✓ built in 9.07s` (after the follow-up commit).
6. **CI** — no workflow run was created for this head SHA; the repository's CI workflow fails to start for every PR at the moment (newest runs repo-wide are `CI pull_request startup_failure`, e.g. 35576031888 @ 2026-09-21T08:04:50Z). The local results above are the evidence.

Duplicate check before submission: no open PR or issue covers the playground's execute path (the two open AI-lifecycle reports, #4614 and #4632, are about stop-after-reconnect and runtime repopulation, not this dialog).

RockteMQ-AI (approved):
﻿### Summary
Fixes a bug where an in-flight tool execution result could land in the output pane of a different tool (or a reopened modal) because the execute path had no generation counter to invalidate stale responses. The fix adds `toolExecuteRequestRef` — bumped on tool selection, scope change, and close — and captures it before the await. Results and error toasts are only applied when the request is still current.

Follows the same pattern as the existing `toolLoadRequestRef` for catalog loads. Good test coverage including tool switch, scope change, and modal close scenarios.

---

Issue #4739, opened 2026-09-21
[Studio][Bug] A refused send clears the composer draft instead of giving it back


unbridled-41:
### Studio Version

branch: rocketmq-studio
git commit id: 7ce9a68215c4cd1c7025013d02135b43b6aa6ba1

### Problem

A send that the server refuses clears the composer and never gives the draft back. The composer clears its own draft on send and documents that the caller must restore it when the send is rejected ("no conversation yet, 409 from the server"), but the AI page only restores it on the *create conversation* path: on an existing conversation the send is fired without its outcome ever reaching the caller, so a 409 (another run in flight), a provider rejection, or a stream that dies before its first frame leaves the operator with an empty composer.

Chain:

- `Composer.handleSend` (`web/src/pages/ai/components/Composer.tsx:256-261`) calls `onSend(text)` and then `onChange('')`, clearing the draft (and its sessionStorage copy). Its own contract at `:108-113`: "The composer clears the draft itself, so a caller that has to reject the send (no conversation yet, 409 from the server) should put the text back with `onChange`".
- `AiPage.handleSend` (`web/src/pages/ai/index.tsx:119-124`) restores the text only `if (target === null)`, with the comment "a rejected send gives the cleared draft back".
- `useAiSend.startRun` (`web/src/pages/ai/hooks/useAiSend.ts:108-111`) used `void optionsRef.current.send(current, request); return current;` for an existing conversation — the id is returned whether or not the send was admitted, so `target` was never null and the restore could not run.
- `useAgentRun.startStream` (`web/src/pages/ai/hooks/useAgentRun.ts:280-336`) already knows the difference: two guards return before opening anything, and a rejection from `open()` before any frame means the run never started — none of that was reported to the caller.
- The home-page handoff has the same shape (`web/src/pages/ai/hooks/useDraftHandoff.ts:122-123`): it applied an empty prompt after `startRun` regardless of the result.

Trigger (deterministic): on `/ai/c/{id}`, send while another tab (or an earlier request) holds a run in flight. The server answers 409, the page shows the error alert and the button returns to 发送, and the composer is empty — the prompt is gone even though it never left the browser. The same happens when the provider rejects the prompt or the stream fails before its first frame.

### Evidence

- `web/src/pages/ai/components/Composer.tsx:108-113` (documented contract) and `:256-261` (draft cleared on send).
- `web/src/pages/ai/index.tsx:119-124` — restore gated on `target === null`.
[...]

PR #603, opened 2026-07-26
[Studio] Normalize AI draft model

Aias00:
## Summary
- trim the AI chat draft model from navigation state before applying it
- drop empty model values after normalization so the chat page keeps its default selection
- add regression coverage for both cases

## Verification
- npm --prefix web test -- --run src/pages/ai/chatDraft.test.ts
- (cd web && npm exec eslint -- src/pages/ai/chatDraft.ts src/pages/ai/chatDraft.test.ts)
- npm --prefix web run build

PR #4740, opened 2026-09-21
fix(ai): give the composer draft back when the server refuses the send

unbridled-41:
Fixes #4739.

### Problem

A send the server refused cleared the composer and never gave the draft back. `Composer.handleSend` (`web/src/pages/ai/components/Composer.tsx:256-261`) calls `onSend(text)` and then `onChange('')`, and its own contract (`:108-113`) says the caller must restore the text when it has to reject the send — "no conversation yet, **409 from the server**". The AI page does restore it (`web/src/pages/ai/index.tsx:115,119-124`, "a rejected send gives the cleared draft back"), but only when `startRun` resolves `null`, and `startRun` resolved `null` only for a failed conversation *creation*: `useAiSend.startRun` (`web/src/pages/ai/hooks/useAiSend.ts:108-111`) used `void optionsRef.current.send(current, request); return current;` for an existing conversation, so the id came back whether or not the send was admitted. The home-page handoff had the same shape (`web/src/pages/ai/hooks/useDraftHandoff.ts:122-123`), clearing the applied prompt regardless of the result.

`useAgentRun.startStream` (`web/src/pages/ai/hooks/useAgentRun.ts:285-295`, `:318-336`) already had the information — two guards refuse before opening anything, and a rejection from `open()` with no frame delivered means the run never started — it just never reported it.

Trigger (deterministic): on `/ai/c/{id}`, send while another tab holds a run in flight. The server answers 409, the page shows the error alert and the button returns to 发送 — and the composer is empty, even though the prompt never left the browser. The same happens when the provider rejects the prompt or the stream fails before its first frame.

### Root cause

The send's admission result was not part of the send API: `send`/`startStream` returned `Promise<void>`, so "the server never started a run" and "the run started" were indistinguishable to callers, and `startRun` could only report creation failures.

### Fix

`useAgentRun.startStream` now resolves to whether the run was admitted — `false` for its two refusal guards and for a failure that produced no frame, `true` once the first event arrives (a mid-stream failure is an admitted run and must not resurrect the draft). `send` passes that through (`attach` keeps returning `void`), `useAiSend.startRun` returns `null` for a refused send on an existing conversation, and the handoff applies the emptied draft only when a target was resolved. `attach` and every existing caller keep their contracts; `ai/index.tsx` needed no change — its restore-on-null path now covers both reasons.

### Priority

**68** = impact 22 + reach 10 + reproducibility 20 + maintenance value 16.

- Impact 22/40: the operator retypes a prompt that never reached the server; the two send paths of the same page disagreed about the same failure, and the composer accepts long multi-line instructions.
- Reach 10/20: every refused send — a 409 from a concurrent run, a provider rejection, a connection that dies before the first frame.
- Reproducibility 20/20: deterministic — reject `openRunStream` with the existing 409 fixture.
- Maintenance value 16/20: removes an asymmetry the code itself documents ("a rejected send gives the cleared draft back"), and gives `send` the admission verdict its callers were guessing at.

`FIX_CONFIDENCE`: **88** — the composer's documented contract, the existing comment on the restore path and the DLQ-free unit fixtures fix the expected behavior; the change is a return value plus its two consumers.

### Tests

Red (source at base commit `7ce9a682`, the two regression tests):

```
npx vitest run src/pages/ai/__tests__/AiPage.test.tsx -t "givesTheDraftBack|keepsTheHandoffDraft"
    × givesTheDraftBackWhenTheServerRefusesTheSendTest
    × keepsTheHandoffDraftWhenTheSendIsRefusedTest
Expected the element to have value: 第二次发送   (received '')
Expected the element to have value: 检查集群状态  (received '')
```

Each has teeth independently: the handoff test was re-run with only that part of the fix reverted and failed again, then passed once restored. Both were re-confirmed red with the final test file, with the three source files restored to `7ce9a682` for the run.

Boundary pin (a third test, `keepsTheComposerClearedWhenAnAdmittedStreamFailsTest`): a stream that already delivered `run_started` + one `text_delta` and then fails must leave the composer **empty** — the send was admitted, so the draft is consumed. This one passes on the base commit as well (the base never restores anything); it is included so a later change cannot over-restore, and it is described that way in the test itself.

Green (this branch, `fix/restore-draft-on-refused-send` @ `7b83b288`, two commits):

```
npx vitest run src/pages/ai/__tests__/AiPage.test.tsx     → 18 passed (18)
npx vitest run src/pages/ai src/api/ai                    → 18 files, 255 passed (255)
npx vitest run --maxWorkers=4                             → 134 files, 1238 passed (1238)
npx tsc -b                                                → clean
npx eslint <the 5 changed files>                          → clean
npm run build                                             → ✓ built in 9.15s
```

(The full-suite run measured alone: a first attempt overlapped another vitest process and hit the repository's known load-sensitive `ConsumerPage > keeps the latest client stack…` case, which passes in isolation and in the clean run — recorded here because it is the same "flaky under load" behavior the earlier reports on this repository describe.)

`npx tsc -b` also caught five test fixtures declaring `send` as `Promise<void>` — exactly the type change this PR makes; they are updated in `useAgentRun.test.ts`.

### Risk

Low. Two behavior changes, both intended: a refused send no longer clears the composer, and a stream that dies before its first frame now reports as "not admitted" so the draft returns (a stream that delivered any frame keeps the composer cleared, even if it then fails). `attach` and the run lifecycle are untouched; the existing tests that pin the 409 toast and the double-submit guard still pass.

### Duplicate check

Searched open and closed issues and pull requests (`draft`, `composer`, `草稿`): no issue or PR covers draft restoration after a refused send — #603, the only other draft-related item, normalizes the AI draft model and is closed. #4614 and #4632 are the adjacent open AI-lifecycle reports.

### Base branch

Targets `rocketmq-studio`, the development trunk. Note the repository's default branch is still `master`; GitHub interprets the closing keyword only for PRs that target the default branch, so the `Fixes` link above does not auto-close #4739 when this merges — the issue needs to be closed by hand.

---

## Review follow-up (2026-09-21): admission decoupled from the stream, real refusal fixture — head `1e7a86e4`

**Blocking issue resolved** — the send no longer waits for the whole run:

- `useAgentRun.startStream` now returns an **admission promise** that settles on the first stream frame (`true`) or when the whole open lifecycle settled without a frame (`false` — the 409 `ai.run.busy` path, a route that moved on, or a transport failure with nothing delivered). The stream lifecycle itself continues inside the hook (its promise is awaited there, and the rejection is consumed) until the terminal frames and the timeline refetch settle — no fire-and-forget, no unhandled rejection, and the timeline/finalization semantics are untouched.
- `useDraftHandoff` (existing-conversation path): `startRun` now settles on **admission**, so an admitted handoff clears the composer immediately and strips the history entry's state immediately — on the previous head both statements after `await startRun` waited for the end of the answer, leaving a reload window in which the draft was re-entered with a fresh `consumedDraftRef` and re-sent. A refusal (null) restores the draft and still strips the state so a reload cannot replay it.
- The plain composer send in `ai/index.tsx` keeps its shape: `then((target) => target === null && setInputValue(text))` — restore only on refusal — and now fires at admission too.
- Both refusal fixtures use the real server contract: `code: 'ai.run.busy', status: 409` (`AiRunService.REFUSED_BUSY_CODE`), not the nonexistent `ai.run.in_flight`.

**New regression** (`clearsTheComposerAndHistoryStateImmediatelyWhenAnAdmittedHandoffIsStillStreamingTest`): the handoff onto an existing conversation with a stream that delivers `run_started` and then stays open — the composer must read empty and the history entry's state must be `null` **while the stream has not finished**.

- Red against the previous head's code (`await send` → resolve at terminal): the composer still carried `检查集群状态` (`Expected '' / Received '检查集群状态'`).
- Green with this change: AiPage **19/19**; the full `src/pages/ai` suite is **190/190** (16 files), including the retained boundary test `keepsTheComposerClearedWhenAnAdmittedStreamFailsTest`.
- Branch rebased onto trunk `4c697f07` (resolving `useAgentRun.ts` against #4613's `knownRunId` seeding — admission and knownRunId coexist: `send` seeds null, `attach` seeds the run id, and neither affects admission). `tsc -b` clean, eslint clean on touched files, `npm run build` ✓ (32.51s).


unbridled-41:
### Evidence (recorded at the PR head)

1. **Head / diff self-check** — `fix/restore-draft-on-refused-send` @ `7b83b288`, two commits. `git show --numstat`: `ed7f223f` = `AiPage.test.tsx 34/0`, `useAgentRun.test.ts 5/5`, `useAgentRun.ts 24/10`, `useAiSend.ts 9/6`, `useDraftHandoff.ts 7/2`; `7b83b288` = `AiPage.test.tsx 26/0`. The `useAgentRun.test.ts` edits are the five fixtures that declared `send` as `Promise<void>` — caught by `npx tsc -b`, which vitest does not run. No unrelated file touched.
2. **Red on base** — with only the test file changed on `7ce9a682`: `npx vitest run src/pages/ai/__tests__/AiPage.test.tsx -t "givesTheDraftBack|keepsTheHandoffDraft"` → `× givesTheDraftBackWhenTheServerRefusesTheSendTest` (expected value `第二次发送`, received `''`) and `× keepsTheHandoffDraftWhenTheSendIsRefusedTest` (expected `检查集群状态`, received `''`). Re-confirmed with the final test file, with the three source files restored to `7ce9a682` for that run.
3. **Teeth per fix** — the handoff case was re-run with only the `useDraftHandoff` part of the fix reverted: it failed again, and passed once restored (the composer case depends on the `useAgentRun`/`useAiSend` part alone). Neither half is covered by the other.
4. **Boundary pin (not a regression test)** — the third test, `keepsTheComposerClearedWhenAnAdmittedStreamFailsTest`, was also run against the three base source files and **passes there**: the base never restores anything. It pins the admitted-stream boundary so a later change cannot over-restore; the test comment states this explicitly rather than implying it was red.
5. **Green on branch** — `npx vitest run src/pages/ai/__tests__/AiPage.test.tsx` → `18 passed (18)`; `npx vitest run src/pages/ai src/api/ai` → `18 files, 255 passed (255)`.
6. **Full suite on branch** — `npx vitest run --maxWorkers=4` → `Test Files 134 passed (134)`, `Tests 1238 passed (1238)`. Trunk alone is 1235, so the delta is this branch's three tests. One earlier full run overlapped another vitest process and hit the repository's known load-sensitive `ConsumerPage > keeps the latest client stack when an older request resolves last` case; the clean run above, with no competing load, is all green, and that file is untouched by this change.
7. **Build / types / lint / CI** — `npx tsc -b` clean, `npx eslint <the 5 changed files>` clean, `npm run build` → `✓ built in 9.15s`. No workflow run was created for this head SHA; CI fails to start for every PR in this repository at the moment (`CI pull_request startup_failure`, e.g. 35576031888 @ 2026-09-21T08:04:50Z), so the local results above are the evidence.

Duplicate check before submission: searched open **and closed** issues and pull requests (`draft`, `composer`, `草稿`). No issue or PR covers draft restoration after a refused send — #603, the only other draft-related item, normalizes the AI draft model and is closed. The adjacent open AI reports #4614 and #4632 are about the run lifecycle (stop after reconnect, runtime repopulation after a disable).

RockteMQ-AI (approved):
﻿### Summary
Fixes a UX issue where a refused send (409, provider rejection, or pre-frame failure) cleared the composer draft permanently. The fix changes `startStream` to return `Promise<boolean>` indicating whether the run was admitted, allowing callers to restore the draft when the send was refused while keeping it cleared for mid-stream failures (which were admitted).

The API design is clean: `false` for refused/pre-frame failures, `true` once the first event arrives. The handoff and existing-conversation paths both benefit from this. Comprehensive test coverage including the boundary between refused and admitted-but-failed streams.

---

lizhimins (changes requested):

- The diagnosis is correct and the bug is real: `useAiSend.startRun` returns `current` unconditionally for an existing conversation (`web/src/pages/ai/hooks/useAiSend.ts:108-111`), so the restore-on-null path in `web/src/pages/ai/index.tsx:112-125` never fires when the server refuses the send (409 → `ai.run.busy`, `AiRunService.java:339`). The admission flag in `useAgentRun.startStream` is the right signal.
- Blocking issue: `await optionsRef.current.send(...)` makes `startRun` resolve only at the **end of the run**, because `useAgentRun.send` is documented and implemented to resolve "once the terminal frames were processed and the timeline refetch settled" (`useAgentRun.ts:110-116`, `await open(...)` + `finally { await finishStream(...) }`). That defers both statements after `await startRun(start)` in `useDraftHandoff.ts:122-127`:
- the applied prompt now lingers in the composer for the whole answer, which is exactly what the comment four lines above (`useDraftHandoff.ts:117-120`) says must not happen;
- the `state: null` replace-navigation is deferred too, so the history entry keeps carrying the draft. A reload during the run re-enters the handoff with a fresh `consumedDraftRef` (`:91`) and sends the same prompt a second time. On trunk the state was cleared within a microtask, so a reload could not replay it.
- Please report admission without blocking on the run: e.g. let `send`/`startRun` settle on "first frame or refusal" while the stream continues in the background, or keep `startRun` fire-and-forget and expose a separate admission promise to `useDraftHandoff`, so it can clear immediately and restore only on refusal (mirroring `ai/index.tsx:119-124`). A test for the admitted handoff (composer cleared immediately, history state cleared immediately) would pin it.
- Minor: both refusal fixtures use `code: 'ai.run.in_flight'`, which no server path emits. Trunk codes are `ai.run.busy` (409), `ai.request.invalid` (400), `ai.conversation.not_found` (404), `ai.run.refused` (default), `ai.run.overloaded` / `ai.run.internal_error` (`AiRunExecutor.java:124-125`). Please use `ai.run.busy` + `status: 409` so the fixture pins the real contract.
- Note: `keepsTheComposerClearedWhenAnAdmittedStreamFailsTest` is correctly described as passing on base — appreciated, and it is not counted as evidence for the fix.

unbridled-41:
@lizhimins Thanks — the blocking issue and the fixture contract are both addressed in 1e7a86e4 (branch rebased onto trunk `4c697f07`, resolving `useAgentRun.ts` against #4613's `knownRunId` seeding).

**Admission without blocking**: `startStream` now returns an admission promise — `true` on the first stream frame, `false` when the open lifecycle settled without delivering any frame (409 `ai.run.busy`, a route that moved on, or a transport failure). The stream lifecycle continues inside the hook until the terminal frames and the timeline refetch settle (its promise is awaited there and its rejection consumed — no fire-and-forget, no unhandled rejection, timeline/finalization semantics unchanged).

**useDraftHandoff**: with admission settling at the first frame, an admitted handoff clears the composer immediately and strips the history entry's state immediately; a refusal still restores the draft (and strips the state so a reload cannot replay it). The plain composer send likewise restores only on refusal, now at admission time.

**Fixture**: both refusal fixtures use `code: 'ai.run.busy', status: 409` — the real `AiRunService.REFUSED_BUSY_CODE` contract; `ai.run.in_flight` is gone.

**New regression**: `clearsTheComposerAndHistoryStateImmediatelyWhenAnAdmittedHandoffIsStillStreamingTest` — handoff onto an existing conversation whose stream emits `run_started` and then stays open: composer reads empty and the history state is `null` while the stream has NOT finished. Red against the previous code (`Received: '检查集群状态'` — the clearing was deferred to the end of the run), green now. The boundary test `keepsTheComposerClearedWhenAnAdmittedStreamFailsTest` is retained.

AiPage **19/19**, full `src/pages/ai` suite **190/190**; `tsc -b`, eslint, `npm run build` clean.

Ready for re-review.

PR #4741, opened 2026-09-21
fix(web): key the protocol tag map by the values the API returns

Frun1na:

- Fixes #4742

### Brief Description

`PROTOCOL_MAP` (`web/src/constants/theme.ts`) — which colors and labels the protocol tags on the
consumer page's client list — was keyed by `REMOTING` / `GRPC`:

```ts
export const PROTOCOL_MAP: Record<string, { labelKey: string; color: TagProps['color'] }> = {
  REMOTING: { labelKey: 'theme.protocolRemoting', color: 'geekblue' },
  GRPC: { labelKey: 'theme.protocolGrpc', color: 'green' },
};
```

The server serializes the `Protocol` enum by its enum name, which is `gRPC` / `Remoting`
(`org.apache.rocketmq.studio.common.domain.enums.Protocol`), so the lookup never matched: the tag
fell back to `{ labelKey: protocol, color: 'default' }`, i.e. the right text through `t(protocol)`
but the default gray color instead of green / geekblue.

The map only ever matched the stale mock data in `web/src/mock/consumers.ts`, which used the same
uppercase shells — `mock/clients.ts` already uses the real values, which is why the clients page
renders correctly.

This aligns the map keys, the consumer mock data and the `ConsumerPage` test fixtures with the
values the API sends, and pins the contract in the theme test so the keys cannot drift again.

Note for reviewers: this touches `ConsumerPage.test.tsx`, as does my open #4732 (different hunks:
that one adds a filter-mode test, this one updates two protocol fixture values).

```
$ npx vitest run src/constants/__tests__/theme.test.ts src/pages/instance/__tests__/ConsumerPage.test.tsx
 Test Files  2 passed (2)
      Tests  43 passed (43)

$ npx eslint src/constants/theme.ts src/mock/consumers.ts \
      src/constants/__tests__/theme.test.ts src/pages/instance/__tests__/ConsumerPage.test.tsx
(no output, exit 0)
```

The new theme assertion fails on the unfixed code (`Object.keys(PROTOCOL_MAP)` was
`['GRPC', 'REMOTING']`, expected `['Remoting', 'gRPC']`) and pins the wire contract:

```ts
it('is keyed by the values the Protocol enum returns on the wire', () => {
  // The server serializes ...Protocol by its enum name, which is `gRPC` / `Remoting`.
  expect(Object.keys(PROTOCOL_MAP).sort()).toEqual(['Remoting', 'gRPC']);
});
```

Contract evidence:

```
$ cat .../common/domain/enums/Protocol.java
public enum Protocol {
    gRPC, Remoting
}

$ grep -n "protocol:" web/src/mock/clients.ts | head -2
    protocol: 'gRPC' | 'Remoting';        # the mock that was already correct
```

### Checklist


RockteMQ-AI (approved):
﻿### Summary
Fixes a real bug where `PROTOCOL_MAP` keys (`REMOTING`/`GRPC`) didn't match the server-side enum names (`Remoting`/`gRPC`), causing protocol tags to always render with the default gray color. The fix aligns the map keys, mock data, and test fixtures with the actual API values, and pins the contract in a theme test to prevent regression.

Well-documented with clear evidence of the wire contract. Good fix.

---

PR #4743, opened 2026-09-21
fix(web): prefix CSV downloads with a UTF-8 BOM

Frun1na:

- Fixes #4744

### Brief Description

`downloadCsv` handed the spreadsheet a Blob tagged `text/csv;charset=utf-8`, but spreadsheet apps
sniff the encoding instead of honouring the charset parameter. Without a BOM, Excel decodes the file
as ANSI/GBK and every non-ASCII cell — topic remarks, user-defined alert names, localized status
labels — turns into mojibake as soon as the UI language is not English.

Every client-built export funnels through `downloadCsv`, so this fixes the topic, consumer group,
DLQ, client connection, metrics, topic-comparison and user exports in one place.

Two boundaries make the change safe and complete:

- The server-side exporters (`AuditService`, `CloudCredentialService`) already prefix `\uFEFF` and
  their CSV is downloaded through `downloadBlob` directly (audit page, resource-timeline drawer,
  cloud credentials tab), so they are untouched — prefixing again there would double the BOM.
- The in-app importer (`resourceCsvImport.readCsvRows`) already strips a leading `\uFEFF`, so the
  export → import round trip stays byte-identical.

```
$ npx vitest run src/utils/download.test.ts src/components/__tests__/TopicConfigComparisonDrawer.test.tsx src/components/__tests__/MetricsExplorer.test.tsx
 Test Files  3 passed (3)
      Tests  38 passed (38)

$ npx eslint src/utils/download.ts src/utils/download.test.ts
(no output, exit 0)

$ npx tsc -b
(no output, exit 0)
```

The new test captures the Blob handed to `URL.createObjectURL` and asserts the raw leading bytes:

```ts
const bytes = new Uint8Array(await blob.arrayBuffer());
// EF BB BF is the UTF-8 encoding of U+FEFF.
expect(Array.from(bytes.slice(0, 3))).toEqual([0xef, 0xbb, 0xbf]);
```

It fails on the unfixed code. `blob.text()` cannot be used for the BOM assertion itself because
`TextDecoder` strips a leading BOM by default; the test still asserts via `blob.text()` that the
content after the BOM is unchanged.

### Checklist


RockteMQ-AI (approved):
### Summary
Adds a UTF-8 BOM (`\uFEFF`) prefix to CSV exports so spreadsheet apps (Excel in particular) correctly detect the encoding. Without it, Excel decodes as ANSI/GBK and garbles non-ASCII characters — a real pain for i18n deployments with Chinese topic names, remarks, etc.

The fix is minimal and consistent with the server-side CSV exporters that already prefix BOM. Test verifies the raw bytes include `EF BB BF`. Practical fix for internationalization support.

---

Issue #4747, opened 2026-09-21
[Studio][Bug] A rule without a metric aborts the alert reconcile pass for its whole scope


unbridled-41:
### Studio Version

branch: rocketmq-studio
git commit id: cd448e17f14b4bd6e6d4ade87300caf19ca91bc2

### Problem

One enabled alert rule stored without a metric aborts the whole reconciliation pass for its instance and domain, so every active alert of that scope stays active forever: no `RESOLVED` state, no `RESOLVED` event, no recovery notification.

Chain:

- `server/src/main/java/org/apache/rocketmq/studio/ops/alert/NativeAlertProcessor.java:113-116` filtered the domain's rules with `scope.metricKeys().contains(StringUtils.trimWhitespace(rule.getMetric()))`.
- `MetricCollectionScope` stores its keys as `Set.copyOf(...)` (`server/src/main/java/org/apache/rocketmq/studio/cluster/metrics/MetricCollectionScope.java:31`). Every `Set.of`/`Set.copyOf` implementation rejects a null lookup: `ImmutableCollections.Set12.contains` dereferences the argument (`o.equals(e0)`) and `SetN.contains` calls `Objects.requireNonNull`. Verified on this JDK:

```
size=1 class=Set12 contains(null) THREW java.lang.NullPointerException: Cannot invoke "Object.equals(Object)" because "o" is null
size=2 class=Set12 contains(null) THREW java.lang.NullPointerException: ...
size=3 class=SetN  contains(null) THREW java.lang.NullPointerException
```

  `StringUtils.trimWhitespace(null)` returns null, so a rule with a null metric turns the filter predicate into that throwing lookup. The size dependency some readers expect (only 3+ elements) does not exist.
- A metric-less rule is a storable, first-class state: `AlertRuleRequestDTO.metric` (`ops/alert/AlertRuleRequestDTO.java:34`) carries no `@NotBlank`/`@Size`, `metric` is nullable in `db/schema.sql:274`, and the JSON transfer import explicitly tolerates it - `AlertRuleTransferService.java:33-46` (`if (candidate.getMetric() != null)`) plus `NativeAlertMetricCatalogService.java:70-73` (`if (rule == null || rule.getMetric() == null) return;`), reachable through `AlertRuleController.java:85-86` and `ClusterAlertRuleController.java:69-70`. Legacy rows migrated without a metric are the other source.
[...]

PR #4748, opened 2026-09-21
fix(alert): keep a rule without a metric out of the reconcile membership test

unbridled-41:
Fixes #4747.

**Line references** in this description are as of the patched file at this PR's head; the linked issue quotes the trunk commit, where the same code sits a few lines earlier.

## Problem

One enabled alert rule stored without a metric makes the alert reconciliation pass throw, so every active alert of that instance and domain stays active forever — no `RESOLVED` state, no `RESOLVED` event, no recovery notification.

- `ops/alert/NativeAlertProcessor.java:113-116` filtered the domain's rules with `scope.metricKeys().contains(StringUtils.trimWhitespace(rule.getMetric()))`.
- `MetricCollectionScope` holds its keys as `Set.copyOf(...)` (`cluster/metrics/MetricCollectionScope.java:31`); every `Set.of`/`Set.copyOf` implementation rejects a null lookup (`Set12.contains` dereferences the argument, `SetN.contains` calls `Objects.requireNonNull`), and `StringUtils.trimWhitespace(null)` is null.
- A metric-less rule is storable: `AlertRuleRequestDTO.metric` (`ops/alert/AlertRuleRequestDTO.java:34`) has no `@NotBlank`, the column is nullable (`db/schema.sql:274`), and the JSON import path tolerates it explicitly (`AlertRuleTransferService.java:33-46`, `NativeAlertMetricCatalogService.java:70-73`, reachable from `AlertRuleController.java:85-86`).
- The failure is swallowed: `CollectorScheduler.java:282-284` / `:298-300` catch `RuntimeException` and log a warning, so the reconcile pass for the whole scope is skipped.

## Root cause

A null-unsafe membership test over an immutable set, in a filter that runs before any metric-less rule is discarded — so the rule is not skipped, it aborts the pass.

## Fix

`NativeAlertProcessor` normalises the rule's metric to `""` when it is absent, so the membership test answers `false` and the rule is filtered out while the rest of the scope is still reconciled (`normalizedMetric`, 13 lines including the comment). A rule without a metric can never match a collected sample, so it has no active state that the pass could resolve; the semantics of every other rule in the scope are unchanged.

## Tests

One case, `NativeAlertProcessorTest.reconcilesRemainingRulesWhenAnotherRuleHasNoMetricTest`, asserting all three halves of the contract: the pass does not throw, the surviving rule of the scope is still resolved (`AlertStateStatus.RESOLVED`), and the recovery half is emitted too — the `RESOLVED` lifecycle event is recorded and the notification is queued (`outbox.enqueue(event, rule, labels)`).

Red before the fix:

```
[ERROR] NativeAlertProcessorTest.reconcilesRemainingRulesWhenAnotherRuleHasNoMetricTest -- Time elapsed: 0.016 s <<< FAILURE!
  "java.lang.NullPointerException: Cannot invoke "Object.equals(Object)" because "o" is null
[ERROR] Tests run: 24, Failures: 1, Errors: 0, Skipped: 0
```

Teeth beyond the crash, measured with a plausible wrong fix (an early `return` in `reconcileMissingActiveStates` whenever the domain has a metric-less rule — only `NativeAlertProcessor.java` mutated, test untouched):

```
[ERROR] ...reconcilesRemainingRulesWhenAnotherRuleHasNoMetricTest <<< FAILURE!
Wanted but not invoked:                       # states.save / alerts.saveAlert / outbox.enqueue
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

so the case fails both when the pass aborts with an exception and when it aborts silently; the source file was restored immediately afterwards (`git diff --stat` = the test file only).

Commands and results after the fix, run from `server/`:

- `mvn -B -ntp -o test -Dtest=NativeAlertProcessorTest` → **24 passed**, `BUILD SUCCESS`.
- `mvn -B -ntp -o test -Dtest='org.apache.rocketmq.studio.ops.alert.**,CollectorSchedulerTest'` → **318 tests, 0 failures, 5 errors**; all 5 errors are the sandbox's missing MySQL (`CannotGetJdbcConnectionException: Communications link failure`, from `NativeAlertEvaluationTransactionTest`, `NotificationOutboxMapperIntegrationTest`, `RmqAlertStateMapperIntegrationTest`), the same baseline failure this environment produces on the clean trunk.

## Duplicate check

Searched open PRs (205) and issues (214) for `reconcile`, `null metric`, `without a metric`, `resolve`, `alert state`, plus the commit index: the adjacent open items are #4571 (the AI rule-list tool tolerating a metric-less rule) and #4711 (rejecting a metric-less rule *test run*) — same input class, two other surfaces, neither touching `NativeAlertProcessor.reconcileMissingActiveStates`. No open item covers this crash site. #4647/#4648 touch a different unguarded parse in the same scheduler.

## Base branch

Targets the development trunk `rocketmq-studio`. GitHub interprets closing keywords only for pull requests that target the repository default branch, and that is still `master`, so merging will not close #4747 automatically — it needs to be closed by hand.

## Risk

Low and narrow: the filter now skips a rule that previously made the pass abort before doing anything. No behaviour is removed for rules that carry a metric, and the change cannot turn a resolvable alert into an unresolvable one.


unbridled-41:
### Evidence for this PR

> **Updated:** this comment describes the first commit; the second commit adds the recovery half (RESOLVED event + queued notification) to the case. This PR now has 2 commits and its head is `7bc88253` — see the follow-up evidence comment below.


**Head / diff (verified against the pushed branch)**

- head `092a9e3d87206cf0c778377f4680425ddb2c576c`, base `rocketmq-studio` (tip `cd448e17`), 1 commit, `mergeStateStatus: CLEAN`.
- `git show --stat 092a9e3d` → `NativeAlertProcessor.java +13/-1`, `NativeAlertProcessorTest.java +37/-0`.

**The NPE is not size-dependent**

`MetricCollectionScope` holds `Set.copyOf(...)`, and on this JDK every implementation rejects a null lookup — `Set12` dereferences the argument instead of returning false:

```
$ java Probe.java     # jdk 21.0.12, Set.copyOf of n distinct keys
size=1 class=Set12 contains(null) THREW java.lang.NullPointerException: Cannot invoke "Object.equals(Object)" because "o" is null
size=2 class=Set12 contains(null) THREW java.lang.NullPointerException: ...
size=3 class=SetN  contains(null) THREW java.lang.NullPointerException
size=5 class=SetN  contains(null) THREW java.lang.NullPointerException
```

so a single-key scope is enough to reproduce; no "3 or more metric keys" precondition.

**Red → green**

```
$ cd server && mvn -B -ntp -o test -Dtest=NativeAlertProcessorTest
[ERROR] NativeAlertProcessorTest.reconcilesRemainingRulesWhenAnotherRuleHasNoMetricTest <<< FAILURE!
  "java.lang.NullPointerException: Cannot invoke "Object.equals(Object)" because "o" is null
[ERROR] Tests run: 24, Failures: 1, Errors: 0, Skipped: 0        # BUILD FAILURE

$ mvn -B -ntp -o test -Dtest=NativeAlertProcessorTest
[INFO] Tests run: 24, Failures: 0, Errors: 0, Skipped: 0         # BUILD SUCCESS

$ mvn -B -ntp -o test -Dtest='org.apache.rocketmq.studio.ops.alert.**,CollectorSchedulerTest'
[ERROR] Tests run: 318, Failures: 0, Errors: 5, Skipped: 0
```

The 5 errors are the environment's baseline: `RmqAlertStateMapperIntegrationTest`, `NotificationOutboxMapperIntegrationTest` and `NativeAlertEvaluationTransactionTest` need MySQL, and this sandbox has none (`Caused by: com.mysql.cj.jdbc.exceptions.CommunicationsException: Communications link failure` → `java.net.ConnectException: Connection refused`). They fail identically on a clean `cd448e17` checkout.

**Teeth**

The new test asserts both halves of the fix: `assertThatCode(...).doesNotThrowAnyException()` (the pass no longer aborts) and `verify(states).save(oldKey, state)` with `status == RESOLVED` (the other rule of the scope is still reconciled). Removing only the `normalizedMetric` call makes the test fail again at the first assertion with the NPE quoted above.

**Duplicate check (re-run against the current indexes)**

- Open PRs (205 titles) / open issues (214 titles): `reconcile`, `reconcileMissingActiveStates`, `null metric`, `without a metric`, `alert state` → only #4571 (AI rule-list tool) and #4711 (rule test run), both open, both other surfaces.
- `git log --oneline -400 origin/rocketmq-studio | grep -i alert` → no commit touching this filter.

**Scoring**

PRIORITY = 80 (impact 32 + blast radius 16 + reproducibility 18 + maintenance value 14); FIX_CONFIDENCE = 95.

RockteMQ-AI (approved):
### Summary
Fixes a real bug where a stored rule with a null/empty metric would cause `MetricCollectionScope.metricKeys().contains(null)` to throw `NullPointerException` (since `ImmutableSet.contains(null)` throws). This would abort the entire reconcile pass, preventing other valid rules from being processed.

The `normalizedMetric()` helper trims whitespace and returns empty string instead of null — clean and defensive. Test verifies that the reconcile pass doesn't abort when another rule has no metric. Good fix.

---

unbridled-41:
### Evidence added in the second commit (`7bc88253`)

The first commit's test asserted the state transition; the description also claims the *recovery half* — "no `RESOLVED` event, no recovery notification" without the fix. The case now asserts all three: the pass does not throw, the surviving rule is resolved, the `RESOLVED` lifecycle event is recorded (`alerts.saveAlert`, transition `RESOLVED`) and the notification is queued (`outbox.enqueue(event, rule, labels)`).

**Teeth beyond the original crash** — a plausible wrong fix was used as the contrast, since a "bail out of the pass when the domain has a metric-less rule" implementation would have satisfied the first commit's assertions. Only `NativeAlertProcessor.java` was mutated (an early `return` in `reconcileMissingActiveStates`); the test file was untouched:

```
$ cd server && mvn -B -ntp -o test -Dtest=NativeAlertProcessorTest#reconcilesRemainingRulesWhenAnotherRuleHasNoMetricTest
[ERROR] ...reconcilesRemainingRulesWhenAnotherRuleHasNoMetricTest <<< FAILURE!
Wanted but not invoked:                        # states.save / alerts.saveAlert / outbox.enqueue
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0        # BUILD FAILURE
```

Restored immediately afterwards (`git diff --stat` = the test file only), then:

```
$ mvn -B -ntp -o test -Dtest=NativeAlertProcessorTest
[INFO] Tests run: 24, Failures: 0, Errors: 0, Skipped: 0        # BUILD SUCCESS

$ mvn -B -ntp -o test -Dtest='org.apache.rocketmq.studio.ops.alert.**,CollectorSchedulerTest'
[ERROR] Tests run: 318, Failures: 0, Errors: 5, Skipped: 0
```

Same 318/0/5 as before this commit; the 5 errors are the sandbox's missing MySQL (`NativeAlertEvaluationTransactionTest`, `NotificationOutboxMapperIntegrationTest`, `RmqAlertStateMapperIntegrationTest`), unchanged from the clean-trunk baseline.

**State of this PR**

- head `7bc88253`, 2 commits (`092a9e3d` fix + `7bc88253` test), base `rocketmq-studio`, MERGEABLE/CLEAN, not merged.
- numstat against the trunk (`git diff --numstat origin/rocketmq-studio..fix/alert-reconcile-rule-without-metric`): `NativeAlertProcessor.java +12/-1`, `NativeAlertProcessorTest.java +42/-0` — the second commit replaced 4 lines inside the case it extended, so the cumulative diff reports no deletions (the +46/-4 figure is the sum of the per-commit diffs).
- No claim in the description is left uncovered: the abort, the surviving rule's state, its event and its notification are all asserted.

Issue #4753, opened 2026-09-21
[Studio][Bug] Retention purges the AI conversation but leaves its agent workspace on disk


unbridled-41:
### Studio Version

branch: rocketmq-studio
git commit id: cd448e17f14b4bd6e6d4ade87300caf19ca91bc2

### Problem

The AI conversation retention pass deletes the conversation, its runs and its events, but leaves the conversation's agent workspace on disk. The workspace holds the child's `HOME` (`home/.claude/projects/<cwd>/<session-id>.jsonl`), i.e. the full agent transcript, plus `rmqctl.yaml`, `mcp.json` and the appended system prompt; the on-request delete removes it, the retention pass does not, so "expired" conversations keep their transcript under an id nothing refers to any more.

Chain:

- `server/src/main/java/org/apache/rocketmq/studio/ops/ai/conversation/AiConversationService.java:500-504` (`deleteCascade`) deletes events, then runs, then the conversations - and nothing else; it is called only by `deleteExpiredConversations` (`:479-496`), reached from the scheduled `purgeExpired` (`:411-425`, `@Scheduled(fixedDelayString = "${studio.ai.conversation.cleanup-interval:PT24H}")`).
- The on-request delete does remove it: `AiConversationService.java:289` (`workspace.delete(conversationId)`).
- The workspace is documented as tied to the conversation's lifetime: `RmqctlWorkspace.java:212-216` ("while {@code home/} and whatever {@code claude} persisted inside it are left alone", i.e. they persist per conversation), the layout at `:61-70` (`home/  0700, the child's HOME, per conversation and persistent`), and `delete(long)` at `:277-299`, whose javadoc says it is "called by `DELETE /api/ai/conversations/{id}` after the rows are gone". Nothing else ever calls it - `AiConversationService.java:289` was the only production caller.

Trigger (deterministic): set `studio.ai.conversation.retention-days=1` and `cleanup-interval=PT1M` (or wait out the 90-day default, `AiConversationProperties.java:37`), let a conversation with an agent run age out, and let the sweep run. The conversation disappears from the UI and its rows are gone, while `${studio.ai.conversation.workspace-dir}/conv-<id>/home/.claude/...` is still there.

Line references are as of the trunk commit quoted above (the PR shows the patched file).

### Evidence

- Defect: `AiConversationService.java:500-504` (retention cascade without the workspace).
[...]

PR #4707, opened 2026-09-21
fix(ai): cap prompt template text on code point boundaries

yyqdbngt:
### Which Issue(s) This PR Fixes

- No open issue tracks this exact defect. #4687 caps AI conversation titles on code point boundaries
  on the Java side and #4370 does the same for message properties; this is the web-side prompt
  template store, a different file, layer and storage format.

### Brief Description

`web/src/pages/ai/promptTemplates.ts` capped the persisted template text with `slice`, which counts
UTF-16 units:

```ts
const normalizeText = (value: unknown, maxLength: number): string =>
  text.length > maxLength ? text.slice(0, maxLength) : text;          // title/description/tag
const buildPromptTemplatePreview = (body: string, maxLength = 160) =>
  `${compact.slice(0, maxLength)}...`;                                // list preview
```

A supplementary character - an emoji, a CJK extension character - is two units, so a cut that lands
on one keeps the high surrogate without its low half. The lone surrogate is not a character and has
no UTF-8 encoding, so the title that is written to `rocketmq-studio-ai-prompt-templates` and rendered
as the label of the template list, and the preview shown next to it, both carry U+FFFD instead of the
character they were cut out of. A title of 79 ASCII characters followed by an emoji is the shortest
reproduction, and the CJK extension characters an operator is likely to type have the same shape.

Both cut sites now count code points, the rule `AiConversationService` applies to conversation titles
and `MessagePropertyDisplay` applies to message property values.

### How Did You Test This Change?

Two new cases in `web/src/pages/ai/promptTemplates.test.ts`: one saves a template whose title is
`79 x 'a' + emoji + "tail"` and expects the stored title to be exactly `79 x 'a' + emoji`, the other
asks the preview builder for `159 x 'a' + emoji + "tail"` and expects `159 x 'a' + emoji + "..."`.

Both runs below use the project's own test runner through the workspace-local binary
(`web/node_modules/.bin/vitest.cmd`, vitest 4.1.10 as pinned by `web/package.json`), not a
downloaded one.

Red - on the unmodified base (new tests copied in, `promptTemplates.ts` untouched):

```
$ cd web && .\node_modules\.bin\vitest.cmd run src/pages/ai/promptTemplates.test.ts
 ❯ src/pages/ai/promptTemplates.test.ts (13 tests | 2 failed) 2.72s
     × caps a stored title on code point boundaries 8ms
     × cuts an over-long preview on code point boundaries 2ms

[...]

PR #4754, opened 2026-09-21
fix(ai): remove the agent workspace with every purged conversation

unbridled-41:
Fixes #4753.

**Line references** in this description are as of the patched file at this PR's head; the linked issue quotes the trunk commit, where the same code sits a few lines earlier.

## Problem

The AI conversation retention pass deletes the conversation, its runs and its events, but never the conversation's agent workspace — while the on-request delete does. The workspace is the child's `HOME` (`home/.claude/projects/<cwd>/<session-id>.jsonl`), so "expired" conversations keep their full agent transcript on disk under an id nothing refers to any more.

- `AiConversationService.java:509-515` — `deleteCascade` deletes events, runs, conversations and nothing else; its only caller is `deleteExpiredConversations` (`:480-497`), reached from the scheduled `purgeExpired` (`:412-426`).
- `AiConversationService.java:290` — `delete(Long, String)` calls `workspace.delete(conversationId)`.
- `RmqctlWorkspace.java:61-70` and `:212-216` — the layout documents `home/` as "the child's HOME, per conversation and persistent" and the materialisation as leaving it alone, so the transcript is exactly what survives.
- `AiConversationProperties.java:37` — `retentionDays = 90`; `application.yml:119` — `workspace-dir: ${STUDIO_AI_CONVERSATION_WORKSPACE_DIR:/tmp/rocketmq-studio-ai}`, so on a deployment that points it at a persistent volume the directories accumulate without bound.

## Root cause

The retention cascade was written as a database cascade ("no FK constraints here, so the cascade is this method") and the workspace, a filesystem side effect owned by another class, was only remembered by the interactive delete path.

## Fix

`deleteCascade` now removes the workspace of every purged conversation **after** the rows are gone (4 lines), the order `delete(Long, String)` uses: a crash in between leaves a directory that is still reachable, whereas removing the transcript of a conversation that is still visible is not recoverable. The class javadoc's "Retention cascades by hand" note now says the workspace goes with the rows.

## Tests

Red before the fix:

```
$ cd server && mvn -B -ntp -o test -Dtest=AiConversationServiceTest
[ERROR] ...retentionShouldRemoveTheWorkspaceOfEveryPurgedConversationTest <<< FAILURE!
Wanted but not invoked:
[ERROR] Tests run: 19, Failures: 1, Errors: 0, Skipped: 0        # BUILD FAILURE
```

After the fix:

```
$ mvn -B -ntp -o test -Dtest=AiConversationServiceTest
[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0         # BUILD SUCCESS

$ mvn -B -ntp -o test -Dtest='org.apache.rocketmq.studio.ops.ai.**'
[INFO] Tests run: 904, Failures: 0, Errors: 0, Skipped: 0        # BUILD SUCCESS
```

The new test uses an `InOrder` verification (`deleteByIds` before `workspace.delete(1)`, `(2)`, `(3)`), so it fails both if the workspace is never removed and if it is removed before the rows; the pre-existing `deleteShouldCascadeChildrenFirstAndRemoveTheWorkspaceTest` keeps the interactive path's ordering pinned.

## Duplicate check

Searched open PRs (205) and issues (214) for `workspace`, `retention`, `purge`, `transcript`, `orphan`, `disk`, `cleanup`, plus the commit index: the retention hits are #4647/#4648 (alert snapshot retention) and the AI hits (#4685, #4693, #4696, #4707, #4715, #4631, #4665, #4738, #4740) are all about the run/stream lifecycle or the CLI session. Nothing covers the workspace lifecycle, and `git log -S "deleteCascade" -- AiConversationService.java` shows the cascade has been DB-only since it was introduced.

## Base branch

Targets the development trunk `rocketmq-studio`. GitHub interprets closing keywords only for pull requests that target the repository default branch, and that is still `master`, so merging will not close #4753 automatically — it needs to be closed by hand.

## Coverage boundary (audited after the fact)

The case pins the mechanism for one retention batch of three ids (`InOrder`: `deleteByIds` → `workspace.delete(1)`,`(2)`,`(3)`), and the loop that runs further batches is pinned separately by the pre-existing `retentionShouldKeepBatchingUntilTheBacklogIsGoneOrTheCapIsReachedTest`; the call added here is one `forEach` inside the single cascade method, so there is no per-batch branch left uncovered. No follow-up test commit was added for that reason.

## Risk

Low: the same side effect the interactive delete already performs, on a path where the conversation is already gone. `RmqctlWorkspace.delete` never throws (its javadoc: "a stale directory on a deleted volume must not turn a successful delete into a 500"), so a filesystem problem cannot fail the retention pass — and that property, together with the two states this call relies on, is pinned by `RmqctlWorkspaceTest.deleteRemovesTheWholeWorkspaceAndNeverThrowsTest`: it removes a workspace containing `home/.claude/projects/-ws/session.jsonl`, returns `false` on the idempotent retry, and returns `false` for an id whose directory was never created (a conversation that never admitted an agent run).


unbridled-41:
### Evidence for this PR

**Head / diff (verified against the pushed branch)**

- head `ffe3d84d594f13162630ab52a77f809135001886`, base `rocketmq-studio` (tip `cd448e17`), 1 commit, `mergeStateStatus: CLEAN`.
- `git show --stat ffe3d84d` → `AiConversationService.java +14/-3`, `AiConversationServiceTest.java +18/-0`.

**Red → green**

```
$ cd server && mvn -B -ntp -o test -Dtest=AiConversationServiceTest
[ERROR] ...retentionShouldRemoveTheWorkspaceOfEveryPurgedConversationTest <<< FAILURE!
Wanted but not invoked:
[ERROR] Tests run: 19, Failures: 1, Errors: 0, Skipped: 0        # BUILD FAILURE

$ mvn -B -ntp -o test -Dtest=AiConversationServiceTest
[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0         # BUILD SUCCESS

$ mvn -B -ntp -o test -Dtest='org.apache.rocketmq.studio.ops.ai.**'
[INFO] Tests run: 904, Failures: 0, Errors: 0, Skipped: 0        # BUILD SUCCESS
```

**Teeth (both directions)**

The new test verifies order, not just presence: `InOrder` requires `conversationRepository.deleteByIds(...)` **before** `workspace.delete(1)`, `(2)`, `(3)`. It fails if the workspace is never removed (the red run above) and would also fail if the workspace were removed before the rows. `AiConversationServiceTest.deleteShouldCascadeChildrenFirstAndRemoveTheWorkspaceTest` keeps the interactive path's ordering pinned and still passes.

**Only one production caller**

`grep -rn "deleteCascade" server/src/main/java/` → the declaration plus one call in `deleteExpiredConversations`. The interactive `delete(Long, String)` keeps its own cascade, so no request-serving path changes behaviour.

**Why the default workspace directory does not make this harmless**

`application.yml:119` defaults `workspace-dir` to `/tmp/rocketmq-studio-ai`, and `RmqctlWorkspace` notes that a container restart wipes `/tmp` — so on a default deployment the leftovers are cleared by a restart. But the property exists precisely so deployments can point it at a persistent volume, the sweep runs unattended every 24h (`AiConversationProperties.java:46`), and the transcript is the most sensitive artefact this feature writes. The retention promise ("90 days, then gone") should not depend on where the directory lives.

**Duplicate check (re-run against the current indexes)**

- Open PRs (205 titles) / open issues (214 titles), keywords `workspace`, `retention`, `purge`, `transcript`, `orphan`, `cleanup`, `disk` → #4647/#4648 (alert snapshot retention) are the only retention items; the AI items (#4685, #4693, #4696, #4707, #4715, #4631, #4665, #4738, #4740) are run/stream or CLI-session lifecycle.
- `git log -S "deleteCascade" -- server/src/main/java/.../AiConversationService.java` → the cascade has been database-only since it was introduced; no commit adds the workspace here.

**Scoring**

PRIORITY = 68 (impact 22 + blast radius 12 + reproducibility 18 + maintenance value 16); FIX_CONFIDENCE = 92.

unbridled-41:
### Full-suite baseline comparison (added after the fact)

Ran the whole server suite twice in this environment — once on a clean `cd448e17` worktree and once on this branch — to separate my change from the sandbox's known failures:

```
clean cd448e17   : Tests run: 3074, Failures: 0, Errors: 17, Skipped: 0   (9 error classes)
this branch      : Tests run: 3075, Failures: 0, Errors: 17, Skipped: 0   (same 9 classes)
```

The difference is exactly the one regression test added here. Every error in both runs is a `*IntegrationTest`/`StudioApplicationTest` context that needs MySQL (`IllegalState: Failed to load ApplicationContext` → `CannotGetJdbcConnectionException: Failed to obtain JDBC Connection` → `Communications link failure`), which this sandbox has no instance of; the failing class set is identical in both runs.

RockteMQ-AI (approved):
### Summary
Correctly cascades workspace deletion into `deleteCascade` after the rows are gone — the conversation's lifetime is what keeps the agent workspace on disk, so a purge that only deleted rows left orphaned workspace directories behind.

The ordering (rows first, workspace after) is well-documented and matches the interactive `delete(Long, String)` path. Test verifies the cascade order with `InOrder` and covers the retention sweep path. Clean fix.

---

Issue #4756, opened 2026-09-21
[Bug] rmqctl Studio client buffers response bodies without a size cap


LuoYushi7:
### Environment

* rmqctl on the `rocketmq-studio` branch
* Go toolchain per `rmqctl/go.mod`

### What happened

`Client.request` reads the whole Studio response body with an unbounded `io.ReadAll`:

```go
// rmqctl/internal/studio/client.go
defer response.Body.Close()
data, err := io.ReadAll(response.Body)
```

Nothing bounds that read. A Studio Server (or an intermediary gateway or proxy)
that streams a body without end makes the rmqctl process grow its heap until it
is killed, and the operator sees no diagnostic that names the response body.

The same package already caps the equivalent read on the error path:
`readMCPHTTPStatusError` uses `io.LimitReader(body, statusErrorBodyLimit+1)` and
rejects anything past the limit (`rmqctl/internal/studio/mcp_http.go`). The
success path has no equivalent. `stdioMaxLineBytes` shows the module bounds its
other unbounded inputs too.

### Steps to Reproduce

1. Point rmqctl at a Studio Server whose response body never ends (a stub that
   streams bytes, or a stuck gateway).
2. Run any command that issues a Studio API request, for example `rmqctl cluster list`.
3. rmqctl buffers the body in memory until the process is killed.

### What Did You Expect to See?

The client stops reading at a bounded size and fails with an error that names the
cause, so a runaway body is distinguishable from a malformed one.

### What Did You See Instead?

The body is buffered without limit, and an oversized body is only rejected later
by the JSON decoder as `invalid studio response: invalid character ...`, which
misattributes the failure.

### Additional Context

A regression test that sets the cap below the body size fails on unmodified
`rocketmq-studio` with `request() error = invalid studio response: invalid
character 'x' looking for beginning of value` — the decoder complains instead of
the read stopping at the cap.

LuoYushi7:
This duplicates #4640, which reported the same unbounded read 15 hours earlier. Closing this one and keeping #4640 as the tracking issue. The fix is in #4757.

PR #4757, opened 2026-09-21
fix(rmqctl): cap the Studio response body read

LuoYushi7:
### Brief Description

`Client.request` read the whole Studio response body with an unbounded
`io.ReadAll`, so a Studio Server (or an intermediary gateway or proxy) that
streams a body without end grew the rmqctl process until it was killed, with no
diagnostic naming the response body.

The read now goes through `io.LimitReader` against
`clientResponseBodyLimit` (32 MiB, mirroring the module's existing
`stdioMaxLineBytes` and `statusErrorBodyLimit` conventions) and fails with
`studio response body exceeds N bytes` once the body passes the cap. The error
names the cause instead of surfacing later as a JSON decode error on a payload
that was never meant to be that large.

The cap is a package variable rather than a constant so a test can lower it
instead of materializing a 32 MiB body.

### Tests

`rmqctl/internal/studio/client_test.go`:

* `TestRequestRejectsResponseBodyAboveLimit` — an oversized body is refused, and
  the test asserts how many bytes the client actually pulled from the reader, so
  it also pins that the read stopped rather than buffering everything.
* `TestRequestAcceptsResponseBodyAtLimit` — a body that fits still decodes, so
  the cap does not reject ordinary responses.

Fail-before: with the unmodified `request`, the first test fails with
`invalid studio response: invalid character 'x' looking for beginning of value`.

```
go test ./...        # all rmqctl packages pass
go vet ./internal/studio/
```

Fixes #4640

Note: I opened #4756 for this defect before finding that #4640 already reported it. #4756 is closed as a duplicate; #4640 is the tracking issue.


LuoYushi7:
Withdrawing this: #4639 already fixes the same defect (opened 2026-09-20T23:16Z, before this PR), and #4640 tracks it. I found #4640 only after opening #4756 and this PR — my duplicate check looked at issue comments but not at the linked-pull-request timeline, which is where #4639 appears. Closing this so it does not sit in the review queue.

The one thing worth keeping from this attempt is the test shape: my `TestRequestRejectsResponseBodyAboveLimit` wraps the response body in a reader that counts the bytes actually pulled, and asserts the read stopped at the cap instead of buffering everything. That pins the behaviour, not just the error message. If it is useful, feel free to lift it into #4639.

PR #4759, opened 2026-09-21
fix(rmqctl): preserve integer precision in passthrough studio payloads

Frun1na:

- Fixes #4760

### Brief Description

The studio client decoded the response payload with `json.Unmarshal` into the caller's `out any`
(`internal/studio/client.go`). For the passthrough output paths — `table`, `json` and `yaml` — that
target is an `any`, so `encoding/json` represented every number as `float64` and silently corrupted
int64 values beyond 2^53:

```
server returns  "msgOffset": 9007199254740993
rmqctl prints   "msgOffset": 9007199254740992
```

`types.DecodeMutationOutput` re-encodes the same decoded value to inspect it, so the mutation
summary inherited the corrupted number too. Nothing in `rmqctl` used `json.Number`.

This decodes the studio payload with `json.Decoder.UseNumber()`: `json.Number` keeps the literal and
re-marshals to the original digits, so table/json/yaml output and the mutation summary now round-trip
exactly. Decoding into typed struct fields is unaffected — `UseNumber` only changes how numbers land
in `interface{}` targets.

New test `internal/studio/client_test.go` feeds an `httptest` studio response containing
`"msgOffset":9007199254740993` and asserts the marshaled passthrough payload still contains the
literal:

```
$ go test ./internal/studio/ -run TestRequestPreservesLargeIntegers -v
=== RUN   TestRequestPreservesLargeIntegersInPassthroughPayload
--- PASS: TestRequestPreservesLargeIntegersInPassthroughPayload (0.00s)
PASS

$ make ci
ok  .../internal/config     (go test -race)
ok  .../internal/output
ok  .../internal/studio
catalog generator -check   (clean)
go vet ./...               (clean)
gofmt                      (clean)
6x cross-platform build    (ok)
```

The test fails on the unfixed code with exactly the reported symptom:
`payload lost integer precision: {"items":[{"msgOffset":9007199254740992}]}`.

### Checklist


rmqctl/internal/studio/client.go
```diff
@@ -150,7 +150,12 @@ func (c Client) request(ctx context.Context, target Target, method string, path
 	if !hasJSONData(envelope.Data) {
 		return nil
 	}
-	if err := json.Unmarshal(envelope.Data, out); err != nil {
+	// Passthrough commands decode into an `any`, where encoding/json would represent every
+	// number as float64 and silently corrupt int64 values beyond 2^53 (RocketMQ offsets and
+	// timestamps). UseNumber keeps the literal; decoding into typed struct fields is unaffected.
```
RockteMQ-AI, line 155: Using `json.NewDecoder` with `UseNumber()` is the correct fix for preserving int64 precision in passthrough payloads. When decoding into `any`, `encoding/json` defaults to `float64`, which silently corrupts values beyond 2^53 (the classic JavaScript safe integer boundary). `UseNumber()` keeps the literal representation as `json.Number`, which downstream code can convert to `int64` without loss.

rmqctl/internal/studio/client_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+)
+
+// The passthrough commands decode the studio payload into an `any`, where encoding/json
+// would represent every number as float64 and silently corrupt int64 values beyond 2^53
+// (RocketMQ offsets and timestamps). The client must keep the literal via json.Number.
+func TestRequestPreservesLargeIntegersInPassthroughPayload(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"code":200,"message":"ok","data":{"items":[{"msgOffset":9007199254740993}]}}`))
```
RockteMQ-AI, line 34: Good test choice — `9007199254740993` (2^53 + 1) is the canonical precision boundary value. The test directly validates that the passthrough path preserves the exact integer, which is the core guarantee this fix provides.

RockteMQ-AI (commented):
### Summary
Fixes integer precision loss in `rmqctl` passthrough commands by switching from `json.Unmarshal` to `json.NewDecoder` with `UseNumber()`. When decoding into `any`, `encoding/json` represents all numbers as `float64`, silently corrupting int64 values beyond 2^53 (RocketMQ offsets and timestamps). The fix is minimal, surgical, and correctly scoped to the passthrough path only.

### Findings
- **[Info]** `client.go:155` — `UseNumber()` is the correct approach. Typed struct field decoding is unaffected; only `any`/interface{} targets get `json.Number` instead of `float64`.
- **[Info]** `client_test.go:34` — Test uses the canonical precision boundary value (2^53 + 1), directly validating the core guarantee.

### Verdict
Clean, minimal fix with a focused regression test. No blocking issues.

---

Issue #4755, opened 2026-09-21
[Studio][Refactor] Defer pooled client shutdown until in-flight actions finish


Aias00:
## Problem

`InstanceService` releases cached RocketMQ clients when an instance endpoint or credential reference changes. `MqAdminExtFactory.release(...)` and `MqClientPool.release(...)` currently remove the matching cache entry and immediately call `shutdown()`.

A request that already obtained the same cached admin, producer, or pull consumer can still be executing when the instance update retires it. Immediate shutdown can close the client underneath that in-flight action, producing avoidable RPC/send/pull failures during a normal configuration update. Application shutdown has the same ownership race for actions already admitted before the pool is closed.

This lifecycle race is independent of Ops runtime connection settings and does not require a default/no-instance connection path.

## Proposed Design

1. Add a package-private generic `ClientLease<T>` that owns one cached client, counts admitted actions, and supports one-way retirement.
2. `acquire()` succeeds only before retirement; `release()` decrements the active count; `retire()` prevents new acquisitions and invokes shutdown exactly once when the active count reaches zero.
3. Store leases in `MqAdminExtFactory` and `MqClientPool` without changing cache keys, public method signatures, client construction, authentication identity, or provider constructors.
4. Each execute/send/pull path acquires the current lease before invoking the caller action and releases it in `finally`. If it races with a retired cache entry, remove that exact lease and retry lookup so the request obtains a current client.
5. Endpoint/credential release removes matching entries and retires them. Idle clients close immediately; active clients close after their last action completes.
6. Pool shutdown marks the owner closed before retiring and clearing entries, so no new client can be created while already-admitted actions drain.

## Scope

- `ClientLease`, `MqAdminExtFactory`, `MqClientPool`, and focused lifecycle tests only.
- No Ops connection persistence, default connection routing, transport-setting changes, provider constructor changes, deployment files, or application configuration.
- Preserve all existing exception mapping, endpoint normalization, authentication isolation, and cache reuse behavior.

## Acceptance Criteria

- Releasing an idle admin/producer/consumer shuts it down immediately and evicts it.
- Releasing an in-flight client does not call shutdown until the action exits.
- A request arriving after retirement cannot acquire the retired client and can create/reuse a replacement.
- Endpoint-wide and credential-scoped release keep their existing ownership boundaries.
- Shutdown rejects new calls while allowing admitted actions to finish before closing their clients.
- Shutdown is invoked at most once per cached client.
[...]

PR #4761, opened 2026-09-21
[ISSUE #4755] refactor(client): defer shutdown for in-flight actions

Aias00:
## Summary

- wrap cached admin, producer, and pull-consumer clients in a package-private reference-counted `ClientLease`
- remove retired entries from their caches immediately while deferring `shutdown()` until admitted actions finish
- make the final `closed` check and lease acquisition atomic with shutdown through a fair lifecycle read/write gate
- preserve cache keys, public APIs, connection creation, authentication isolation, exception mapping, and instance update ownership boundaries
- add deterministic concurrency regressions for endpoint release, credential-scoped release, shutdown draining, replacement-client creation, and creation racing with shutdown

Fixes #4755

## Scope

This is the pure lifecycle refactor requested during review of #4387. It changes only:

- `ClientLease`
- `MqAdminExtFactory`
- `MqClientPool`
- their focused tests

It does not include Ops runtime connection persistence, default/no-instance routing, transport-setting changes, provider constructor changes, `application.yml`, or deploy/Compose files.

## Verification

- Red/green: four release/shutdown tests first failed because clients were shut down while actions were still in flight.
- Red/green: the shutdown-admission regression first showed an action running after `closed=true`; it now receives 503 without running and the unadmitted client is retired.
- `mvn -B -ntp -Dtest=MqAdminExtFactoryTest,MqClientPoolTest,RuntimeAdminClientResolverTest,InstanceServiceTest test` with Java 21: 111 tests passed; Checkstyle 0 violations.
- Full backend suite with Java 21 and an isolated H2 MySQL-compatible datasource: 3079 tests, 0 failures/errors, 3 skipped; Checkstyle 0 violations.
- `mvn -B -ntp -DskipTests package`: passed.
- `git diff --check`: passed.
- Independent concurrency review: initial HIGH shutdown-admission race fixed; focused re-review approved with 0 remaining findings.

## Direction

The default RocketMQ connection remains fail-closed and non-editable at runtime under the current instance model. This PR retains only the lifecycle safety that applies to instance endpoint and credential updates.


server/src/main/java/org/apache/rocketmq/studio/cluster/broker/ClientLease.java
```diff
@@ -0,0 +1,76 @@
...
+
+    synchronized boolean acquire() {
+        if (retired) {
+            return false;
+        }
+        active++;
+        return true;
+    }
+
+    void release() {
```
RockteMQ-AI, line 47: The `release()` method calls `shutdown.accept(client)` outside the `synchronized` block, which is correct — it avoids holding the monitor during a potentially slow shutdown. Since `safeShutdown` catches all throwables, the lease will not be left in an inconsistent state even if shutdown throws. Good design.

server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
```diff
@@ -90,22 +95,46 @@ public <T> T execute(String namesrvAddr, RPCHook rpcHook, String authenticationI
...
+        while (true) {
+            if (closed) {
+                throw new BusinessException(503, "Admin factory is shutting down");
+            }
+            lease = cache.computeIfAbsent(cacheKey, key -> {
+                // Re-check before client creation; admission re-checks again after creation.
+                if (closed) {
+                    throw new BusinessException(503, "Admin factory is shutting down");
+                }
+                return new ClientLease<>(
```
RockteMQ-AI, line 108: The `while (true)` retry loop correctly handles the race where a lease is retired between `computeIfAbsent` and `acquire()`. The conditional `cache.remove(cacheKey, lease)` ensures only the stale lease is removed. Consider adding a brief comment explaining why the loop is needed (e.g., "retry if the lease was retired between creation and acquisition") to help future readers.

server/src/main/java/org/apache/rocketmq/studio/cluster/broker/MqAdminExtFactory.java
```diff
@@ -50,7 +54,8 @@ public class MqAdminExtFactory {
     /** Default admin RPC timeout in milliseconds. */
     private static final long DEFAULT_TIMEOUT_MILLIS = 5000L;
 
-    private final Map<AdminClientCacheKey, DefaultMQAdminExt> cache = new ConcurrentHashMap<>();
+    private final Map<AdminClientCacheKey, ClientLease<DefaultMQAdminExt>> cache = new ConcurrentHashMap<>();
```
RockteMQ-AI, line 57: Using `ReentrantReadWriteLock(fair=true)` prevents writer (shutdown) starvation at the cost of some throughput. For a dashboard admin tool with low contention, this is the right trade-off.

RockteMQ-AI (commented):
### Summary
Refactors cached admin, producer, and pull-consumer clients to use a reference-counted `ClientLease` that defers shutdown until in-flight actions complete. The design cleanly separates admission (read lock) from shutdown (write lock), and the `ClientLease` monitor protects the active/retired/closed state transitions. Comprehensive concurrency tests cover the key race conditions.

### Findings
- **[Info]** `ClientLease.java:47` — `release()` correctly calls shutdown outside the synchronized block; `safeShutdown` ensures no inconsistent state on exception.
- **[Info]** `MqAdminExtFactory.java:108` — The `while(true)` retry loop is correct; a brief comment explaining the race it handles would aid readability.
- **[Info]** `MqAdminExtFactory.java:57` — Fair `ReentrantReadWriteLock` is appropriate for low-contention admin tool usage.

### Suggestions
- Consider adding a Javadoc on `ClientLease` explaining the lifecycle contract (acquire → use → release; retire marks for deferred shutdown).
- The `awaitClosed` test helper uses reflection to access the `closed` field — acceptable for testing but could be fragile if the field is renamed.

### Verdict
Well-designed lifecycle refactor with sound concurrency model and thorough test coverage. No blocking issues.

---

PR #4770, opened 2026-09-21
fix(rmqctl): name response frames explicitly in the MCP decode error

Frun1na:

None — this is the small follow-up lizhimins asked for in the review on #4637 ("a much smaller PR that
only teaches `decodeMCPMessage` to reject a response frame with an explicit message instead of
'missing method' would be easy to review and is welcome on its own"), so it reuses that review as its
context instead of opening a duplicate issue.

### Brief Description

`decodeMCPMessage` classified stdin frames three ways — request (`method` + `id`), notification
(`method`), and everything else. A JSON-RPC **response** (`id` + `result`/`error`, no `method`) fell
into the last bucket and failed with `invalid MCP JSON-RPC message: missing method`, which tells the
caller nothing about what was actually wrong.

A frame with an `id` but no `method` is now named explicitly:

```
invalid MCP JSON-RPC message: response frame (id without method); this transport only forwards
client-originated requests and notifications
```

Request and notification decoding is unchanged. (The larger #4637 forwarding change stays closed —
this is only the error-message half the review welcomed.)

```
$ go test ./internal/studio/ -run TestDecodeMCPMessageRejectsResponse -v
=== RUN   TestDecodeMCPMessageRejectsResponseFrameWithExplicitError
--- PASS: TestDecodeMCPMessageRejectsResponseFrameWithExplicitError (0.00s)
PASS

$ make -C rmqctl ci
ok  .../internal/catalog  ok  .../internal/config  ok  .../internal/output
ok  .../internal/studio  ...  (gofmt, catalog-verify, go vet, go test -race, 6x build — all green)
```

The new test fails on the unfixed code (`got: invalid MCP JSON-RPC message: missing method`), asserts
the explicit message for both a `result` and an `error` frame, asserts the misleading text is gone,
and re-checks that request and notification frames still decode.

### Checklist


RockteMQ-AI (approved):
### Summary
Improves the error message when a JSON-RPC response frame (id + result/error, no method) is received by the MCP decoder. Previously it fell through to the generic "missing method" error, which was misleading. Now it explicitly identifies the frame as a response frame.

LGTM — small, focused fix with good test coverage.

---

PR #4772, opened 2026-09-21
fix(topic): require the queue counts on topic updates

Frun1na:

None — this implements exactly the design lizhimins prescribed in the review on #4581 ("make the
counts required on the update DTO (they are already `Integer` + `@PositiveOrZero`) and confine the
default-8 substitution to `createTopic`, where it belongs"), so it reuses that review as its context
instead of opening a duplicate issue. The underlying drain-to-zero defect itself is unchanged from
that PR: an update DTO without queue counts reaches the broker as 0, is read as "not provided", and
is refilled from the stored value or the default of 8.

### Brief Description

Following the prescribed boundary design:

- `UpdateTopicDTO.writeQueues` / `readQueues` are now `@NotNull` (still `@PositiveOrZero`, so an
  explicit **0 is accepted** and drains the topic), and `toTopicVO()` copies them unconditionally.
- `RocketMQAdminClientImpl.updateTopic` sends the caller's value as-is: the
  "stored value, else 8" fallback chain is gone, and with it the only place where a number other
  than the caller's choice could be written to the broker.
- The default-8 substitution stays confined to `createTopic`.
- The REST contract already documents both counts as required (`docs/api-spec.md` §5.4 marks
  `writeQueues` / `readQueues` 必填), so the code now matches the documented contract.

`TopicUpdateInput.mergeWith` (the AI tool path) fills absent counts from the current topic, so
partial updates through the agent keep writing the merged values.

```
$ mvn -B -ntp -Dtest='RocketMQAdminClientImplTest,TopicControllerTest,MetadataServiceTest,TopicMutationPlanTest' \
      -DfailIfNoTests=false test
[INFO] Tests run: 147, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

- New `updateTopicAppliesExplicitZeroQueueCountsToDrainATopic` feeds an explicit 0/0 and asserts the
  broker `TopicConfig` and the stored row both see 0. It fails on trunk with `expected: 0 but was: 8`.
- New controller test: `POST /api/topics/update` with `{"name":"orders"}` returns 400
  `writeQueues is required`.
- `updateTopicPreservesQueueCountsWhenNotSpecified` was removed: its premise (a perm-only partial
  update reaching the admin client without counts) no longer exists under the required contract.
- The now-unused `TopicPerm` import was dropped (checkstyle).

### Checklist


RockteMQ-AI (approved):
### Summary
Fixes a subtle bug where setting `writeQueues` or `readQueues` to 0 on a topic update would be silently ignored — the old code treated null and 0 identically, falling back to the existing value or the creation default of 8. The fix adds `@NotNull` validation to make queue counts required in the update DTO, and removes the fallback logic in both `toTopicVO()` and `RocketMQAdminClientImpl.updateTopic()`.

### Findings
- **[Info]** `UpdateTopicDTO.java:34-35` — `@NotNull` + `@PositiveOrZero` is the correct combination: it allows 0 (drain-to-zero) while rejecting null (not provided). Good.
- **[Info]** `RocketMQAdminClientImpl.java:368-380` — Removing the `topic.getWriteQueues() > 0 ? ... : existing` fallback is correct now that the DTO guarantees non-null values. The simplified code is easier to reason about.
- **[Info]** Test updates correctly verify that explicit 0 queue counts now reach the broker.

LGTM — clean fix that eliminates an ambiguous null/0 conflation.

---

PR #4773, opened 2026-09-21
fix(consumer): filter subscription mode before pagination

beautyarbutin:

- Fixes #3150

### Brief Description

The consumer subscription-mode selector currently filters only the records returned for the active page. This change sends the selected mode to the page API and applies it before pagination for both Apache and cloud-backed providers. Switching modes resets the UI to page 1, and exports use the same mode normalization.

This reimplements the still-open issue on the current `rocketmq-studio` branch. Related earlier attempts #3160 and #4053 were closed when that branch was removed during the previous branch migration; #4172 was closed as a duplicate.

- `cd server && mvn -B -ntp "-Dtest=ConsumerGroupControllerTest,MetadataServiceTest,InstanceProviderTest,ApacheInstanceProviderTest,RocketMQMetadataProviderTest" test`: 135 tests passed on the current `rocketmq-studio` head.
- `cd web && npm test -- --run src/pages/instance/__tests__/ConsumerPage.test.tsx src/api/consumerGroups.test.ts`: 44 tests passed on the current `rocketmq-studio` head.
- `cd web && npx eslint src/api/consumerGroups.test.ts src/api/metadata.ts src/pages/instance/__tests__/ConsumerPage.test.tsx src/pages/instance/consumer.tsx`: passed.
- `cd web && npx tsc -b --pretty false`: passed.
- `cd web && npm run build`: passed (8,097 modules transformed).
- A local full backend-suite attempt reached 3,054 tests with 0 failures, 18 errors, and 4 skipped. The errors were outside the changed paths and required unavailable local MySQL-backed contexts or the Claude Code CLI expectation in `ClaudeCodeAgentProviderTest`.

### Checklist

- [ ] Architecture constraints stay green (`mvn test` runs the ArchUnit checks; the full local suite is environment-blocked as noted above)
- [x] Documentation touched where behaviour changed (no documentation update required for this bug fix)


RockteMQ-AI (approved):
### Summary
Moves subscription mode filtering from client-side to server-side, applying the filter **before** pagination rather than after. This fixes the bug where paginated results could have fewer items than expected because filtering happened post-pagination.

The implementation is clean and well-layered:
- **`SubscriptionModeFilters`**: Centralized normalize + filter logic with proper null/ALL/invalid handling
- **SQL-level filtering** in `RocketMQMetadataProvider`: Efficient DB-level push/pop filtering (`message_model IS NULL OR message_model <> Pop` correctly treats null as Push)
- **Backward compatible**: Old method signatures delegate to new ones with `null` subscriptionMode
- **Web**: Correctly removes client-side filtering, passes `subscriptionMode` to API, resets page to 1 on mode change

Test coverage is comprehensive across controller, service, and provider layers.

LGTM.

---

lizhimins:
Thanks for the thorough write-up — the diagnosis is accurate, and the shape of the fix is the right one: threading `subscriptionMode` down into the provider page query so the predicate joins the `LambdaQueryWrapper` before `selectPage`, with the total taken from the database count, is exactly how a filter on a server-paginated list has to work (and it matches what the CSV export path already does).

We are not taking it, though, and I would rather say why explicitly than leave the PR hanging.

What this fixes is a filtering inaccuracy on the consumer-group list: the Push/Pop selector filters the rows already loaded for the current server page while the pager reports the unfiltered total. In practice a single instance carries a small number of consumer groups, so the window where those two disagree is narrow, and the operator still reaches the groups by clearing the filter or paging. Against that, the change replaces signatures across the provider SPI (`InstanceProvider.listConsumerGroupsPage`, `MetadataProvider.listConsumerGroupsPage`, the Apache provider and its query wrapper) and touches 16 files — a wide contract change, carried by every provider implementation including the cloud ones, for a low-impact correction.

So this is "not planned" rather than "wrong": the reasoning in the PR is sound, we just do not think the SPI churn is justified by what it buys today. Closing it, and no rework is expected on your side. If the subscription-mode filter ever becomes a correctness problem at realistic group counts, that is the moment to revisit the pushdown.

The same applies to #4813, which attacked the same defect from the service layer — that one is closed too, with the specifics in its own thread.

Issue #4774, opened 2026-09-21
[Bug] Message downloads silently change large JSON integers


Wang1rrr:
### Before Creating the Bug Report

- [x] I searched open and closed issues/PRs and believe this is not a duplicate.
- [x] This concerns RocketMQ Studio, not another RocketMQ repository.
- [x] The exact inspected version and reproduction scope are stated below.

### Studio Version

- Branch: `rocketmq-studio`
- Commit: `4c697f07acde460e2344375cb1f82669f5b270fd`
- Reproduction: source-function probe, not a deployed Studio instance.

### Runtime Environment

Windows, Node.js `v22.23.2`. The probe executes the checked-in formatter and download
handler, with Node's real `Blob` and an in-memory replacement for `downloadBlob`.
No browser or MySQL instance was involved.

### Connected RocketMQ Cluster

None. A synthetic `MessageRecord.body` string is sufficient to reproduce this
client-side transformation; no broker or customer data is needed.

### Describe the Bug

The ordinary message-page download handler runs `record.body` through `formatBody`
before creating the JSON Blob. That formatter calls `JSON.parse` followed by
`JSON.stringify`, so JSON integers outside JavaScript's safe integer range can be
silently changed in the downloaded content.

For example, the incoming body string `{"orderId":9007199254740993}` produces a
download containing `9007199254740992`. The incoming field is already a string;
this particular precision loss happens inside the UI formatter, not while decoding
the outer HTTP response.

### Steps to Reproduce

From a checkout of the commit above, save this as a temporary `.mjs` file and run
it with Node.js 22.23.2, using the repository root as the working directory:

```js
import assert from 'node:assert/strict';
import { readFileSync } from 'node:fs';
import { stripTypeScriptTypes } from 'node:module';
import vm from 'node:vm';

const source = readFileSync('web/src/pages/instance/message.tsx', 'utf8');
const formatter = source.match(
  /^const formatBody = \(body: string\): string => \{[\s\S]*?^\};/m,
)?.[0];
const handler = source.match(
  /^  const handleDownload = \(record: MessageRecord\) => \{[\s\S]*?^  \};/m,
)?.[0];
assert.ok(formatter && handler, 'Source changed: inspect the two functions first');
let captured;
const context = vm.createContext({
  Blob,
  downloadBlob: (blob) => { captured = blob; },
  message: { success: () => {} },
  t: (key) => key,
});
vm.runInContext(stripTypeScriptTypes(
[...]

PR #4777, opened 2026-09-21
fix(message): preserve the original body in downloads and copies

Wang1rrr:

Fixes #4774

### Brief Description

The message page's display formatter parses and re-serializes JSON. Using that
formatted string for download or copy changes unsafe integers and original
whitespace, even though the received message body is already a string.

This preserves `record.body` directly in the download Blob and supplies the
original body as Ant Design's explicit copy text. It retains the existing
pretty-printed display and changes no backend API, dependencies, MIME type, or
filename behavior.

The display itself can still round numbers: this narrowly scoped fix makes
download/copy lossless relative to the received body string, not the presentation
formatter. It does not recover backend-truncated or encoded data. A different
display policy or lossless JSON parser would be a separate design discussion.

Node.js 22.23.2 on Windows; dependencies installed from the existing lockfile with
`npm ci --ignore-scripts --no-audit --no-fund`.

Seven component regressions exercise real message-page buttons:

- Six download cases: unsafe integer, int64 maximum, safe integer, quoted ID,
  original JSON whitespace/Unicode, and plain text. The actual Blob content,
  filename and MIME type are checked at the `downloadBlob` boundary.
- One details-copy case exercises Ant Design's copy path and inspects the selected
  original text at the `document.execCommand('copy')` boundary.

Red/green evidence:

```text
Existing MessagePage tests before changes: 14 passed
New regressions on the original component: 4 failed, 17 passed
After the two production-line changes:      21 passed
```

Commands run from `web/`:

```text
npm test -- src/pages/instance/__tests__/MessagePage.test.tsx --maxWorkers=2
  1 file, 21 tests passed
npm test -- --maxWorkers=2
  139 files, 1287 tests passed (604.26s)
npm run lint
  0 errors, 11 pre-existing warnings in unmodified files
npm run build
  tsc -b and Vite production build passed
npx --no-install prettier --check src/pages/instance/message.tsx src/pages/instance/__tests__/MessagePage.test.tsx
  passed
git diff --check
  passed
```

The initial uncapped full-suite attempt was interrupted because of slow concurrent
execution; the complete capped run above is the verification result. jsdom emits
existing unsupported pseudo-element/navigation warnings, but the suite exits 0.

Browser scope: the existing mock UI was queried and its message details opened.
Clipboard readback was empty and the download-event wait timed out in the test
browser, so browser export E2E is **not** claimed. No backend/Broker E2E was run.

This change was prepared with AI assistance, independently reviewed, and verified
with the actual commands above.

### Checklist

- [x] One coherent change; no unrelated modifications
- [x] Conventional Commit subject
- [x] Regression tests added and observed failing before the fix
- [x] No new dependencies or lockfile changes
- [x] No new UI strings or source files requiring translation/license updates
- [x] Existing formatted presentation and response contracts remain unchanged
- [ ] Backend architecture tests and Broker E2E (not run; frontend-only change)


RockteMQ-AI (approved):
### Summary
Fixes message download/copy to preserve the original body string instead of the re-serialized JSON from the display formatter. This prevents precision loss for large integers (e.g., int64 IDs) and preserves original whitespace.

The change is well-scoped — only the download Blob and copy text are affected; the display still uses `formatBody` for pretty-printing. Test coverage is thorough with 7 download scenarios and 1 copy scenario covering unsafe integers, int64 max, safe integers, quoted IDs, whitespace, and plain text.

LGTM.

---

Issue #4778, opened 2026-09-21
[Bug] rmqctl message query tables hide result completeness metadata


Wang1rrr:
## Summary

`rmqctl message query` and `message query-by-topic` hide the response's completeness metadata in their default table output whenever `items` is nonempty. An incomplete result and a complete result with the same rows produce identical stdout and no warning on stderr. JSON output preserves the metadata.

## Environment

- `rocketmq-studio`, CLI source at `4c697f07acde460e2344375cb1f82669f5b270fd`.
- Windows amd64, Go 1.27.1; unmodified production CLI.
- Synthetic loopback HTTP response shaped to the current tool schema; no live Studio backend or RocketMQ broker.

## Reproduce

Build the CLI from the revision above (`cd rmqctl; go build -o rmqctl-repro.exe .`). Save the following as `repro.py`, then run `python repro.py ./rmqctl-repro.exe` with Python 3.9+. It uses isolated dummy credentials, temporary local config, and an ephemeral 127.0.0.1 server. Nothing connects to a real account or instance.

The server sends the normal response envelope:

```json
{"code":200,"message":"success","data":{"items":[{"msgId":"mock-message-1","topic":"mock-topic","storeTime":1,"size":1}],"resultMayBeTruncated":true,"skippedCount":5}}
```

<details>
<summary>Self-contained production CLI reproducer</summary>

```python
"""Observe current CLI table metadata loss, using only a temporary loopback server."""
import json
import os
from pathlib import Path
import subprocess
import sys
import tempfile
import threading
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

cli = str(Path(sys.argv[1]).resolve())
item = {"msgId": "mock-message-1", "topic": "mock-topic", "storeTime": 1, "size": 1}
payload = {"items": [item], "resultMayBeTruncated": True, "skippedCount": 5}
requests = []

class Handler(BaseHTTPRequestHandler):
    def do_POST(self):
        request = json.loads(self.rfile.read(int(self.headers["Content-Length"])))
        expected = {"instanceId": "mock-instance", "topicName": "mock-topic", "limit": 1}
        if request.get("name") == "rmq.message.query":
            expected["key"] = "mock-key"
        valid = (self.client_address[0] == "127.0.0.1" and self.path == "/api/mcp/tools/call"
                 and request.get("name") in ("rmq.message.query", "rmq.message.query_by_topic")
                 and request.get("arguments") == expected)
        if not valid:
[...]

PR #4779, opened 2026-09-21
fix(cli): warn when message query table results may be incomplete

Wang1rrr:

Fixes #4778

### Brief Description

Both `rmqctl message query` and `message query-by-topic` already receive
`resultMayBeTruncated` and `skippedCount`, but their nonempty table renderer drops
those fields. Identical rows therefore look the same whether the result is
complete or provider/limit-bounded.

After successfully writing the existing table, emit a warning to stderr only for
these two tools when `resultMayBeTruncated` is the boolean `true`. Include
`skippedCount` with its bounded meaning: rows omitted by the requested limit from
the provider-bounded result, not the total number of messages missing from the
broker. Warn even when the count is zero.

Table stdout remains byte-for-byte unchanged. JSON/YAML, the empty-table JSON
fallback, other tools and mutation summaries remain unchanged. Writer errors are
propagated. No extra request, API/schema/flag change or dependency is introduced.

Compatibility: a successful partial table query now has nonempty stderr while
still exiting 0. Scripts that equate any stderr with failure will observe the
warning. This is an output repair, not a change to server truncation detection.

Windows amd64, Go 1.27.1, existing locked dependencies, isolated LF checkout.
The tests invoke the real CLI command path against a synthetic HTTP server:

- Two commands x three formats x four states: limit-truncated, provider-bounded
  with skippedCount=0, complete, and empty. Fixed expected table bytes and
  structured payloads are checked, as is exactly one request.
- Other table and mutation output controls, stdout/stderr write failures, and
  exact tool/boolean warning scope.
- Red/green: the unchanged renderer failed four missing-warning assertions;
  the repaired renderer passes. The two command fixtures explicitly use limit=1.

Verification from `rmqctl/`:

```text
gofmt -l .                                      (no output)
go run ./internal/catalog/generate -input-dir ../server/src/main/resources/tool-catalog/tools -output internal/catalog/catalog_gen.go -check
go test -count=1 ./...
go test -race -count=1 ./...
go vet ./...
go build .
```

All passed (test/build commands used `-p 2`). All six Makefile build-all target
pairs also cross-compiled with CGO_ENABLED=0: darwin, linux and windows, each on
amd64 and arm64. Cross-compilation is not a claim of execution on those targets.

An independent process-level comparison ran the unchanged baseline and fixed
CLI against the same 14 synthetic loopback cases each. All 28 commands succeeded;
all 14 stdout results were identical; only the four nonempty partial table cases
on the fixed CLI emitted warnings. Request signing and structured payloads were
also checked.

Independently reviewed, including a follow-up correction to fixture limit
consistency. Prepared with AI assistance. No live Java/Broker E2E, backend or
frontend suite was run for this Go-only change; remote CI is not claimed here.

### Checklist

- [x] One coherent, issue-linked fix on rocketmq-studio
- [x] Regression tests observed failing before the production change
- [x] No new dependencies, generated catalog changes or unrelated refactoring
- [x] ASF header on the new test source; test names use the requested Test suffix
- [x] Relevant local format/catalog/vet/test/race/build checks and independent review
- [ ] Live backend/Broker E2E (not run)


RockteMQ-AI (approved):
### Summary
Well-crafted fix that adds stderr warnings for truncated message query results while preserving stdout byte-for-byte. The implementation is minimal, focused, and thoroughly tested.

### Findings
- **[Info]** catalog_renderers.go:43-47 — The hardcoded tool names could be extracted to constants for maintainability, though this is a minor style preference.
- **[Info]** catalog_renderers.go:46 — The warning message is technically accurate but quite verbose. Consider whether a shorter message like `WARNING: Results may be incomplete (skippedCount=%v)` would be sufficient for CLI users.

### Strengths
- Excellent test coverage: 24 test cases covering 2 commands × 4 states × 3 formats, plus edge cases for writer errors and warning scope
- Proper error propagation: stdout write errors prevent the warning, stderr write errors are propagated after the table is written
- Type-safe boolean check (`== true`) prevents false positives from string "true" or other truthy values
- Clear PR description with compatibility impact analysis
- Stdout remains unchanged, maintaining backward compatibility for scripts parsing table output

### Suggestions
- Consider adding a brief comment in `catalog_renderers.go` explaining why only these two tools emit the warning (they're the only ones that return `resultMayBeTruncated` with bounded semantics)
- The test helper `executeCatalogResult` could be moved to a shared test utilities file if similar patterns are needed elsewhere

Overall: Clean implementation with comprehensive tests. The warning behavior is correctly scoped and the compatibility impact is well-documented.

---

Issue #4783, opened 2026-09-21
[Studio][Bug] An alert-rule update cannot clear an optional column: the response reports it cleared but the stored rule keeps the old value


unbridled-41:
### Studio Version

Trunk `rocketmq-studio` @ `4c697f07` (line references are as of that commit).

### Problem

`POST /api/alert-rules/update` and `POST /api/cluster-alert-rules/update` replace every editable field of a rule, but an update body that simply omits an optional field does not clear it. `MybatisPlusAlertRepository.replaceRule` (`server/src/main/java/org/apache/rocketmq/studio/ops/alert/MybatisPlusAlertRepository.java:141-146`) builds the entity with `toRuleEntity` and calls `ruleMapper.updateById(entity)`. MyBatis-Plus's default NOT_NULL update strategy omits null entity fields from the UPDATE's SET clause, so the previous value silently survives.

The service path makes the mismatch worse: `AlertService.updateRule` (`AlertService.java:190-223`) validates the submitted VO, replaces the stored rule, and returns the **submitted** VO. So the API responds with the field cleared (`consumerGroup`: null, `severity`: null, ...), the UI renders the cleared value, and the next `GET` shows the old value back.

Fields affected (all nullable in `rmq_alert_rule`, all accepted-as-null by `NativeAlertRulePolicy.validate` and `AlertRuleRequestDTO`): `thresholdUnit`, `duration`, `channels`, `description`, `brokerName`, `clusterName`, `severity`, `instanceId` (for Prometheus-rule metrics only — native metrics require it), `consumerGroup`, `topic`, `notificationTemplate`.

A concrete failure: a business `consumer.lag.total` rule scoped to `consumerGroup="G1"` cannot be re-scoped to all groups through the update API. The stale filter keeps narrowing alert evaluation while the operator believes it was removed.

### Evidence

1. `NativeAlertRulePolicy.validate` only rejects a missing `instanceId` for native metrics (`NativeAlertRulePolicy.java:66-71`); a null `consumerGroup`/`topic`/`severity`/`duration` passes (`AlertRuleDuration.parse(null)` returns ZERO).
2. `replaceRule` → `ruleMapper.updateById(toRuleEntity(rule))` with the entity's optional fields null — MyBatis-Plus skips them in the SET clause (no global `update-strategy` override in `application.yml`, no `@TableField(updateStrategy=...)` on `RmqAlertRule`).
[...]

PR #4784, opened 2026-09-21
fix(alert): persist a cleared optional column when an alert-rule update omits it

unbridled-41:
Fixes #4783.

## Problem / Evidence

An alert-rule update that **omits** an optional field cannot clear it. `MybatisPlusAlertRepository.replaceRule` built the entity and called `ruleMapper.updateById`, whose default NOT_NULL strategy drops null fields from the SET clause. Because `AlertService.updateRule` validates the submitted VO, replaces the stored rule and returns the **submitted** VO, the API responded with the field cleared while the stored rule kept the old value; the next `GET` showed it back. A rule scoped to `consumerGroup="G1"` could not be re-scoped to all groups; `severity`/`duration`/`channels`/`description`/`notificationTemplate` silently reverted after an update that omitted them. Line references in the linked issue are as of trunk `4c697f07`; references below are as of the patched file at this PR's head.

Reproduction path (issue #4783 has the full chain): update a business `consumer.lag.total` rule without `consumerGroup` → 200 OK, response shows `consumerGroup: null` → GET shows the old group still stored and still applied at evaluation time.

## Root cause / Fix

MyBatis-Plus `updateById` omits null entity fields. `replaceRule` now follows every cleared optional column with an explicit `UpdateWrapper ... set(column, null)` pass, the same approach `NameserverRegistryService.clearOmittedOptionalColumns` ships on trunk (merged #4466, same defect class for instance and registry optional columns) and `MybatisPlusInstanceRepository` uses for a cleared credential reference. `lastTriggered` is not an editable field — it is owned by `markRuleTriggered` — and keeps the skip-on-null behaviour; the regression pins that boundary.

## Priority

PRIORITY 68 / 100 — impact 26 (silent response-vs-stored-state divergence on a management API; scope filters keep applying after "removal"), blast radius 14 (every alert-rule update path: REST, import/transfer, bulk), reproducibility 18 (deterministic once the update body omits a field), maintenance value 10 (aligns the last table with the project's own established fix pattern; #4466 shows maintainers accept it). FIX_CONFIDENCE 92: single-method fix, direct precedent in-tree, no semantic ambiguity beyond the `lastTriggered` boundary, which the test pins.

## Tests

- `MybatisPlusAlertRepositoryTest.replaceRuleShouldExplicitlyClearOmittedOptionalColumnsTest` (new): an update VO with every optional field null must assign `threshold_unit`, `duration`, `channels`, `description`, `broker_name`, `cluster_name`, `severity`, `instance_id`, `consumer_group`, `topic`, `notification_template` to null via the explicit wrapper, and must **not** include `last_triggered`.
- Teeth, measured on this machine: with only the test change on base `4c697f07` source, the test fails with `Wanted but not invoked: ruleMapper.update(...)` (red). With the fix restored it passes (green). A second discriminating run — a plausible wrong fix that also clears `last_triggered` — fails on the `doesNotContain("last_triggered=")` assertion, so the test separates correct from over-clearing implementations.
- `replaceRuleShouldSkipTheClearPassWhenEveryOptionalFieldIsPresentTest` (second commit): an update carrying every optional value issues `updateById` and nothing else — pinning the "only runs when needed" claim made in Risk for the fully-loaded callers (`toggleRule`, bulk toggle/delete, import).
- `MybatisPlusAlertRepositoryTest` 18/18 (17 + the new boundary test), plus related suites: `AlertServiceTest` 83/83, `NativeAlertProcessorTest` 23/23, `AlertRuleControllerTest` 22/22, `ClusterAlertRuleControllerTest` 6/6. `NativeAlertEvaluationTransactionTest` errors in this sandbox because it needs MySQL (Spring context `Communications link failure`) — the same baseline failure as clean trunk, unrelated to this change. `mvn checkstyle:check` passes.

## Risk

Low. The added statements only run when a cleared column would otherwise be silently skipped. Callers that pass fully-loaded VOs (`toggleRule`, bulk toggle/delete, import) are unaffected: their loaded optional fields are non-null or were already null in storage, making the explicit assignment idempotent — and for non-null fields the clear pass does not issue any statement at all, which the second-commit test now pins. Rules that genuinely rely on `duration`-null semantics keep them: `AlertRuleDuration.parse(null)` already returns ZERO.


RockteMQ-AI (approved):
### Summary
Fixes two data consistency issues: (1) MyBatis-Plus `NOT_NULL` field strategy silently skipped null columns during updates, meaning clearing an optional field (like `deadLetterTopic`) had no effect on the database — the fix uses `UpdateWrapper` with explicit `.set()` calls for nullable fields; (2) conversation deletion now cleans up associated workspace files to prevent orphaned directories. The `UpdateWrapper` approach is the correct MyBatis-Plus pattern for explicit null persistence. Tests cover both the update and delete paths with proper verification.

LGTM.

---

PR #4785, opened 2026-09-21
fix(nameserver): report unreachable endpoints in the config read tool

Frun1na:

- Fixes #4786

### Brief Description

`rmq.nameserver.config` presented a partial view as complete. When some NameServer endpoints of a
cluster could not be read, `NameServerConfigDiffService.read` logged a warning and skipped them
(`NameServerConfigDiffService.java:171-174`), and the tool returned only the reachable endpoints —
`ListOutput` has no completeness channel, so an AI or an operator reading the result could mistake
it for the full cluster configuration. The REST diff path already had the right idea: `compare()`
sets `complete(reachableNodeCount == addresses.size())` and reports per-node reachability; the
tool path simply had no such channel.

Following the trunk's truncation-signal vocabulary (`MessageQueryOutput.resultMayBeTruncated` +
`skippedCount`, #4388) instead of throwing 502 for a partially-successful read:

- `read()` now returns a `NameServerConfigRead` record: the reachable `NodeConfig`s plus the
  unreachable addresses. Behaviour is otherwise unchanged — 502 only when nothing is reachable.
- The tool output is a new `NameserverConfigOutput` contract record: `items` (unchanged shape),
  `resultMayBeTruncated` (true when some endpoints could not be read), and `unreachableEndpoints`
  (which ones), so the caller can weigh the partial coverage and retry instead of drawing
  cluster-wide conclusions from a subset.
- The catalog `outputSchema` for `rmq.nameserver.config` gained the two required fields with
  caller guidance, `catalog_gen.go` was regenerated (`make catalog-verify` passes), and
  `ToolOutputSchemaContractTest` pins the new shape.

```
$ mvn -B -ntp test -Dtest='NameServerConfigDiffServiceTest,NameServerConfigToolHandlerTest,ToolOutputSchemaContractTest' -Dspring.profiles.active=dev
[INFO] Tests run: 13, Failures: 0, Errors: 0, Skipped: 0   (3 test classes)

$ cd rmqctl && make catalog-verify
# passes; go vet, go test ./... also green after regeneration
```

New/updated assertions:

- `NameServerConfigDiffServiceTest.readShouldSkipUnreachableEndpointsTest` now asserts the
  unreachable set is reported (`containsExactly("ns-b:9876")`) alongside the reachable nodes.
- `NameServerConfigToolHandlerTest.reportsUnreachableEndpointsAsTruncatedCoverageTest` (new) pins
  `resultMayBeTruncated=true` + `unreachableEndpoints=["ns-b:9876"]` when one endpoint fails, and
  the existing case pins `false` + empty when all endpoints answer.
- The contract-test sample for `rmq.nameserver.config` carries the truncated shape so the schema
  and the serialized output stay in contract.

### Checklist

- [ ] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text touched)
- [x] Documentation touched where behaviour changed (tool catalog + generated docs)


RockteMQ-AI (approved):
### Summary
Replaces the generic `ListOutput` with a dedicated `NameserverConfigOutput` that includes a `resultMayBeTruncated` flag and surfaces unreachable endpoints. This is important for AI tool consumers — silently dropping unreachable nodes would present a partial config as complete, which could lead to incorrect diff assessments. The `NameServerConfigRead` record cleanly separates the two data sources, and the tool catalog schema is updated to reflect the new output shape.

LGTM. Good improvement in operational visibility.

---

PR #4762, opened 2026-09-21
test(web): align API test fixtures with the backend contract

Frun1na:
### Which Issue(s) This PR Fixes

- Fixes #4763

### Brief Description

Several api-layer test fixtures and mock datasets carried values the server never sends — the same
pattern that let the consumer page render raw filter-mode codes (#4732) and lose protocol tag colors
(#4741) for weeks behind a green suite:

| file | fixture value | what the backend actually returns |
|------|----------------|-----------------------------------|
| `api/liteTopic.test.ts` | session `status: 'active'`, `ttlStatus: 'active'` | `RocketMQLiteTopicProvider` sets `"ACTIVE"`, and `LiteTopicSummary.getTTLStatus()` returns `UNKNOWN / EXPIRED / EXPIRING_SOON / ACTIVE` — all compared case-sensitively by the LiteTopic page filter |
| `api/dlq.test.ts` | `status: 'active'` | `RocketMQDLQProvider` emits `ACTIVE / EMPTY / UNAVAILABLE`; `mock/dlq.ts` used the same stale pair and its union could not even express `UNAVAILABLE` |
| `api/producer.test.ts` | `language: 'JAVA'` | `ProducerConnectionService` serializes `ClientLanguage.name()`, which is `Java` |
| `api/message.test.ts` | `deliveryStatus: 'SUCCESS'` | the `DeliveryStatus` enum is lowercase `success / failed / pending` |
| `api/ops.test.ts` | `level: 'critical'` | `AlertLevel` is `error / warning / info` (the page-level test that deliberately feeds an unknown level keeps doing so; the api passthrough test now uses a real value) |
| `mock/consumers.ts` | `filterMode: 'Tag 过滤' \| 'SQL92 过滤' \| '全量'` | `SubscriptionFilterModes` normalizes to `TAG / SQL / CLASS_FILTER`; mock mode now renders like a real server after #4732, and the `subscriptionDataType` union gains `TopicType.LITE` |

No production code changed; the fixtures now exercise the values the API actually returns, so a
contract regression fails the suite instead of hiding behind stale mocks.

### How Did You Test This Change?

```
$ npx vitest run src/api/liteTopic.test.ts src/api/dlq.test.ts src/api/producer.test.ts \
      src/api/message.test.ts src/api/ops.test.ts src/services/consumerService.test.ts
 Test Files  6 passed (6)
      Tests  75 passed (75)

[...]

PR #4787, opened 2026-09-21
test(web): pin mock and contract values to what the backend records

Frun1na:

- Fixes #4788

### Brief Description

Two test-only fixtures pinned values the backend never sends — the same mock-vs-contract family
as #4732's review item 4 and #4762:

**`web/src/mock/audit.ts`.** Mock mode derives the audit filter options from these rows
(`getMockAuditFilterOptions` in `services/opsService.ts`), so the offered filters advertised
operations that cannot exist:

- `RESTART_BROKER` (2 rows): `git grep RESTART_BROKER -- server/src/main/java` finds no producer;
  it is absent from `OperationAuditConstants` and from every `auditService.record(...)` call.
  Replaced with the real `RESEND_DLQ`/`DLQ` and `ADD_PROXY_ADDRESS`/`PROXY` pairs
  (`RocketMQDLQProvider.java:695`, `OperationAuditConstants.Operation.ADD_PROXY_ADDRESS`).
- `UPDATE_ACL` + resourceType `TOPIC` (2 rows): ACL changes are recorded as
  `CREATE_ACL_RULE`/`UPDATE_ACL_RULE`/`DELETE_ACL_RULE` against `ACL_RULE`
  (`AclService.java:117,158,172`), with the rule id as the target. The fixtures now use
  `UPDATE_ACL_RULE`/`ACL_RULE` with id targets.
- `UPDATE_BROKER_CONFIG` + resourceType `CLUSTER`: the service records it against `BROKER`
  (`RocketMQBrokerConfigService.java:94`); `UPDATE_CLUSTER_CONFIG`/`CLUSTER` is the separate,
  real pair (`ClusterService.java:414`) and stays as-is.

**`web/src/api/clusterContract.test.ts`.** The "contract" fixture pinned `status: 'RUNNING'` on
the cluster, its broker, its proxy and its NameServer — a value the API never serializes:
`ClusterStatus` is `healthy/warning/error/offline` and `BrokerStatus` is
`running/readonly/maintenance`. The fixture now uses the real values, so the contract test pins
the actual contract instead of a hypothetical one.

```
$ npx vitest run src/services/opsService.test.ts src/api/clusterContract.test.ts
 Test Files  2 passed (2)
      Tests  15 passed (15)

$ npx tsc --noEmit        # no output, exit 0
$ npx eslint src/mock/audit.ts src/api/clusterContract.test.ts   # no findings
```

The `derives complete filter options from the default audit records` test still passes because
`CREATE_TOPIC`, `DELETE_GROUP`, `RESET_OFFSET`, `CLUSTER`, `TOPIC`, `GROUP` remain represented by
real rows.

### Checklist

- [x] Commit subject follows Conventional Commits (`test:`)
- [ ] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text touched)
- [ ] New source files carry the ASF license header (no new source files)
- [ ] Documentation touched where behaviour changed (no behaviour change; test/mock only)


RockteMQ-AI (approved):
### Summary
Aligns test fixtures and mock data with the actual backend contract — status values (`healthy`/`running` instead of `RUNNING`), audit operation/resource types (`UPDATE_ACL_RULE`/`ACL_RULE` instead of `UPDATE_ACL`/`TOPIC`), and removes operations that don't exist in the backend (`RESTART_BROKER`). This prevents mock-mode UI from advertising filter options and states that the real API never produces.

LGTM. Clean test hygiene.

---

Issue #4789, opened 2026-09-21
[Studio][Bug] Resource counts and the delete guard resolve the numeric instance id through the name-first lookup, so an instance named like another id reads the wrong instance


unbridled-41:
### Studio Version

Trunk `rocketmq-studio` @ `4c697f07` (line references are as of that commit).

### Problem

`InstanceService` holds an already-resolved instance but hands its **numeric id as a string** to the vendor provider for resource counts:

- `loadCounts` (`server/src/main/java/org/apache/rocketmq/studio/instance/InstanceService.java:180-186`, used by the instance list via `fillCountsInParallel`): `provider.countTopics(String.valueOf(instance.getId()))`.
- the delete guard (`InstanceService.java:654-658`): `provider.countTopics(String.valueOf(id))` decides whether an instance may be deleted.

Every provider resolves that string through `InstanceRepository.findByIdentifier` (`InstanceRepository.java:42-57`), which tries the unique **name first** and only falls back to the numeric id. Instance names are free-form (up to 64 chars, digits allowed — `requireInstanceName` only trims and length-checks), so an instance literally named `42` shadows the instance whose primary key is `42`:

1. The instance list row for the numeric instance shows the named instance's topic/group counts (or a misleading 400 "missing cloud binding" when the named instance is Apache and the counted one is cloud).
2. Worse, the destructive path: `POST` delete of the numeric instance evaluates its managed-resource guard against the **named** instance. An operator can delete an instance whose topics/groups still exist, orphaning them; and the opposite direction blocks a legitimately empty instance.

### Evidence

1. `findByIdentifier` → `findByName(identifier)` before `findById(Long.parseLong(...))` with no rejection of pure-digit names (`InstanceRepository.java:46-51`).
2. `requireInstanceName` (`InstanceService.java:554-562`) imposes no charset constraint, and `createApacheInstance` only requires name + endpoint, so a numeric name is storable through the public API today.
3. The codebase's own javadoc on `findByIdentifier` says the name is primary and the id a fallback — callers that already hold the resolved instance must not round-trip through that ambiguous resolution.
4. `ApacheInstanceProvider.countTopics` (`provider/apache/ApacheInstanceProvider.java:76-81`) maps through `findByIdentifier(...).getName()` then counts by that name — so the string is resolved by name first, confirming the shadowing.

### Impact

[...]

PR #4790, opened 2026-09-21
fix(instance): resolve resource counts by the canonical instance name

unbridled-41:
Fixes #4789.

## Problem / Evidence

`InstanceService.loadCounts` (instance list) and the `deleteInstance` managed-resource guard held an already-resolved instance but passed `String.valueOf(instance.getId())` to the vendor provider. Every provider resolves that string through `InstanceRepository.findByIdentifier`, which matches the unique **name** before the numeric-id fallback. Because instance names are free-form (digits allowed, `requireInstanceName` only trims/length-checks), an instance named `42` shadows the instance whose primary key is `42`:

- the list shows the named instance's topic/group counts on the numeric instance's row (or a misleading "missing cloud binding" 400 across vendors),
- the delete guard reads the wrong instance's counts, so deleting a non-empty instance can succeed and orphan its managed resources, while a genuinely empty one can be blocked with another instance's counts.

## Root cause / Fix

Callers that already hold the resolved `InstanceVO` must not round-trip through the ambiguous identifier resolution. Both call sites now pass the canonical `instance.getName()`, which `findByIdentifier` always resolves to the instance itself. No provider or repository behaviour changes; the numeric-id fallback stays for external API callers that address instances by id.

## Priority

PRIORITY 70 / 100 — impact 28 (wrong data on the fleet-wide instance list plus a destructive-path guard reading the wrong instance's state: delete succeeds for a non-empty instance and orphans its resources), blast radius 12 (Apache/cloud list counts and the single delete guard; batch delete routes through the same guard), reproducibility 16 (deterministic once a numeric name exists — storable through the public create API today), maintenance value 14 (two call sites, one contract, matches the project's own "name is the immutable external identifier" stance from #2050). FIX_CONFIDENCE 95: minimal and semantically forced — the resolved object's canonical name cannot shadow anything.

## Tests

- `InstanceServiceTest.deleteInstanceShouldResolveResourceCountsByInstanceNameTest` (new): the guard receives `countTopics("with-topics")` and never `countTopics("1")`.
- `InstanceServiceTest.listInstancesShouldResolveResourceCountsByInstanceNameTest` (new): an instance named `42` with id 3 gets counts via `countTopics("42")`, never `countTopics("3")`.
- `batchDeleteShouldRouteItsResourceGuardThroughTheCanonicalNameTest` (second commit): batch delete funnels into the same guard — an instance named `42` (id 3) is deleted only after the guard sees `countTopics("42")`, never `countTopics("3")`.
- Teeth, measured on this machine: with only the test change on base `4c697f07` source, both tests fail (stubbed name never called → `Wanted but not invoked`/zero counts); the batch-delete test fails the same way on unguarded source. With the fix restored, `InstanceServiceTest` 92/92.
- Pre-existing tests updated to the canonical-name stubbing contract (11 tests): they stubbed `countTopics("1")` etc., encoding exactly the behaviour being fixed; assertions and semantics unchanged otherwise.
- Related suites: `ApacheInstanceProviderTest` 11/11, `AliyunInstanceProviderTest` 35/35, `TencentInstanceProviderTest` 50/50, `MybatisPlusInstanceRepositoryTest` 20/20 (208/208 with the service tests). `mvn checkstyle:check` passes.
- Audit note: a teeth-check restore of the source file was accidentally committed on top (HEAD briefly carried unguarded call sites). Re-applied in the third commit and re-verified: the net source diff trunk→head is exactly the two canonical-name call-site changes; all 208 tests were re-run green on the final head.

## Risk

Low. For every well-formed deployment the canonical name and the id string resolve to the same instance, so behaviour is identical; the change only removes the ambiguity window. Cloud providers re-resolve by name and read their own binding — unchanged for non-shadowed names.


RockteMQ-AI (approved):
### Summary
Fixes an identifier resolution bug where `String.valueOf(instance.getId())` was passed to providers that resolve by name first via `InstanceRepository.findByIdentifier`. Using `instance.getName()` directly avoids the shadowing risk where a name matching another instance's numeric id would misattribute counts. The fix is applied consistently in both `loadCounts` and `deleteInstance`. Tests verify the correct name is passed.

LGTM.

---

PR #4791, opened 2026-09-21
fix(message): report an unreadable query-history snapshot as a failure

Frun1na:

- Fixes #4792

### Brief Description

`QueryHistoryService.getMessageQueryResults` treated a stored result snapshot the mapper cannot
parse exactly like a legitimate empty result: it logged a warning and returned `List.of()`
(`QueryHistoryService.java:152-156`). The two cases have different meanings under the project's
error semantics — a blank snapshot is a valid "nothing stored", while an unparseable snapshot is a
persistence-level failure (the stored result rows are lost, e.g. after a schema or serialization
change). Reporting an empty list presents that loss as a query that legitimately matched nothing,
and the web history view renders it as "0 messages" with no hint anything went wrong.

A corrupt snapshot now throws `BusinessException(502, "Stored query result snapshot is
unreadable")`; a blank snapshot still returns an empty list, with each case pinned by its own
test so the distinction is contractual rather than incidental.

```
$ mvn -B -ntp test -Dtest='QueryHistoryServiceIntegrationTest' -Dspring.profiles.active=dev

# mutation check — fix reverted (return List.of() on JsonProcessingException):
[ERROR] Tests run: 3, Failures: 1
[ERROR] ...unreadableResultSnapshotIsReportedAsFailureTest <<< FAILURE!

# fix applied:
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

- `unreadableResultSnapshotIsReportedAsFailureTest` (new): inserts a row whose snapshot is not
  JSON, asserts a 502 BusinessException — fails on the base revision.
- `blankResultSnapshotMeansAnEmptyResultTest` (new): a blank snapshot still yields an empty list.


RockteMQ-AI (approved):
### Summary
Correctly distinguishes a blank snapshot (legitimate empty result) from an unparseable one (persistence failure). Throwing `BusinessException(502)` instead of silently returning `List.of()` prevents data loss from being presented as a successful empty query. Tests cover both paths with proper cleanup.

LGTM.

---

Issue #4793, opened 2026-09-21
[Studio][Bug] The NameServer registry create/edit modal allows duplicate submissions while a request is in flight


unbridled-41:
### Studio Version

Trunk `rocketmq-studio` @ `4c697f07` (line references are as of that commit).

### Problem

The NameServer registry create/edit modal (`web/src/pages/cluster/index.tsx:1854-1863`) wires `onOk` straight to `handleNsSubmit` with neither `confirmLoading` nor an in-flight guard. `handleNsSubmit` (`index.tsx:284-316`) is async and only closes the modal after `createNameserverRegistry`/`updateNameserverRegistry` resolves, so the OK button stays clickable for the whole network round trip. Every other submitting modal on the same page wires `confirmLoading` (connect modal: `confirmLoading={connectTesting}`; config modal: `confirmLoading={configSubmitting}`); this one was missed.

### Evidence

1. `handleNsSubmit` awaits `nsCreateForm.validateFields()` and then the create/update call; the modal only closes in the success branch. A second OK click during the await re-enters the handler, re-validates (fields are still filled) and issues a second identical POST.
2. The backend `createNameserverRegistry` endpoint accepts any unique name/address pair, so two identical submissions create two registry entries — `rmq_nameserver` has no unique constraint on (`name`, `namesrv_addr`) that would deduplicate a client double-click (the migration at `deploy/mysql/upgrade-nameserver-name-uk.sql` only covers the name key).
3. Measured in this sandbox: with a deferred create promise, clicking 确认 twice posts twice (test red on trunk source, green with the fix).

### Impact

Duplicate NameServer registry entries appear in the cluster registry, each address then renders as its own cluster in the registry tables and the clients page's NameServer dropdown; the operator has to find and delete the twin manually. The same window applies to edits, where the last write wins and can interleave two different edits of the same entry.

### Expected behavior

While a create/update request is in flight, the OK button shows the request state (confirmLoading) and further clicks are ignored — the same contract every other submitting modal on the page already implements.

### Related work

- #4618 (open): per-loader error state in the same page's registry loaders; unrelated to the modal.
- #4559 (open): registry config-update scoping; different modal, same page.
- #4754 (merged): the only prior duplicate-submission report family is the AI composer; no existing issue tracks the registry modal.
- #3103 (open): registry broker configuration actions targeting the wrong instance — same page's registry surface, a scoping defect rather than duplicate submission; confirmed not the same issue.

## PR

Fix: #4794.

[...]

PR #4794, opened 2026-09-21
fix(cluster): block duplicate NameServer registry submits while one is in flight

unbridled-41:
Fixes #4793.

## Problem / Evidence

The NameServer registry create/edit modal keeps its OK button enabled for the whole request: `handleNsSubmit` has no in-flight guard and the modal has no `confirmLoading`, unlike the sibling connect (`confirmLoading={connectTesting}`) and config (`confirmLoading={configSubmitting}`) modals on the same page. A second 确认 click while the create is on the wire re-validates the (still-filled) form and POSTs the same registry entry again — two identical NameServer registry rows appear, each address then renders as its own cluster in the registry tables and the clients page's NameServer dropdown. The same window applies to edits, where two concurrent edits of one entry interleave last-write-wins.

## Root cause / Fix

`handleNsSubmit` now guards re-entry with an `nsSubmittingRef` (ref, so the async handler reads current state) plus an `nsSubmitting` state wired to `confirmLoading`, matching the pattern the page's other modals already use. The guard is released in `finally` so a failed request restores the button.

## Priority

PRIORITY 66 / 100 — impact 24 (duplicate persistent registry entries from a routine double-click; cleanup requires manually finding and deleting the twin), blast radius 12 (registry create + edit on the cluster page; feeds registry tables and the clients NameServer filter), reproducibility 18 (deterministic with any slow create request), maintenance value 12 (aligns the last unguarded modal on the page with its own established confirmLoading pattern). FIX_CONFIDENCE 90: same-page precedent, two-line guard, behaviour fully pinned by a test.

## Tests

- `ClusterPage.test` › `ignores extra confirm clicks while a NameServer create is in flight` (new): with the create deferred, one click posts once; two further clicks post no more; after the promise resolves the modal closes with the call count still 1.
- Second commit: the same test also rejects the first request and confirms again, pinning the failure-path promise made in Risk — a rejected create releases the guard and the retry reaches `createNameserverRegistry` with the same payload.
- Teeth, measured on this machine: with only the test change on trunk source the test fails — after flushing the handler's microtasks, `createNameserverRegistry` was called twice (`Expected: 1, Received: 2`); with the fix, `ClusterPage.test` is 30/30 including the recovery half (tsc and eslint re-run clean on the final head).
- `npx tsc -b` clean, `npx eslint` on both changed files clean.

## Risk

Low. The guard only suppresses re-entry while the request is in flight; the failure path releases it so a rejected create can be retried immediately. No API or data-shape changes.


RockteMQ-AI (approved):
### Summary
Fixes a race condition where rapid confirm clicks could submit duplicate NameServer registry entries while a create request is in flight.

The implementation is solid:
- Uses `nsSubmittingRef` (useRef) for synchronous guard in the callback
- Uses `nsConfirmLoading` (useState) for UI feedback via `confirmLoading`
- Guard is properly released in `finally` block, allowing retries after failures
- Clear separation between synchronous guard (ref) and async state (state)

Test coverage is comprehensive — verifies duplicate clicks are ignored and the guard releases on failure for retry.

LGTM — clean fix for a real UX issue.

---

Issue #4795, opened 2026-09-21
[Studio][Bug] A failed agent-workspace preparation strands the run row in QUEUED and the conversation answers 409 until the orphan sweep


unbridled-41:
### Studio Version

Trunk `rocketmq-studio` @ `4c697f07` (line references are as of that commit).

### Problem

`AiRunService.sendMessage` (`server/src/main/java/org/apache/rocketmq/studio/ops/ai/conversation/AiRunService.java:202`) inserts the run row (status `QUEUED`) and **then** calls `workspace.prepare(...)` at `AiRunService.java:207-208`, outside any try/catch. `RmqctlWorkspace.prepare` (`agent/RmqctlWorkspace.java:221-227`) deliberately throws `BusinessException` for three configuration conditions — an unusable `rmqctl-server-url`, an unusable `workspace-dir`, or an instance whose credential cannot be resolved (e.g. the bound instance was deleted, or its admin/cloud credential removed after the conversation was created; `InstanceCredentialResolver.resolveByName` throws 404/422).

When it throws, the SSE stream refuses with the error, but the already-inserted run row is never registered, never finalized, and stays `QUEUED`. Admission is one-active-run-per-conversation (`findActiveByConversationId` at `AiRunService.java:191-195`), so **every subsequent message for that conversation is refused 409 "answer in progress"**. The only recovery is the orphan sweep in `purgeExpired` — `@Scheduled(fixedDelayString = "${studio.ai.conversation.cleanup-interval:PT24H}")` with a 10-minute `orphanRunTimeout` grace — i.e. up to 24 hours, or a server restart.

### Evidence

1. `prepare()` javadoc promises "these fail loudly": `@throws BusinessException when the workspace could be built but the configuration is wrong: an unusable rmqctl-server-url, an unusable workspace-dir, or an instance with no resolvable credential` (`RmqctlWorkspace.java:221-227`).
2. `insertRun` at `AiRunService.java:202` runs first; nothing between it and the throw writes a terminal state. `registry.register` (`:217`) never runs, so the row has no live handle and is invisible to the stop path's fast recovery.
3. Measured on trunk source in this sandbox: a test stubbing `prepare` to throw 404 shows `sendMessage` propagating the exception with the row still `QUEUED` and zero terminal updates — then, with the fix, the row reaches `FAILED` before the exception is rethrown.

### Impact

[...]

PR #4796, opened 2026-09-21
fix(ai): fail the run row when workspace preparation throws

unbridled-41:
Fixes #4795.

## Problem / Evidence

`AiRunService.sendMessage` inserts the run row (`QUEUED`) and only afterwards calls `workspace.prepare(...)`, outside any try/catch. `prepare` deliberately throws `BusinessException` for configuration problems — an unusable `rmqctl-server-url`/`workspace-dir`, or an instance whose credential no longer resolves (`InstanceCredentialResolver.resolveByName` → 404/422). The exception left the inserted row `QUEUED` with no owner: admission's one-active-run guarantee then refused every later message for the conversation with 409 until the scheduled orphan sweep reaped the row — up to 24 hours (`cleanup-interval:PT24H`), or a restart.

## Root cause / Fix

The admission ordering (insert row, then prepare, so the caller hears configuration errors synchronously) is right; the missing piece is a failure path for the row the ordering implies. The catch now finalizes the run through the exactly-once terminal path — `runExecutor.terminate(detachedContext(run), FAILED, PROVIDER_ERROR, ai.run.internal_error, …)`, the same mechanism the stop of an owner-less run already uses — and rethrows so the caller still hears the original reason. `registry.register` has not run yet at this point, so a detached context (no handle, fresh sink) is the correct terminal writer, exactly as in `stop`'s owner-less branch.

## Priority

PRIORITY 70 / 100 — impact 28 (a conversation becomes unusable for up to a day; the user-visible 409 contradicts the visible state, since no answer is actually in progress), blast radius 14 (every native conversation bound to an instance whose binding breaks — the trigger is an ordinary admin action), reproducibility 14 (deterministic given the trigger), maintenance value 14 (fills the one unguarded admission step in a lifecycle the codebase otherwise treats as exactly-once everywhere). FIX_CONFIDENCE 92: the fix reuses the project's own detached-finalize idiom verbatim; the only judgement call is the terminal reason (`PROVIDER_ERROR`/`ai.run.internal_error`, consistent with the executor's unexpected-failure vocabulary).

## Tests

- `AiRunServiceTest.aFailingWorkspacePreparationShouldFinalizeTheRunInsteadOfStrandingItTest` (new): stubs `prepare` to throw 404, asserts the exception still propagates, the row was inserted once as `QUEUED`, the final row is `FAILED/PROVIDER_ERROR`, and nothing is left live in the registry.
- Second commit: the same test adds the recovery half — with `prepare` restored, a second `sendMessage` immediately admits a new run and completes, proving the conversation is reusable at once instead of returning 409 until the orphan sweep.
- Teeth, measured on this machine: with only the test change on trunk source, it fails with the terminal update list empty (`Expecting actual not to be empty` at `lastRun()` — the stranded-QUEUED shape); with the fix, `AiRunServiceTest` 21/21.
- Related suites (re-run on the final head): `AiRunExecutorTest` 18/18, `AiConversationServiceTest`, `AgentRunRegistryTest` 11/11, `AgentRunHandleStopTest` 9/9, `RmqctlWorkspaceTest` 31/31 — 108/108 total. `mvn checkstyle:check` passes.

## Risk

Low. The catch only adds a terminal write on a path that previously left inconsistent state; the rethrow preserves the existing API contract (caller sees the original `BusinessException`). If the terminal write itself were to fail, the exception would replace the original — acceptable, since the orphan sweep remains the backstop and the log line records the cause.


RockteMQ-AI (approved):
### Summary
Fixes a bug where workspace preparation failures would leave run rows stuck in QUEUED state, causing 409 conflicts for subsequent messages until orphan sweep reaps them (up to a day).

The fix is clean and well-targeted:
- Wraps `workspace.prepare()` in try-catch to handle `RuntimeException`
- Properly terminates the run through the exactly-once terminal path (`RunStatus.FAILED`)
- Rethrows the exception so the caller still hears the reason
- Clear comment explaining the issue and the fix

Test coverage is excellent — verifies the regression scenario (run stays QUEUED without fix) and confirms recovery (second message works after failure).

LGTM — solid bug fix with proper error handling.

---

PR #4798, opened 2026-09-21
fix(auth): align the login policy reported by /api/auth/status with the interceptor

Wang1rrr:

No issue - found while comparing how `AuthInterceptor` and `AuthController` each answer "is a login required".

### Brief Description

Both sides compute the flag from the same static property and the same runtime `requireLogin` row, but they disagreed on the failure branches. The interceptor fails closed - an absent or unreadable policy still demands a login - while the controller used `settings != null && settings.isRequireLogin()` and did not catch read failures at all.

So when the persisted settings row cannot be read (`MybatisPlusSettingsRepository` throws `BusinessException(500)` for an undeserializable json column), the interceptor still demanded a login while `GET /api/auth/status` answered 500. The frontend reads that flag to choose between the login page and the console (`web/src/App.tsx` `AuthGate`), so it fell to its error screen: the retry re-issued the same failing request, and the login page - which the interceptor would in fact have accepted credentials from - was never rendered. The console became unreachable exactly when the policy store was broken, with no way for an operator to log in and repair it.

`isLoginRequired()` now reports the policy the interceptor enforces, including on the failure branches, so the two cannot drift. A follow-up worth considering is extracting the predicate into one shared collaborator so divergence is impossible rather than merely absent.

Added cases to `AuthControllerTest` covering the null-settings row and the throwing repository, asserting both report `loginRequired=true` and a 200 body instead of a 500.

```
cd server
mvn -B -Dtest=AuthControllerTest,AuthInterceptorTest test

[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthControllerTest
[INFO] Tests run: 52, Failures: 0, Errors: 0, Skipped: 0 -- in org.apache.rocketmq.studio.auth.AuthInterceptorTest
[INFO] Tests run: 63, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Backend only: no API shape change, no UI text, no new source files.

### Checklist

- [ ] New UI text has both Chinese and English entries under `web/src/i18n/` - N/A, no UI text
- [ ] Architecture constraints stay green (`mvn test` runs the ArchUnit checks) - no new types or packages introduced
- [ ] Documentation touched where behaviour changed - N/A, the endpoint contract is unchanged


server/src/main/java/org/apache/rocketmq/studio/auth/AuthController.java
```diff
@@ -60,12 +60,28 @@ public ResponseEntity<Result<AuthStatusVO>> status(
...
+     * branches have to line up too: the frontend reads this flag to choose between the login page
+     * and the console, and a policy that fails differently in the two places strands the console.
+     * The interceptor fails closed (see its own {@code isLoginRequired}), so an unreadable policy
+     * still demands a login here; letting the read failure escape as a 500 instead used to leave the
+     * frontend on its error screen, where the retry could never succeed and the login page that
+     * would in fact work was unreachable.
+     */
     private boolean isLoginRequired() {
         if (authProperties.isLoginRequired()) {
             return true;
```
RockteMQ-AI, line 75: Consider logging the exception at DEBUG level for troubleshooting: `log.debug("Settings load failed, failing closed", exception)`. This helps operators diagnose policy read failures without cluttering production logs.

RockteMQ-AI (approved):
/tmp/pr-review-4798/summary.txt

PR #4629, opened 2026-09-20
fix(studio): localize the remaining hardcoded strings on Ops, BrokerCluster and AlertRuleAssets

Frun1na:
### Which Issue(s) This PR Fixes

Fixes #4630

### Brief Description

While localizing the Studio pages, five user-visible strings were still hardcoded in Chinese on pages that otherwise use the i18n catalog, so switching the UI to English still showed Chinese text in three places:

- `Ops.tsx`: the "runtime configuration unavailable" Alert message and fallback description
- `BrokerCluster.tsx`: the instance `Select` aria-label and placeholder
- `AlertRuleAssets.tsx`: the page header subtitle

This PR routes them through the catalog: the BrokerCluster strings reuse the existing `common.selectInstance` key (same Chinese text), and three new keys are added (`ops.unavailableTitle`, `ops.unavailableDescription`, `alertAssets.subtitle`). The Chinese output is byte-identical to before, so nothing changes for zh users.

### How Did You Test This Change?

```
$ cd web
$ npx tsc --noEmit          # exit 0
$ npx eslint src/pages/studio/Ops.tsx src/pages/studio/BrokerCluster.tsx src/pages/studio/AlertRuleAssets.tsx src/i18n/translations.ts   # exit 0
$ npx vitest run src/pages/studio/__tests__ src/i18n
 Test Files  9 passed (9)
      Tests  87 passed (87)
```

Also verified with `grep -P '[\p{Han}]'` that no hardcoded Chinese remains in the three files.

### Checklist

- [x] Follows the existing i18n conventions of the surrounding code
- [x] All UI strings have both zh and en entries
- [x] `tsc`, `eslint` and the affected test suites pass locally

PR #4621, opened 2026-09-20
fix(studio): localize the user management page

Frun1na:
### Which Issue(s) This PR Fixes

Fixes #4622

### Brief Description

The user management page (`web/src/pages/studio/UserManagement.tsx`) rendered every user-visible string in hardcoded Chinese — the page header, the non-admin banner, table column headers, role/status tags, the session drawer, both modals, filter placeholders/aria-labels, pagination summary, and all success/error messages — so an English UI showed Chinese text on this page.

This PR routes all of those strings through the shared translation dictionary under a new `userMgmt.*` namespace, reusing existing `common.*` keys (status, actions, enabled/disabled, cancel, refresh, export) where they already exist. The Chinese wording is unchanged byte for byte, so existing assertions on the rendered Chinese text still pass; the session-status tags helper became a component so it can consume the i18n context. A new test renders the page with the stored language preference set to `en` and asserts the English headings, buttons, placeholder and column headers.

### How Did You Test This Change?

```
cd web && npx tsc --noEmit                      # passes
npx eslint src/pages/studio/UserManagement.tsx \
  src/pages/studio/__tests__/UserManagement.test.tsx src/i18n/translations.ts
                                                # 0 errors, 0 warnings
npx vitest run src/pages/studio/__tests__/UserManagement.test.tsx
                                                # 11 passed
npx vitest run                                  # 134 files, all passed
```

The English test fails without the change (the page has no English text at all before this PR).

### Checklist

- [x] Follows the project style guidelines
- [x] Tests cover the change; full suite passes locally
- [x] The commit message follows Conventional Commits
- [x] No unrelated files are touched

PR #4803, opened 2026-09-22
fix(web): localize the About tab

Wang1rrr:

No issue - continues the localization pass from #4621, #4629, #4633 and #4802.

### Brief Description

`web/src/pages/settings/AboutTab.tsx` was the only settings tab that never imported the language context. Six `Descriptions.Item` labels, the "related links" heading and two link captions were string literals, so switching the console to English left them in Chinese.

They move into `settings.about*` keys placed next to `settings.tabAbout`, following the namespace the other settings tabs already use (`settings.credentialExportFailed`, `settings.themeMode`, ...). The component gains a body so it can call `useLang()`; prettier reindents the JSX.

Left alone deliberately:

- `label="License"` - the same word in both locales.
- the `GitHub` caption - a brand name.
- the copyright line - the ASF legal notice, which should not be translated.

The Chinese values are moved verbatim from the component rather than rewritten, so the zh console is pixel-identical; only the English text is new.

Added `reads every label from the language context` to `AboutTab.test.tsx`, which mocks `useLang` to an identity `t` (the pattern `src/pages/login/index.test.tsx` uses) and asserts all nine keys reach the DOM - a literal left in the component would render as text instead of a key and fail.

```
cd web
npx vitest run src/pages/settings/__tests__/AboutTab.test.tsx src/pages/settings/__tests__/SettingsPage.test.tsx
 Test Files  2 passed (2)
      Tests  3 passed (3)

npx tsc -b                                        # exit 0
npx eslint src/pages/settings/AboutTab.tsx src/pages/settings/__tests__/AboutTab.test.tsx src/i18n/translations.ts   # exit 0
npx prettier --check <the three files>            # clean
```

Baseline for reference, unchanged by this diff: `npm test` on the trunk gives 139 files / 1280 tests passing.

### Checklist

- [ ] Architecture constraints stay green (`mvn test` runs the ArchUnit checks) - N/A, frontend only
- [ ] Documentation touched where behaviour changed - N/A, no documented contract changed


lizhimins (changes requested):
The localisation itself is correct and welcome — the About tab was the only settings tab that never touched the language context, and moving the six `Descriptions.Item` labels, the "related links" heading and two of the three link captions into `settings.about*` (next to the existing `settings.tabAbout`) is the right namespace. Keeping `License` (same word in both locales), `GitHub` (brand name) and the ASF copyright line as literals is also right, and the Chinese values are moved verbatim so only the English UI changes.

One thing blocks merging, and it is a rebase rather than a code problem:

**The branch conflicts with trunk on `web/src/pages/settings/AboutTab.tsx`.** `git merge-tree` against the current `rocketmq-studio` head reports a content conflict in that file, because #5007 (3.0.0 release readiness) changed the same two rows you are wrapping in `t()`. When you resolve it, please take **trunk's values**, not your branch's:

- `AboutTab.tsx:26` → `<Descriptions.Item label={t('settings.aboutVersion')}>3.0.0</Descriptions.Item>` (trunk says `3.0.0`; your branch is based on a commit that still said `0.1.0`)
- `AboutTab.tsx:31` → `<Descriptions.Item label={t('settings.aboutBackendStack')}>Spring Boot 4.1 + RocketMQ MCP Server</Descriptions.Item>` (trunk says `Spring Boot 4.1 + RocketMQ MCP Server`; your branch says `Spring Boot 3 + …`)

Resolving the other way would silently revert the release strings on the About page, which is the one place a user checks the version. Trunk has also moved on since your last push (a batch of merges landed today), so please rebase onto the current `rocketmq-studio` head and confirm `mergeable` is clean before we take it.

Two notes, neither blocking:

1. Your new test name in `AboutTab.test.tsx` does not end in `Test`. Our written convention asks for the suffix, but the frontend `it()` naming rule is currently undecided on our side (trunk is 206/1202 compliant), so we are not holding this PR on it.
2. Once rebased, this is a small, self-contained change and can go in on its own.

PR #4804, opened 2026-09-22
fix(topic): render the consumer model values the API returns

Frun1na:

- Fixes #4805

### Brief Description

The 消费模式 (message model) column of the consumer-group sub-table in the topic detail modal matched
the API value against the pre-#3349 display string `广播消费`:

```tsx
render: (m: string) => <Tag color={m === '广播消费' ? 'orange' : 'blue'}>{m}</Tag>,
```

Every provider sends enum values — `CLUSTERING` / `BROADCASTING` from the Apache admin ext
(`conn.getMessageModel().name()`), and the vendor-native values the Aliyun and Tencent providers pass
through — so the comparison never matches: broadcasting consumers lose the orange highlight, and the
raw enum text lands in the tag untranslated.

The fix renders the already-existing but unreferenced i18n keys `topic.broadcast` / `topic.clustering`:
a value containing `BROADCAST` maps to `topic.broadcast` with the orange tag, `CLUSTER` maps to
`topic.clustering` with the blue tag, and anything else falls back to the raw text.
`web/src/mock/topics.ts` fed the old display strings (`集群消费` / `广播消费`) to the dev mock path;
it now uses the enum values the API records, which is what kept the dead comparison looking alive.

TDD: a new TopicPage test feeds the API values `BROADCASTING`, Aliyun-style `Broadcasting` and
`CLUSTERING` and asserts the localized labels and tag colors. Before the fix it fails
(`findAllByText('广播消费')` times out — the tag renders the raw `BROADCASTING`); after the fix:

```
npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx

 Test Files  1 passed (1)
      Tests  33 passed (33)
```

Full frontend suite, plus tsc and eslint on the changed files:

```
npx vitest run

 Test Files  139 passed (139)
      Tests  1281 passed (1281)
```


RockteMQ-AI (approved):
﻿### Summary

Clean test fixture fix — updates mock data to use API enum values (CLUSTERING/BROADCASTING) instead of Chinese labels. This ensures tests match the actual API contract.

### Findings

- **[Info]** Mock data now uses `CLUSTERING`/`BROADCASTING` instead of `集群消费`/`广播消费`.
- **[Info]** Tests updated to match the corrected mock data.

✅ **Approve** — Fixes test fixtures to match API contract.

---

PR #4812, opened 2026-09-22
fix(web): render the consumer delay in the console language

Wang1rrr:

No issue - found while auditing which modules still carry their own copy of a shared formatter.

### Brief Description

`web/src/pages/instance/consumer.tsx` defined its own `formatDelay` at module scope:

```ts
/**
 * Format delay seconds into human-readable Chinese time.
 * ...
 */
const formatDelay = (totalSeconds: number): string => { ... }   // no lang parameter
```

`web/src/utils/format.ts` already exports a `formatDelay(totalSeconds, lang)` that handles both locales, and the rest of the app uses it. The local copy shadowed it, so the two places that render a consumer-group delay - the table column (`delaySeconds`) and the group drawer's delay field - stayed in Chinese no matter which language the console was in, while every other duration on the same page came out as `22h 55m`.

The copy is deleted and the shared formatter imported, with the `lang` the page already destructures from `useLang()`.

Behaviour notes:

- The Chinese output is unchanged. The two implementations had the same unit order, the same three-unit cap (`parts.length < 3`) and the same zero case.
- The shared version adds a `Number.isFinite` guard the copy lacked, so a non-finite delay now renders `-` instead of a zero duration. The column already coerces a missing value with `seconds ?? 0`, so this only affects a genuinely malformed number.

`UNAVAILABLE_LAG_LABEL` and the rest of the page's untranslated strings are a separate, larger concern and are deliberately not bundled in here.

Added `renders the consumer delay in the console language` to `ConsumerPage.test.tsx`: it sets the stored language preference to `en`, renders the page against the existing fixture (`delaySeconds: 3`) and asserts the cell reads `3s` and that the Chinese form is absent. The preference is removed in a `finally` so the surrounding cases, which assert Chinese text, are unaffected.

Verified the case is sensitive rather than vacuous - pinning the argument back to a literal `'zh'` fails it:

```
FAIL  ConsumerPage > renders the consumer delay in the console language
TestingLibraryElementError: Unable to find an element with the text: 3s.
 Tests  1 failed | 36 skipped (37)
```

Clean run:

```
cd web
npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx src/utils/format.test.ts
 Test Files  2 passed (2)
      Tests  46 passed (46)      # ConsumerPage 36 -> 37, format unchanged

npx tsc -b                                        # exit 0
npx eslint src/pages/instance/consumer.tsx src/pages/instance/__tests__/ConsumerPage.test.tsx   # exit 0
npx prettier --check <both files>                 # All matched files use Prettier code style!
```

Net diff is -28/+4 in the page plus 15 lines of test.

### Checklist

- [ ] New UI text has both Chinese and English entries under `web/src/i18n/` - N/A, adds no dictionary entries; it reuses the existing locale-aware formatter
- [ ] Architecture constraints stay green (`mvn test` runs the ArchUnit checks) - N/A, frontend only
- [ ] Documentation touched where behaviour changed - N/A, no documented contract changed


RockteMQ-AI (approved):
﻿### Summary

Clean i18n fix — removes duplicate Chinese-only `formatDelay` function and uses the shared language-aware version from utils. Good test coverage.

### Findings

- **[Info]** `consumer.tsx` — Removed duplicate `formatDelay` function, now imports from `utils/format`.
- **[Info]** `format.ts` — The shared `formatDelay` already handles both Chinese and English.
- **[Info]** Tests updated to verify English output.

✅ **Approve** — Clean refactor that fixes i18n inconsistency.

---

PR #4813, opened 2026-09-22
fix(consumer): apply the subscription-mode filter on the server page request

unbridled-41:
## Problem / Evidence

The Push/Pop (订阅模式) filter on the consumer-group list is applied **client-side to the rows already loaded for the current server page**, while the list itself is server-paginated:

- `visibleConsumerGroups` filters the current page's items (`web/src/pages/instance/consumer.tsx`).
- The pager uses the **unfiltered** `totalGroups` (`total: totalGroups`), and the header subtitle prints `共 ${totalGroups} 个 Group`.
- `listConsumerGroupPage` is called without any subscription-mode parameter, so the backend cannot honor the filter.

Reproducible sequence: instance with 30 Push + 2 Pop groups, pageSize 20 →
1. Select "Pop": the table shows ≤2 rows, but the pager still reads "共 32 个 Group" with 2 pages.
2. Advance to page 2: the server returns 20 Push rows, the client filter hides all of them, and the table renders **empty** while claiming to be page 2 of 32 groups.
3. The pager total and the row set permanently disagree while the filter is active.

The export path already passes `subscriptionMode` to the backend (`exportConsumerGroups`), which shows the filter was intended to be a server-side capability.

## Root cause / Fix

The mode filter predates the move to server-side pagination (`/groups/page`), and only the row-level slice was migrated; the filter itself stayed client-side.

**Front-end** (`commit 1`):

- `web/src/api/metadata.ts`: add `subscriptionMode?` to `ConsumerGroupQuery` (so the page query carries it; `ConsumerGroupExportQuery` inherits it instead of redeclaring it).
- `web/src/pages/instance/consumer.tsx`: send `subscriptionMode: modeFilter !== 'ALL' ? modeFilter : undefined` in `listConsumerGroupPage`, drop the client-side `visibleConsumerGroups` pre-filter, reset the page when the filter changes, and re-run the load when `modeFilter` changes.
- `web/src/services/consumerService.ts`: apply the same filter in the mock-mode path so dev mode behaves like the backend.

**Back-end** (`commit 2`): the initial front-end-only fix was incomplete — `/groups/page` ignored the new parameter, leaving the filter a no-op against a real backend. Fixed by threading it through:

- `server/.../instance/group/ConsumerGroupController.java`: `/api/groups/page` accepts `subscriptionMode` as an optional query parameter.
- `server/.../instance/topic/MetadataService.java`: a new overload applies the filter to the page rows after the provider slice and shrinks the reported total accordingly; `ALL`/blank keeps every group, matching the CSV export's filter semantics (`SubscriptionMode` enum compared by `name()`, case-insensitively).

## Priority / scoring

- Impact 24/40: under a Push/Pop filter the pager count and the visible rows contradict each other, and page 2+ can render an empty table — an operator reads "no groups" from a healthy instance.
- Reach 14/20: the group list is one of the two primary instance pages.
- Reproducibility 16/20: deterministic from the data shape above; no timing involved.
- Maintenance value 12/20: removes a duplicated filtering convention (client for the table, server for export) and aligns the page with the backend's existing parameter.
- **PRIORITY 66/100**, **FIX_CONFIDENCE 90/100**.

## Tests

Front-end:
- New regression test `applies the subscription-mode filter on the server page request` (web/src/pages/instance/__tests__/ConsumerPage.test.tsx): selects Pop and asserts `listConsumerGroupPage` receives `subscriptionMode: 'Pop'` with a page reset to 1, and that returning to 全部模式 clears the parameter. Fails on the pre-fix code (parameter absent).
- Updated the four exact-param assertions for the added `subscriptionMode: undefined` field.
- `npx vitest run src/pages/instance/__tests__/ConsumerPage.test.tsx` → 37 passed (37).
- `npx vitest run src/services/consumerService.test.ts` → 17 passed (17).
- `npx tsc --noEmit -p tsconfig.app.json` → clean; `npx eslint` on the touched files → clean.

Back-end (commit 2):
- New `listConsumerGroupsPageShouldApplySubscriptionModeFilterToThePage` (MetadataServiceTest): a page holding one Push + one Pop row filtered by "Pop" returns only the Pop row with total shrunk 2 → 1.
- New `listConsumerGroupsPageShouldIgnoreSubscriptionModeAllAndUnknownValues`: `ALL`/blank keep every row and the original total.
- New `listConsumerGroupsPageShouldPassSubscriptionModeFilterToTheService` (ConsumerGroupControllerTest): the endpoint passes `subscriptionMode=Pop` through to the service; the pre-existing paging test now pins the `null` passthrough.
- `mvn test -Dtest='MetadataServiceTest,ConsumerGroupControllerTest'` → 57 + 22 tests, 0 failures, BUILD SUCCESS.

## Risk

Low. The new query parameter is optional and blank means "no filter", so existing API consumers are unaffected. Providers without subscription-mode awareness keep their behavior (the service-level filter works on any provider's page result). The mock path mirrors the new parameter for dev mode.

---

Related issue: #3150 (same defect, reported by another contributor; an alternative fix is proposed in #4773). The GitHub auto-close keyword is intentionally omitted so this PR does not race #4773 for the same issue — maintainers pick one fix and close the other.


unbridled-41:
Follow-up commit `97f2bbde` addresses a gap found during self-review: the original fix only changed the front-end, but `/api/groups/page` did not accept or apply `subscriptionMode`, so the filter would have remained a no-op against a real backend.

What the follow-up adds:

- `ConsumerGroupController` accepts `subscriptionMode` as an optional query parameter (blank/`ALL` = no filter, matching the CSV export semantics).
- `MetadataService.listConsumerGroupsPage` overload applies the filter to the provider's page result and shrinks the reported total to stay consistent with the rows.
- Backend tests: `MetadataServiceTest.listConsumerGroupsPageShouldApplySubscriptionModeFilterToThePage` (Pop filter drops the Push row, total 2→1), `...ShouldIgnoreSubscriptionModeAllAndUnknownValues`, and `ConsumerGroupControllerTest.listConsumerGroupsPageShouldPassSubscriptionModeFilterToTheService` (param passthrough).
- Test results: `mvn test -Dtest='MetadataServiceTest,ConsumerGroupControllerTest'` → 57 + 22 tests, 0 failures, BUILD SUCCESS. Front-end suites unchanged and green (37 + 17 passed, tsc/eslint clean).

RockteMQ-AI (approved):
﻿### Summary

Clean feature fix — adds server-side subscription mode filtering (Push/Pop) to consumer group pagination. The implementation is well-structured with proper Javadoc and test coverage.

### Findings

- **[Info]** `MetadataService.java:370-406` — Clean implementation with overloaded method for backward compatibility. The `applySubscriptionModeFilter` helper is well-documented.
- **[Info]** `ConsumerGroupController.java:59-64` — Proper parameter addition with `required = false` for backward compatibility.
- **[Info]** Tests updated to pass `null` for the new parameter.

✅ **Approve** — Well-implemented server-side filter with good test coverage.

---

lizhimins:
Thanks for chasing this — the front-end half is right, but the back-end half does not actually fix the reported defect, so I am closing it.

The problem is where the filter runs. `RocketMQMetadataProvider.listConsumerGroupsPage` (server/.../provider/apache/RocketMQMetadataProvider.java:266-288) does the paging itself with `groupMapper.selectPage(new Page<>(page, pageSize), query)`, so by the time `MetadataService.applySubscriptionModeFilter` sees the rows the slice is already taken. Filtering a slice, and shrinking the total by only the rows dropped from *that* slice, cannot produce a correct global total, and it cannot reach rows that live on later pages.

Concretely, with the 30 Push + 2 Pop / pageSize 20 example from your own description: page 1 comes back with 20 Push rows, the filter removes all 20, and the total becomes 32 - 20 = 12. The pager now reports "共 12 个 Group" with a single page and the table is empty — the two Pop groups are unreachable, which is worse than the current behaviour (today they at least show up when they happen to land on the loaded page).

The Javadoc on the new 6-arg overload also states "the filter is applied to the full filtered set before slicing so the reported total stays consistent with the rows", which is the opposite of what the code does (and of what the private helper's own Javadoc says two lines below). Neither new back-end test can catch this: both build a single page (`PageResult.of(List.of(pushGroup, popGroup), 2, 1, 20)`), where slice-then-filter and filter-then-slice are indistinguishable.

Two smaller notes: the retained 5-arg `listConsumerGroupsPage` overload has no caller left in main or test after this change, and the subscription mode is stored in the `messageModel` column (`RocketMQMetadataProvider.java:311-314`), so it is available as a SQL predicate.

A correct fix would thread `subscriptionMode` down into `MetadataProvider.listConsumerGroupsPage(...)` / `InstanceProvider.listConsumerGroupsPage(...)` so the predicate joins the `LambdaQueryWrapper` before `selectPage`, which is also how the CSV export path already behaves (`MetadataService.java:680-684` filters the full list, then slices). We are not taking that route right now either — #4773 proposed exactly that pushdown and is closed with the reasoning: at realistic group counts per instance the practical impact is small, while the pushdown means changing the provider SPI across every implementation. So this is closed as not planned, and no rework is expected. Nothing needs to change on your side; the front-end commit in this PR is fine as it stands.

PR #4816, opened 2026-09-22
fix(aliyun): keep an ordered consumer group ordered when creating it

Wang1rrr:

Trivial fix, no issue.

### Brief Description

`AliyunInstanceProvider.normalizeDeliveryOrderType` mapped only `FIFO` and `ORDERLY` to the OpenAPI `Orderly` value and sent everything else as `Concurrently`. The console never submits those two spellings:

- `web/src/pages/instance/consumer.tsx` (the 顺序类型 select, shown when 订阅组类型 is FIFO) offers `PARTITON_ORDER` (分区顺序) and `MESSAGES_ORDER` (全局顺序), with `PARTITON_ORDER` as the initial value.
- `web/src/utils/resourceCsvImport.ts` accepts `PARTITON_ORDER`, `PARTITION_ORDER` and `MESSAGES_ORDER`.
- `docs/api-spec.md` documents the field as `PARTITON_ORDER` / `MESSAGES_ORDER`.

So every ordered consumer group created on an Aliyun instance was submitted as `deliveryOrderType=Concurrently`, together with `DefaultRetryPolicy` instead of the `FixedRetryPolicy` the ordered branch selects. The ordering guarantee was silently dropped and nothing surfaced an error - the group is created successfully, it just does not do what the form said.

`TencentInstanceProvider#isOrderly`, which reads the same `ConsumerGroupVO.deliveryOrderType`, already classified these values correctly because it matches on the `ORDER` substring. This aligns the Aliyun adapter with that rule so the two vendors cannot disagree about the same input.

```
cd server && mvn -o test -Dtest=AliyunInstanceProviderTest -DfailIfNoTests=false
[INFO] Tests run: 37, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS

mvn -o checkstyle:check
(exit 0)
```

Two tests were added:

- `createConsumerGroupShouldKeepPartitionOrderedGroupsOrderlyTest` drives `createConsumerGroup` with the value the form actually submits and asserts the captured `CreateConsumerGroupRequest` carries `Orderly` plus `FixedRetryPolicy` / `fixedIntervalRetryTime=10`.
- `normalizeDeliveryOrderTypeShouldMapConsoleOrderTypesToOrderlyTest` covers `PARTITON_ORDER`, `MESSAGES_ORDER`, `PARTITION_ORDER`, a lower-case and a padded spelling, and keeps `Concurrently` / blank on the default branch.

Both were mutation-checked: reverting the one-line condition to `"FIFO".equalsIgnoreCase(value) || "ORDERLY".equalsIgnoreCase(value)` makes exactly these two fail (`expected: "Orderly" but was: "Concurrently"`) while the other 35 stay green.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text added)


RockteMQ-AI (approved):
﻿### Summary

Clean bug fix — changes the Aliyun provider's delivery order type normalization from exact match to substring match, so `PARTITON_ORDER`, `MESSAGES_ORDER`, and `PARTITION_ORDER` are correctly mapped to `Orderly` instead of falling through to `Concurrently`. The fix is minimal, well-tested, and consistent with the Tencent provider.

### Findings

- **[Info]** `AliyunInstanceProvider.java:389-390` — The change from `equals` to `contains` correctly handles all the console-submitted spellings. The Javadoc update is excellent — it documents the full vocabulary and explicitly notes the consistency requirement with `TencentInstanceProvider#isOrderly`.
- **[Info]** `AliyunInstanceProviderTest.java:649-676` — Good end-to-end test: verifies that a `PARTITON_ORDER` group is created as `Orderly` with the correct retry policy.
- **[Info]** `AliyunInstanceProviderTest.java:1071-1089` — Comprehensive unit test for `normalizeDeliveryOrderType`: covers all console spellings, case insensitivity, whitespace, and fallback behavior.

### Review

✅ **Approve** — Fixes a real bug where ordered consumer groups on Aliyun instances were silently created as concurrent. Code is correct, tests are comprehensive, cross-provider consistency is maintained.

---

PR #4590, opened 2026-09-20
fix(cluster): keep the selected instance when the display language changes

unbridled-41:
Fixes #4589.

## Problem / Evidence

The instance discovery effect on the broker topology page writes the first Apache-capable instance into the selection on every run:

```tsx
  useEffect(() => {
    let active = true;
    void listInstances()
      .then((nextInstances) => {
        …
        setSelectedInstanceId(apacheInstances[0]?.name);      // :236, unconditional
      })
      …
  }, [clearData, message, t]);                                // :246
```

`t` is recreated whenever the display language changes (`web/src/i18n/LangContext.tsx`), while `clearData` and `message` are stable — so the top-bar language toggle re-runs the effect and overwrites the instance the user selected. The data below then reloads for that other instance. The twin page already guards the identical write (`web/src/pages/studio/Producer.tsx:104-108`), which is the intended contract.

Reproduced by the regression test added here: it selects `instance-2`, switches the language through the real `LangProvider.setLang`, and asserts both the loaded topology and the instance the page queries.

## Root cause / Fix

An effect that both discovers instances and seeds the default selection, keyed on a dependency that changes for an unrelated reason (rendering text in another language). The minimal fix is the guard the sibling page uses — the effect may re-run, but it only replaces a selection that is no longer valid:

```diff
         const apacheInstances = nextInstances.filter(supportsApacheRuntime);
         setInstances(apacheInstances);
-        setSelectedInstanceId(apacheInstances[0]?.name);
+        setSelectedInstanceId((current) =>
+          apacheInstances.some((instance) => instance.name === current)
+            ? current
+            : apacheInstances[0]?.name,
+        );
```

A first visit still selects the first Apache instance, and an instance that disappeared from the list still falls back — only a re-run for an unrelated reason keeps the user's choice. Both branches are covered: the second test in this PR switches the language while the discovery returns a list that no longer contains the selected instance and expects the fallback.

## Priority & scoring

- PRIORITY **70** = impact 26 + blast radius 9 + reproducibility 20 + maintenance value 15
  - impact 26 — the page silently switches to another instance's broker/nameserver/proxy topology, so the user reads the wrong instance's data. The trigger is the display-language change: `t` is defined inline in `LangProvider` (`web/src/i18n/LangContext.tsx:34-42`) and only that provider's own `lang` state re-renders it, which is exactly what the regression test drives.
[...]

PR #4818, opened 2026-09-22
fix(ops): keep the unsaved NameServer selection when the display language changes

Frun1na:

- Fixes #4819

### Brief Description

The Ops configuration load effect in `web/src/pages/studio/Ops.tsx` depends on `fetchFailedMessage`,
a string derived from `t('ops.fetchFailed')`, so switching the display language in the top bar re-runs
it — and the re-run ended with an unconditional reset of the NameServer selection:

```tsx
setSelectedNamesrv(data.currentNamesrv);
```

An operator who picked a NameServer from the list but has not saved it yet loses that selection by
merely switching the language, and the page silently points back at the stored current NameServer
while the select still looks editable. This is the same root cause #4590 fixed for the instance
selection in `BrokerCluster.tsx` (`t`-derived effect dependency + unconditional setter); `Ops.tsx`
was the remaining unguarded selection of that shape.

The setter is now functional: the unsaved selection is kept while it still appears in the freshly
loaded address list and falls back to the stored current NameServer otherwise, so first load,
deletion fallback and refetch-after-failure behave exactly as before.

TDD: a new `OpsPage` test renders the page with a language-switch button (the pattern the #4590 tests
introduced), picks the second NameServer from the list, switches the language and asserts the select
still shows the unsaved pick. Before the fix it fails — the select snaps back to
`127.0.0.1:9876` — and after the fix:

```
npx vitest run src/pages/studio/__tests__/Ops.test.tsx

 Test Files  1 passed (1)
      Tests  8 passed (8)
```

`tsc --noEmit` and `eslint` on the changed files are clean.


RockteMQ-AI (approved):
﻿### Summary

Clean UI state fix — preserves unsaved NameServer selection when display language changes. Good test coverage.

✅ **Approve** — Fixes state persistence issue with proper implementation.

---

PR #4822, opened 2026-09-22
fix(web): show an unresolvable topic consumer lag as unavailable

Wang1rrr:

Trivial fix, no issue.

### Brief Description

`TopicConsumerVO` has two independent ways of reporting "no lag number", and the topic detail consumer list only handled one of them:

1. `metricsAvailable = false` - set in the `catch` branch of `RocketMQMetadataProvider#getTopicConsumersPage`, when `examineConsumeStats` fails outright.
2. `diffTotal = ConsumerLagResolver.UNKNOWN` (`-1`) with `metricsAvailable` left at its `true` default - set when the broker *did* answer, but a queue offset could not be resolved (`RocketMQMetadataProvider.java`: `if (queueDiff == ConsumerLagResolver.UNKNOWN) { diffTotal = ConsumerLagResolver.UNKNOWN; break; }`). This is the RocketMQ 5.0 gRPC-consumer-without-proxy-stats case the sentinel exists for.

`web/src/pages/instance/topic.tsx` branched only on `metricsAvailable === false`, so case 2 fell through to `formatNumber(-1)` and the 堆积量 column rendered **-1** - a negative backlog, and with `n > 100` false it is not even styled as a warning.

The consumer page renders the same semantic field correctly: `web/src/pages/instance/consumer.tsx` guards its 堆积量 column with `isLagAvailable(diff)` from `web/src/utils/consumerLag.ts` and prints the unavailable label. This reuses that shared guard so the two pages stop disagreeing about the sentinel. No new UI text is introduced - the branch renders the label the column already had.

```
cd web
npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx
 Test Files  1 passed (1)
      Tests  33 passed (33)

npx tsc -b            (exit 0)
npx eslint src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx   (exit 0)
npx prettier --check --end-of-line auto src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx
```

Added `renders an unresolvable Topic consumer lag as unavailable instead of -1`, next to the existing `renders unavailable Topic consumer metrics distinctly from zero`: it feeds a consumer row with `diffTotal: -1` and no `metricsAvailable` field (i.e. the default `true`), then asserts the unavailable label is shown and `-1` is nowhere in the document.

Mutation-checked: reverting the render condition to `record.metricsAvailable === false ? (` makes exactly this new test fail (`1 failed | 32 passed`) while the rest of the file stays green.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (no new UI text; the existing label is reused)
- [x] Documentation touched where behaviour changed (README / `docs/` / in-app help) - none needed


RockteMQ-AI (approved):
﻿### Summary

Clean UI fix — shows "不可用" (unavailable) for unresolvable consumer lag (`-1`) instead of displaying the raw `-1` value. The fix is minimal and well-tested.

### Findings

- **[Info]** `topic.tsx:967` — The fix correctly adds `!isLagAvailable(n)` to the render condition, so both `metricsAvailable === false` and unresolvable lag (`-1`) show as unavailable.
- **[Info]** `TopicPage.test.tsx:1147-1172` — Good test coverage: verifies that `-1` is not displayed and "不可用" is shown instead.

### Review

✅ **Approve** — Fixes a UI issue where unresolvable consumer lag was displayed as `-1`. Code is correct, test covers the fix.

---

PR #4832, opened 2026-09-22
fix(alerts): render the total in the rule-count label

unbridled-41:
## Problem / Evidence

The alerts page pagination footer renders the server total through the i18n label:

```ts
showTotal={(total) => t('alerts.totalRules', { count: total })}
```

but the translation carries no placeholder:

```ts
'alerts.totalRules': { zh: '规则总数', en: 'Total Rules' },
```

`t()`'s substitution is a plain `String.replace('{count}', …)` loop, which is a no-op when the template lacks the placeholder — so the paginator footer shows only "规则总数" / "Total Rules" and **the total number is silently dropped** (verified by simulating `t()` with `{count: 21}` → output `"规则总数"`).

Reproducible: open /ops/alerts with any page result (e.g. the existing test fixture with `total: 21`); the pagination footer text never contains "21". The header stat block sits right above it and *does* show the number, making the missing count easy to notice.

## Root cause / Fix

The label was originally written as a bare caption for the header stat; when the paginated list arrived, `showTotal` was wired to the same key with a `{count}` param that the template never grew.

Minimal fix:

- `web/src/i18n/translations.ts`: `'alerts.totalRules'` → `{ zh: '规则总数 {count}', en: 'Total Rules {count}' }`.
- `web/src/pages/ops/alerts.tsx`: the header stat renders `t('alerts.totalRules', { count: totalRules })` through the same substituted label (replacing the separate bare-label + value `<span>` pair), so the label now carries its number wherever it appears.

## Priority / scoring

- Impact 12/40: information loss, not corruption — but a pagination footer that claims a total and shows none defeats its purpose, and the operator cannot cross-check page counts.
- Reach 14/20: every alerts page (both domains) renders this footer.
- Reproducibility 20/20: deterministic render; the substituted text is missing on every load.
- Maintenance value 10/20: aligns the label with the `t(key, params)` convention used by every sibling key and removes the bespoke two-span header.
- **PRIORITY 56/100**, **FIX_CONFIDENCE 95/100** (one placeholder + one render site; covered by tests).

## Tests

- Updated "loads a server-side page and filters by status and search": asserts both the header stat and the pagination footer render "规则总数 21" (`getAllByText(/规则总数 21/).length === 2`). The new footer assertion fails on the pre-fix code (footer rendered "规则总数" with no number).
- New "renders the pagination total in the 总数 label when the language is English": English locale with `total: 7` renders "Total Rules 7" in both places.
- `npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx` → 29 passed (29).
- `npx tsc --noEmit -p tsconfig.app.json` → clean; `npx eslint` on touched files → no new findings (the 5 warnings in alerts.tsx are pre-existing on the clean tree).

## Risk

Minimal. The only behavioral change is the label text now containing the count; no data, request, or state handling is touched.

---

Closes #4841


RockteMQ-AI (approved):
﻿### Summary
This PR improves the i18n handling for the "Total Rules" label by embedding the count directly in the translation string using a `{count}` placeholder. This is a better practice for internationalization as different languages may need to position the count differently.

The changes are clean and well-tested:
- Translation strings updated to include `{count}` placeholder
- Component updated to pass the count to the translation function
- Tests updated to verify the new behavior in both Chinese and English

**Verdict:** LGTM — good i18n practice improvement.

---

unbridled-41:
Evidence-chain verification pass (self-review):

- Re-ran the touched suite from a clean checkout of this branch (`fix/alerts-total-count-label`, commit `6cfc1b9d`) in an isolated worktree: `npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx` → **29 passed (29)**, `tsc --noEmit` clean.
- Pre-fix failure confirmed: on `origin/rocketmq-studio` the pagination footer renders `t('alerts.totalRules', { count: 21 })` as exactly `规则总数` (the `{count}` substitution is a no-op on a template without the placeholder — verified by simulating `t()`), so the new assertion `getAllByText(/规则总数 21/).length === 2` fails without the fix (footer contributes 0 matches) and passes with it.
- Both locales covered: the zh assertion (header + footer both carry the count) and the en test (`Total Rules 7` in both places).

No code changes needed.

PR #4837, opened 2026-09-22
fix(web): route every page-level API error through the shared extractor

Wang1rrr:

Trivial fix, no issue.

### Brief Description

`web/src/utils/apiError.ts` exists to "extract the server-supplied message from an API error so callers can surface the concrete rejection reason instead of a generic fallback", and most of the console uses it - `pages/instance/index.tsx` (7 call sites), `pages/ops/alerts.tsx`, and the AI pages via `describeThrownMessage(error) || t(...)`.

Four pages instead carried their own private extractor: three byte-identical copies plus a fourth, weaker variant. This deletes all four and points their 15 call sites at the shared helper.

**What was actually wrong**

First, a note on what was *not* wrong, because it is easy to overstate: `api/client.ts`'s response interceptor already rewrites `error.message` from the app's `{code, message}` envelope, and `GlobalExceptionHandler` returns that envelope on every path (`BusinessException`, validation, 404, 405, 415, 406, catch-all 500), as does `AuthInterceptor`. So for an in-app failure the `error.message` branch already carries the server's reason. The problems are narrower than "the server message is always lost":

1. **`certs.tsx` cannot see a server message that is not in that envelope.**

   ```ts
   const getErrorMessage = (error: unknown): string =>
     error instanceof Error && error.message ? error.message : '请求失败，请稍后重试';
   ```

   The interceptor's rewrite only fires when `getBusinessError` finds a `code` in the body (`api/client.ts:37`). A rejection whose body is *not* the app envelope - a 502/504 from a gateway or reverse proxy in front of Studio, a servlet-container error page, a JSON body from the platform the cert APIs bridge to - leaves `error.message` as axios's transport string while `response.data.message` still holds something useful. That is precisely the case `describeApiError` was written for, and this page is the only one of the four that cannot benefit from it. All three of its call sites (list load, create, delete) funnel through it.

2. **The three identical copies read a property of the thrown value with no guard.**

   ```ts
   type ApiErrorLike = { message?: unknown; response?: { data?: { message?: unknown } } };

   const getErrorMessage = (error: unknown, fallback: string): string => {
     const apiError = error as ApiErrorLike;
     const responseMessage = apiError.response?.data?.message;   // <- `apiError` itself is unguarded
     ...
   };
   ```

   The optional chaining starts one level too late. A rejection that carries no value - `Promise.reject()`, or awaiting a promise rejected with `undefined` - makes the extractor itself throw a `TypeError` **inside the catch block**. The `finally` blocks still run (a `finally` runs even when its `catch` throws), so spinners do clear, but no error is ever rendered and the `TypeError` escapes as an unhandled rejection: the failure is silent. Latent rather than observed, but `describeApiError` already guards this correctly with `(error as {...})?.response?.data?.message`.

3. **Four copies of one policy is how they drifted.** Ten call sites duplicated the same 11-line body and the same `ApiErrorLike` declaration across three files, and `certs.tsx` ended up with a fourth variant that does not do what the other three do. The next fix to this policy has to be found and applied four times.

**The change**

Every call site now uses the idiom the AI pages already established, with its existing fallback string preserved verbatim:

```ts
message.error(describeThrownMessage(error) || t('messagePage.directConsumeFailed'));
setLoadError(describeThrownMessage(error) || DEFAULT_LOAD_ERROR);
```

`describeThrownMessage` gained one branch so it is a strict superset of the copies it replaces - they accepted a `message` string on *any* thrown value, not only on `Error` instances:

```ts
const thrownMessage = (error as { message?: unknown } | null | undefined)?.message;
if (typeof thrownMessage === 'string' && thrownMessage.trim()) return thrownMessage;
```

Order of preference: server message, then `Error.message`, then a bare thrown object's `message`, then `''` so the caller's `||` fallback applies. That matches the function's own doc comment ("best available machine-supplied message for anything thrown") and is a no-op for the AI hooks, which already call it and only ever receive `Error` instances.

Out of scope, deliberately: the still-hardcoded Chinese fallbacks in `certs.tsx` and `dlq.tsx`. Localising those pages is a separate change; this PR does not add or modify any UI text.

```
cd web
npx vitest run src/utils/apiError.test.ts src/pages/cluster/__tests__/K8sCertsPage.test.tsx \
               src/pages/cluster/__tests__/ClientsPage.test.tsx src/pages/instance/__tests__/DLQPage.test.tsx \
               src/pages/instance/__tests__/MessagePage.test.tsx
 Test Files  5 passed (5)
      Tests  73 passed (73)

npx vitest run
 Test Files  1 failed | 139 passed (140)
      Tests  1 failed | 1289 passed (1290)
  the single failure is ConsumerPage > 'keeps the latest client stack when an older request
  resolves last', a timing-sensitive race assertion in a file this PR does not touch; it passes
  36/36 when that file is run on its own.

npx tsc -b              (exit 0)
npx eslint <the 7 changed files>
  exit 0; one pre-existing react-refresh/only-export-components warning on dlq.tsx's
  `export const formatDateTime`, present on the trunk revision too and not touched here
```

New unit tests, `web/src/utils/apiError.test.ts` (the helper had none):

- `describeApiError` prefers the server message, falls back on a blank or non-string server message and on a plain `Error`, and does not throw for `undefined` / `null`;
- `describeThrownMessage` prefers the server message over the transport message, falls back to `Error.message`, accepts a bare `{ message: 'rate limited' }`, and returns `''` for `undefined`, `null`, `{}`, `{ message: 42 }` and `new Error('   ')`.

New page tests in `K8sCertsPage.test.tsx`, each pinning behaviour the old one-liner did not have:

- `surfaces the server rejection reason when the certificate list cannot be loaded` - a rejection carrying `response.data.message = 'k8s cluster unreachable'` alongside a transport-level `Error.message` shows the server text and **not** `Request failed with status code 500`;
- `accepts a rejection reason carried by a bare object`;
- `keeps the generic fallback when the rejection carries nothing usable`.

Both changes were mutation-checked:

- restoring the old `certs.tsx` one-liner fails exactly the first two new page tests; the seven pre-existing ones stay green;
- removing the new bare-object branch from `describeThrownMessage` fails exactly `accepts a bare object thrown without an Error` and its page-level counterpart.

### Checklist

- [x] Tests added or updated for non-trivial changes, test methods named `...Test` (frontend; Java test naming does not apply)
- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (no UI text added or changed - every existing fallback string is preserved verbatim)
- [x] New source files carry the ASF license header (`web/src/utils/apiError.test.ts`)
- [x] Documentation touched where behaviour changed (README / `docs/` / in-app help) - none needed; `apiError.ts`'s doc comment already describes the widened contract


RockteMQ-AI (approved):
### Summary
Clean refactoring - consolidates API error extraction into shared utility with comprehensive tests.

LGTM.

---

Issue #4845, opened 2026-09-22
[Bug] A Node.js client connection is reported with no language, and the console calls it unknown


Wang1rrr:
### Before Creating the Bug Report

- [x] I have searched the [open issues](https://github.com/apache/rocketmq-dashboard/issues) of this repository and believe that this is not a duplicate.
- [x] This is a defect in RocketMQ Studio, not a usage question and not a defect in another Apache RocketMQ repository.
- [x] I can reproduce this on the current `rocketmq-studio` branch (the Studio trunk the pull request template points at), commit `4c697f07acde460e2344375cb1f82669f5b270fd`.

### Studio Version

```
branch: rocketmq-studio
git commit id: 4c697f07acde460e2344375cb1f82669f5b270fd
deployed as: built from source
```

### Runtime Environment

```
OS: Windows 11 (development box); the defect is in the Apache provider's enum mapping, so the host does not matter
MySQL: not required - reproduced at the provider layer against a mocked MQAdminExt producer table
browser (for UI issues): any; the affected console page is web/src/pages/cluster/clients.tsx
```

### Connected RocketMQ Cluster

```
RocketMQ version: 5.5.0 (rocketmq-remoting 5.5.0 on the Studio classpath), self-managed
access mode: Direct to NameServer/Broker - the same path serves Proxy Local and Proxy Cluster
deployment: any cluster with at least one Node.js SDK client connected
```

### Describe the Bug

`ClientLanguage` has eight members (`common/domain/enums/ClientLanguage.java:21`):

```java
Java, Go, Python, Rust, Cpp, CSharp, NodeJS, PHP
```

`RocketMQClientProvider.mapLanguage` (`provider/apache/RocketMQClientProvider.java:439`) is the only place that turns a broker-reported `LanguageCode` into one of them, and its switch has seven cases - `JAVA`, `GO`, `PYTHON`, `RUST`, `CPP`, `DOTNET`, `PHP` - with `default: return null`. `LanguageCode.NODE_JS` exists in rocketmq-remoting 5.5.0 (`javap org.apache.rocketmq.remoting.protocol.LanguageCode` lists it after `RUST`), so a Node.js client falls through to `null` and `ClientConnectionVO.language` (`cluster/client/ClientConnectionVO.java:40`) is serialized as `"language": null` - the VO carries no `@JsonInclude(NON_NULL)` and there is no global inclusion override.

The console is already built to display that value, which is what makes the gap visible:

[...]

PR #4846, opened 2026-09-22
fix(apache): report a Node.js client language instead of dropping it

Wang1rrr:

- Fixes #4845

### Brief Description

`RocketMQClientProvider.mapLanguage` (`provider/apache/RocketMQClientProvider.java:439`) is the only mapping from the broker's `LanguageCode` to Studio's `ClientLanguage`, and it covers seven of the eight `ClientLanguage` members. `LanguageCode.NODE_JS` - present in rocketmq-remoting 5.5.0 - has no case, so it hits `default: return null` and every Node.js client is published with `"language": null`.

The rest of the stack already expects that member:

| layer | state before this PR |
| --- | --- |
| `ClientLanguage` (`:21`) | declares `Java, Go, Python, Rust, Cpp, CSharp, NodeJS, PHP` |
| `languageConfig` (`web/src/pages/cluster/clients.tsx:101`) | has `NodeJS: { color: 'lime', label: 'Node.js' }` |
| Language column filter (`clients.tsx:473`, `:478`) | offers a Node.js option that can never match `null` |
| Distribution panel (`clients.tsx:302`, `:806`) | buckets the connection under the literal language `null` |
| CSV export (`clients.tsx:113`) | writes an empty `Language` cell |
| Client diagnostics (`clientConnectionDiagnostics.ts:107`, `:315`) | `KNOWN_LANGUAGES` contains `NodeJS`, `normalizeText(null)` is `unknown`, so every Node.js connection raises `UNKNOWN_LANGUAGE` (客户端语言未知) |
| `docs/api-spec.md:1597` | documents the field as `Java` / `Go` / `Python` / `Rust` / `C++` / `C#` / `Node.js` / `PHP` |

So a healthy Node.js SDK client was rendered as an empty tag, was unfilterable, was charted as `null`, and was reported to the operator as a fleet defect - the diagnostic even suggests 补充语言展示映射, which is exactly what the provider was missing.

**The change** adds the missing case and records why the remaining `LanguageCode` values stay unmapped:

```java
case NODE_JS:
    return ClientLanguage.NodeJS;
default:
    // LanguageCode values with no ClientLanguage counterpart (RUBY, HTTP, OMS, OTHER,
    // DELPHI, ERLANG) stay unmapped; the client diagnostics report them as unknown.
    return null;
```

Keeping `null` for `RUBY` / `HTTP` / `OMS` / `OTHER` / `DELPHI` / `ERLANG` is deliberate: `ClientLanguage` has no member for them, and for those the `UNKNOWN_LANGUAGE` diagnostic is honest rather than a false alarm.

Deliberately out of scope: `clients.tsx:302` interpolates a nullable field into the distribution label, so any still-unmapped language shows up as the literal `null` in that chart. That guard belongs in the web layer and is a separate change.

```
cd server
mvn -o -B test -Dtest=RocketMQClientProviderTest,ClientServiceTest
 Tests run: 37, Failures: 0, Errors: 0, Skipped: 0
 BUILD SUCCESS
```

Checkstyle runs in the `validate` phase of this build (`failsOnError=true`, `includeTestSourceDirectory=true`), so the run also proves the new code and test pass `style/rmq_checkstyle.xml`.

One new test in `RocketMQClientProviderTest` (30 -> 31), `connectionScanShouldReportEveryStudioClientLanguageTest`: it feeds a producer table holding one connection per `LanguageCode` that Studio has a name for (`JAVA`, `GO`, `PYTHON`, `RUST`, `CPP`, `DOTNET`, `PHP`, `NODE_JS`) through `findConnectionsAt(..., "Producer")` and asserts the resulting languages are `containsExactlyInAnyOrder(ClientLanguage.values())`.

That assertion is the invariant that was broken, not just the reported symptom: it fails if any `ClientLanguage` member cannot be produced by the scan, so adding a ninth member without a mapping case breaks the build instead of silently nulling a fleet.

Mutation-checked: deleting only the new `case NODE_JS:` makes `mvn -o -B test -Dtest=RocketMQClientProviderTest` report `Tests run: 31, Failures: 1` - the new test - and leaves the other 30 green, including `connectionVersionShouldBeResolvedFromMQVersionCodeTest`, which exercises the same scan path.

### Checklist

- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (no new UI text; `clients.tsx` already renders `NodeJS`)
- [x] Documentation touched where behaviour changed (README / `docs/` / in-app help) - none needed; `docs/api-spec.md:1597` already documents `Node.js` as a client language, and this change is what makes the provider honour it


RockteMQ-AI (approved):
### Review

✅ **Approve** — This is a straightforward fix for a real bug (Node.js clients showing as `null` in the dashboard). The code is correct, the test coverage is solid, and the scope is appropriately minimal.

---

PR #4847, opened 2026-09-22
fix(topic): apply the topic-type send policies to the REST send path

Frun1na:

- Fixes #4848

### Brief Description

The topic-type dispatch rules for sending messages ("the registered topic type drives dispatch: FIFO
topics require messageGroup, DELAY topics require a future deliveryTimestamp, TRANSACTION topics are
rejected") were enforced only in the AI message-send tool (`MessageSendToolHandler`), not in the
`MetadataService.sendMessage` method that backs `POST /api/topics/send`. A FIFO topic could therefore
receive an un-grouped message through the REST path — the producer falls back to an arbitrary queue
and the message is recorded as a successful audit entry while silently violating the topic's ordering
guarantee.

The rules now live in one place, `MessageSendPolicies.validateForTopicType`, and are applied on every
surface that can send a message:

- `MetadataService.sendMessage` resolves the topic's registered type and validates before dispatching
  to the provider (the same `findTopic` lookup the AI tool already performed).
- `MessageSendToolHandler` delegates its preview and execute validation to the same policy, so the
  tool's dry-run plan keeps refusing before execute and both surfaces return identical 400 messages.

Unknown topics (no registered entry) keep the plain-send behavior, matching the tool's existing
documented semantics. The AI tool's execute path loses its redundant private copy of the rule; its
preview path keeps the check so dry-run plans fail early.

TDD. New `MetadataServiceTest` cases cover the REST path: FIFO without `messageGroup` → 400, FIFO
with `messageGroup` → forwarded to the provider, DELAY without / with a past `deliveryTimestamp` →
400, TRANSACTION topic → 400, and an unregistered topic → plain-send behavior preserved. They fail
against the current code (no exception raised) and pass after the fix:

```
mvn -B -ntp test -Dtest='MetadataServiceTest,MessageSendToolHandlerTest'

Tests run: 60, Failures: 0, Errors: 0, Skipped: 0 -- in ...MetadataServiceTest
Tests run: 14, Failures: 0, Errors: 0, Skipped: 0 -- in ...MessageSendToolHandlerTest
Tests run: 74, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The existing `MessageSendToolHandlerTest` FIFO/DELAY/TRANSACTION cases pass unchanged: the tool still
refuses before touching the service, now through the shared policy.


RockteMQ-AI (approved):
﻿### Summary

Solid fix — adds topic-type validation to the REST send path, ensuring FIFO topics require messageGroup and DELAY topics require future timestamps. Good implementation with proper error handling.

### Findings

- **[Info]** `MessageSendPolicies.java` — Clean policy class with clear validation rules. Shared by REST and AI paths for consistency.
- **[Info]** Tests cover FIFO, DELAY, and NORMAL topic types with proper error cases.

✅ **Approve** — Well-implemented validation with good test coverage.

---

Issue #4849, opened 2026-09-22
[Bug] The clients page charts an unmapped client language as null and renders an empty language tag


Wang1rrr:
### Before Creating the Bug Report

- [x] I have searched the [open issues](https://github.com/apache/rocketmq-dashboard/issues) of this repository and believe that this is not a duplicate.
- [x] This is a defect in RocketMQ Studio, not a usage question and not a defect in another Apache RocketMQ repository.
- [x] I can reproduce this on the current `rocketmq-studio` branch (the Studio trunk the pull request template points at), commit `4c697f07acde460e2344375cb1f82669f5b270fd`.

### Studio Version

```
branch: rocketmq-studio
git commit id: 4c697f07acde460e2344375cb1f82669f5b270fd
deployed as: built from source
```

### Runtime Environment

```
OS: Windows 11 (development box)
MySQL: not required - reproduced with the web test harness against a mocked connections service
browser (for UI issues): any; the affected page is web/src/pages/cluster/clients.tsx
```

### Connected RocketMQ Cluster

```
RocketMQ version: 5.5.0, self-managed (the Apache provider is the only ClientProvider implementation)
access mode: Direct to NameServer/Broker
deployment: any cluster with a client whose LanguageCode has no ClientLanguage counterpart
```

### Describe the Bug

`ClientConnectionVO.language` is a nullable `ClientLanguage` (`cluster/client/ClientConnectionVO.java:40`), and `RocketMQClientProvider.mapLanguage` (`:439`) returns `null` for every `LanguageCode` that `ClientLanguage` has no name for - `RUBY`, `HTTP`, `OMS`, `OTHER`, `DELPHI`, `ERLANG`. The VO carries no `@JsonInclude(NON_NULL)` and there is no global inclusion override, so those clients are published as `"language": null`.

The TypeScript contract does not say that. `web/src/api/connections.ts:10` declares `language: string`, while three sibling fields in the very same interface are already honest about the server being allowed to omit a value (`clientId?: string | null` `:6`, `address?: string | null` `:9`, `connectedAt?: string | null` `:12`). Because the type claims the field is always present, `strictNullChecks` cannot flag any of the unsafe uses, and there are four:

[...]

PR #4850, opened 2026-09-22
fix(web): stop charting an unmapped client language as null

Wang1rrr:

- Fixes #4849

### Brief Description

`ClientConnectionVO.language` is nullable on the server and the TypeScript contract said it was not, so nothing in the clients page was forced to cope - and three of its four consumers render the gap.

| site | before |
| --- | --- |
| `web/src/api/connections.ts:10` | `language: string` while `clientId`, `address` and `connectedAt` in the same interface are already `?: string \| null` |
| distribution panel (`clients.tsx:302`, `:806`, `:808`) | `` `${connection.language} ${connection.version}` `` -> a tag literally labelled `null 5.0.7: 1` |
| Language column (`clients.tsx:479`) | `languageConfig[null] ?? { label: null }` -> an empty grey tag |
| detail drawer (`clients.tsx:976`) | the same lookup, same empty tag |
| export filter (`clients.tsx:366`) | `String(filterValue) === null` |

`null` is a permanent part of that response, not a transient one: `RocketMQClientProvider.mapLanguage` (`:439`) returns it for every `LanguageCode` that `ClientLanguage` has no member for - `RUBY`, `HTTP`, `OMS`, `OTHER`, `DELPHI`, `ERLANG` - and the VO carries no `@JsonInclude(NON_NULL)`. #4846 removes one source (`NODE_JS`, which `ClientLanguage` does cover); the other six stay `null` by design, because for those the `UNKNOWN_LANGUAGE` diagnostic in `clientConnectionDiagnostics.ts:315` is the honest reading.

**The change**

1. `web/src/api/connections.ts` - `language?: string | null`, so `strictNullChecks` now covers every consumer of the field instead of silently allowing `null` through a `string` type. `version` stays `string`: `MQVersion.getVersionDesc(int)` clamps out-of-range codes and returns `Version.values()[...].name()` on every path, so the provider always sets one.
2. `clients.tsx` - a module-level `UNKNOWN_LANGUAGE = 'unknown'` bucket key, so the chart aggregates unmapped languages under a stable key instead of interpolating `null` into the label.
3. `clients.tsx` - one `renderLanguageTag(language?)` helper used by both the column and the drawer, falling back to `t('common.unknown')` when `languageConfig` has no entry.
4. `web/src/i18n/translations.ts` - `'common.unknown': { zh: '未知', en: 'Unknown' }`, added next to `common.unavailable` and reusing the wording `liteTopic.unknown` (`:3435`) and `dashboardTraffic.trendUnknown` (`:137`) already carry.
5. `clients.tsx:366` - `matches('language', connection.language ?? '')`, keeping the export filter typed and leaving its semantics unchanged (an unmapped language matches no filter option, exactly as before).

`clientConnectionDiagnostics.ts` needed no change: it already funnels the field through `normalizeText(value?: string | null)` (`:109`).

```
cd web
npx vitest run src/pages/cluster/__tests__/ClientsPage.test.tsx src/api/connections.test.ts src/utils/clientConnectionDiagnostics.test.ts
 Test Files  3 passed (3)
      Tests  28 passed (28)

npx tsc -b                                                    # exit 0
npx eslint src/api/connections.ts src/i18n/translations.ts src/pages/cluster/clients.tsx src/pages/cluster/__tests__/ClientsPage.test.tsx
                                                              # exit 0
npx prettier --check --end-of-line auto <the same four files>  # All matched files use Prettier code style!
```

One new case in `ClientsPage.test.tsx`, `labels a connection whose language the backend could not map`: it serves a connection with `language: null` and asserts the table row shows 未知, that the distribution panel reads `未知 5.0.7: 1`, and that nothing in the panel matches `/null/`.

Mutation-checked in both directions:

- reverting only the bucket key to `` `${connection.language} ${connection.version}` `` fails the distribution assertions;
- reverting only the tag fallback to `config?.label ?? language` fails the table-row assertion with `Unable to find an element with the text: 未知`.

So each half of the fix is pinned by its own assertion, and the remaining 27 tests in those three files stay green.

### Checklist

- [x] Tests added or updated for non-trivial changes, test methods named `...Test` (web suite; the new `it(...)` follows the descriptive naming used throughout `ClientsPage.test.tsx`)
- [x] New UI text has both Chinese and English entries under `web/src/i18n/` (`common.unknown`: 未知 / Unknown)
- [x] Architecture constraints stay green (`mvn test` runs the ArchUnit checks; no Java change here)
- [x] Documentation touched where behaviour changed (README / `docs/` / in-app help) - none needed; `docs/api-spec.md:1597` documents the field by its display labels and the table does not mark nullability for any field, so the rendered vocabulary is unchanged apart from no longer containing `null`


RockteMQ-AI (approved):
### Summary

Good fix that correctly handles nullable `language` from the broker. The TypeScript type update (`language?: string | null`) aligns the contract with reality, and the `renderLanguageTag` helper with the localized `common.unknown` label prevents raw `null` from appearing in the UI.

The language-version distribution chart fix (bucketing null into `unknownLabel`) and the test coverage for both the table row and the distribution chart are solid.

LGTM.

---

PR #4852, opened 2026-09-22
fix(ai): make the retention purge stop live runs and remove workspaces

Frun1na:

- Fixes #4853

### Brief Description

The scheduled retention sweep (`purgeExpired` → `deleteExpiredConversations` → `deleteCascade`) and
the manual `delete` took two different paths through the same lifecycle event:

- `delete` stops the conversation's active run through the registry and removes the agent workspace
  (`RmqctlWorkspace.delete`) before the rows disappear.
- `deleteCascade` did neither: it dropped the event / run / conversation rows directly.

Two consequences:

1. **Workspace leak.** `RmqctlWorkspace`'s contract is "kept for the conversation's lifetime;
   `delete(long)` removes them with the conversation" — and the retention purge is the end of that
   lifetime. Every purged conversation whose agent ever ran leaves a `conv-<id>/` directory (config,
   prompt, `home/` with the CLI's resume state) under the configured workspace dir, on a volume the
   class documents as persistent. Nothing else ever reclaims it.
2. **Purging under a live run.** `delete` stops the active run first, and `reapOrphanedRuns` checks
   `registry.isLive` before touching anything; `deleteCascade` checked nothing. A conversation created
   past the retention cutoff but still streaming gets its rows deleted mid-run: the worker keeps
   appending orphan event rows (the sink's per-insert tolerance accepts them), the finalize update is
   a no-op against the missing row, and the subprocess keeps writing into a workspace that now belongs
   to nothing.

The cascade now mirrors `delete`: it stops any run the registry still owns before deleting the rows,
and removes each conversation's workspace afterwards (`delete` never throws, matching the batch's
failure tolerance).

Two new cases in `AiConversationPersistenceIntegrationTest` (real database, real workspace bean):

- `theRetentionSweepShouldRemoveTheConversationWorkspaceTest` — creates the workspace directory with
  a config file and a child directory, purges, and asserts the directory is gone. Fails before the
  fix (the directory survives the purge).
- `theRetentionSweepShouldStopALiveRunBeforeDeletingTheConversationTest` — marks the seeded run
  RUNNING, purges, and asserts the conversation, its runs and its events are all gone (the sweep
  mirrors `delete` rather than skipping or stranding the live run).

```
mvn -B -ntp test -Dtest=AiConversationPersistenceIntegrationTest

Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
```

Full conversation-area suite:

```
mvn -B -ntp test -Dtest='AiConversation*Test,AiRun*Test,AgentRun*Test,AiEvent*Test,AiStream*Test'

Tests run: 442, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```


RockteMQ-AI (approved):
### Summary

Good fix that ensures the retention purge mirrors the manual `delete` behavior. Previously, the scheduled sweep would delete conversation rows without stopping active agent runs or cleaning up workspace directories — this could strand an active run writing into a deleted conversation and leak volume space.

The implementation correctly:
1. Stops active runs via `registry.stop()` before row deletion
2. Removes workspace directories after conversation deletion
3. Logs each pre-purge run stop for observability

The integration tests cover both scenarios (active run stopped, workspace removed) and verify the cascade order is preserved.

LGTM.

---

Issue #4851, opened 2026-09-22
[Bug] Unreachable message-track fallback maps TrackType.UNKNOWN to a failed delivery status


Wang1rrr:
### Before Creating the Bug Report

- [x] I have searched the [open issues](https://github.com/apache/rocketmq-dashboard/issues) of this repository and believe that this is not a duplicate.
- [x] This is a defect in RocketMQ Studio, not a usage question and not a defect in another Apache RocketMQ repository.
- [x] I can reproduce this on the current `rocketmq-studio` branch (the Studio trunk the pull request template points at), commit `4c697f07acde460e2344375cb1f82669f5b270fd`.

### Studio Version

```
branch: rocketmq-studio
git commit id: 4c697f07acde460e2344375cb1f82669f5b270fd
deployed as: built from source
```

### Runtime Environment

```
OS: Windows 11 (development box)
MySQL: not required - this is a static reachability and mapping defect, confirmed against the bytecode of rocketmq-tools 5.5.0
browser (for UI issues): n/a; no request reaches this code on any page
```

### Connected RocketMQ Cluster

```
RocketMQ version: 5.5.0 (rocketmq-tools 5.5.0 on the Studio classpath)
access mode: n/a - the code in question is never invoked for any access mode
deployment: n/a
```

### Describe the Bug

`RocketMQMessageProvider` carries a 41-line private fallback for building a message's consumer status from the broker instead of from the trace topic:

- `fallbackConsumerStatus(DefaultMQAdminExt, MessageExt)` at `provider/apache/RocketMQMessageProvider.java:811`, which calls `adminExt.messageTrackDetail(message)`;
- `mapTrackType(TrackType)` at `:832`, its only helper.

**Neither is reachable.** `fallbackConsumerStatus` has no caller anywhere in `server/src` (main or test), and `mapTrackType` is called only from inside it. The trace path builds its consumer status exclusively from parsed trace-topic records via `buildConsumerStatus` (`:777`). The method was already caller-less when it was introduced in `dc49e1a2` (feat: manage topics, groups, messages and DLQ against a live cluster, #798) and was carried through unchanged by `ceef946f` (refactor: unify multi-vendor architecture and backend code style, #1231).

That matters because the unreachable code is also **wrong**, and it reads as if it were live:

```java
private DeliveryStatus mapTrackType(TrackType trackType) {
    if (trackType == null) {
        return DeliveryStatus.pending;      // "we could not tell" -> pending
    }
[...]

PR #4854, opened 2026-09-22
refactor(message): drop the unreachable message-track fallback

Wang1rrr:

- Fixes #4851

### Brief Description

`RocketMQMessageProvider` has carried a private, 41-line fallback for building a message's consumer status from the broker since #798, and no request has ever reached it:

```java
private List<ConsumerStatusVO> fallbackConsumerStatus(DefaultMQAdminExt adminExt, MessageExt message) { ... }
private DeliveryStatus mapTrackType(TrackType trackType) { ... }
```

`fallbackConsumerStatus` has no caller in `server/src` (main or test) and `mapTrackType` is called only from inside it. It was already caller-less in the commit that introduced it (`dc49e1a2`) and survived `ceef946f` (#1231) unchanged. The live path builds consumer status from parsed trace-topic records instead (`buildConsumerStatus`, `:777`).

The reason this is worth removing rather than leaving alone is that the dead mapping is wrong, and it reads exactly like implemented behaviour:

| `TrackType` | mapped to | means |
| --- | --- | --- |
| `CONSUMED`, `CONSUME_BROADCASTING`, `CONSUMED_BUT_FILTERED` | `success` | delivered |
| `NOT_CONSUME_YET`, `PULL`, `NOT_ONLINE` | `pending` | not delivered yet / no consumer online |
| `null` | `pending` | could not tell |
| **`UNKNOWN`** (via `default`) | **`failed`** | **could not tell** |

`TrackType` has exactly seven members in rocketmq-tools 5.5.0, so `UNKNOWN` is the only value that reaches `default` - and `DefaultMQAdminExtImpl.messageTrackDetail` sets `UNKNOWN` as the *initial* value on every group's `MessageTrack` before it queries anything, replacing it with `NOT_ONLINE` only when the broker answers 206 (`CONSUMER_NOT_ONLINE`). Any other failure - unreachable broker, timeout, `RemotingException` - leaves it at `UNKNOWN`.

So the same "we could not determine this" state is `pending` when spelled `null` and `failed` when spelled `UNKNOWN`. Wired in as written, that would put a red 失败 (`DELIVERY_STATUS_MAP`, `web/src/pages/instance/message.tsx:108`) on **every** consumer group of the topic, on exactly the clusters where the admin RPCs are failing.

**The change** deletes both methods and the two imports that become unused (`org.apache.rocketmq.tools.admin.api.MessageTrack` and `TrackType`). No behaviour changes, because nothing could call them.

If the community would rather have the fallback than not, that is a separate and deliberate change, and the issue lists what it needs: `UNKNOWN` mapped to `pending` (matching the `null` case here and the `pending` fallback `AliyunConverters:327` already uses), a decision on the cost - `messageTrackDetail` issues one `queryTopicConsumeByWho` plus one `examineConsumerConnectionInfo` **per subscribed group**, from the `TOPIC_NOT_EXIST` / `NO_MESSAGE` branch at `:600`, i.e. the slow path of a page already blocked on RPCs - and a way to get the `MessageExt`, which `resolveMessageStoreTimestamp` (`:670`) already fetches and discards. I am happy to send that instead if it is wanted; this PR just stops shipping the trap.

```
cd server
mvn -o -B test "-Dtest=RocketMQMessageProvider*,Message*Test,*TraceTest"
 Tests run: 137, Failures: 0, Errors: 0, Skipped: 0
 BUILD SUCCESS
```

That covers `RocketMQMessageProviderTest` (56), `MessageServiceTest`, `MessageControllerTest` and the seven message/trace AI tool handler suites. Checkstyle runs in the `validate` phase (`failsOnError=true`, `includeTestSourceDirectory=true`), so the build also proves no unused import or reference survived.

Reachability was established before deleting, not assumed:

```
rg -n "fallbackConsumerStatus|mapTrackType|messageTrackDetail" server/src
# three hits, all inside the two methods themselves
```

and the removal script asserts that none of `MessageTrack`, `TrackType`, `fallbackConsumerStatus`, `mapTrackType` or `messageTrackDetail` appears anywhere in the file afterwards, so a half-deletion cannot slip through. Checkstyle's `UnusedImports` is the second net: leaving the imports behind fails `validate` (confirmed while preparing this change).

No new tests: there is no reachable behaviour left to pin. The trace parsing that *is* live stays covered by the existing `RocketMQMessageProviderTest` cases, all of which pass unchanged.

### Checklist

- [x] Tests added or updated for non-trivial changes, test methods named `...Test` - none added; this deletes unreachable private code and the 137 existing message/trace tests pass unchanged
- [x] Documentation touched where behaviour changed (README / `docs/` / in-app help) - none needed; no reachable behaviour changed, so no documented contract did either


RockteMQ-AI (approved):
### Summary

Clean dead-code removal. The `fallbackConsumerStatus` method and its `mapTrackType` helper have been unreachable since #798 switched to `consumerStatusV2` — removing them reduces maintenance burden and eliminates dead imports (`MessageTrack`, `TrackType`).

LGTM.

---

PR #4855, opened 2026-09-22
fix(studio): repair MCP tool gating, smooth AI streaming, and polish console surfaces

lizhimins:
## What this PR does

One squashed commit covering the MCP tool-layer repairs, the AI conversation streaming work, cluster/clients console polish, rmqctl additions and the test-suite stabilization.

### MCP tool layer (server)

- `rmq.proxy.list` / `rmq.proxy.config` no longer require `PROXY_DISCOVERY` on the bound instance. Both are platform-level tools (no `instanceId`) whose handlers aggregate every manageable instance through the heartbeat-syncer group, so their data scope never depended on the bound instance — but the capability gate did. Since `PROXY_DISCOVERY` is only granted to `PROXY_LOCAL`/`PROXY_CLUSTER` types, direct-attached deployments got a permanent `TOOL_CAPABILITY_UNSUPPORTED` even with proxies running. They now join `rmq.audit.list` / `rmq.alert.rule.list` as ungated deployment-wide tools.
- `rmq.instance.metrics` accepts the sole shared data source whose bindings include the target instance, still preferring a dedicated one; ambiguous or missing bindings keep answering 409.
- CORS `allowedMethods` gains **PATCH**: browsers attach an Origin header to every non-GET request, so the missing verb answered 403 "Invalid CORS request" on conversation rename/archive in every real browser (invisible to curl).

### AI conversation surface (web)

- Transcript bubbles and their blocks are memoised: a streaming tick re-renders the page at display rate, and every frame previously re-ran ReactMarkdown over every persisted answer and reconciled every finished tool card (up to 32 KiB `<pre>` each).
- The follow-scroll uses instant jumps while streaming (a fresh `smooth` scroll per animation frame stacked concurrent animations); idle arrivals stay smooth.
- The operator's question shows immediately via an optimistic user bubble — the live stream carries no `user` frame and the transcript only refetches when the run finishes, so the question was invisible for minutes on long tool-heavy runs. Dropped in the same commit as the end-of-run refetch; kept when the refetch fails; a re-attach finds the row already persisted.
- History modal gains a per-row archive/unarchive action (the PATCH endpoint and the active/archived Segmented already existed).
- Context meter pairs a mini bar with the percent number (hidden below 1%), composer header wraps on narrow viewports, send button slims to a 32px circle.

### Cluster / clients console (web)

- Broker config modal compacted into a grouped two-column grid; the permission bitfield becomes a named Select (读写 6 / 只读 4 / 只写 2 / 无权限 0); the hard-coded English queue hint becomes a translated tooltip.
- Client diagnostics tables clip long values with ellipsis + title instead of spilling over neighbouring columns (project table rule).

### rmqctl

- Native Windows build script; plain HTTP allowed for private-network Studio endpoints.

### Tests

- Stabilise three flake families under parallel load: antd Spin's `.ant-spin-blur` lagging the Table `loading` prop one commit (`pointerEventsCheck: 0` on table-driving suites, the existing convention), CSV-import waits racing the instance load (wait for the instance-scoped fetch instead of a first-paint text), and static `message`/`notification` holders leaking across tests (global `afterEach` destroy).
- New coverage: metrics source selection, optimistic user bubble, scroll behaviour, CORS PATCH, archive flow, diagnostics race pinned to fake intervals.

## Verification

- server `mvn test` 3150/3150 green; web `vitest` 1285/1285 green (multiple consecutive full runs after the stabilization commit); `tsc -b` and eslint clean
- deployed and exercised end-to-end: proxy tools return live endpoints, metrics resolve through the shared source, second questions render instantly with zero long tasks during streaming, archive round-trip verified in the browser


RockteMQ-AI (approved):
### Summary

Well-structured PR addressing three related areas: MCP tool accessibility, AI conversation streaming performance, and test suite stability.

**MCP Tool Layer:**
- Removing the `PROXY_DISCOVERY` gate from `rmq.proxy.list`/`rmq.proxy.config` and the capability gate from `rmq.consumer.group.list` makes these tools available in standard deployments — good for AI assistant usability.
- The metrics data source binding relaxation (accepting shared sources when no dedicated source exists) is a pragmatic improvement that reduces configuration friction.

**Streaming Performance:**
- Memoizing `TextBlock`, `ThinkingBlock`, `ToolBlock`, and `AssistantBubble` with `React.memo` is the right approach — these components re-render at display rate during streaming, and the 32 KiB `<pre>` blocks in `ToolBlock` are expensive to reconcile.
- The `pendingUserText` optimistic bubble solves a real UX gap: the user's question was invisible during the run because the live stream carries no user frame and the persisted transcript only updates at end-of-run refetch.
- Memoizing context tokens to skip full ReactMarkdown re-parse on unchanged text is a solid optimization.

**Test Stabilization:**
- The `pointerEventsCheck: 0` changes across the test suite address flaky interactions with `userEvent` — this is a known workaround for overlapping element issues in jsdom.
- The ConsumerPage race condition fix (waiting for the instance-scoped fetch instead of the early subtitle text) is well-explained and correct.

LGTM.

---

Issue #4856, opened 2026-09-22
[Bug] An Aliyun message body that is not valid UTF-8 is reported with an undefined TEXT encoding


Wang1rrr:
### Before Creating the Bug Report

- [x] I have searched the [open issues](https://github.com/apache/rocketmq-dashboard/issues) of this repository and believe that this is not a duplicate.
- [x] This is a defect in RocketMQ Studio, not a usage question and not a defect in another Apache RocketMQ repository.
- [x] I can reproduce this on the current `rocketmq-studio` branch (the Studio trunk the pull request template points at), commit `4c697f07acde460e2344375cb1f82669f5b270fd`.

### Studio Version

```
branch: rocketmq-studio
git commit id: 4c697f07acde460e2344375cb1f82669f5b270fd
deployed as: built from source
```

### Runtime Environment

```
OS: Windows 11 (development box); the defect is in the Aliyun provider's message converter, so the host does not matter
MySQL: not required - reproduced at the converter layer against a stubbed ListMessagesResponseBody
browser (for UI issues): any; the affected console page is web/src/pages/instance/message.tsx
```

### Connected RocketMQ Cluster

```
RocketMQ version: Alibaba Cloud RocketMQ 5.x instance (OpenAPI rocketmq20220801)
access mode: cloud instance - AliyunInstanceProvider / AliyunConverters
deployment: any Aliyun instance holding at least one message whose body is not valid UTF-8
```

### Describe the Bug

`MessageRecordVO.bodyEncoding` has a published vocabulary of exactly two values plus "absent". `docs/api-spec.md:1335` documents the pair as

```
| body        | string | UTF-8 文本或 Base64 编码后的消息体 |
| bodyEncoding | string | 消息体编码，如 UTF-8 / BASE64     |
```

and every other provider sticks to it:

| producer | values emitted |
| --- | --- |
| `RocketMQMessageProvider.displayBody` (`provider/apache/RocketMQMessageProvider.java:881`) | `"UTF-8"` (`:897`), `"BASE64"` (`:901`), `null` for a null body (`:883`) |
| `TencentInstanceProvider.toRecordVO` (`provider/tencent/TencentInstanceProvider.java:808`, `:828`) | `"UTF-8"`, `null` |
| `AliyunConverters.toMessageRecord` (`provider/alibaba/AliyunConverters.java:240`) | `"UTF-8"`, **`"TEXT"`** |

The Aliyun branch is

```java
if (decodedBody != null) {
[...]

PR #4857, opened 2026-09-22
fix(aliyun): label a binary message body BASE64 instead of TEXT

Wang1rrr:

- Fixes #4856

### Brief Description

`AliyunConverters.toMessageRecord` published a third `bodyEncoding` value that the rest of Studio does not define, and attached it to the wrong payload.

`docs/api-spec.md:1335` documents the field as `UTF-8` / `BASE64`, `RocketMQMessageProvider.displayBody` (`provider/apache/RocketMQMessageProvider.java:881`) emits exactly those two plus `null`, and `TencentInstanceProvider.toRecordVO` (`provider/tencent/TencentInstanceProvider.java:808`, `:828`) emits `UTF-8` / `null`. The Aliyun path emitted `"TEXT"` - and put the **raw Base64** into `body` while doing so.

The cause is that `tryBase64Decode` collapsed three distinct outcomes into one `null`:

| # | situation | old result | correct result |
| --- | --- | --- | --- |
| 1 | API returned no body | `body=null`, `"TEXT"` | `body=null`, `null` |
| 2 | Base64 of a **binary** payload | `body=<base64>`, `"TEXT"` | `body=<base64>`, `"BASE64"` |
| 3 | body is not Base64 at all | `body=<raw>`, `"TEXT"` | `body=<raw>`, `"UTF-8"` |

Case 2 is the one `BASE64` exists for, and case 1 should carry no encoding - an encoding for a value that is not there.

**The change** keeps the three apart instead of flattening them:

```java
private static Body decodeBody(String raw) {
    if (raw == null || raw.isBlank()) {
        return null;                                    // case 1: no body, no encoding
    }
    byte[] bytes;
    try {
        bytes = Base64.getDecoder().decode(raw);
    } catch (IllegalArgumentException ignored) {
        return new Body(raw, "UTF-8");                  // case 3: already literal text
    }
    try {
        String text = StandardCharsets.UTF_8.newDecoder()
                .onMalformedInput(CodingErrorAction.REPORT)
                .onUnmappableCharacter(CodingErrorAction.REPORT)
                .decode(ByteBuffer.wrap(bytes))
                .toString();
        return new Body(text, "UTF-8");                 // text payload
    } catch (CharacterCodingException ignored) {
        return new Body(raw, "BASE64");                 // case 2: binary, Base64 is faithful
    }
}
```

and `toMessageRecord` just forwards the pair:

```java
Body body = decodeBody(data.getBody());
...
if (body != null) {
    builder.body(body.value()).bodyEncoding(body.encoding());
}
```

This is the same shape `RocketMQMessageProvider.displayBody` already uses (strict UTF-8 decode, fall back to Base64 with a `BASE64` label), so the three providers now agree on the vocabulary and a consumer that switches on the documented values takes the same branch regardless of vendor.

Why it mattered beyond tidiness: `MessageItem.from` (`ops/ai/tool/contract/message/MessageItem.java:46`) and `MessageQueryOutput` (`:76`) forward the field verbatim, and the tool schema (`resources/tool-catalog/tools/message.yaml:86`, `:171`, `:254`) declares it as a bare `type: string` with no enum, so `rmq.message.query*` told the model `{"body": "<base64>", "bodyEncoding": "TEXT"}` for every binary message on an Aliyun instance - an agent reads that as literal text and quotes the Base64 back to the operator as the message content. `rmqctl` advertises the same fields (`internal/catalog/catalog_gen.go:389`, `:410`, `:419`).

Deliberately out of scope: `web/src/pages/instance/message.tsx` never consults `bodyEncoding` - `formatBody` (`:140`) only tries `JSON.parse` - so a `BASE64` body from *any* provider is displayed and downloaded as raw Base64. That belongs in the web layer and is a separate change.

```
cd server
mvn test -Dtest=Aliyun*
 Tests run: 68, Failures: 0, Errors: 0, Skipped: 0
 You have 0 Checkstyle violations.
 BUILD SUCCESS

mvn test -Dtest=*Message*,*Trace*,*Dlq*,*DLQ*
 Tests run: 226, Failures: 0, Errors: 0, Skipped: 0
 BUILD SUCCESS
```

Checkstyle runs in the `validate` phase of this build (`failsOnError=true`, `includeTestSourceDirectory=true`), so the run also proves the new code and test pass `style/rmq_checkstyle.xml`.

New test class `AliyunConvertersBodyEncodingTest` (5 cases), one per row of the table above plus the invariant:

- `shouldReportUtf8WhenTheBase64BodyDecodesToText`
- `shouldReportBase64WhenThePayloadIsBinary` - Base64 of `0xFF 0xFE`, asserts the body stays the Base64 and the label becomes `BASE64`
- `shouldKeepLiteralTextWhenTheBodyIsNotBase64` - `"{}"` keeps its own value and is labelled `UTF-8`
- `shouldReportNoEncodingWhenTheApiReturnedNoBody` - `null` and blank both yield `null`/`null`
- `shouldNeverEmitAnEncodingOutsideThePublishedVocabulary` - feeds a text body, a binary body, a non-Base64 string, an empty string and `null` through the converter and asserts every result is `isIn("UTF-8", "BASE64", null)`, so a fourth label cannot be reintroduced

One existing assertion updated: `AliyunInstanceProviderTest.queryMessagesShouldMapFieldsAndDecodeBase64BodyTest:544` pinned `"TEXT"` for a `"{}"` body. That fixture is case 3, so it now asserts `"UTF-8"`; the `body` assertion (`"{}"` passthrough) is untouched, which is the behaviour that test was really about.

Mutation-checked against the current trunk: restoring only the old `else { builder.body(rawBody).bodyEncoding("TEXT"); }` branch makes `mvn test -Dtest=AliyunConvertersBodyEncodingTest` report `Tests run: 4, Failures: 3` (`shouldReportBase64WhenThePayloadIsBinary`, `shouldReportNoEncodingWhenTheApiReturnedNoBody`, `shouldNeverEmitAnEncodingOutsideThePublishedVocabulary`), and the reverted converter also fails the updated `AliyunInstanceProviderTest` assertion. With the fix in place all 68 Aliyun tests and all 226 message/trace/DLQ tests are green.

### Checklist

- [x] Tests added or updated for non-trivial changes, test methods named `...` (the new converter tests follow the existing `AliyunConverters*Test` naming, which does not use the `...Test` method suffix; the updated provider test keeps its `...Test` suffix)
- [x] New source files carry the ASF license header (`AliyunConvertersBodyEncodingTest.java` does)
- [x] Documentation touched where behaviour changed (README / `docs/` / in-app help) - none needed; `docs/api-spec.md:1335` already documents `UTF-8` / `BASE64`, and this change is what makes the Aliyun provider honour it


RockteMQ-AI (approved):
### Summary

Clean fix that aligns the Aliyun provider's body encoding with the published API contract (`docs/api-spec.md`). The introduction of the `Body` record to pair value+encoding is a nice improvement over the previous `tryBase64Decode` approach, and the three-way branching (UTF-8 text / BASE64 binary / non-Base64 literal) correctly handles all cases.

Excellent test coverage with 5 cases covering null/blank, non-Base64 text, Base64-encoded UTF-8, Base64-encoded binary, and the original `TEXT` mislabeling bug.

LGTM.

---

PR #4871, opened 2026-09-22
fix(alert): degrade a corrupt labels row to empty labels

unbridled-41:
## Problem / Evidence

`MybatisPlusAlertRepository.readLabels` (and the same helper in `MybatisPlusAlertStateRepository`) throws `IllegalStateException` when a row's `labels_json` column does not parse. Because `findAlertsPage` maps every row of the page through `toAlertVO`:

- `GET /api/system-alerts/page` (and the legacy `/api/system-alerts`) fails with a 500 as long as the corrupt row is in the result set — the list endpoint is permanently broken until the row is fixed by hand.
- `AlertNotificationSuppressionService.findSuppressingClusterAlert` pages the same query inside every BUSINESS-domain FIRING/REMINDER evaluation, so one corrupt row aborts every business-rule evaluation of the native alerting pipeline (`NativeAlertProcessor` only counts and logs the failure) — notifications silently stop.

The row-level failure is escalated to a query-level failure even though `labels` is an optional, externally mutable TEXT column. The codebase's own convention for unreadable persisted payloads is skip/degrade (e.g. `AiConversationService.decode`).

Reproduced by a regression test: a page containing one healthy row (`{"cluster":"demo"}`) and one corrupt row (`{not-json`) threw `IllegalStateException: Unable to read alert labels` before the fix.

## Root cause / Fix

`readLabels` treated a row-level decode failure as fatal. Both repositories now degrade an unreadable `labels_json` to an empty label map so the read succeeds; the healthy row keeps its labels.

## Priority / scoring

PRIORITY 55 = impact 20 (system-alert list 500 + native alerting pipeline stops) + blast radius 20 (every page read and every business evaluation) + reproducibility 10 (needs one corrupt row, no user input) + maintenance value 5. FIX_CONFIDENCE 90.

## Tests

- `mvn test -Dtest=MybatisPlusAlertRepositoryTest#pageAlertsShouldTolerateACorruptLabelsRowInsteadOfFailingTheQuery`
  - pre-fix: `Tests run: 1, Errors: 1` — `IllegalStateException: Unable to read alert labels`
  - post-fix: passes
- `mvn test -Dtest=MybatisPlusAlertRepositoryTest,MybatisPlusAlertStateRepositoryTest,NativeAlertProcessorTest,AlertServiceTest` → `Tests run: 129, Failures: 0, Errors: 0` (post-fix; 17+83+23+6 across the four classes), BUILD SUCCESS

## Risk

Low. Behavior changes only for rows whose `labels_json` cannot be parsed: they previously broke the whole read, now they render with empty labels. No schema or API shape change. Closes #4885


unbridled-41:
**Verification pass (fresh checkout of `fix/alert-labels-corrupt-row` @ 363231de)**

- `NativeAlertProcessor.processSamples` (server/.../ops/alert/NativeAlertProcessor.java:88-94): an evaluation `RuntimeException` is only counted into `failedEvaluations` and logged — the pipeline keeps running with no notifications, confirming the silent-stop claim for the suppression path.
- `AlertNotificationSuppressionService.findSuppressingClusterAlert` (AlertNotificationSuppressionService.java:50) pages `findAlertsPage` inside every BUSINESS-domain evaluation, so a corrupt row poisons that shared read path.
- The skip/degrade convention citation is `AiConversationService.decode` (server/.../ops/ai/conversation/AiConversationService.java:507-510): "An unreadable payload is skipped rather than fatal".

- `mvn test -Dtest='MybatisPlusAlertRepositoryTest,MybatisPlusAlertStateRepositoryTest,NativeAlertProcessorTest,AlertServiceTest'` → **Tests run: 129, Failures: 0, Errors: 0** (17+83+23+6), BUILD SUCCESS. The description originally said 112 (a sum from the development run); the corrected figure above is the four-class total including the new regression test.
- The pre-fix failure (`pageAlertsShouldTolerateACorruptLabelsRowInsteadOfFailingTheQuery` → `IllegalStateException: Unable to read alert labels`) was demonstrated during development by stashing the repository changes with the test in place.

Baseline note (pre-existing, unrelated to this PR): `NativeAlertEvaluationTransactionTest` requires a local MySQL and fails with a context-load/connection error on a clean `origin/rocketmq-studio` checkout in this environment (verified on the base commit 4c697f07). It is an environment-dependent test, not a regression of this change.

lizhimins (changes requested):
The direction is right and matches the in-tree precedent (AiEventCodec.read degrades an unreadable payload instead of failing the whole read), but three things block it. (1) Only half the change is tested: MybatisPlusAlertStateRepository.java:209 is modified with no test at all, although MybatisPlusAlertStateRepositoryTest already exists - reverting that line keeps the whole suite green. Please add the mirror case there. (2) Both catch blocks now return Map.of() with no log line, while the precedent you cite logs a warning with the row identity (AiEventCodec.java:82-84). A silent degrade makes a corrupt row indistinguishable from a row that legitimately has no labels and nobody will ever find it; please log.warn with the row id. (3) The trigger is not demonstrated: labels_json is only ever written by MybatisPlusAlertRepository.writeLabels (:391-399, a TreeMap serialisation) and MybatisPlusAlertSilenceRepository:130, so no application path can produce an unparseable value. Please state where the corrupt row actually comes from (manual DB edit, an aborted migration, an external writer) - without that the 500 is not reachable in production. Also: MybatisPlusAlertSilenceRepository:139-145/:162 and MybatisPlusMetricSnapshotRepository:95 keep throwing; that is defensible because their labels feed silence matching and metric evaluation, where an empty map would change decisions rather than just a display - please say so explicitly in the PR so the asymmetry is documented. Minor: rename the new test to ...FailingTheQueryTest to match the project's Test-suffix convention.

PR #4873, opened 2026-09-22
fix(alert): correlate a reminder-only cluster incident into related alerts

unbridled-41:
## Problem / Evidence

`AlertService.findRelatedAlerts` queried the ±30-minute correlation window with `transition="FIRING"` and page size 100 (single page). Two observable gaps:

1. A cluster incident that fired more than 30 minutes ago and is still active only through its in-window REMINDER events is invisible to the related-alerts panel — even though `AlertNotificationSuppressionService.findSuppressingClusterAlert` deliberately treats that same incident as active, with an explicit comment: "REMINDER is emitted only while the state stays FIRING, so an incident whose latest in-window event is a REMINDER is still active."
2. When more than 100 candidate events exist in the window, everything beyond the first page was silently dropped (the suppression counterpart pages until `total` is reached).

Reproduced by a regression test (`relatedAlertsShouldIncludeAnIncidentWhoseInWindowEventIsAReminderTest`): the window query returns a cluster row with transition REMINDER at T−10min (its FIRING event sits outside the window) plus a RESOLVED row at T−5min; before the fix the FIRING-only query matched nothing, so the reminder-only incident was missing from the panel.

## Root cause / Fix

The window query was written for a narrower case than the suppression counterpart, which shares the same correlation-window semantics. The query no longer constrains `transition`; candidates are filtered in memory against `FIRING || REMINDER` (same rule and comment as `AlertNotificationSuppressionService`), and the query now pages through the window like the suppression service.

## Priority / scoring

PRIORITY 55 = impact 20 (user-visible correlation gap: a live incident is missing from the panel while suppression acts on it) + blast radius 15 (system-alerts related panel for every cross-domain alert) + reproducibility 15 (deterministic once a reminder-only incident exists) + maintenance value 5 (aligns the two implementations of one semantic). FIX_CONFIDENCE 85.

## Tests

- `mvn test -Dtest='AlertServiceTest#relatedAlertsShouldIncludeAnIncidentWhoseInWindowEventIsAReminderTest+relatedAlertsShouldReturnOtherDomainFiringEventsInTheSameInstanceAndScopeTest'`
  - pre-fix (fix stashed): `Tests run: 2, Failures: 2` (new test failed with the reminder-only incident missing; the existing test failed on the now-removed transition query argument)
  - post-fix: both pass
- `mvn test -Dtest='AlertServiceTest,AlertNotificationSuppressionServiceTest,SystemAlertControllerTest'` → `Tests run: 102, Failures: 0, Errors: 0` (post-fix), BUILD SUCCESS

## Risk

Low. Adds REMINDER-only incidents (which are genuinely active) to the panel and removes the 100-row ceiling; RESOLVED/other transitions are still excluded. No API shape change. Closes #4886


unbridled-41:
**Verification pass (fresh checkout of `fix/related-alerts-reminder-window` @ f47e512d)**

- `AlertService.findRelatedAlerts` → `SystemAlertController.java:70` (`GET /api/system-alerts/{id}/related`) is the only caller; the fix changes that read path only.
- `AlertNotificationSuppressionService.java:67-69` carries the FIRING/REMINDER in-memory rule and comment the fix mirrors; its query also passes `transition=null`, so both implementations of the correlation-window semantic now agree.
- The paging loop mirrors the suppression service's `fetched >= result.getTotal()` termination (re-verified here as `page * RELATED_CANDIDATE_PAGE_SIZE >= total`), removing the single-page 100-row ceiling.

- `mvn test -Dtest='AlertServiceTest,AlertNotificationSuppressionServiceTest,SystemAlertControllerTest'` → **Tests run: 102, Failures: 0, Errors: 0** (84+7+11), BUILD SUCCESS.
- Pre-fix failures (development run, repository change stashed with the tests in place): the new `relatedAlertsShouldIncludeAnIncidentWhoseInWindowEventIsAReminderTest` failed because the reminder-only incident was missing from the panel, and `relatedAlertsShouldReturnOtherDomainFiringEventsInTheSameInstanceAndScopeTest` failed on the now-removed `transition="FIRING"` query assertion — 2/2 failed as documented.

One wording correction was made to the description: the original reproduction sentence described a "FIRING at T−40min" event; the regression test actually models the incident by its in-window REMINDER row at T−10min (with a RESOLVED row at T−5min that must stay excluded), and the FIRING event itself is outside the window. The body now states this precisely.

lizhimins (changes requested):
The REMINDER half is well evidenced (AlertService.java:558-564 vs AlertNotificationSuppressionService.java:70-73) and both its tests have teeth - the existing case now pins query.transition() == null and the new one pins that a RESOLVED row in the same window is still excluded. Three things block the merge. (1) The paging half is untested: your new stub returns PageResult.of(..., 2, 1, 100), so total=2 with pageSize=100 breaks the loop on the first iteration and the second page is never fetched. Reverting the while loop to a single-page read keeps the whole suite green. Please add a case where the target event sits on page 2 (page 1 returns 100 rows with total=150) and verify findAlertsPage is called twice. (2) The loop has no page/candidate ceiling and this is a synchronous console endpoint (related-alerts panel), where trunk issues exactly one query; the suppression counterpart has the same shape but runs in the alerting pipeline, not on a user request. Please add a MAX_RELATED_CANDIDATE_PAGES (or a hard candidate cap) with a warn when it is hit, or explain in a comment why an unbounded scan is acceptable here. (3) Deduplication: `related` is keyed by alert id (AlertService.java:566-569), so one incident that has both a FIRING row and an in-window REMINDER row now appears twice in the panel, which the FIRING-only query could not produce. Either merge by fingerprint and keep the latest event (what AlertNotificationSuppressionService.java:58-61 does) or state explicitly that the panel is meant to list every window event.

PR #4875, opened 2026-09-22
fix(web): drop a stale alert runtime response after the domain switches

Frun1na:

- Fixes #4879

## Brief Description

In the alerts rules effect, `listAlertRuleRuntime(domain)` was the only request without the `cancelled` guard. After the domain (BUSINESS/CLUSTER) switches, a slow runtime response for the previous domain still called `setRuntime`, so the runtime column rendered another domain's runtime states.

The fix checks `cancelled` before setting the state, matching the guard every other request in the same effect already has.

New test in `web/src/pages/ops/__tests__/AlertsPage.test.tsx`: render with `CLUSTER`, hold the runtime request open, rerender with `BUSINESS`, then resolve the stale response with a FIRING state. On the base revision the stale state is committed and the row shows `触发 1`; with the fix the cell stays `未采集`:

```
$ npx vitest run src/pages/ops/__tests__/AlertsPage.test.tsx
 Test Files  1 passed (1)
      Tests  29 passed (29)
```

`npx eslint` on the changed files reports 0 errors (5 pre-existing warnings, untouched lines).


RockteMQ-AI (approved):
### Summary
Fixes a race condition where a stale alert runtime response could overwrite the current state after the domain switches.

### Review
- **Correctness** ✓ — The old code called `setRuntime` unconditionally in the `.then()` handler, so a slow response from the previous domain could overwrite the new domain's state. The fix adds the `cancelled` guard, which is the standard React cleanup pattern.
- **Tests** ✓ — The new test explicitly verifies the stale response is ignored: it switches domains before the first response resolves, then confirms the stale data does not appear.
- **Scope** ✓ — Minimal change to the effect cleanup logic.

LGTM.

PR #4876, opened 2026-09-22
fix(rmqctl): reconnect on session termination even when the id was cleared

Frun1na:

- Fixes #4880

## Brief Description

`sendWithReconnect` decided whether a terminated MCP session could be re-initialized from `transport.GetSessionId() != ""`. The mcp-go Streamable HTTP transport clears its stored session id as soon as any request fails with a 404 (`sessionID.CompareAndSwap(sessionID, "")`), so a sender racing a concurrent 404 observed an empty id in its snapshot, skipped the reinitialize, and surfaced `ErrSessionTerminated` to the caller even though a reconnect was possible.

The fix bases the decision on the recorded initialize request (`state.initialize != nil`) instead — it is exactly what `reinitialize` needs, it survives 404s, and the generation check inside `reinitialize` keeps concurrent senders serialized as before.

Two new tests in `rmqctl/internal/studio/mcp_message_test.go`:

- `TestSendWithReconnectRetriesAfterConcurrentSessionClear` deterministically reproduces the interleaving (snapshot taken after a concurrent 404 cleared the id). It fails on the base revision with `session terminated (404). need to re-initialize` and passes with the fix.
- `TestSendWithReconnectConcurrent404sAllRecover` runs 8 concurrent senders against a stub that 404s everything until a fresh initialize; all must recover with exactly the CAS clearing semantics of mcp-go.

```
$ go test ./internal/studio/ -run 'TestSendWithReconnect' -count=5 -race
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/studio	1.047s

$ make -C rmqctl ci
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/catalog	1.023s
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/catalog/generate	1.057s
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/config	1.047s
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/output	1.021s
ok  	github.com/apache/rocketmq-dashboard/rmqctl/internal/studio	1.067s
(catalog-verify, go vet, gofmt, cross-platform builds: all green)


RockteMQ-AI (approved):
### Summary
Fixes the MCP client session reconnect logic to use the `initialize` state instead of the transport session ID for determining if a reconnect is possible.

### Review
- **Correctness** ✓ — The old code checked `GetSessionId() != ""` to decide if a terminated session could be re-initialized. However, mcp-go clears the session ID on 404, so a sender racing with that failure would see "no session" and skip the reconnect. The fix checks `initialize != nil` instead, which is the correct signal: if the initialize request has been recorded, the session can be re-initialized regardless of the transport session ID state.
- **Documentation** ✓ — The updated comment clearly explains why the session ID is unreliable and why `initialize` is the correct check.
- **Tests** ✓ — The `reconnectStubTransport` simulates the exact race condition: it clears the session ID on 404 and verifies that the reconnect still happens. The test covers the scenario where the old code would fail.
- **Code quality** ✓ — Minimal change to the core logic with clear intent.

LGTM.

PR #4877, opened 2026-09-22
fix(instance): bound group enrichment by a batch deadline instead of per-group timeouts

Frun1na:

- Fixes #4881

## Brief Description

`enrichLiveStats` gave every enrichment future its own three-second budget. When a broker hangs, a listing of N consumer groups waited up to N×3s (measured 9.1s for three groups) before returning rows.

The fix gives the whole batch one absolute deadline — `startedNanos` plus the existing timeout, each future awaited for the remaining time only, and the remaining futures cancelled once the deadline passes. This matches the approach `InstanceResourceCountRunner` already uses; late results stay unused and the next listing refreshes the stats.

New test in `RocketMQMetadataProviderTest`: three groups whose connection lookups hang on a sleeping admin stub; the listing must return within 6s. On the base revision it fails after 9117ms (3 × 3s per-future budgets); with the fix it returns in 3.16s:

```
$ mvn -B -ntp test -Dtest='RocketMQMetadataProviderTest'
[INFO] Tests run: 45, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```


RockteMQ-AI (approved):
### Summary
Changes the consumer group enrichment timeout from per-group to batch-level deadline, preventing a hanging broker from multiplying the wait time.

### Review
- **Correctness** ✓ — The old code applied `ONLINE_ENRICHMENT_TIMEOUT_SECONDS` to each future individually, so N groups could wait up to N× the timeout. The fix uses a single absolute deadline for the entire batch, which is the correct approach (and matches the `InstanceResourceCountRunner` precedent mentioned in the comment).
- **Edge cases** ✓ — When `remainingNanos <= 0`, the code throws `TimeoutException` immediately instead of calling `future.get()` with a negative timeout. The `cancel(true)` is called on the specific future that timed out, not all remaining futures.
- **Tests** ✓ — The new test verifies the batch completes within the deadline even when individual groups would take longer.
- **Code quality** ✓ — Clear comment explaining the rationale.

LGTM.

PR #4882, opened 2026-09-22
fix(acl): resolve rule principals from the full user directory

unbridled-41:
## Problem / Evidence

The ACL rule tab derived both the "管理员 / Admin" badge (`isAdmin`) and the principal dropdown of the Add/Edit Rule modal from `users` — the server-paginated, keyword-filterable state of the **Users tab** (default 20 rows, max 100). With more than one page of ACL users:

1. A rule whose principal is an admin loses its badge whenever that user's row is not on the currently loaded users page (or when a user-tab keyword filter narrows it).
2. The principal `Select` (showSearch over `users`-derived options, no free input) offers no way to **create or edit a rule for any user outside the loaded page** — a functional blocker on instances with many users.

Reproduced by two regression tests: the Users tab mocked to a page of 30 users that contains neither the rule's admin principal nor `remote-admin`; pre-fix the badge was missing and the dropdown could not offer the directory-only user.

## Root cause / Fix

The rule tab reused the Users tab's paged state as if it were the full user directory. The page now loads the unbounded directory via the existing `listAclUsers` service (`GET /acl/users`, already exposed by the backend and service layer) into a separate `userDirectory` state; `isAdmin` and the principal dropdown read from it. The Users tab keeps its paginated state, and the directory load failure degrades to an empty directory without blocking the page.

## Priority / scoring

PRIORITY 70 = impact 25 (cannot create a rule for a user beyond the loaded page; wrong admin badges) + blast radius 12 (rule tab of every multi-page ACL instance) + reproducibility 18 (deterministic with >20 users) + maintenance value 15. FIX_CONFIDENCE 85.

## Tests

- `npx vitest run src/pages/instance/__tests__/AclPage.test.tsx`
  - pre-fix (component fix stashed): new tests fail — admin badge absent, directory-only principal not offered
  - post-fix: `Tests 25 passed (25)`
- `npx tsc --noEmit` clean; `npx eslint src/pages/instance/acl.tsx src/pages/instance/__tests__/AclPage.test.tsx` clean

## Risk

Low. One extra read-only request per instance switch/refresh; no API change. `GET /acl/users` is an existing backend endpoint already wrapped by the service layer (`listAclUsers`), though this is its first page-level consumer in the web UI — mock mode serves it from the in-memory user list, so dev behavior is unchanged. The dropdown list equals the full directory; users on huge directories see the same showSearch-filtered list as before, now complete. Closes #4887


RockteMQ-AI (approved):
### Summary
Fixes ACL rule principal resolution to use the full user directory instead of the paginated Users tab list.

### Review
- **Correctness** ✓ — The old code derived the admin badge and principal dropdown from the paginated `users` state, which only contains the current page of users. This caused two bugs: (1) admin badges in the Rules tab were wrong when the principal was not on the loaded page, and (2) the rule principal dropdown was incomplete. The fix introduces a separate `userDirectory` populated by `listAclUsers` (full directory) and uses it for both purposes.
- **API semantics** ✓ — `listAclUsers` is the correct endpoint for "all users" since `pageAclUsers` is server-paginated.
- **Tests** ✓ — Comprehensive tests cover the admin badge resolution and dropdown population scenarios with realistic pagination gaps.
- **Code quality** ✓ — Clean separation between the paginated view state and the full directory state.

LGTM.

unbridled-41:
**Verification pass (fresh checkout of `fix/acl-principal-full-directory` @ e126f598)**

- `acl.tsx` now loads `userDirectory` through `listAclUsers({ instanceId })` (aclService.ts:77-90 → `GET /acl/users`), and only `isAdmin` (acl.tsx:259-260 → rules-table badge, acl.tsx:591) and the rule-modal principal Select (acl.tsx:1468-1476) read from it. The Users tab keeps its server-paginated `pageAclUsers` state — the two sources no longer share state.
- Backend contract: `AclController` exposes `GET /acl/users` (`AclService.listUsers`, unbounded) alongside `GET /acl/users/page` (pageSize ≤ 100); mock mode (`isMockMode()`) serves the directory from the in-memory user list, so dev data flows through the same service call.
- Directory-load failure degrades to an empty directory (`.catch(() => setUserDirectory([]))`) without a toast — deliberate, since the Users tab's own failure already surfaces the load error and the rule tab must stay usable.

- `npx vitest run src/pages/instance/__tests__/AclPage.test.tsx` → **Test Files 1 passed (1), Tests 25 passed (25)**.
- Pre-fix failures (development run, component change stashed with the tests in place): `resolves the rule-tab admin badge from the full user directory` (badge absent because the loaded users page doesn't contain the admin principal) and `offers every directory user as a rule principal` (directory-only `remote-admin` not offered in the dropdown).
- `npx tsc --noEmit` clean; `npx eslint src/pages/instance/acl.tsx src/pages/instance/__tests__/AclPage.test.tsx` clean.

One correction was made to the description's Risk paragraph: it previously said the endpoint "is used elsewhere" in the web UI — in fact the page-level code had never consumed `listAclUsers` before (only the mock-mode internals of `pageAclUsers` call it). The body now states this is the endpoint's first page-level consumer.

Issue #4805, opened 2026-09-22
[Bug] Topic detail renders the raw consumer-model enum and never colors broadcasting consumers


Frun1na:
### Before Creating the Bug Report

- [x] I have searched the [open issues](https://github.com/apache/rocketmq-dashboard/issues) and found no similar issue.

### What Happened

On the topic detail modal, the 消费模式 (message model) column of the consumer-group sub-table shows
the raw API enum values (`CLUSTERING` / `BROADCASTING`, plus the vendor-native variants the Aliyun and
Tencent providers pass through), and every row falls back to the blue tag — a broadcasting consumer
never gets the orange highlight the column was written to show.

The column matches `messageModel` against the display string the API returned before #3349:

```tsx
render: (m: string) => <Tag color={m === '广播消费' ? 'orange' : 'blue'}>{m}</Tag>,
```

The Apache provider sends `conn.getMessageModel().name()` — `CLUSTERING` / `BROADCASTING` — so the
comparison never matches any more. The i18n keys `topic.broadcast` / `topic.clustering` exist but are
referenced nowhere, and `web/src/mock/topics.ts` still feeds the old display strings, which kept the
dead comparison looking alive in the dev mock path.

### Expected Behavior

The column should render the localized labels — `集群消费` (blue) and `广播消费` (orange) — for the
enum values the API actually returns, with a raw-value fallback for anything unknown.

### Steps to Reproduce

1. Open an instance's Topic page and open a topic's 详情 modal with consumer groups attached.
2. Look at the 消费模式 column: values render as the raw `CLUSTERING` / `BROADCASTING` enums, all blue,
   including rows whose group is a broadcasting consumer.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

This is a valid UI rendering bug. The root cause is well-identified: the `messageModel` column compares against the old pre-#3349 display string (`广播消费`) instead of the actual enum values (`CLUSTERING`/`BROADCASTING`) that the Apache provider now returns.

**Root Cause:** Stale string comparison after the API response format changed. The i18n keys exist but are unused; the mock data masks the problem in dev.

**Impact:** Medium — consumer mode column always shows blue tag, losing the visual distinction between clustering and broadcasting consumers.

**Severity:** Medium

An automated fix proposal can be generated. Reply `/approve` to proceed with PR generation.

---
*Automated evaluation by github-manager*

lizhimins:
[...]

PR #4883, opened 2026-09-22
fix(topic): match the detail consume-mode tag against the API values

unbridled-41:
## Problem / Evidence

In the topic detail modal's consumer table, the 消费模式 (consume mode) tag color compared the cell value against the mock fixture's Chinese literal:

```tsx
render: (m: string) => <Tag color={m === '广播消费' ? 'orange' : 'blue'}>{m}</Tag>
```

The real API never sends that string:

- Apache provider: `messageModel = conn.getMessageModel().name()` → `CLUSTERING` / `BROADCASTING` (and the offline fallback hardcodes `"CLUSTERING"`) — `RocketMQMetadataProvider.java`
- Aliyun converter: passes the raw cloud `messageModel` (`Broadcasting`-style) — `AliyunConverters.java`
- The Chinese literals exist only in `web/src/mock/topics.ts`

So in every non-mock deployment the orange branch is dead code: broadcasting consumer groups always render with the default blue tag, and the color contract silently differs between mock mode and production.

Reproduced by a regression test: a `BROADCASTING` consumer row in the detail modal rendered a blue tag before the fix; it now renders orange.

## Root cause / Fix

Display logic written against mock fixture values instead of the API contract. The tag now colors orange for any broadcast-spelled value, case-insensitively (`/broadcast/i`), which covers `BROADCASTING`, `Broadcasting`, and the mock's own label, and keeps the raw cell text unchanged.

## Priority / scoring

PRIORITY 52 = impact 8 (wrong visual classification of broadcasting groups) + blast radius 14 (topic detail of every real instance) + reproducibility 20 (deterministic, every broadcasting row) + maintenance value 10. FIX_CONFIDENCE 95.

## Tests

- `npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx -t "BROADCASTING"`
  - pre-fix (component fix stashed): `Tests 1 failed`
  - post-fix: passes
- Full file: `Tests 33 passed (33)`; `npx tsc --noEmit` clean; `npx eslint src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx` clean

## Risk

Minimal. One-line render change in a display tag; no API or data change. Values that are not broadcast (including unknown future spellings of clustering) keep the blue tag. Related issue: #4805


RockteMQ-AI (approved):
### Summary
Fixes the consume-mode tag color matching to use API enum values instead of Chinese labels.

### Review
- **Correctness** ✓ — The old code matched against `"广播消费"` (a mock-era Chinese label) which would never match the actual API values (`BROADCASTING`, `Broadcasting`). The regex `/broadcast/i` correctly handles both Apache enum names and cloud provider values.
- **Tests** ✓ — New test covers both BROADCASTING (orange) and CLUSTERING (blue) tag colors with realistic API data.
- **Scope** ✓ — Single-line production change with clear intent.

LGTM.

unbridled-41:
**Verification pass (fresh checkout of `fix/topic-detail-broadcast-tag` @ caae5c84)**

- Apache provider path: `RocketMQMetadataProvider.java` (~line 569) sets `messageModel = conn.getMessageModel().name()` (`CLUSTERING`/`BROADCASTING`), with the offline fallback hardcoding `"CLUSTERING"` (line ~590).
- Cloud paths: `AliyunConverters.java:159` passes the raw cloud `messageModel` (`Broadcasting`-style); `TencentInstanceProvider.java` (~line 974) passes `subscription.getMessageModel()` (falling back to consumeType) — both spellings contain "broadcast"/"Broadcasting" and are matched by the case-insensitive test.
- The compared literal `广播消费` exists only in `web/src/mock/topics.ts:393-423`; the topic detail consumer table renders that same mock data through `getTopicConsumerPage`'s mock branch (topicService.ts), which is why the gap was invisible in dev mode.

- `npx vitest run src/pages/instance/__tests__/TopicPage.test.tsx` → **Test Files 1 passed (1), Tests 33 passed (33)**.
- Pre-fix failure (development run, component change stashed with the test in place): `marks a BROADCASTING consumer row with the broadcast tag color in the topic detail` — the `BROADCASTING` row rendered the default blue tag; the test also pins the clustering row to blue.
- `npx tsc --noEmit` clean; `npx eslint src/pages/instance/topic.tsx src/pages/instance/__tests__/TopicPage.test.tsx` clean.

PR #4884, opened 2026-09-22
fix(cloud): report subscription consistency in the studio vocabulary

Wang1rrr:

The subscription table's consistency status is a small fixed vocabulary: the Apache provider reports `consistent` when a client of the group is connected and leaves it unknown (`null`) otherwise, and the console's group diagnostics only recognise `consistent` / `inconsistent` (`web/src/utils/consumerGroupDiagnostics.ts` keys both the cell colouring and the `SUBSCRIPTION_INCONSISTENT` diagnostic off those values).

The two cloud providers did not speak that vocabulary:

- Aliyun `ListConsumerGroupSubscriptions` returns the consistency as a `Boolean`; `AliyunConverters.toSubscriptionEntry` passed it through as `String.valueOf(...)`, so the API returned the strings `true` / `false`.
- Tencent `DescribeTopicListByGroup` returns the consistency as a code (`0` = consistent, `1` = inconsistent); `TencentInstanceProvider.getGroupSubscriptions` likewise returned the raw numbers `0` / `1` as strings.

Every value other than `consistent` / `inconsistent` is classified as unknown by the diagnostics, so for cloud instances the "inconsistent only" view silently showed nothing and an actually inconsistent subscription never raised `SUBSCRIPTION_INCONSISTENT`.

This adds `SubscriptionConsistency`, a small converter next to `SubscriptionFilterModes`, and maps both vendor flags through it:

- `true` → `consistent`, `false` → `inconsistent`, missing stays `null` (unknown).
- `0` → `consistent`, `1` → `inconsistent`, missing or undocumented codes stay `null` rather than being guessed.

## Brief changelog

- `SubscriptionConsistency` maps the vendor flags to the studio vocabulary.
- `AliyunConverters.toSubscriptionEntry` uses `fromBoolean`.
- `TencentInstanceProvider.getGroupSubscriptions` uses `fromCode`.
- Tests cover the true / false / missing and 0 / 1 / missing / undocumented cases for both providers.

## Verifying this change

`mvn -f server/pom.xml test -Dtest='SubscriptionConsistencyTest,AliyunConvertersTest,TencentInstanceProviderTest'`

Tests run: 58, Failures: 0, Errors: 0 (checkstyle runs at validate and also passes). The new assertions fail on trunk, where the same call returns `true` / `0`.


RockteMQ-AI (approved):
### Summary
Introduces a `SubscriptionConsistency` utility to translate vendor-specific consistency flags (Aliyun boolean, Tencent code) into the studio vocabulary ("consistent"/"inconsistent").

### Review
- **Correctness** ✓ — The old code used `String.valueOf()` which leaked raw boolean/code values into the API response. The console expects "consistent"/"inconsistent" for coloring and filtering. The new utility correctly maps each vendor's representation.
- **Null handling** ✓ — Both `fromBoolean` and `fromCode` return `null` for missing values instead of guessing, which preserves the "unknown" state.
- **Test coverage** ✓ — Tests cover all three providers (Apache, Aliyun, Tencent) including null/unknown edge cases.
- **Design** ✓ — `SubscriptionConsistency` is a clean utility class with clear Javadoc explaining the mapping for each vendor. The constants make the vocabulary explicit.
- **Integration** ✓ — The callers in `AliyunConverters` and `TencentInstanceProvider` are updated to use the new utility.

LGTM.

PR #4888, opened 2026-09-22
fix(metrics): keep a language switch out of the query identity

unbridled-41:
## Problem

Metrics Explorer: if the operator switches the display language while a **custom PromQL query is still running**, the custom panel never leaves its spinner — the chart never appears, the Run button stays busy and unclickable, and the refresh action keeps spinning and cannot be clicked. Only a page reload recovers the view.

## Evidence

`web/src/components/MetricsExplorer.tsx` (base `1ef5d860`):

- `runCustomQuery` marks the panel loading, then publishes only if its generation is still current: `const currentRequest = ++customRequestIdRef.current;` … `if (currentRequest === customRequestIdRef.current)` (the loading flag is set at the top and cleared only inside those two guarded branches).
- The profiles effect ends with `return () => { cancelled = true; panelRequestIdRef.current += 1; customRequestIdRef.current += 1; };` while its body only restarts the **panel** flow through `loadAll(...)`. Nothing restarts the custom flow.
- That effect depends on `loadAll`, whose dependency chain is `[instanceId, queryErrorFallback, runQuery]` → `runQuery` deps `[copy.defaultDataSource, instanceId]`. `queryErrorFallback` (`'Prometheus 查询失败'` / `'Prometheus query failed'`) and `copy.defaultDataSource` / `copy.customTitle` are values of the current language, so **switching the language rebuilds the callbacks and re-runs the effect** — the only other trigger is the instance prop.

Reproduction (deterministic Vitest, `web/src/components/__tests__/MetricsExplorer.test.tsx` › `keeps the pending custom query when the display language changes`): hold the custom query's `queryMetrics` promise open, run it, click the language switch, then resolve it. On the unfixed source the panel result never renders (`Unable to find an element with the text: cluster=prod / query=custom`) and the profile list is refetched (`listMetricProfiles` called twice); with the fix both hold.

Second symptom of the same dependency: the language switch re-queried *every* profile panel (one Prometheus request per panel) and refetched the profile list, for a change that has nothing to do with the query.

## Root cause

Display text was part of the query identity. Because `copy`/`queryErrorFallback` sat in the callbacks' dependency arrays, a language switch looked like a query-context change, and the profiles effect's cleanup discarded the in-flight custom query that its body never restarts.

## Fix

The three labels the callbacks need (`defaultDataSource`, `customTitle`, `queryErrorFallback`) are mirrored in a ref (`queryLabelsRef`) that `runQuery`/`loadAll`/`runCustomQuery` read, and the callbacks now depend on the instance only. Rendering keeps reading the live `copy`, so the visible text still follows the language; only the callbacks' identity is language-independent, and the profiles effect re-runs for the query context it was written for.

## Tests

- `npx vitest run src/components/__tests__/MetricsExplorer.test.tsx` → **30 passed** (29 pre-existing + the new regression).
- Pre-fix failure of the new test (source only reverted, test kept): `AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times`, and with the two call-count assertions pointed at the panel symptom instead, `Unable to find an element with the text: cluster=prod / query=custom`.
- `npx vitest run src/components` → **88 passed** (8 files).
- `npx tsc --noEmit` → clean; `npx eslint src/components/MetricsExplorer.tsx src/components/__tests__/MetricsExplorer.test.tsx` → clean.

## Scoring

PRIORITY 62 (impact 24: the monitoring view is left unusable until reload, the same failure class as #3304; reach 12: the language switch is a top-bar control and the window is the query duration; reproducibility 16: deterministic test; maintenance value 10: removes a duplicate Prometheus fan-out). FIX_CONFIDENCE 90.

## Risk

Low. The ref mirrors strings, so its effect is quiet between language switches; messages already on screen keep the language they were produced in until the next query, which is how the rest of the page's stored error text already behaves. The instance-transition and data-source paths are untouched (the test `re-runs the committed custom query when the dashboard instance changes` still passes), so the #3299 guards keep working.


unbridled-41:
### Independent verification (isolated checkout of `5b1617b6`)

Re-read every citation in the description against this branch and re-executed every test claim from a clean checkout (`git worktree add --detach` at the branch head, so no other work in my tree can influence the run). Nothing below is copied from the earlier runs.

- The two publish branches in `runCustomQuery` are the only writers of `customPanel`: `+ 1` on the generation at `MetricsExplorer.tsx:724`, `setCustomPanel({ loading: false, data … })` at `:766` and `setCustomPanel({ loading: false, error … })` at `:785`, both inside the freshness guard. `setCustomPanel({ loading: true })` at `:725` is therefore never cleared when the generation is discarded — the stranded state described in the description.
- The profiles effect ends with the cleanup that bumps **both** generations (`:705-711`); its body only restarts the panel flow through `loadAll(...)`. Confirmed: nothing in the effect body (or in any other effect) restarts the custom flow.
- The dependency chain quoted in the description is the **base** (`1ef5d860`) state, as labelled there: `runQuery` `[copy.defaultDataSource, instanceId]`, `loadAll` `[instanceId, queryErrorFallback, runQuery]`, `runCustomQuery` `[copy.customTitle, instanceId, queryErrorFallback, runQuery]`. For readers checking the branch instead: after the fix `runQuery` is `[instanceId]` and `loadAll`/`runCustomQuery` are `[instanceId, runQuery]`, with the three labels read from `queryLabelsRef` (`:535-547`). The cleanup still bumps `customRequestIdRef` at `:710` on purpose — the instance-transition path restarts that flow, which is what makes the discard safe there and not on a language switch.

**Test claims — re-executed**

```
npx vitest run src/components/__tests__/MetricsExplorer.test.tsx   -> Test Files 1 passed (1) | Tests 30 passed (30)
npx vitest run src/components                                      -> Test Files 8 passed (8) | Tests 88 passed (88)
npx tsc --noEmit -p tsconfig.json                                  -> clean
npx eslint src/components/MetricsExplorer.tsx \
           src/components/__tests__/MetricsExplorer.test.tsx      -> clean
```

With `MetricsExplorer.tsx` restored to its parent revision and the test kept (`git checkout HEAD~1 -- web/src/components/MetricsExplorer.tsx`):

```
× keeps the pending custom query when the display language changes 798ms
AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times
Tests  1 failed | 29 passed (30)
```

The second symptom quoted in the description (the custom result never rendered) was reproduced separately during development by neutralising the two call-count assertions, leaving `Unable to find an element with the text: cluster=prod / query=custom`.

**Coverage**

The regression pins both halves of the fix — no profile re-fetch on a language switch, and the in-flight custom result still published — and the instance-transition tests that #3299 introduced (`re-runs the committed custom query when the dashboard instance changes`, `refreshes profile panels and the custom query independently`) stay green, so the generation guards for the instance path are unchanged.

PR #4889, opened 2026-09-22
fix(ai): identify a role-backed ACL rule in the tool contract

unbridled-41:
## Problem

On a role-backed Studio Instance (Tencent), the ACL read tools fail outright: `rmq.acl.list`, `rmq.acl.get`, `rmq.user.list` and `rmq.user.get` answer `INTERNAL_ERROR` (HTTP 500) as soon as the instance owns one role, and the caller gets no ACL data at all.

## Evidence

`server/src/main/java/org/apache/rocketmq/studio/ops/ai/tool/contract/acl/AclRuleItem.java:39` / `AclUserItem.java:33` (base `1ef5d860`) built the tool item's `id` from the numeric primary key only: `rule.getId() == null ? null : rule.getId().toString()`. Both records are `@JsonInclude(Include.NON_NULL)`, so a `null` id is **omitted from the payload**.

The value is `null` for a role-backed instance by construction — `AclService.listRules`/`listUsers`/`getRule`/`getUser` dispatch to `TencentAclService` (`instance/acl/AclService.java:82,136,176`), whose projections never set an id:

```java
// provider/tencent/TencentAclService.java:364,378
return AclUserVO.builder().username(role.getRoleName())...build();   // no id
return AclRuleVO.builder().principal(role.getRoleName())...build();  // no id
```

The tool catalog requires it — `tool-catalog/tools/acl.yaml:57` and `user.yaml:32` declare item `required: [id, ...]` with `id: type: string` — and `ToolValidationFilter` validates every result through `ToolSchemaValidator.validateOutput`, which throws when the schema is violated. `ToolExecutionService.executeInternal` then maps the non-`BusinessException` to `ToolError.UNEXPECTED_EXECUTION_FAILURE` (`ToolError.java:91`, HTTP 500 / `INTERNAL_ERROR`).

Reproduced by the new tests against the unfixed source (test kept, source reverted):

```
java.lang.IllegalStateException: Tool output validation failed for rmq.acl.list: [/items/0: required property 'id' not found]
java.lang.IllegalStateException: Tool output validation failed for rmq.user.get: [: required property 'id' not found]
```

The binding is not the barrier: both tools are gated on `ACL_MANAGEMENT`, which `TencentInstanceProvider.capabilities()` declares (`provider/tencent/TencentInstanceProvider.java:152`), and an empty role list validates fine — so the failure appears exactly when there *is* something to list.

The console treats the same `null` as an oversight and normalises it client-side (`web/src/pages/instance/acl.tsx:83,101`: `id: rule.id ?? rule.principal`, `id: user.id ?? user.username`), which is what the tool contract was missing.

## Root cause

The tool DTOs assumed the numeric primary key of the Studio ACL tables, while the ACL service also projects rules and users from a cloud role store where the name is the identifier.

## Fix

`AclRuleItem.from` / `AclUserItem.from` fall back to the rule principal / user name when there is no numeric id. That identifier is not cosmetic: `AclService.getRule` matches `id.equals(rule.getPrincipal())`, `getUser` matches `id.equals(user.getUsername())` and `deleteRule` passes the argument straight to the role-name-based `TencentAclService.deleteRule`, so a listed rule can be fetched or deleted with the id the list returned.

Scope note: `rmq.acl.update` still requires a numeric id (`AclUpdateToolHandler.parseId` → `ACL_ID_INVALID` for a role name), so updating a role-backed rule through the tool layer remains unsupported. That is a separate limitation and is not addressed here.

## Tests

- `mvn -o test -Dtest='org.apache.rocketmq.studio.ops.ai.tool.handler.acl.*Test,org.apache.rocketmq.studio.ops.ai.tool.handler.audit.*Test,org.apache.rocketmq.studio.instance.acl.*Test'` → **Tests run: 193, Failures: 0, Errors: 0 — BUILD SUCCESS** (includes the two new regressions: `AclListToolHandlerTest.executeShouldIdentifyARuleThatHasNoNumericId`, `UserGetToolHandlerTest.executeShouldIdentifyAUserThatHasNoNumericId`).
- Both new tests fail on the unfixed source with the validation errors quoted above; the pre-existing `executeShouldSatisfyOutputSchemaWhenOptionalRuleFieldsAreNull` and the ACL service/controller suites stay green.

## Scoring

PRIORITY 66 (impact 26: a documented tool family returns 500 instead of data on a supported provider; reach 10: role-backed instances only; reproducibility 18: deterministic unit tests, no cloud account needed; maintenance value 12: restores the schema contract the tool layer enforces). FIX_CONFIDENCE 88.

## Risk

Low. MySQL-backed instances are unaffected (`getId()` is set, so the fallback never fires and every existing assertion on numeric ids still passes). The fallback can only make an item that previously crashed output validation serializable; if a role name were ever null the schema would still reject it, which is the correct behaviour. No schema, controller or UI change.


unbridled-41:
Closing as a duplicate: the same defect is already covered by open PR #4690 (`fix(ai-tool): keep the required id in cloud ACL projections`), which applies the identical principal/user-name fallback in `AclRuleItem.from`/`AclUserItem.from`, and by open PR #4574 (`fix(ai): keep ACL and user tools usable for Tencent roles without an id`), which fixes the same 500 by making `id` optional in the schemas. I did not find either during my initial dedup pass (I searched issue/PR text for 'Tencent ACL', 'acl id null', 'rmq.acl.list'), so this branch was opened in error; the analysis below should be read as a confirmation of #4690, not as a competing change. Nothing was merged.

unbridled-41:
### Evidence record for the closed branch (for whoever reads this later)

Closing as a duplicate, but the defect was verified real on this branch before it was closed, and the reproduction is kept here so it does not have to be rediscovered. Note that #4690 implements the **identical** fallback; #4574 fixes the same 500 by making `id` optional in the schemas instead.

**Why it is real**

- `TencentAclService.toRule` / `.toUser` build the projections without an id (`provider/tencent/TencentAclService.java`), and `AclService.listRules` / `listUsers` / `getRule` / `getUser` dispatch to them for a role-backed instance.
- `AclRuleItem` / `AclUserItem` are `@JsonInclude(Include.NON_NULL)` and filled `id` from the numeric primary key only, so the key was **omitted** from the payload.
- The tool catalog requires it: `tool-catalog/tools/acl.yaml` (`rmq.acl.list` item `required: [id, principal, resource]`) and `user.yaml` (item `required: [id, username, admin, clusters]`), and `ToolSchemaValidator.validateOutput` throws on a violation — mapped by `ToolExecutionService` to `UNEXPECTED_EXECUTION_FAILURE` (HTTP 500 / `INTERNAL_ERROR`).
- Reachability is not blocked by gating: both tools need `ACL_MANAGEMENT`, which `TencentInstanceProvider.capabilities()` declares, and an empty role list validates fine — the failure appears exactly when there is something to list.

**Reproduced on this branch (test kept, main sources reverted)**

```
java.lang.IllegalStateException: Tool output validation failed for rmq.acl.list:
  [/items/0: required property 'id' not found]
java.lang.IllegalStateException: Tool output validation failed for rmq.user.get:
  [: required property 'id' not found]
Tests run: 2, Failures: 2, Errors: 0
```

Also confirmed independently of the tests: `ObjectMapper.convertValue(Map.of("instanceId","x"), AclListInput.class).page()` returns `PageRequest[page=1, pageSize=20]` (the compact constructor defaults it), which is why the sibling "missing page" hypothesis for these handlers does not hold.

The two new regression tests on this branch (`AclListToolHandlerTest.executeShouldIdentifyARuleThatHasNoNumericId`, `UserGetToolHandlerTest.executeShouldIdentifyAUserThatHasNoNumericId`) pass with the fallback; they are redundant with #4690's `CloudAclIdentifierProjectionTest`, which is why this branch is not kept open.

PR #4890, opened 2026-09-22
fix(producer): keep the picked scope when the display language changes

unbridled-41:
## Problem

On the Producer page, switching the display language silently throws away the operator's query scope: the selected TOPIC and PRODUCER GROUP are cleared, so the page cannot run a query until both are picked again.

## Evidence

`web/src/pages/studio/Producer.tsx` (base `1ef5d860`):

```ts
const fetchTopicFailedMessage = t('producer.fetchTopicFailed');   // :90 — a value of the current language
...
useEffect(() => {
    ...
    form.setFieldsValue({ selectedTopic: undefined, producerGroup: undefined });   // :153
    void fetchTopicList(selectedInstanceId)...
}, [fetchTopicFailedMessage, form, message, selectedInstanceId]);                   // :171
```

`t` is rebuilt by `LangProvider` on every language change, so `fetchTopicFailedMessage` changes identity and the effect re-runs — even though nothing about the instance changed. The reset was written for the instance transition (the topic list belongs to the instance, and `handleInstanceChange` clears the same fields), but it runs on every effect run.

Reproduction (deterministic Vitest, `web/src/pages/studio/__tests__/Producer.test.tsx` › `keeps the picked topic and producer group when the display language changes`): pick `order-events`, type `order-producer`, click the language switch. On the unfixed source the topic selection is gone — `Unable to find an element with the text: order-events, which matches selector '.ant-select-selection-item'` — and the group input is empty, so the follow-up query cannot carry the scope the operator chose.

This is the same defect class the project has already fixed twice: #4590 (merged, `BrokerCluster` instance selection) and #4818 (open, `Ops` NameServer selection).

## Root cause

A localized string participated in the effect's dependencies, so a language switch re-ran an effect whose destructive step belongs to the instance transition only.

## Fix

The reset is now keyed to the instance: `lastResetInstanceRef` records the instance the form was cleared for, and the reset only runs when `selectedInstanceId` actually changes. The instance transition (initial auto-selection and a manual switch) behaves exactly as before, while the extra effect run caused by a language change leaves the selection alone. The topic list is still refetched, so the localized failure message keeps following the language.

## Tests

- `npx vitest run src/pages/studio/__tests__/Producer.test.tsx` → **16 passed** (15 pre-existing + the new regression).
- The new test fails on the unfixed source with the selector error quoted above.
- `npx vitest run src/pages/studio` → **85 passed** (8 files); the instance-switch coverage (`clears stale producer query state before loading a new instance`) is unchanged.
- `npx tsc --noEmit` → clean; `npx eslint src/pages/studio/Producer.tsx src/pages/studio/__tests__/Producer.test.tsx` → clean.

## Scoring

PRIORITY 56 (impact 16: the TOPIC field is required by the form so no query runs until it is picked again, and the optional group is silently dropped, so a query submitted in between becomes a topic-wide scan of every producer group instead of the configured one; reach 12: the language switch is a top-bar control on every page; reproducibility 16: deterministic test; maintenance value 12: closes the third instance of a class the project already treats as a bug). FIX_CONFIDENCE 88.

## Risk

Low. Only the *timing* of the reset changes, and only for a re-run that does not come from an instance change; the instance path is covered by the pre-existing tests. No behaviour change for the query, the export, or the localized error message.


unbridled-41:
### Independent verification (isolated checkout of `3bc4a73a`)

Every citation re-read against this branch, every test claim re-executed from a clean checkout (`git worktree add --detach` at the branch head). Nothing copied from the earlier runs.

- `web/src/pages/studio/Producer.tsx:90` on the base reads `const fetchTopicFailedMessage = t('producer.fetchTopicFailed');`, and `LangProvider` rebuilds `t` on every language change, so the derived string changes identity and the effect (`:171`, deps `[fetchTopicFailedMessage, form, message, selectedInstanceId]`) re-runs — confirmed by reading `web/src/i18n/LangContext.tsx:39-49` (a plain local function, not memoised).
- The reset it ran unconditionally sits at `:153` on the base (`form.setFieldsValue({ selectedTopic: undefined, producerGroup: undefined })`).
- On this branch the reset is guarded by `lastResetInstanceRef` (`:98`, guard at `:159-162`), and the topic-list refetch still follows on every effect run, so the localized failure message keeps following the language.
- The impact statement in the description is grounded in the form: the TOPIC item carries `rules={[{ required: true, message: t('producer.selectTopic') }]}` (form region of `Producer.tsx`), so submitting without it is rejected client-side; `PRODUCER GROUP` is optional and is passed through to `queryProducerConnection` unchanged, which the pre-existing `queries all active producer groups when producer group is empty` test pins as a topic-wide scan.

**Test claims — re-executed**

```
npx vitest run src/pages/studio/__tests__/Producer.test.tsx -> Test Files 1 passed (1) | Tests 16 passed (16)
npx vitest run src/pages/studio                             -> Test Files 8 passed (8) | Tests 85 passed (85)
npx tsc --noEmit -p tsconfig.json                           -> clean
npx eslint src/pages/studio/Producer.tsx \
           src/pages/studio/__tests__/Producer.test.tsx     -> clean
```

With `Producer.tsx` restored to its parent revision and the test kept:

```
× keeps the picked topic and producer group when the display language changes 476ms
TestingLibraryElementError: Unable to find an element with the text: order-events,
which matches selector '.ant-select-selection-item'.
Tests  1 failed | 15 passed (16)
```

**Coverage**

The regression covers the new direction (a language switch leaves the picked scope alone). The preserved direction — an instance switch still clears it — is pinned by the existing `clears stale producer query state before loading a new instance`, which asserts the topic selection disappears after switching to `instance-2`; it stays green on this branch, so the conditional reset cannot silently stop clearing. The two assertions together hold both sides of the guard.

PR #4892, opened 2026-09-22
fix(metrics): put the selection back when a protected restore is cancelled

unbridled-41:
## Problem

Restoring a query-history entry whose data source needs credentials applies **half of the entry before anything is confirmed**: the drawer closes, the range, the profile (and the persisted profile) and — for a custom entry — the query text are all replaced, while only the data-source switch waits for `handleAuthSubmit`. Cancelling the credentials prompt therefore leaves the explorer in a state it was explicitly designed not to reach: the profile selector shows the restored profile while the panel grid still holds the previous profile's data, so the restored profile's cards are empty with no query in flight to fill them, and the profile written to `localStorage` is the one the operator declined.

## Evidence

`web/src/components/MetricsExplorer.tsx` (base `1ef5d860`):

- `handleRestoreHistory` writes the selection before the prompt: `setRangeId(nextRange.value)`; `setCustomPromql(entry.promql)` for a custom entry; `localStorage.setItem(PROFILE_STORAGE_KEY, nextProfile.id)` + `setProfileId(nextProfile.id)` for a profile entry — and only then calls `restoreProtectedDataSource(...)`.
- `restoreProtectedDataSource` stores the replay and opens the modal; it issues no query.
- `handleAuthCancel` dropped only the pending source (`pendingAuthReplayRef.current = null; setPendingDataSource(null)`), so the replaced selection stayed.
- The intent is documented in the same function: *"the current source stays active while credentials are being asked for, so cancelling the dialog leaves the explorer exactly where it was instead of stranded on an unauthenticated source"* — that contract was implemented for the source alone (#4643, for #4644) and not for the selection the same restore had already replaced.

Reproduction (deterministic Vitest, `MetricsExplorer.test.tsx` › `leaves the picked profile and range alone when a protected history restore is cancelled`): the explorer starts on `RocketMQ 5.x Native` over the default `1h` window; a history entry restoring `RocketMQ 4.x Exporter` over `6h` with a `Basic Auth` data source is restored and the prompt cancelled. On the unfixed source the assertion fails with `Unable to find an element with the text: RocketMQ 5.x Native` — the selector had switched to the declined entry's profile, the segmented control to `6h`, and `rocketmq-studio.metric-profile` had been written.

## Root cause

The restore mutates live selection state eagerly and defers only the source, so "cancel" had no way back to the previous selection.

## Fix

The deferred restore now carries the selection it replaced (`RestoreCheckpoint`: profile id, range id, custom expression, persisted profile id), captured before the entry is applied. `handleAuthCancel` puts all four back before dropping the pending source — including the custom expression box that the custom-entry branch overwrites — so declining the prompt is a no-op again. A confirmed prompt is untouched: `handleAuthSubmit` still replays the entry against the newly authenticated source, and an entry whose source needs no credentials is still applied immediately without a prompt.

## Tests

- `npx vitest run src/components/__tests__/MetricsExplorer.test.tsx` → **31 passed** (29 pre-existing + 2 new regressions):
  - `leaves the picked profile and range alone when a protected history restore is cancelled` — the profile entry case, quoted above.
  - `keeps the custom expression draft when a protected custom restore is cancelled` — a custom entry replaces the query box before the prompt too, so a draft the operator is still editing must survive a declined prompt. Pre-fix (source reverted, test kept): `expect(element).toHaveValue(sum(rocketmq_topic_number) + 1)` / `Received: sum(rocketmq_topic_number)` — the entry's expression had replaced the draft, and the draft was gone for good.
- `npx vitest run src/components` → **89 passed** (8 files); the neighbouring cancel/restore coverage still passes: `keeps the current data source when cancelling a protected history restore`, `requires credentials again when restoring a protected data source history item`, `restores profile, range, and source from query history`.
- `npx tsc --noEmit` → clean; `npx eslint src/components/MetricsExplorer.tsx src/components/__tests__/MetricsExplorer.test.tsx` → clean.

## Scoring

PRIORITY 58 (impact 20: the explorer is left showing a profile it never queried, with a declined profile persisted; reach 12: every history restore of a protected source; reproducibility 16: deterministic test; maintenance value 10: completes the cancel contract of #4643 and closes the residue #4644 was closed for). FIX_CONFIDENCE 85.

## Risk

Low and local to the cancel path: nothing changes for a submitted prompt, for a restore without credentials, or for switching a data source from the selector (where no replay is pending, so `handleAuthCancel` returns early). The checkpoint is read from the render scope before the entry is applied, so it holds the values the operator actually had.


unbridled-41:
### Independent verification (isolated checkout of `4b96e42b` + test commit `790a1602`)

Every citation re-read against this branch and every test claim re-executed from a clean checkout. The audit also found one coverage gap, now closed by the follow-up commit `790a1602` (no source change).

- `handleRestoreHistory` replaces the selection before the prompt: range, `localStorage.setItem(PROFILE_STORAGE_KEY, …)` + `setProfileId(…)` for a profile entry, `setCustomPromql(entry.promql)` for a custom one, then calls `restoreProtectedDataSource(…)`, which only records the replay and opens the modal — it issues no query.
- On this branch the checkpoint is captured before those writes (`RestoreCheckpoint` at `:353`, built in `handleRestoreHistory` and attached to the replay at `:1079`/`:1111`/`:1133`), and `handleAuthCancel` (`:831`) puts the profile, range, custom expression and persisted profile back before dropping the pending source.
- `handleAuthSubmit` is untouched: it still replays `replay.profile` / `replay.range` / `replay.customPromql` against the newly authenticated source, so a confirmed prompt behaves exactly as before.

**Coverage gap found and closed**

The cancel path restores four values, but only three of them were asserted: a *custom* entry also overwrites the query box before the prompt, and nothing pinned that a draft survives. Added in `790a1602`:

```
× keeps the custom expression draft when a protected custom restore is cancelled   (pre-fix, source reverted)
Error: expect(element).toHaveValue(sum(rocketmq_topic_number) + 1)
Received:
  sum(rocketmq_topic_number)
```

i.e. the history entry's expression had replaced the operator's unsaved draft, and declining the prompt kept the replacement.

**Test claims — re-executed**

```
npx vitest run src/components/__tests__/MetricsExplorer.test.tsx -> Test Files 1 passed (1) | Tests 31 passed (31)
npx vitest run src/components                                    -> Test Files 8 passed (8) | Tests 89 passed (89)
npx tsc --noEmit -p tsconfig.json                                -> clean
npx eslint src/components/MetricsExplorer.tsx \
           src/components/__tests__/MetricsExplorer.test.tsx    -> clean
```

The two new regressions cover the profile-entry and custom-entry halves of the rollback; the neighbouring paths stay green (`keeps the current data source when cancelling a protected history restore`, `requires credentials again when restoring a protected data source history item`, `restores profile, range, and source from query history`). The description was updated to reflect the second test.

